PR #20, opened 2015-01-12
New Login design

songwutk:
I try to change 

login.php (view)
login.css

add new logo.gif (images)

change new look


jekkos:
I ddi some cleanup on the language files, which can be generated from CSV's containing all of the labels and translations per language for each module. I might need to regenerate these after merging in this pull request.

Doesn't seem to be fast forward so first I will create a different branch so I can try out the new UI mods.

jekkos:
You forgot to add a couple of javascript files for the logon screen modifications (slider.js). Where can I get it exactly? It does look better thanks for that!

songwutk:
here is my prototype login design
http://designscrazed.org/css-html-login-form-templates/

Javascript file not in master, I forgot it that I undecided will it "slide to unlock" like iphone ?
But I think it not make sense for desktop POS, so that code remain in file.

PR #23, opened 2015-01-14
Fixing session-table-name bug

marteserede:
It seemed like, that if you change the dbprefix-value in database.php from ospos_ to something else, you will get an error if you don't edit the sess_table_name manually.
It seems like that if just the table with only the value of $config['sess_table_name'] as a name is not found, it will try it with the dbprefix set in database.php in front of it.
Therefore I changed it to just "sessions", so that there is no need to edit this value manually if you change the dbprefix.


jekkos:
ok verified this change on my local install, looks ok.

Issue #25, opened 2015-01-22
Bug: When deleting receiving, stock of item does not change


marteserede:
I noticed that when I receive items, of course, after the receiving the quantity is higher according to the receiving.
If I now go into reports and choose Detailed Reports->Receivings and click on the receiving's ID and click "delete entire sale", the receiving gets deleted, but the quantity of the products which were received does not change...
So when I make a mistake in a receiving, notice it, and delete it, I have to manually change the quantity which can be difficult if I don't write down the quantity which was received...
Or is there a different way to delete receivings?

Same with sales: If I delete a sale through the Sales-Report, the quantity stays the same as if the item was sold...

jekkos:
I have reviewed yout patch and everything looks ok for the quantity corrections when deleting a receiving. For sales however you forgot to invert the correction for both inventory and item quantities. In this case the * -1 should thus be removed.

PR #26, opened 2015-01-22
Fix for bug #39, error in excel_import

marteserede:
- fixed excel import: if there was missing a location-id in the import-csv, then there would not have been created an entry into item_quantities for the item and location.

- fixed also the inventory info-part: before in the inventory-table there was no quantity inserted and data[10] as the location_id


marteserede:
This pullrequest now includes also the commit for fixing bug #25

PR #30, opened 2015-02-02
Adding Indonesian Language

marteserede:
Well, I don't know if it is possible to only select certain commits for the pull request, because the commit from the 30th of Jan you don't need to pull. I don't know why but at I had some problems with this license file, that git always said either I deleted the licenes folder or the license file, so there were always changed files which git wanted me to commit, so I just added the extension to the license file and now everything is fine...


jekkos:
ok thanks for the work here. I don't know if you have noticed the translations folder in the meanwhile. All of the labels are centralized in there now. It's easier to maintain as you see every language in one place (one CSV file for all translations). When you finish with the CSV, just go to the project root, open a terminal and hit 'php generate_languages.php'

all language files will be generated now.

marteserede:
Oh, ok. I didn't know that. I saw the folder, but I was not sure what it was good for. Now I know.
Indeed easier to maintain.

jekkos:
Looks good now. One thing I'd like to say is that when adding config options we'll need to adapt the database scripts (database.sql and version upgrade script) as well (just add default values in there), otherwise things will break on new installations. You can leave this to me as it isn't that much work (will need to create a new upgrade script for 2.3.2 anyway).

marteserede:
Okay, I'll keep it in mind the next time to update the sql-script as well.

Oh yes, I see that regarding the csv file. it is the common-, giftcards- and reports-language-csv file. shall I correct them and commit it again or what shall i do?

jekkos:
That would be nice.. I will look into this pull request as a last thing for 2.3.2 (probably this week).

jekkos:
this pull request was merged in ade913f0e477c1fe848283987c6be931aa2c397d
It might be best to pull from my repo again before making any other changes.

thanks.

PR #93, opened 2015-07-14
Add better local config support

daN4cat:
Here the pull request for what I described in Feature Request 86 (Sourceforge)


jekkos:
Looks good. I'll pull it in later this week and do some testing. Actually I was about to bump the version number anyway (so I guess I'll do that first). Will keep you updated on the progress.

daN4cat:
Thanks, if I find the time later today I'll try to sort out the rules at submit time or remove the text in red if not required.
But I cannot guarantee I will be able to do because I'm flying away for my holidays tomorrow and not back until end of the month.
If you spot a bug I cannot help until end of the month and I know it's not nice to leave to others bugs to fix. It's your choice.

Also, I have more things to submit from my old fork based on 1.1. I never had the time to catch up because having a child changes your life a lot for few years :-)

daN4cat:
By the way, I found the graphical reports returning a 404 error page. Is that correct or do I have an issue with my installation?

jekkos:
Do you have this issue in the latest version? I did do some modifications to the report routing (to add stock locatoin filtering) so maybe this is indeed a bug.

daN4cat:
Yes, the latest.

daN4cat:
The non graphical reports are fine

jekkos:
Ok I'll look into that first then later today. Ok good to know you won't be there, I'll integrate it anyway (hope there won't be too much issues). I'd be glad to see what other work you have done and would be happy to try and integrate it in this current version.

daN4cat:
My branch 1.1 has it more but trying to sync it to newer versions step by step didn't go that far. At certain point I cannot get pages to list anything or searches going wrong so I decided to work the other way round. I take the latest and add bit a bit the features I've added.
I've done work on the reports part adding more features.

However to be honest I had something else in mind for the reports which is to embed Reportico in OSPOS. Reportico: https://sourceforge.net/projects/reportico/?source=directory

I use it externally against OSPOS database and I've created templates for reports we need (my wife's shop).
If we add reportico to OSPOS and basic templates in XML people could contribute with more templates or customise freely for their needs without asking code maintainers to touch the php code.

Moreover Reportico doesn't require flash for the diagrams which is a good thing because Flash is becoming a security issue and browsers like Firefox now block it.

jekkos:
Yes I noticed your branch a couple of days ago. The changes with the current master branch are probably too big to be managable.

In this case it might indeed be better to integrate the functionality step by step. This will take more time, but it allows to do better review of what was built and cherry pick relevant functionality for the current version. I did the same for my modification some time ago. Advantages are that I can keep my own up-to-date with the latest release now (as all functionality I need is in there now)

I agree on the reporting part. For these kind of projects it might be best to keep the original source untouched and integrate the reporting by creating templates or using webservices. This will make the whole update proof (as new versoins of this reportico software will appear). I would curious to see some template examples for the current database layout in any case.

daN4cat:
We'll take the Reportico stuff when I'm back.
The templates I have are linked to the database layout I have modified too.
My reports come with databse modifications because my wife shop is a clothing one, so at the time there was no custom1..10 story and I've added the fields colour, size, season.
Also it's important to know cost and all the markup business, so I show cost in OSPOS on different places.

Most likely in my fork I'll need to drop those custom1..10 fields and readd the ones I have otherwise I need to rework all the Reportico templates again. Too much pain :-)

jekkos:
Yes similar thing here, I have done these things for a clothing sports shop as well. I added a category system that would allow a user to select from a set of attributes depending on the selected category for that item. This is a feature that will need some more rework (I have done some integration for this feature already on a private branch).

Short example: not every item needs similar values for a size attribute. That's why it's useful to link those attributes to a category, like 'running shoe men'. This item will only allow to choose from sizes that apply to that category. This is probably the same for season and colour.

The only work remaining here is the configuration part for this customizable category. It works through conifguring the data directly in the db, but this is probably not sufficient to integrate it in the current master.

daN4cat:
Hi,

I fixed in the Locale config the text in red and added a time format 12h and 24h. Somehow I put h instead of H and was showing time not in 24h format as I wanted.
Probably you want to take that latest file as I'm not sure I can add to this pull? I'll check if possible later.

jekkos:
No worries, Ill pull the changes in directly from your branch, as this PR isn't fast forwarding anyway. Otherwise you'd have to rebase onto my latest master and recreate the PR.

daN4cat:
Sorry, I've done an extra change. I relayout the general config to have a bit of logic with the list of settings.
Maybe would be better to split the company info from the customisations part having an extra tab? No time now for that next round ;-)

best999:
I need custom field in customer data, because i need to include customer cloth size. Can you help me?

jekkos:
Merged in 8bcd2a69. Did a quick sanity check and most functionality seems ok.

daN4cat:
Great, have you merged the changes in your master?
Somehow I don't see on GitHub the merge and shows still differences between my master and your master.

daN4cat:
Disregard my comment, I've just checked your master and see my changes. OK, I'm not that familiar with GitHub and I guess I need to understand how to rebase my master now.

jekkos:
Regarding the reportico reporting, would you mind to share a couple of templates with me? I might want to look into integrating it any time soon.

daN4cat:
To get them to work I would need to update the database adding a couple of fields, otherwise they won't work. But I can look to generate few against the current database.
However I'm on holidays as of Friday for a week, so it won't be until I'm back,

PR #95, opened 2015-08-01
Rebase with 2.3.3

daN4cat:


jekkos:
Think it's already integrated, if you check the latest commit in master (pr #93). As you can see all your work was squashed into one commit (which is cleaner in the history).

PR #96, opened 2015-08-04
Update to suppliers and fixed invoice date bug + license file clash

daN4cat:
Hi,

I've updated the Suppliers part adding an Agency field because often brands are sold via agents or distributors. Maybe Company should be renamed to Brand because that should be the correct label in my opinion. If you agree I'll fix that in a next round of changes.
I put the first name first and family name second because I think it's more polite to have it in that way, but if you disagree please revert that change.
I also added in the Suppliers view the ID because when it comes to import items via the excel sheet you need to know the ID but there is no easy way to get it.
Database file is fine, needs delta file generation before issuing a new release.
Translation files are fine in English and French I believe but lacking rest of translations.

I fixed a date bug in the invoice date that was due to the format selection feature I added. I didn't test the email though.

Please disregard the mess with the license files, I use Windows and there is a problem with having LICENSE file and license dir. Probably LICENSE should be rename to LICENSE.txt or .md or something that doesn't clash with the dir name.


daN4cat:
The CI fails somewhere else, but I'm not familiar with this tool. I would need to learn a bit about it before I can help with this.

daN4cat:
I've fixed the LICENSE and license/ clash happening on Windows renaming LICENSE to COPYING, an allowed name for the file. I also removed license/license.txt to avoid confusing duplication and conflicting statements.

jekkos:
OK just got back from holidays myself. 

In this case travis-ci is failing because of a committed merge conflict (I put a comment at that point in the diff). The phantomjs setup is still a bit sensitive to timing (tests do run ok on my local machine) but fail most of the time on the travis environment. Think this might be improved by moving to their containerized infrastructure (need to check that one out tho)

I might want to pull in all the changes you did after my last rebase (I think starting 1 aug) otherwise history will become polluted, as I have integrated those changes earlier through rebasing your branch with the master.

If you want I can help you out with this but then you'll have to reapply newer changes to the latest commit in master that I'll push out.

daN4cat:
Hi, you go first and I'll rebase from your master sorting out merges. If you do before the weekend I can find some time between Sat and Sun.

jekkos:
Ok I have integrated your last commits, will run the tests locally as a last thing now and then push to master.

Also I suppose that there will need to be a new database script to cope with the new supplier attribute? I'll create it anyway.

daN4cat:
I fixed the merge mistake in the test script.

I've updated the overall database.sql, but I didn't generate an upgrade one to 2.3.3.
I'll look into that later, but we should add to ospos_suppliers -> `agency_name` varchar(255) NOT NULL,

jekkos:
I have merged in your latest work now. I did do a couple of small fixes / mods 

* I would prefer to keep the first name - last name order. Think it's consistent with the other modules this way
* Same for the supplier id, I understand that it can indeed be helpful to display this, but then I would like to have it in each module that supports import (items, customers and employees)
* I added the update script for the change you did
* Also I increased the colspan for the empty result warning message in suppliers screen

It might be best to continue working on a branch, based on the latest commit in master now. The next pull request you make will not include already integrated functionality in that case.

daN4cat:
I noticed that LICENSE is still in master, that should be removed because replaced by COPYING.

daN4cat:
With regards to supplier id, I'm not sure I can follow the rationale of the exclusion.

When you import in items you need to add in the .csv template the supplier id.
There is no view showing that and only in suppliers view you could get that id.

You don't need in customer import any such id, there is none in customer_template.csv.
There is no import in employees view as far as I can see.

For a test click in items on Excel import and then download the template. Now how can a user set a value on the 4th column Supplier ID?

I'll regenerate a pull request with reintroduction of this supplier id display and removal of the LICENSE.

PR #103, opened 2015-08-07
Reintroduction of supplier ID and removal of LICENSE file

daN4cat:
Please read my last two comments in pull request #96 to understand why I'm resubmitting this.


jekkos:
Ok I'll

jekkos:
Ok I understand the inconvenience in this case. I just thought it would be more consistent to have the items and customers display their id's as well. 

Most of the time I derive the supplier id by hovering the edit link, which has the id at the end of the url. This is often needed to correct wrong data or to do direct lookups in the database if something went wrong.

I'll reintroduce this change. After this commit both our master branches should be aligned and I think it'll be easier if submit future pr's based on a fresh pull of the master? I see that the duplicate history is still there (you have both my rebase commits + the original commits you did).

jekkos:
Should be fixed in cc06d93d3048c24a72e68cbc1e40a23f7883440d

daN4cat:
Thank you for accepting the pull request.

With regards to merging, I have cloned your repository a while back and then I do a pull from your one to mine. I follow GitHub instructions going through a merging branch.

I'll check if there is an alternative way to do a rebase without going through a pull?
Alternatively I should create a branch out of your repository to add my changes? I'll look into that as well.

daN4cat:
Yes, it would have been easier to use git rebase jekkos/master .... silly me.
As said I'm learning a bit about Git as I used ClearCase for years, then Perforce for years and in recent years I didn't touch code that much.
I played with GitHub just to store my stuff in one straight code line, not doing merging and else.

jekkos:
Yes most of the time pull requests are sent based on feature branches (based on the latest commit in master of the origin repo). So in this case you would for example copy the master branch locally, and then rebase your changes coming from your own devleopment brnach on top of it (so that the history becomes one straight line). 

doing interactive rebase gives you some fancy features like squashing merge commits or removing or reordering them

then make a pull request for this branch so that you can continue working on your own development branch. This will prevent the pr to become polluted with extra unrelated commits.

PR #105, opened 2015-08-10
Solved to issues in this request. Spanish support, and temp table creation.

gabrielperezs:


jekkos:
This 'if not exists' check might indeed be needed depending on the mysql settings.

PR #108, opened 2015-08-14
More work around Suppliers/Brand

daN4cat:
- Added company_name to summary reports (name of the people didn't make sense in the reports)
- Added search by company_name (brand) in items
- Added company_name as column in items tabular view

### DIFF
--- a/application/models/item.php
+++ b/application/models/item.php
@@ -35,6 +35,7 @@ class Item extends CI_Model
 	function get_found_rows($search,$stock_location_id=-1,$low_inventory=0,$is_serialized=0,$no_description=0,$search_custom=0,$is_deleted=0)
 	{
 		$this->db->from("items");
+		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id');
 		if ($stock_location_id > -1)
 		{
 			$this->db->join('item_quantities','item_quantities.item_id=items.item_id');
@@ -47,6 +48,7 @@ class Item extends CI_Model
 				$this->db->where("(name LIKE '%" . $search . "%' OR " .
 					"item_number LIKE '" . $search . "%' OR " .
 					$this->db->dbprefix('items').".item_id LIKE '" . $search . "%' OR " .
+					"company_name LIKE '" . $search . "%' OR " .
 					"category LIKE '%" . $search . "%')");
 			}
 			else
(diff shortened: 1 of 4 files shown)


application/models/item.php
```diff
@@ -35,6 +35,7 @@ function get_total_rows()
 	function get_found_rows($search,$stock_location_id=-1,$low_inventory=0,$is_serialized=0,$no_description=0,$search_custom=0,$is_deleted=0)
 	{
 		$this->db->from("items");
+		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id');
```
jekkos: I suppose I might need to change this into a left join, otherwise there won't be any result if an item has no supplier defined?

application/models/item.php
```diff
@@ -104,6 +107,7 @@ function get_all($stock_location_id=-1, $rows = 0, $limit_from = 0)
 	function get_info($item_id)
 	{
 		$this->db->from('items');
+		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id');
```
jekkos: Same remark as above here

application/models/item.php
```diff
@@ -135,8 +139,9 @@ function get_info($item_id)
 	function get_item_id($item_number)
 	{
 		$this->db->from('items');
+		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id');
 		$this->db->where('item_number',$item_number);
-        $this->db->where('deleted',0); // Parq 131226
+        $this->db->where('ietms.deleted',0); // Parq 131226
```
jekkos: think this was ought to be 'items.deleted'?

jekkos:
Ok pull request is clean now, thanks for that. Will integrate it first on my local and then push back to github.

jekkos:
Integrated in 4ab2ec.

Might be best to sync with my master branch again (I made the joins into left joins, and corrected the deleted query error in this commit).

application/models/item.php
daN4cat: yes, that's a typo. Strange that I didn't spot it whilst testing.

daN4cat:
Thanks, I'll sync my delivery branch and master too. I have another change done to deliver but I was waiting for this pull to be completed.

PR #109, opened 2015-08-17
Added Cost & Count to reports, improved Items suggest, added copyrights and some fixes

daN4cat:
** Added cost and count to reports **

- Added cost in temp table, reports at bottom of page and in each
summary and detailed report
- Added count in all the reports
- Renamed some report fields to make it generic and consistent across views
- Fixed few issues here and there
- Adjusted some code indentation

NOTE: Proper translation of word Cost is required in all the languages but English

** Improved Items suggestion **

- Added select to restrict results fields
- Added distinct in suppliers instead of the array match (this reverts one fix)
- Added $is_deleted in model part but is not still effective (the controller seems not passing the value coming from the view checkbox, the idea would be to search in deleted items if the checkbox is selected)

** Add copyright statements in COPYING license file **

- Added jekkos and FrancescoUK / daN4cat

NOTE: Other people that contributed should add their statements.

### DIFF
--- a/application/controllers/items.php
+++ b/application/controllers/items.php
@@ -88,13 +88,13 @@ class Items extends Secure_area implements iData_controller
 	*/
 	function suggest()
 	{
-		$suggestions = $this->Item->get_search_suggestions($this->input->post('q'),$this->input->post('limit'));
+		$suggestions = $this->Item->get_search_suggestions($this->input->post('q'), $this->input->post('limit')/*, $this->input->post('is_deleted')*/); // TODO: is_deleted doesn't seem to be effective...
 		echo implode("\n",$suggestions);
 	}
 	
 	function item_search()
 	{
-		$suggestions = $this->Item->get_item_search_suggestions($this->input->post('q'),$this->input->post('limit'),'warehouse');
+		$suggestions = $this->Item->get_item_search_suggestions($this->input->post('q'), $this->input->post('limit'));
 		echo implode("\n",$suggestions);
 	}
 
--- a/COPYING
+++ b/COPYING
@@ -1,6 +1,8 @@
 The MIT License (MIT)
 
 Copyright (c) 2014 pappastech
+Copyright (c) 2014-2015 jekkos
+Copyright (c) 2015 FrancescoUK aka daN4cat
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/application/controllers/reports.php
+++ b/application/controllers/reports.php
@@ -69,6 +69,7 @@ class Reports extends Secure_area
 				to_currency($report_data['subtotal']),
 				to_currency($report_data['total']),
 				to_currency($report_data['tax']),
+				to_currency($report_data['cost']),
 				to_currency($report_data['profit']),
 				$report_data['payment_type'],
 				$report_data['comment']);
@@ -118,7 +119,7 @@ class Reports extends Secure_area
 
 		foreach($report_data as $row)
 		{
-			$tabular_data[] = array($row['sale_date'], to_currency($row['subtotal']), to_currency($row['total']), to_currency($row['tax']),to_currency($row['profit']));
+			$tabular_data[] = array($row['sale_date'], $row['quantity_purchased'], to_currency($row['subtotal']), to_currency($row['total']), to_currency($row['tax']), to_currency($row['cost']), to_currency($row['profit']));
 		}
 
 		$data = array(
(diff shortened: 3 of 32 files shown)


jekkos:
Looks good, I'll try to integrate it later tonight. Thanks.

daN4cat:
Bit a bit I'm trying to port my changes done some years ago to the main codeline.

BTW, I wanted to add the suggest of just deleted items is the Deleted box in Items is selected, but it looks like that in controllers/items.php the is_deleted is not valid.
I didn't have that much time to work more on this, but maybe you can help?
The model/item.php is ready to search items if deleted or not.

On a different topic, I managed to get my master in line with yours removing old commits.
At certain point the rebase went broken, I noticed what I rebased my delivery branch that I was getting again the old confusing story.
I created a delivery2 adding my latest changes and force master with an hard reset to go back your latest and then I merged delivery2 to master.
All seems fine, the only problem is that I lost history of my commits.
No big deal, going forward I'll work in a different way.

application/controllers/items.php
```diff
@@ -88,13 +88,13 @@ function pic_thumb($pic_id)
 	*/
 	function suggest()
 	{
-		$suggestions = $this->Item->get_search_suggestions($this->input->post('q'),$this->input->post('limit'));
+		$suggestions = $this->Item->get_search_suggestions($this->input->post('q'), $this->input->post('limit')/*, $this->input->post('is_deleted')*/); // TODO: is_deleted doesn't seem to be effective...
```
jekkos: I'll have a look using my php debugger. Checkboxes don't post any value except if they're checked.

daN4cat: OK, if you find a means to pass the Deleted checkbox value there then you get suggestions only for deleted items or suppliers which will be handy.

This is one of those things I always need to go to myPhpAdmin to change the flag because in the shop they made a mistake...

daN4cat:
With regards to PHP debugging, I was wondering about this tool that maybe I could install: https://github.com/barbushin/php-console however it will require something to be added to ospos. Worth considering?

jekkos:
For debugging I currently use XDebug in IntellJ, which works really well. CI has some logging functionality built in. Do you have any specific reasons to consider using phpconsole?

jekkos:
By the way I fixed the is_deleted bug, will push your commit and mine later today.

daN4cat:
Good. I'm curious to see what the solution was to the is_deleted bug.

Re debugging, I'll have a look into XDebug when time permits.

jekkos:
I've pushed the integrated version. I added some code to cope with the is_deleted fix. However I have found two remaining issues
* the jquery autocomplete is caching the search suggestions. So once you pull up a suggestion for a deleted item, it will keep appearing even if you uncheck the deleted checkbox in the filter list. I might fix that by clearing that cache once the checkbox's state is changed
* the deleted checkbox in the item details didn't seem to be set correctly when opening an item's details. I'll need to look into that further
* Also I didn't have the time yet to do some testing on the reports part, so might want to look into that later.

daN4cat:
Testing the reports means going through all of them one by one. I did it before delivering but it's better to check multiple times.

jekkos:
Yes it would be nice to have some kind of automated test for that (using the selenium / phantomjs setup on travis) so errors are spotted more quickly.

daN4cat:
I was wondering if the best way to discuss topics would be to create an issue, so we keep history of what we talk about.

I think we have one case for PHP debugging (XDebug as you proposed), one case for the Test Automation / CI.

jekkos:
Currently I think the project is a bit underdocumented, and I've been thinking on improving this for some time 

It might also be a nice idea to summarize our conclusions and store it in a github pages repository for this project. Good thing is that it can be easily managed, as other people can make pull requests to update/add documentation as well (using git).

daN4cat:
Documentation sounds a good idea. New issue for that topic :-)

jekkos:
I've pushed two more fixes for the cache clear and also one change that fixes the delete checkbox. The left join was making the deleted field ambiguous (one for items and one for suppliers table).

daN4cat:
ok, I'll re-sync. I was about to take the latest and test.

daN4cat:
OK, I still don't see it....

daN4cat:
You sure you pushed the changes?

daN4cat:
OK, got it

daN4cat:
I just tried but at least on Google Chrome it doesn't seem to work.
I click Deleted then I get the list of deleted items and if I search by name let say even the first item in the list it's not suggested.
Even if I search a supplier that I deleted it does not appear in the suggestion box.

daN4cat:
Also I noticed that in model/item in get_search_suggestions you added 			
if (!in_array($row->company_name, $suggestions)) {
		$suggestions[]=$row->company_name;
}

But I tried to avoid the issue adding 		$this->db->distinct(); earlier.

Maybe you could try to remove the if(...) and see if the distinct does the trick?

jekkos:
Did you do a hard refresh on the javascript? ctrl-shift-r might help (at least on ff). The in_array check might not be needed for one field's result indeed. I have actually added it to prevent duplicated results across different fields (eg name and description)

Issue #124, opened 2015-09-02
Summary Payment Report


odiea:
Is there a way for all the gift cards to be combined into 1 amount instead of all the gift cards showing in the report.

daN4cat:
When I have time I'll look into this. It should be a sum under generic giftcard indeed.
I guess even the daily takings I worked on is affected by this issue. I'll need to test.

daN4cat:
I fixed this, please read #110 and latest commit.

odiea:
Works great. Thanks

Issue #110, opened 2015-08-17
Change Overview from Sale page to be a daily view


daN4cat:
It would be handy to change the Overview from Sale page to be a daily view reporting even the payment amounts per transaction and a payment total.
This overview could be an handy end of day report that shop keeper can use to countercheck PDQ machine reports and cash drawer amount.

If no objections I would do that change.

jekkos:
You mean like a 'cut Z' report? There are some performance issues in this screen when a large number of rows are present in the database. However I think it's quite handy if you want to search for a customer's sale records in the past.

Maybe we should try to add this daily report to the reports section only? I know that it's not really consistent to have the manage view separated from the regular reports either, but currently I think it's a nice and convenient way to edit/delete sales and regenerate invoices/receipts for customers.

daN4cat:
I find handy to have it from sales because that's what they tend to use in the shop and a click shortcut to see the summary of the day will be appreciated.
A daily report should not have tons of transactions and surely less than what's at the moment. I didn't check if logging out the view is reset or else?

jekkos:
Yes you're right, most of the time we only use the first 1 - 2 pages of this overview. Loading all of the data in that view makes it terribly slow as well. Maybe we can add a couple of filter options (set it to daily by default), while still allowing to search through the whole set of records. 

How exactly would you display the transactions then? Similar as in the report section?

daN4cat:
I can first make it "today" and add the payment amounts and totals.
I can add later a day selection or range on the top bar.

jekkos:
Exactly what I had in mind.

daN4cat:
You can assign this ticket to me if you are fine and mark it as enhancement.

jekkos:
I had to grant push access to be able to assign issues, so now you have access to this repo as well . Howevver I would like to keep the pull request way of working (which allows us to do review on new features and fixes more easily).

daN4cat:
Yes, no worries. Push and pull of code stays the same as I prefer as well. Now I sorted the polluted history issue and I'll keep two branches development and master (for delivery).

daN4cat:
Just checking the code around this sales/manage area and there are mistakes.
In sales.php function manage gets $payment_type as parameter, which is then passed to Sale->get_all. But Sale get_all take $only_invoices as parameter.
This is either all always 0 and always ignored or it's a nonsense.

daN4cat:
[...]

PR #125, opened 2015-09-02
#110 Change Overview from Sale page to be a daily view

daN4cat:
I've done the changes as part of #110, but it's not completed and I need some help (see end of text).

- The Sale/Overview is now using the sale_temp_table to speed up things
- It fetches the daily transactions showing amounts, change and value per payment type (it's now also in the sale detail view as it used to be some years ago)
- There is a total line for the amounts
- There is a table with the totals per payment (the table column is dynamic according to the payment types)
- The change due is removed from the cash total (it's a strong assumption but I would say safe one because there is no case you give a change to a credit card, it's not allowed by the credit/debit card companies as far as I am aware)
- I removed the search box
- general tidy up of code

There is an issue that I'm not able to solve easily because I'm not that familiar with UI stuff (html & js) - yet.
I've left a checkbox to select the only_invoices but it doesn't trigger a view refresh.
However if you put /1 or /0 at the end of the URL you can see that the selection works, so it's really a matter of js magic to get it to work properly.
To do a nice job all the search & suggest related stuff in that form between js & html should go and just change when the user ticks a checkbox.
Maybe even the +/- drop down should be removed and leave the checkboxes visible all the time.

If you could help with that then I could add some check box to have a view for yesterday or some other date.


odiea:
I tried your version and the other problem it appears to have is in the overview it does not show returns. I like the additional feature . Can a date picker(like the reports have) be used in your new feature.

daN4cat:
It shows the overview of the daily transactions and yes, the intention would be to have a date selector when I sorted with the html + js issue, hopefully a bit with jekkos help.

application/helpers/table_helper.php
```diff
@@ -20,52 +22,69 @@ function get_sales_manage_table($sales,$controller)
...
 
-	if($sales->num_rows()==0)
+	if($table_data_rows == '')
+	{
+		$table_data_rows .= "<tr><td colspan='8'><div class='warning_message' style='padding:7px;'>".$CI->lang->line('sales_no_sales_to_display')."</div></td></tr>";
+	}
+	else
 	{
-		$table_data_rows.="<tr><td colspan='8'><div class='warning_message' style='padding:7px;'>".$CI->lang->line('sales_no_sales_to_display')."</div></tr></tr>";
+		// empty line
```
jekkos: colspan will probably need to be 10 now (if I'm not mistaken then 2 columns were added).

jekkos:
As a replacement for the search functionality I would then like to add a link to specific_customer reports in the customer section. Any suggestions on where this link should be positoined are

* In the overview itself (I guess there should be enough room for that at this moment, this will be consistent with eg. the three links in the items table)
* In the detailed customer form (might not be the best from a usability point of view)

application/helpers/table_helper.php
daN4cat: Yep, I'm rubbish with HTML & js wizardry, so much to learn...

daN4cat:
Thanks for the help.

Not sure I get the customer report point though. Do you want to show something specific to a customer?

Please note that I will need to submit a fix to the payment summary / gift card story, but that will fix both issues in the new overview and payment_summary report #124.

daN4cat:
Please note that I fixed #124 with the last commit and also the colspan='8' is now colspan='10'.

jekkos:
Ok I've done some testing and found the following

* saving edit form for sales item doesn't work properly anymore, Ill have a look at it (just open and then save, an error occurs)
* using multiple tables in one page causes issues when updating rows. Javascript seems to bind to both of them. I'd prefer to have a simple text overview of the payment methods at the bottom (like in the other reports) which might look a little cleaner
* I did manage to fix the invoice search by now and made the start/end dates optional. The filtering needs the search method (it used ajax) so I added it back again specifically for this. However this code could be moved out and we can try to use a GET instead of POST (which would do the job as well)
* pagination code is still in there, but maybe not functional anymore? Could still be useful to keep it as we rely on making overviews of the invoices through this screen (for longer period of time)
* maybe we can add some extra date filtering options?

daN4cat:
Re:

- saving edit form for sales item doesn't work properly anymore, Ill have a look at it (just open and then save, an error occurs)

works fine for me. The only thing I can notice is that I didn't change the file for the datepicker removal story.

daN4cat:
For the payment summary I can change to be a list at the end, not great but if it causes issue with js that's what it is.

daN4cat:
Please note that the search method is still in the controller and reads the checkbox value. But I'll have a loom at your change to understand what you mean.

jekkos:
No prob for the payment method, I made it like in the other reports no. Maybe check and let me know if you like it.

jekkos:
Ok I've pushed my changes on top of yours in 9a2cafc1b11ed2cd91f10605b75b9985d781c231 (on a separate branch for now). Edit is currently not working completely. At first there I saw there was a method renamed in table_helper and sales.php had a still a reference to the old name. It does open up the form, but once you click save, then the saved data is not refreshed correctly in the overview.

If you want I can look into that further tomorrow.

daN4cat:
Yes, if you don't mind please fix that. I don't know how come that I don't get the issue.

jekkos:
Ok I've added some fixes in a624d21fcde3500429d1738f2251c10d46908487. Did you intent to show all the payment methods on each line for the sales overview? In case there are more than one then the view might become a little cluttered.

daN4cat:
Thanks for the changes, please see my comments.
Please don't comment out the payments detail per line because it's vital in a real shop!

Date ranges don't look that right to me, but I would need to test. 
Again, we should add the entry boxes in the page for the date range.

daN4cat:
We should rename Overview to be Takings. That's what they use as word normally.

daN4cat:
Something more, this report needs to be printable because each close of day takings report is stapled with the PDQ machine report to a daily register, So a Print button next to Delete one will do.

jekkos:
For printing maybe it's more important to have a proper printing stylesheet?

Printing itself can  be done using the jsPrint plugin, but only on firefox. Beside that, javascript only provides a window.print() which just pops up the default print dialog.

daN4cat:
Yes, I had in mind as first step just the "window.print()". In the end what they use today is detail sale report and print that.
More advanced stuff can be for a second step.

jekkos:
Ok I've merged your last changes in master now. The summary bug is solved, search and filtering seems to work ok at first sight. Maybe better to close this PR then instead of adding commits?

daN4cat:
Let me check your changes a bit now and then we can close.

Things went incrementally because we stormed a bit around this story.

daN4cat:
Something is wrong with the language files.

daN4cat:
I get a failure now and cannot load anything.

daN4cat:
It looks like you committed *.orig_lang.php files that are puzzling things now.

daN4cat:
data range doesn't seem to work properly. I even get two date pickers when click on the date box. One with a top orange bar and then overimposed another one white.

daN4cat:
BTW, the date range doesn't work. Better to check into this tomorrow with fresh mind. Tired now.

odiea:
I got 4 different results from 4 different browsers. Indeed the date picker does not work . Firefox, Internet Explorer, Google  and Pale moon.

daN4cat:
Date picker works fine, but the cash calculation at the summary is wrong.

jekkos:
In which situation exactly? I tested that yesterday for most of the scenarios and everything seemed ok.

daN4cat:
See pictures to understand what's going on. Initial total is wrong, subsequent totals applying filtering are wrong as well. Might be the same root cause, checking the code....

image

image

image

daN4cat:
Hold on, probably I found. It's my test case...

daN4cat:
OK, the issue is with my test case because I put a due amount against a check which is not something you really do. That was deducted from the first view against the cash because of the assumption that due amounts are cash only (unfortunately we don't have a means to track the type of due payment).

daN4cat:
Forget about what I said, I must be sleeping this morning.... had an hectic week and worked until late for this feature in the past days. My brain is full of too many different things. Time to relax over the weekend...

daN4cat:
Thank you for the merge and for working to improve and finish off this feature. Much appreciated.

Let's close this thread.

daN4cat:
Is there any reason why the table sorting got disabled?

I don't think this is now correct:

		$("#sortable_table").tablesorter(
		{
			sortList: [[1,0]],
			dateFormat: 'dd-mm-yyyy', 
			headers:
			{
			    0: { sorter: 'datetime'},
				8: { sorter: 'invoice_number'}
			}
		});

Originally I set 0 & 8 to false because they are the checkbox and and links, all the rest was sortable.
Also the dateFormat now changes according to the one selected in the config.

daN4cat:
ok, 8 is by chance correct, because the links are in position 9. But anyway it doesn't work.

jekkos:
yes maybe I was a little too quick on that one. I think we'll need to change the dateformat to the format configured in ospos. I've added a function to convert a php dateformat to jquery one.

also when reverting my changes sorting isn't working either, trying to look into it now

daN4cat:
Please see bug #131 for the sorting issue.

Issue #111, opened 2015-08-17
Add EAN8 or EAN13 barcode generation in case none is available


daN4cat:
I have already this feature working in my old fork that I could add to master:
 in case the field is empty and the user selects the checkbox and option Generate barcode, a EAN8 barcode is generated using item_id.
The barcode label can then be used to mark products and it's unique for the system since item_id is progressive.

If no objections I would proceed in implementing this feature.

jekkos:
If I'm not mistaken then there should already be something in place to handle this situation. In 2.3.2 you can select a preferred method for barcode generation in the config screen (item_id or upc number). If you select upc and there's no value, then code will fall back to the item_id and generate the barcode using code39 or code128

Are there any other advantages in using EAN-8? Would be nice to have it implemented in a similar way as the other generators now, depending on the amount of work involved.

jekkos:
Maybe the item_id used for barcode generation should be filled in in the upc field as well then? Any suggestions are welcome.

daN4cat:
OK, I missed something there. 
For sure the generated barcode should go in the upc field because if you are sync'ing items with an external system you may want to use that field.
But let me play a bit more with what's implemented already and come back.

daN4cat:
I now remember why EAN8. In my wife's shop they use a Casio label printer which natively supports printing EAN8 and EAN13 barcodes. I used EAN8 because it's shorter but I see a point in having that barcode support.

daN4cat:
You can assign this issue to me if you are fine.

daN4cat:
An example of Label printer with native EAN8 & EAN13 support: https://www.casio.co.uk/products/label-products/Product/KL-7400/

daN4cat:
I quickly checked the current barcode generation classes and they come from a github that has only js code now. I wonder where those php files were coming from.
It will take some time to reuse the same class hierarchy to generate EAN codes because of the maps and etc.

daN4cat:
I could find from the github the EAN13 js code, not EAN8.
I think I will proceed with EAN13 and discard EAN8. In the end EAN13 is supported by Casio label printer.

daN4cat:
OK, I implemented the EAN8 and EAN13  barcode generators using the same base class and structure as for Code39 and Code128. The selection is done from the same config page.

I have still to sort out an issue about setting the correct pix per bar or centre the barcode in the image, either will do.

However what I really need at the end of this is to set a code in the UPC field based on the item_id, if empty.

[...]

PR #148, opened 2015-09-15
Add EAN8 and EAN13 barcode support and fill in empty item_number

daN4cat:
As discussed in #111 here the implementation of the feature adding EAN8 and EAN13 barcode generation support.

Together with that there is a refactoring of the barcode library and classes.

As agreed in case of Code39/Code128 the empty item_number is left.

Also there are small adjustments here and there in terms of code indentation and spacing to make things a bit more readable.

There is in Items view a checkbox to allow selection of empty UPC entries to speed up generation of EAN barcodes.

Barcode is selectable in the configuration. There is no case to support mixed cases and if somebody selects EAN barcode and attempts to print other UPC codes it will produce a EAN barcode. It will then not work to lookup the item stored with a different UPC code.

Language translations are provided but all in English at the moment.

EAN8 and EAN13 barcodes where tested using the Android app "Barcode Scanner" to verify the pictures are matching the UPC code.


jekkos:
Looks good. I did find that some of the EAN barcode are a little off alignment sometimes.

What you think about the 2.3.3 release scope? Think we already have quite a bit of new things to include now. Maybe try to stabilize and fix the last outstanding bugs?

daN4cat:
Strange, I tested various items and EAN barcodes are centered.

The only case where it might be off is if you use an existing UPC code and seed that to make a EAN, but that's not what we are supposed to do, only use empty UPC fields because this is like a "auto sku" story.

I think 2.3.3 scope should now include just some fixes and then baseline. 
But would be good to get the translations sorted if possible, not sure you can pull some string there to get people to help and sort that out in other languages?

Issue #145, opened 2015-09-14
sale/manage (Takings): move Print button above the bars on right side


daN4cat:
The print button is placed in the wrong bar and might lead to confusion because it's the one interacting with any selected item.

It's better to have a Print button above the bars, on the right side at the same level of the view title, like for instance the Excel import button in Items view.

daN4cat:
This is fixed and it will be available in the next pull request after the EAN codes one is closed.

PR #151, opened 2015-09-15
Move Print button in sale/manage (Takings) to top right

daN4cat:
Fix to #145 issue.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,8 +1,11 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 pappastech
+Copyright (c) 2012-2014 pappastech
+Copyright (c) 2013 Rob Garrison
+Copyright (c) 2013 Parq
+Copyright (c) 2013 Ramel
 Copyright (c) 2014-2015 jekkos
-Copyright (c) 2015 FrancescoUK aka daN4cat
+Copyright (c) 2015 FrancescoUK (aka daN4cat)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/application/controllers/config.php
+++ b/application/controllers/config.php
@@ -24,32 +24,32 @@ class Config extends Secure_area
 		
 		$batch_save_data=array(
 		'company'=>$this->input->post('company'),
-		'address'=>$this->input->post('address'),
-		'phone'=>$this->input->post('phone'),
-		'email'=>$this->input->post('email'),
-		'fax'=>$this->input->post('fax'),
-		'website'=>$this->input->post('website'),
-		'default_tax_1_rate'=>$this->input->post('default_tax_1_rate'),		
-		'default_tax_1_name'=>$this->input->post('default_tax_1_name'),		
-		'default_tax_2_rate'=>$this->input->post('default_tax_2_rate'),	
-		'default_tax_2_name'=>$this->input->post('default_tax_2_name'),		
-		'return_policy'=>$this->input->post('return_policy'),
-        'tax_included'=>$this->input->post('tax_included'),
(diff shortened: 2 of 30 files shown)


jekkos:
Ok I've pulled this change in to my local repo alright. 

Did you test the barcode generation on the receipts in the previous pr? Currently there's a problem appearing sales completion.

jekkos:
Honestly I liked the more discrete print button a little more. But I get the point saying that it doesn't apply to printing selected rows, which would be a nice to have anyway. Currently users might think the print does a customized version of the takings layout, which in this case isn't really true (it just uses the default stylesheet for printing)

daN4cat:
Sorry, I forgot to rebase and therefore I pushed things that were from your master by mistake.

I'll have a look into the receipt issue with barcode.

daN4cat:
Crap... sale & receiving use the same barcode function but that cannot be EAN8 & 16.
We need to agree whether it's Code19 or Code128 as those ones are the only we can use for the receipts.

I need to add a new function for that.

daN4cat:
Even item_kits need a fix.

daN4cat:
Ok, but the item_kits issue is something else, at least it looks like that.

daN4cat:
Pushed even a fix to #156 fix.

I hope you can now merge everything to master.

jekkos:
There's another example in the sales model for searching the dates. You'll see that there are no more outer quotes (which is error prone anyway) because the escape function already does this step for us.

daN4cat:
I used the last_query to print the query and it looked ok to me.
But I'll check again this evening and take into consideration your point, thanks for the feedback.
I simply reused that way copying what is used in customers and personnel for instance.
So if we have a better way I would need to make a change across other models to be consistent, but that in a separate step otherwise it gets too big.

daN4cat:
Quickly check sale model and I see the following:

$this->db->where('sale_date BETWEEN '. $this->db->escape($inputs['start_date']). ' AND '. $this->db->escape($inputs['end_date']));

So you mean that item query should be:

'name LIKE %' . $this->db->escape_like_str($search) . '% OR ' .

I need to check better the % and what exactly the function escape_like_str does. As said it was a copy and paste from customer:

last_name LIKE '%".$this->db->escape_like_str($search)."%' or

I would need to fix that one too to remove the " as said earlier.

jekkos:
I tihnk that it should be 

     last_name LIKE ".$this->db->escape('%'.$search.'%')." AND

daN4cat:
Reading from here: http://www.codeigniter.com/user_guide/database/queries.html 

3. $this->db->escape_like_str() This method should be used when strings are to be used in LIKE conditions so that LIKE wildcards (‘%’, ‘_’) in the string are also properly escaped.

$search = '20% raise';
$sql = "SELECT id FROM table WHERE column LIKE '%" .
    $this->db->escape_like_str($search)."%'";

Based on the documentation what I've done seems correct, what do you think?

daN4cat:
Pushed an extra fix to the language issue #156. By mistake I added a - to the Indian translation of "Last" in common_lang.php

daN4cat:
Pushed fix to #155

daN4cat:
Jekkos, can you please check and merge to master this batch?
I believe there is enough now to close this pr.

daN4cat:
Fix to #152.

Jekkos, could you please check, merge and close this pr.

daN4cat:
Jekkos, could you please check, merge and close this pr.

jekkos:
Got it merged in now. I did have one last change that I had already on top of the latest github master and I rebased your changes on top of it, so hopefully this won't cause you too much trouble pulling this repo's changes back in?

daN4cat:
I'm afraid this merge messed up things again....

daN4cat:
Not sure I can see generate_receipt_barcode

PR #169, opened 2015-09-22
Fix dateformat value in SQL-Query app_config

marteserede:
I got an error when trying to edit a sale from the detailed sales-report:
It always said "please enter a valid date" because the date format was d-m-Y. I looked into the settings, changed the format and noticed that there it is d/m/Y which was working fine.
I changed the corresponding SQL-Queries so noone who updates or installs OSPOS and tries to edit the date of a sale gets this error...

### DIFF
--- a/database/2.3.2_to_2.3.3.sql
+++ b/database/2.3.2_to_2.3.3.sql
@@ -2,5 +2,5 @@ ALTER TABLE `ospos_suppliers`
    ADD COLUMN `agency_name` VARCHAR(255) NOT NULL;
 
 INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
-   ('dateformat', 'm-d-Y'),
+   ('dateformat', 'm/d/Y'),
    ('timeformat', 'H:i:s');
--- a/database/database.sql
+++ b/database/database.sql
@@ -68,7 +68,7 @@ INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
 ('default_sales_discount', '0'),
 ('lines_per_page', '25'),
 ('show_total_discount', '1'),
-('dateformat', 'm-d-Y'),
+('dateformat', 'm/d/Y'),
 ('timeformat', 'H:i:s'),
 ('currency_symbol', '$');
 


jekkos:
This fix is probably just a band aid currently. Problem here is that javascript can't deal with custom dateformats easily, and the validator only accepts dateformats that are recognized by the javascript Date object (among them Y-m-d). If the user chooses to customize this format, then 
* we could disable the javascript validation for the sale/receiving edit, but will then allow the user to enter invalid dates causing errors the application won't handle
* do an ajax server side validate of the date and reject the update if the date is not correct
* add an ajax jquery validation validator (might not be that performant) to validate the format server side
* keep the american dateformat in there (Y-m-d) which will be inconsistent with the formatting in the other parts of the application.

jekkos:
In this case I actually prefer the second option. Maybe we can just show the red feedback bar if an incorrect date was entered (and show a message that it wasn't updated). The formats are already converted in php anyway.

marteserede:
I agree, that would be the best option, if we don't want to change the form... but why don't we change input field to 6 drop-down-menus where one can select day, month, year, hour... or single input fields like in wordpress (see screenshot) which is more comfortable than dropdowns (in my opinion), but also assures that the the format is correct and makes it hard to make mistakes accidently (like deleting a dash ("-") or slash ("/") in the input field)...
what do you think?
bildschirmfoto 2015-09-28 um 10 15 52

daN4cat:
I think formatting of date should not be exposed to users and drop down could do. It would be better a date pick up as in Overview/Takings or Items.
But the database format used to log sales, inventory and similar should not be touched as all the code that doesn't deal with the UI part uses that format. It is also an issue for backward compatibility with existing databases.

daN4cat:
I would suggest to close this merge with a "not accepted", said in a kindly way of course :-)

marteserede:
Well, although this merge is not the solution we later want to have, it solves the problem that when installing ospos or upgrading and running the sql-query, the dateformat in the config is a format which is recognized by the javascript date object. m-d-Y seems not to be recognized but m/d/Y does, or did I test it the wrong way?
Nevertheless I suggest and agree that we should create a new issue for the whole date-input-thing...

daN4cat:
Agreed, please help in creating a new issue for the data input story.

jekkos:
Any votes for integrating this pr?

odiea:
Is there any benefit to changing a sale date once it's completed?

daN4cat:
The change is in line with what I've added for the dateformat in view/config:

	<?php echo form_dropdown('dateformat', array(
		'd/m/Y' => 'dd/mm/yyyy',
		'm/d/Y' => 'mm/dd/yyyy',
		'Y/m/d' => 'yyyy/mm/dd',
		'd/m/y' => 'dd/mm/yy',
		'm/d/y' => 'mm/dd/yy',
		'y/m/d' => 'yy/mm/dd'), 
		$this->config->item('dateformat'));
		?>

So it's fine to integrate. That said the possibility to select a date & time with a JQuery plguin would be nice to have. I've tried in a first attempt but something is still not right. Hopefully when I find the time I'll complete the job.

daN4cat:
I now have something working with the a date and time picker in sales/form, but from time to time I get an error saying to enter a valid date due to a rule. The date seems fine to me but I need to understand why that error is so random. There doesn't seem to be any pattern or logic behind it.

jekkos:
There's a custom jquery validator for the date field in this form's javascript, maybe just disable it for the time being (as the datepicker widget will fill in only valid dates anyway)

daN4cat:
Yes, that's already disabled. What's left is the following:

		rules: 
		{
			date: 
			{
				required: true
			},
			invoice_number:
			{
				invoice_number: true
			}
		},
		messages: 
		{
			date: 
			{
				required: "<?= $this->lang->line('sales_date_required'); ?>",
			}
		}

Maybe I can remove the rule date completely but I need to check that there are no scenarios where the user might be able to save something with an empty field.

jekkos:
That should be ok if the widget doesn't allow to edit the field directly. Maybe also remove the custom validator (added to the plugin above)

daN4cat:
Yes, custom validator code is already removed, it was attached to the rule time: inside the date: { .. }

I'll remove the rule completely and test it a bit.

I then need to solve the min.js issue once this is fine and working.

daN4cat:
I've done the change but needs more testing (e.g. different time formats) and also sorting out the min.js.

I decided to make the test input readonly so a user can just edit date & time from the datetimepicker and not make any direct entry. There is an issue with rules and datetimepicker and if you don't set a rule user can remove part of time numbers and submit wrong data. OK, you get a server error if you do that, but thought that make the text input read only will do anyway.

I need to check more, but there is a strange thing with this jQuery plugin, even if not included it looks like working which puzzles me a lot. It looks like that plugin is already present?

jekkos:
I'll pull your changes in later today (or tomorrow) and have a look, maybe I can sort those last things out. Also I'll add the picker to the receivings form.

daN4cat:
Good point, the receiving needs to be modified too. Forgot about that....

daN4cat:
Anyway, I tested with different time formats (from config/locale) and all seems to work fine to me.

The code I submitted in my branch seems fine and what was really needed was the .css of that jquery plugin. I don't know why but the datetimepicker is already available although I cannot see it explicitly in min.js. Is that downloaded from a CDN?

daN4cat:
I'll check this evening the receiving/form and apply the same changes.

jekkos:
maybe it's included through the default jquery - ui import. When I added jquery ui I assembled a custom version which included a couple of widgets already. Chances are that it's already included and so we won't need to reimport the separate file again.

odiea:
Where is this new form located? In sales Takings and I go to edit receipt I get this
2015-10-15 04_31_37-once again thrift -- powered by os point of sale
and it doesn't go away until I press enter then the old date field shows?

daN4cat:
Strange, I tested my code with two different browsers and making sure one was completely clean from any cookie, cache or else it's fine.

Are you sure you used my latest code?

jekkos:
maybe force refresh the javascript files?

daN4cat:
Just re-tried using Incognito tab which means nothing is cached and works fine to me.

odiea:
I got the code yesterday and then rechecked today to see if there were any changes. In the following pic is this where the new date picker is suppose to be?
2015-10-15 04_53_32-once again thrift -- powered by os point of sale

odiea:
This was my last pull request other than this morning.
2015-10-15 05_31_53-program manager

daN4cat:
For the date and time picker just click the Sale date input box, a popup will open like in takings date range selection but + time in this case.

odiea:
No luck with that showing. If I click submit I get a new blank page with the following.
{"success":true,"message":"Sale successfully updated","id":"223"}

odiea:
Have you tried downloading your zipped file and then install that under a different name to see how it works. I am getting the same results from your site osposmaster.zip

daN4cat:
I'll check this evening, but I zipped application, js and css and unzipped this morning before doing a retest with different date & time formats.

But I'll need to retest again to make sure nothing is unusual.

odiea:
I don't know if you get my updated comments so I will send this one along.
If I check the element while the animation.gif is stuck I get the following error in Firefox
 Use of getPreventDefault() is deprecated. Use defaultPrevented instead. opensourcepos.min.js:3:0
 TypeError: $(...).datetimepicker is not a function opensourcepos.min.js:21:1

daN4cat:
I tested with Chrome and Opera. But that point was what puzzled me.
OK, I'll try to make the change to receiving and Jekkos can help in regenerating the min.js which should contain the jQuery plugin. It's everything already there, just I cannot run the gruntfile.js properly because I don't have everything installed.

daN4cat:
Added the datetimepicker to receivings/form too and submitted to my master.

I'll go for a PR and jekkos can work on the min.js then

daN4cat:
getPreventDefault() comes from jquery-1.8.3.js which is probably outdated.

I guess jquery needs an upgrade, but I don't know if it's a trivial and straight forward change or there are impacts to be considered?

daN4cat:
odiea, could you try my latest master and see if you still get the TypeError: $(...).datetimepicker is not a function opensourcepos.min.js:21:1 ?

That should be sorted now.

odiea:
Here are the current results I am getting.
[13:43:05.605] Expected 'none' or URL but found 'alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/ospos.css:112
[13:43:05.617] Selector expected.  Ruleset ignored due to bad selector. @ http://localhost/dan4cat4/css/receipt.css:82
[13:43:05.617] Unexpected end of file while searching for closing } of invalid rule set. @ http://localhost/dan4cat4/css/receipt.css:82
[13:43:05.618] Expected 'none' or URL but found 'alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:40
[13:43:05.618] Unknown property '-moz-opacity'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:41
[13:43:05.618] Error in parsing value for 'height'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:47
[13:43:05.618] Error in parsing value for 'margin-top'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:64
[13:43:05.618] Error in parsing value for 'margin-top'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:137
[13:43:05.618] Expected 'none' or URL but found 'alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:147
[13:43:05.618] Unknown property '-moz-opacity'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:148
[13:43:05.618] Error in parsing value for 'height'.  Declaration dropped. @ http://localhost/dan4cat4/css/thickbox.css:156
[13:43:05.619] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.structure.css:56
[13:43:05.619] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.theme.css:142
[13:43:05.619] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.theme.css:149
[13:43:05.619] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.theme.css:153
[13:43:05.620] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.theme.css:401
[13:43:05.620] Expected 'none' or URL but found 'Alpha('.  Error in parsing value for 'filter'.  Declaration dropped. @ http://localhost/dan4cat4/css/jquery-ui.theme.css:408
[13:43:06.955] Use of getPreventDefault() is deprecated.  Use defaultPrevented instead. @ http://localhost/dan4cat4/dist/opensourcepos.min.js?rel=3806e1c0ff:3
--
[13:43:14.299] TypeError: $(...).datetimepicker is not a function @ http://localhost/dan4cat4/dist/opensourcepos.min.js?rel=3806e1c0ff:26
This is from a pull I did about 10 minutes ago. This is from Pale Moon a version of firefox  that I set to store no info or history.

odiea:
Ok I just downloaded your current zipped version and the date picker is there in Takings edit but where is it in Receivings?

odiea:
I see that on the Date picker that the picker is using the Date of the receipt and that the Now button changes the Date and time to the current Date and Time. Good job Glad you got it worked out. but what about the pull I did earlier?

odiea:
I re-cloned your git and that is working out to.

Issue #132, opened 2015-09-05
Items search by submission time


odiea:
It would be great to add a date picker to the Items list. That way a person is not lost adding new items on a daily basis.

daN4cat:
Please note that there is no date recorded in Items table. The only tracking is done via inventory, but then you get all the changes you do against an item.

What is the real use case here for a date field in Items?

The only case I can think of where you want to record some time is season in fashion/clothing retailing, like my case.

odiea:
We own a thrift store and enter 800 to 1400 items a day. The employees would get lost trying to add Items then finding the Items to print tags or labels.

odiea:
<iframe src="http://free.timeanddate.com/clock/i4u71ug5/n75/fs12/fcfff/tct/pct/ftb/tt0/tw1/tm1/th2" frameborder="0" width="224" height="17" allowTransparency="true"></iframe>

daN4cat:
This is not coming easily like the sales transition. I'll have a look into what's there but personally I cannot commit any time for this feature at the moment.

You can check with Jekkos is he wants to implement this support for a fee.

daN4cat:
Going back to this, the idea would be like in takings to have a date range so you can narrow down the search to what was inserted in a shorter period of time.

The default would all time, then one can narrow down the search. I asked my wife and indeed knowing what was inserted in a specific day or week would be useful to track down mistakes between orders / deliveries and packing list.

I'll look a bit in the code and see how difficult is to implement this. One issue is that the option bar is full so we would need to have two lines.

daN4cat:
I've done an initial part of this which is more on the front end. Now I need to find a way to narrow down the list of items according to the time range against the inventory table.

jekkos:
A query for the last inventory modification time for any item ? Something like

    select max(trans_date), ospos_items.* from ospos_items join ospos_inventory on trans_items = item_id having max(sale_time) between " . $start_date . " and " . $end_date

daN4cat:
Yes, something down that line. It's working but I need to test it a bit.

daN4cat:
OK in my master and submitted a PR. Please test and see if that fits the bill.

odiea:
Date filtering not working for current date,Tax 1 is now showing in tax 2 when going to edit the item. Items entered today may or may not show up in items unfiltered list but Items added will show in suggestions. Looks like more work for you.

jekkos:
Works fine so far on the other hand, no obvious errors or anything.

[...]

PR #170, opened 2015-09-22
Add date range to Items view #132

daN4cat:
Use inventory date to narrow touched items based on selected date range #132

### DIFF
--- a/application/controllers/items.php
+++ b/application/controllers/items.php
@@ -20,9 +20,11 @@ class Items extends Secure_area implements iData_controller
 		$lines_per_page = $this->Appconfig->get('lines_per_page');
 		$items = $this->Item->get_all($stock_location, $lines_per_page, $limit_from);
 		$data['links'] = $this->_initialize_pagination($this->Item, $lines_per_page, $limit_from);
-
-		$start_of_time = date($this->config->item('dateformat'), 0);
+		
+		// assume year 2010 as starting date for OSPOS
+		$start_of_time = date($this->config->item('dateformat'), mktime(0,0,0,1,1,2010));
 		$today = date($this->config->item('dateformat'));
+
 		$start_date = $this->input->post('start_date') != NULL ? $this->input->post('start_date', TRUE) : $start_of_time;
 		$start_date_formatter = date_create_from_format($this->config->item('dateformat'), $start_date);
 		$end_date = $this->input->post('end_date') != NULL ? $this->input->post('end_date', TRUE) : $today;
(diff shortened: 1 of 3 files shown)


odiea:
on Initial date search items appear ok. The item search feature pulls up a suggestion but no item. When I added a new item with the date picker set to todays date the new item does not show until I go back into items. It is not being picked up with the date picker at all.

daN4cat:
OK, I'll have a look into that. Probably I touched something by mistake at the last minute because it was fine (the search and pull the item).
And I'll look into the other issues as well.

jekkos:
Thanks for that, nice feature btw.

daN4cat:
By mistake something went in in model item that was not supposed to be in.

I tried the search, I can see suggestion and the item is fetched.

Bear in mind that search takes always priority, so there is no search of item in narrowed time range to avoid conflicts that might complicate matter.

Otherwise I need to add a checkbox before the time range to make it effective or not.

All the other checkboxes apply to the selected time range.

I'm travelling abroad, so not sure I can look into this more next couple of day, therefore apologies in advance for the quick fix & test and maybe another failure.

jekkos:
I'll look into it when I have the time. The idea of disabling date filtering on search is good, I'd like to introduce that to takings as well.

daN4cat:
I found the time to check why the selected range was missing the end date items and fixed.

odiea:
The update worked well for the items to show now. But I am still getting this issue . The first pic is the intial item inout. The second is the edit item pic. Notice the tax rate.
2015-09-23 14_29_18-once again thrift -- powered by os point of sale
2015-09-23 14_30_06-once again thrift -- powered by os point of sale

odiea:
I just noticed in reports date range, it looks like you forgot to include the language file for some  items

daN4cat:
Could you please create a new issue for the Tax bug? It would help to write down the exact sequence to reproduce that. That for sure has nothing to do with my changes related to date range in Items.
I will look into that under a separate bug.

With regards to Date Range and language translations, it's all submitted as far as I can see: https://github.com/jekkos/opensourcepos/blob/master/translations/reports_lang.csv and search for instance for "reports_this_month_last_year". It's there.

daN4cat:
jekkos, it looks like the feature is fine since I reverted the extra file submission and fixed the data range boundary issue.

After your review maybe you can integrate unless you want to have some other issue/suggestion?

jekkos:
I'll integrate it later today. Thanks for the work.

daN4cat:
:+1:

Issue #165, opened 2015-09-20
Wrong value in sales/complete item total


glassen:
When adding more than one of an item, the item total on the receipt displayed in sales/complete only shows the total for one item. When reviewing the receipt from Takings, the total is correct. This is using master branch from git, checked out earlier today (latest commit eee7f137e).

daN4cat:
I've just tested with my latest (there is a pending pull request that needs to go in see #164 ) and I don't see the problem you describe in the receipt. Please check if this is a correct test scenario?

![image](https://cloud.githubusercontent.com/assets/636048/9982552/3a7fc7a0-5fd8-11e5-9ec6-8421e8d23ae0.png)

glassen:
Mine shows like this (I've started translating to Swedish as well) but only in sales/complete. It is correct when reviewing from takings. I will try again tomorrow with a clean slate an directly from jekkos repo without my changes to translations.
![screenshot](https://cloud.githubusercontent.com/assets/14387561/9982637/389a9310-5fe3-11e5-8757-ba2a0f93b67a.png)

daN4cat:
Translation are not the issue here.

Could you please take the latest from daN4cat/master (my repo) and give a try?
I'm pending a review before merging it in jekkos/master but mine is working and stable with several fixes.

glassen:
Tried your repo this morning with the same results from sales/complete 
![screenshot 2015-09-21 05 45 39](https://cloud.githubusercontent.com/assets/14387561/9985047/5728b67c-6024-11e5-887e-cce302305ba6.png)

But good results from sales/receipt
![screenshot 2015-09-21 05 47 38](https://cloud.githubusercontent.com/assets/14387561/9985054/7f19b78a-6024-11e5-8dc7-7b24dac3fc4f.png)

I also note that there are decimals in quantity when reviewing the receipt from Takings (also applies to invoices discount and quantity). Also note the date difference on invoice from sales/complete and sales/invoice

![skarmavbild 2015-09-21 kl 05 53 40](https://cloud.githubusercontent.com/assets/14387561/9985101/478436d2-6025-11e5-8b69-35cfa01f981f.png)

![skarmavbild 2015-09-21 kl 05 57 28](https://cloud.githubusercontent.com/assets/14387561/9985120/ac35de46-6025-11e5-96eb-f2931af34886.png)

daN4cat:
I cannot really reproduce this issue. I'll check next days to see if I can find any condition triggering the problem.

[...]

PR #176, opened 2015-09-25
#165 Fix invoice wrong date on sale completion

daN4cat:
Fix to invoice wrong date on sale completion and fixed the wrong font for "amount due" shown in the invoice.


daN4cat:
I merged this because it was a small fix to the invoice issue on sale completion.

PR #178, opened 2015-09-26
Feature assets url

souparno:
the helper assets url function will load the assets from the assets path mentioned in the config, this will make it easier to move assets in folder/sub-folder in the future if needed.


jekkos:
the header.php is generated by grunt (grunt-script-link-tags). There's a placeholder in the file that grunt uses to enumerate all the js files in the scripts folder and insert them at that position. If you'd want to use this assets function then the plugin config should be adapted as well to make it work properly.

PR #191, opened 2015-10-05
Better Navigation with TAB key on Sales page

asakpke:
Added better navigation with TAB key on sales page i.e add item and press TAB key one time to edit quantity, TAB key two time to enter paid amount and press enter to add payment, and TAB key two time again and press enter to finish the sale.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -7,6 +7,7 @@ Copyright (c) 2013 Parq
 Copyright (c) 2013 Ramel
 Copyright (c) 2014-2015 jekkos
 Copyright (c) 2015 FrancescoUK (aka daN4cat)
+Copyright (c) 2015 Aamir Shahzad (aka asakpke), RoshanTech.com
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -51,8 +51,13 @@ if (isset($success))
 	<label id="item_label" for="item"> 
 		<?php echo $this->lang->line('sales_find_or_scan_item_or_receipt'); ?>
 	</label>
-
-	<?php echo form_input(array('name'=>'item','id'=>'item','size'=>'40'));?>
+	
+	<?php
+	// SAS
+	//echo form_input(array('name'=>'item','id'=>'item','size'=>'40'));
+	echo form_input(array('name'=>'item','id'=>'item','size'=>'40','tabindex'=>'1'));
+	// EAS
+	?>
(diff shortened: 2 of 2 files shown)


daN4cat:
Thanks for the contribution.

Please remove all the SAS EAS comments if those are to identify you and your changes. We already tidied up the code and you can add your name in the COPYING file.
If everybody adds an initial name it becomes a mixup, just leave the comments and even please remove the commented out lines.

asakpke:
Extra comments removed and added name etc to the COPING file.
@daN4cat Thanks for the point out, is there any guide available for OSPOS coding?

daN4cat:
Documentation, coding and all the wiki part is at the moment work in progress. Maybe after this set of improvements for 2.3.3 we can find the time to document things a bit more.

jekkos:
the tab cycling only works when tabbing the first time through the screen. Afterwards focus cycles through all the present items and tabindex seems to be ignored in firefox, is this expected behavior?

asakpke:
Let me check and come back to you. Right now I'm outside and can't confirm what's wrong.

asakpke:
@jekkos It's working fine on Chrome & FireFox. On first tab it should be on Qty, maybe user need to change Qty and press enter. And user adds other items so on. Two consecutive tabs moves control to the "Amount Tendered", it means that he is going to finish the sale, maybe total amount is 95 and customer given the amount 100. So user should enter 100 and press enter. Now for closing the sale press tab two times to move control on the "Complete Sale" button and press enter two times to finish sale.

asakpke:
@jekkos It is showing following error after the sale instead of showing receipt
"
Server error
500
Hide details
The website encountered an error while retrieving http://localhost/opensourcepos/index.php/sales/complete. It may be down for maintenance or configured incorrectly.
Reload this webpage
Press the reload button to resubmit the data needed to load the page.
"
I have checked data is going to the DB and visible in the reports section.

odiea:
I am also getting strange behavior from the tab key. It will eventually tab thru all my icons even in the header. I also had an issue after using the tab key and looking in takings that apparently the tab key ended up into logout so the screen was showing logout instead of the edit form with the datepicker.

odiea:
Sorry my wife taught me years ago how to act like an average user and how easy it is to break stuff that doesn't have much help info about how to use the feature properly.

asakpke:
@jekkos I think above error (Server error 500) is due to some mistake in sales receipt.

asakpke:
opensourcepos v2.3.1 is working fine on sales recceipt

asakpke:
@jekkos Have you figured out what's wrong with sales receipt?

Issue #192, opened 2015-10-06
Comments and Suspend 


odiea:
It appears that since the Suspend Button was moved now the comments don't work when you suspend the sale even if you go to add payment then enter a comment and it's then suspend it's not being stored in the unsuspend form . Just 0 shows. If you complete a normal sale it functions properly. Like you said one thing is fixed but it might break other things.

jekkos:
Comments should be save with an ajax call when changing the field itself (jquery.change). Maybe it's just not copied properly to the suspend table? Will need to look into that further.

daN4cat:
Fixed and in my master.

The issue was that for a strange reason the comment was fetched directly from the form instead of using the sale_lib->get_comment like in the rest of the code.

So my changes spotted a bug in the code.

odiea:
It now stores a comment but if there is not a comment it still shows a 0. I don't know why that bothers me.

daN4cat:
OK, the 0 is a different matter. It has been there for ages, at least since late 2011.
I'll have a look and see if there is something I can do to remove it. It's in any comment that has not be filled in, either suspended or completed. It's a constant.

daN4cat:
OK, I fixed the comment with 0 issue and it's in my master. I'm going to do a PR.

odiea:
Thanks for finding and fixing this issue.

jekkos:
Maybe we should also add an update statement that removes these 'empty' comments from the database.

odiea:
That would be a good idea . Makes things look a lot cleaner.

daN4cat:
You mean clean the all the 0's. Yes it sounds a good idea indeed. But we need to make sure it's =0 not LIKE %0% otherwise we might mess up existing comments having a 0 in it (just to state the obvious here)

jekkos:
Yes indeed, only update comments that exactly equal 0.

daN4cat:
odiea, could you please check the latest fix with sale and receiving, with or without suspend for sales?

I tested and it seemed fine with me, I also checked that suspending and unsuspending the other fields are correct. But again, you never know.

If all fine after odiea test we can close this issue.

daN4cat:
I must admit that this 0 issue bugged me for many years but never found the time to fix it.

Just one line of code in the end....

odiea:
[...]

PR #194, opened 2015-10-07
#192 Remove 0 from comments

daN4cat:
Checked in library get_comment if the string is null and return a '' if that's the case. That removes the 0 from sale comments

### DIFF
--- a/application/controllers/sales.php
+++ b/application/controllers/sales.php
@@ -204,9 +204,13 @@ class Sales extends Secure_area
 		if ( $this->form_validation->run() == FALSE )
 		{
 			if ( $this->input->post( 'payment_type' ) == $this->lang->line( 'sales_gift_card' ) )
+			{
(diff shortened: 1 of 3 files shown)


jekkos:
OK thanks for the pr. I guess the only real change here is in sale_lib?

Maybe it might be better to reformat the code for the whole project all at once, because now it's kind of hard to see what has actually changed (beside the obvious reformatting of this code)

daN4cat:
Yes, the real change is in sale_lib and receiving_lib (did the same on both sides).

But I also removed some non-sense from controller sales.php.

I noticed that there is in one case data['comments'] and in another case data['comment'], it should be probably the same and 'comments' but that needs to change on the view side too.
I'll do later when I find the time.

Yes, you are right. The reformatting should happen at once. I do it in a kind of automatic way when I read the code and trying to make it less "compressed".

jekkos:
I'll integrate it now then (haven't tested it but suppose it'll be ok)

PR #203, opened 2015-10-15
Added datetime picker to sales & receivings forms (see #169)

daN4cat:
See discussion in #169 for more details.


application/views/partial/header.php
```diff
@@ -32,7 +33,7 @@
 	<!-- end js template tags -->
     <?php else : ?>
     <!-- start minjs template tags -->
-    <script type="text/javascript" src="dist/opensourcepos.min.js?rel=3806e1c0ff" language="javascript"></script>
+    <script type="text/javascript" src="dist/opensourcepos.min.js?rel=c3fc0e5615" language="javascript"></script>
```
jekkos: Looks like grunt cache breaker did run as the hash in the query string has changed?

daN4cat: I run grunt that run your Gruntfile.js. 
I don't know the details of it, but I suppose that if you touch a file (having added a new jQuery plugin there is more now) an hashing needs to be updated. Right?

jekkos: yes indeed, the fact that there's a new query string parameter present will assure that a user's cache will be refreshed (thanks to url rewriting). So a new hash means new minified javascript and a forced cache refresh.

daN4cat: OK, so all good and makes sense.

Once you tested it a bit, please integrate and we close this a bit long story of broken plugin, new plugin and etc.

BTW, jQuery core is old and maybe should be updated. Any idea if it's a complicated matter? I'll create an issue for it later.

jekkos: Yes I have looked into that beginning this year. At that time jQuery was a

odiea:
Got a little carried away on that time picker did you?

jekkos:
hehe yes.. I've added jquery ui sliders and now there's no more select boxes for that, this works out of the box.

odiea:
Did you see js print setup has been updated in the last few months?

daN4cat:
not sure about the slider thingy, would it work with a touch screen no problem?
That's why I avoided it.

odiea:
Yes it works. I just tried it on my home screen and both the slider and dropdown in receivings work with a touch screen.

daN4cat:
ok, good. Case closed then.

daN4cat:
Almost closed... receiving is with the select not slider. Correct?

odiea:
2015-10-18 11_55_15-once again thrift -- powered by os point of sale
Yes this is my image from report receiving

PR #208, opened 2015-10-20
Update register.php

songwutk:
colorize button


jekkos:
These kind of updates should be done in line with the new theming approach. Will close this pr for now.

PR #244, opened 2015-11-30
Fixed Severity: Notice Message: Only variable references should be returned by reference Filename: core/Common.php Line Number: 257

kamaroly:
Severity: Notice Message: Only variable references should be returned by reference Filename: core/Common.php Line Number: 257


jekkos:
Ok I see that in CI this change was applied as well.

https://github.com/bcit-ci/CodeIgniter/issues/3302

Issue #212, opened 2015-10-21
2.3.3 release ready?


jekkos:
I think it's time to think about releasing 2.3.3.

Are there any open bugs / features that need to be included? Only thing I want to retest is the takings pagination (didn't find an opportunity to do that yet).

odiea:
Currently I do not see any issues with this build. Takings pagination appears to be working good. I have tested it with it showing 6 pages.

daN4cat:
There are some issues tagged with 2.3.3, probably we should review them and device 2.3.3 or 2.3.4 or 2.4.
For instance the Takings print was one thing I wanted to do, but I also need to sort out the bacode story with is half done.

A bit of config tab retouch too, just moving few bits around and add a tab related to the DB.

Maybe the conversion of 5 custom fields to be colour, size and season can be in a 2.3.4, but that's key for my needs as I cannot move my wife's app yet to the newer because of that issue.

jekkos:
Good point, indeed there's still 3 issues open for this release, let's try to have them solved and then wrap up. I will have a look at #187. Then the only one left is indeed the barcode, best to solve that in this one as the code is already integrated.

daN4cat:
If you could help with #187 that would be great. I found that partial input search very handy in the past.

I'll try to finish the barcode and the adjustment to the retouch to config tabs. We'll see for the Takings print.

But as said I need to delay of one week any effort as I'm away.
BTW I'll be in Brussels. I go there from time to time because part of my wife's family lives there.

jekkos:
We could do a 2.3.4 with only minor changes and bugfixes after releasing this one? More like a bugfix release?

Scope of 2.4 could be the following

* CI 3
* multistore enhancements (working on that module now)
* configurable categories (quantity units branch?)

Or depending on the effort to move to CI3, maybe migrate it in a minor release?

jekkos:
OK cool I live near to Brussels. Will go there as well to night (concert there in AB).

daN4cat:
Maybe we can leave the Takings print, search story and other extra bits to 2.3.4.
I would just need to fix the barcode story.

In 2.4 I would need to convert 5 custom fields to 3 as previously mentioned.

daN4cat:
Maybe next time I'm Brussels we could have a beer? I'm in Brx just for few days.
I knew you live in Leuven.

jekkos:
[...]

PR #246, opened 2015-12-07
Partial fix to #111 issues

daN4cat:
As discussed in #212 here the partial fix.

We can leave the full feature to a 2.3.4, this fix solves part of the problem anyway.

### DIFF
--- a/application/controllers/items.php
+++ b/application/controllers/items.php
@@ -345,7 +345,7 @@ class Items extends Secure_area implements iData_controller
 			if ($item['item_number'] == '' || $item['item_number'] == null)
 			{
 				// get the newly generated barcode
-				$item['item_number'] = $this->barcode_lib->get_barcode($item, $config);
+				$item['item_number'] = $this->barcode_lib->get_new_barcode($item, $config);
 				
 				// remove from item any suppliers table info to avoid save failure because of unknown fields
 				// WARNING: if suppliers table is changed this list needs to be upgraded, which makes the matter a bit tricky to maintain
--- a/application/libraries/Barcode_lib.php
+++ b/application/libraries/Barcode_lib.php
@@ -138,8 +138,16 @@ class Barcode_lib
 			else
 			{
 				$barcode->setData($item['item_id'], true);
+
+				$code = $barcode->getData();
+
+				// in case no new code is generated like in Code39 and Code128 return an empty string because we don't want to override it with a pure item_id				
+				if( $code == $item['item_id'] )
+				{
+					$code = null;
+				}
 				
-				return $barcode->getData();
+				return $code;
 			}
 		} 
 		catch(Exception $e)
--- a/application/libraries/Barcode_lib.php
+++ b/application/libraries/Barcode_lib.php
@@ -121,7 +121,7 @@ class Barcode_lib
 		}
 	}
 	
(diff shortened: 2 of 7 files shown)


jekkos:
I've done some research and I don't think that it's a good idea to start generating EAN13 codes ourselves. They should be unique and registered with a central authority first. 

Maybe we can solve the ean formatting problem with a sane validation on the involved format of the determined barcode input for the configured barcode type. If this validation fails, then just fallback to code128 which will always work.

jekkos:
I've added the validation + fallback code in 83436d93ce3d8bf22a6c195979c1e2a1f2f053f6.

jekkos:
Might be good to trigger the validation code in the item form as well (using jQuery).

daN4cat:
Have you removed the EAN8 generation?

daN4cat:
If that's the case, thank you very much!

You made the feature I surely needed completely useless and I need to scrap what you have done.

jekkos:
Problem was that your scanner only accepts ean barcodes? My idea was that generating them might cause collisions with other existing barcodes out there as they're not registered in any official way.. 

If you like then I'll reintroduce it but only after enabling a config option to generate instead of using fallback.

daN4cat:
That was the point I made already, having a config to generate only, not scrapping the code somebody spent time on it!!!

To be honest I found this approach quite disappointing :-(

The issue is not the scanner, the issue is that items come without any barcode and there are handheld printers that produce a small barcode to be attached to small items. EAN8 was there for that purpose because supported by those label printers.

I can agree that there might be collisions, I wan under the impression that padding the front with 0 was excluding the collision issue after having read in a forum that suggestion (it's ike a "192.168.x.x").

daN4cat:
The use case was described in #111

jekkos:
Yes I suppose that there will be some kind of range that's free to use, will look into that later today. So I'll add the following then

 * config switch for generating in the requested format if the item_number is empty. If this switch is not set then the barcode will be validated against the chosen format, and fallback mechanism will be activated if needed
 * extra jquery validation on the used barcode format for ean13 and ean8 to prevent input of invalid codes once a format is selected

daN4cat:
I'm not sure I follow bullet point 2, the issue here is to generate barcodes if the field is empty. And the first option is just for that. If empty generate it with EAN{8, 13}. If no option the seed is always used.

We can print all the barcodes in code{39. 128} and that will do. There is no requirement to print the barcode in EAN{8, 13}, only if an external printer supports that can print in EAN{8, 13}. The scanner will detect the two barcodes no matter the "graphical format" and match the UPC code.

That's what I wanted to do, splitting generation out of printing.

Of course we could always leverage a recognition of the barcode and be smart in producing the correct barcode printout (in case somebody has a EAN only scanner, but I don't think there is such case).

So what is that jquery needs to validate?

jekkos:
I added this validation bit in the item number check code. So once a user enters a new item and has ean8 selected as format, he should enter a ean8 formatted barcode in the item_number or else leave it empty. This will still allow the generation to work (only if the item_number is empty) and prevent the user from entering invalid barcodes

jekkos:
Ok so in this maybe the first part will do. If I add the generation only for ean then we'll be fine. Ill post a patch for that hope you can take a look and see if it's ok for you.

Issue #247, opened 2015-12-10
Change the store\'s configuration 


odiea:
Will someone please change this desc in Modules.Change the store's configuration

RyanfaeScotland:
To what? Why? Where does it appear in the front end (I know where it is in the code but can't find it on the front end to test changes)

odiea:
It appears on the Home page when you first log into ospos. I believe it is the bottom Item. To me it is just an annoyance but I believe someone else brought it up too.

RyanfaeScotland:
Ah yes I see it and see the issue. I never scroll down that far in normal usage, preferring the buttons up top!

That's it fixed on my branch. Just in the middle of trying to get a test merge/review done with some of my work, hopefully this will spur it on. ;)

odiea:
Total Customers Wanting Stuff (Ignore the Pound Sign!): $2.00 What does this reference in your comments report?

RyanfaeScotland:
Just a count of the customers on the screen. Only the customers with comments are pulled onto the report as we use the comments for customer requests. This report lets us see how many customers are looking for items and what items they are looking for.

The formating of the currency sign is hard coded into the summary so for now I've just marked it as Ignore the Pound Sign (it is dollars in yours) but eventually I want to remove the formatting from the base class and have it render what it is passed.

odiea:
Ok. Thanks for the info.

RyanfaeScotland:
No worries. If I ever figure out how to get these gh_pages working I'll write this up as I go.

Inventory Facebook is a public ready inventory list. Only shows items that have > 1 quantity along with selling price, category and quantity. Great for copy and pasting to Facebook to let people know what you currently have in stock! (Although I plan on removing category and quantity as they make it a bit messy)

jekkos:
Duplicate of #138

PR #248, opened 2015-12-10
add customer comments + inventory reports

RyanfaeScotland:
Some extra useful reports and a bug fix for #247


application/language/en/reports_lang.php
```diff
@@ -6,6 +6,8 @@
 $lang["reports_categories_summary_report"] = "Categories Summary Report";
 $lang["reports_category"] = "Category";
 $lang["reports_comments"] = "Comments";
+$lang["reports_comments_customer"] = "Customer Comments";
+$lang["reports_comments_customer_report"] = "Customer Comments Report";
```
jekkos: Did you change the language csv file first? These files are auto generated with the generate_langauges.php script in the root of the project.

RyanfaeScotland: No I edited them direct. Feel free to reject and I'll redo. Do we have instructions written up for this process or will I write them as I go?

jekkos: Might be good indeed to have some reference about this, as new people often edit the generated files instead of the CSV's. Just a matter of opening them in excel and then editing the columns for each language.

afterwards the script has to be run as foillows

     php generate_languages.php

RyanfaeScotland: Never hurts to have auto-generated files declare themselves as such with a little auto-generated line at the top of the file. Will look to get on it this weekend.

application/models/reports/inventory_facebook.php
```diff
@@ -0,0 +1,48 @@
...
+		return array($this->lang->line('reports_item_name'),
+					$this->lang->line('reports_category'),
+					$this->lang->line('reports_count'),					
+					$this->lang->line('reports_unit_price'));
+	}
+	
+	public function getData(array $inputs)
+	{
+	    $this->db->from('items');
+        $this->db->join('item_quantities','items.item_id=item_quantities.item_id');
```
daN4cat: Would be good to have the same indentation

RyanfaeScotland: I think there is a file I need to add that sets my indentation to be updated automatically. Now that I know my commits can be merged with others I'll focus more on keeping myself consistent with you guys.

application/models/item.php
```diff
@@ -49,7 +49,7 @@ public function search($search, $filters, $rows=0, $limit_from=0)
 	{
 		$this->db->from('items');
 		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
-		$this->db->join('inventory', 'inventory.trans_items = items.item_id');
+		$this->db->join('inventory', 'inventory.trans_items = items.item_id', 'left');
```
jekkos: Might be better to make a separate pr for this issue. You can do that by adding this change on a separate branch, starting from master.

application/views/reports/listing.php
```diff
@@ -51,6 +52,7 @@
 		<?php 
 			show_report('', 'reports_inventory_low');	
 			show_report('', 'reports_inventory_summary');
+			show_report('', 'reports_inventory_facebook');
```
jekkos: maybe rename the last report to inventory_customer? Which extra fields does it show in the report? If I find the time later today then I'll have a closer look.

application/models/item.php
jekkos: Otherwise if too hard I can pull in your master and cherry-pick this change + the typo correction already. But then you'll have the risk of running into trouble when syncing back with my repo afterwards.

application/models/reports/inventory_facebook.php
jekkos: Think ospos is using spaces as indentation currently, similary as in codeigniter. Might be good to have git convert these automatically.

application/views/reports/listing.php
RyanfaeScotland: Yeah I was thinking that as well but needed it to be clear for my none-techie partner who was also using the system and at that point I wasn't sure how the naming worked. Going to get that revised.

Annoyingly (very annoyingly) I've already had to shut the business down due to accounting issues. I'll still commit to this because I think it is a great project but I'll have to deal with the fall out from the business first.

daN4cat:
What do we want to do with this PR?

Not sure we want to include this report or not, I'm even not sure what the purpose for it was...

RyanfaeScotland:
Hey @daN4cat , completely up to you guys of course but here was my thoughts behind it. We we're using the comments to record customer requests for specific items and once a week we'd print off the report and see if we could source any of them.

Eventually I planned to have a requested items field that would let me mark items from our potential inventory (i.e. items that are out of stock but may come back in) to a customer so that it would automatically flag when an item a customer was looking for came back in but in the meantime this manual solution was doing the job.

daN4cat:
I suggest to close this PR and not take the code in. The use case would be interesting if fully automated, this solution is manual and seems to me a bit ad hoc.

Issue #271, opened 2016-01-04
I just got this error after the latest pull.


odiea:
Parse error: syntax error, unexpected 'if' (T_IF) in C:\xampp\htdocs\tabbed\application\controllers\sales.php on line 21

daN4cat:
I can see, there is typo: 		$person_id = $this->session->userdata('person_id');git 

see the git at the end?

daN4cat:
try to remove it, I'm testing and then commit a fix.

odiea:
I did see that and it eliminated that error but now look at my Register screen.
![2016-01-04 14_59_47-ugly sweater store -- powered by os point of sale](https://cloud.githubusercontent.com/assets/10962177/12101810/fa2014fc-b2f3-11e5-975e-4c5611638f48.png)

daN4cat:
I guess that if you grant access to Takings it's sorted, right?

daN4cat:
I don't get that problem: 

![image](https://cloud.githubusercontent.com/assets/636048/12101918/6730a008-b32f-11e5-94de-d2ac83349d19.png)

odiea:
I do not see any new permissions for takings.

daN4cat:
It's the same as reports/sales

odiea:
Yes I already have those permissions. I will revert back and see if the problem is still there.

daN4cat:
ok, but the spurious git was the root cause of the initial bug you had

odiea:
I reverted back to my original and this is how it should look.
![2016-01-04 15_12_13-ugly sweater store -- powered by os point of sale](https://cloud.githubusercontent.com/assets/10962177/12102077/b4cdf944-b2f5-11e5-8ba1-cde1c8cf0584.png)

daN4cat:
OK, I pushed the fix with PR #272 and it's now in jekkos/master.
I'm sure @jekkos doesn't mind this quick one as it's a typo from one of his changes and removes a break with xampp from master.

daN4cat:
For the UI issue, check the changes in application/views/sales/register.php

odiea:
The only thing I can figure is it was a bad pull. I downloaded the latest zip and got the pages from there and its all ok now.

PR #272, opened 2016-01-04
#271 fix to error witn xampp due to a spurious "git"

daN4cat:


daN4cat:
Sorry I just quickly push through this as it's a small one

PR #274, opened 2016-01-06
added de-CH

nano-riff:
I added new language. Please note that "generate_languages.php" does not handle "form_validation_lang.php".


jekkos:
Thanks for the PR. I would like to rebase all these into one commit for cleaner history.

jekkos:
Yes I know that some of the files do have quotes to delimit the fields in the CSV. Depends on the editor whether they are added or not.

daN4cat:
OK, clear.

jekkos:
With escape I guess I meant that once you use comma's, you need to use quotes around the whole sentence. Not sure how csv would handle this otherwise. 

Anyway I check the latest commit and there's still some issues left (mainly encoding problems), so not sure if we can do something about that.

nano-riff:
I'll make a last try using a dedicated CSV Editor with UTF-8 encoding. We'll see.

nano-riff:
original common_lang.csv seems to be corrupt in line 24

nano-riff:
In order to work with de-CH I cheated (my) the system by copying all translations into "en" and select English in config. This way everything works. While selecting 'de-CH' in locales will not allow submits...strange, allthough all entries are the same.

jekkos:
Ok works fine now. Thanks for the contribution.

jekkos:
Btw submit works on my local setup, even with swiss language selected.

daN4cat:
Just tested and looks fine to me, German-CH is fetched and fine.

BTW, @jekkos, should be Dutch be listed as Dutch (Flemish) or Dutch (Belgium)?
We should try to be consistent here if possible.
I know English is generic, probably there should be US and UK versions at least but I never really paid attention to it. I'll check in the future.

daN4cat:
English US/UK, I spotted one. I wrote at the time "Localisation", that's UK, US version should be "Localization".

PR #279, opened 2016-01-11
Update common_lang.csv

asra1st:
indonesian language update


daN4cat:
Hi @asra1st , thank for the language update. Have you checked all the translations? Not sure when Indonesian was updated last, but there are other things that were added in 2.3.3, for instance Agency, then Takings.
Just to make sure everything is covered.

PR #283, opened 2016-01-13
php7 fix

ais-one:
Some fixes to make it work on PHP7 without errors (warning still occurs).


daN4cat:
Good one

ais-one:
In PHP 7 under development mode, the warnings are causing problems with ajax

in index.php

    case 'development':
         error_reporting(E_ALL);
    break;

Suggest change E_ALL to E_STRICT

Error message displayed in E_ALL was:

> A PHP Error was encountered
> Severity: Warning
> Message: Declaration of Customer::save(&$person_data, &$customer_data, $customer_id = false) should be compatible with Person::save(&$person_data, $person_id = false)
> Filename: models/customer.php
Line Number: 277

Issue #258, opened 2015-12-30
Migrate from old version phppos to ospos


lutfiajib:
I have phppos 11.3 installed
How to migrate to current version?
I dont have any SQL file prior 2.2.1
Anyone can help me?

daN4cat:
I assume that with phppos you mean http://phppointofsale.com/.

I don't know which database schema they have, if they have the same tables and so on.
Most likely items, sales, and all basic stuff is the same. But I suspect you would need to do quite some manual work.

I will face something similar soon because I need to migrate a DB from OSPOS 1.1 to the latest 2.3.3.
In my initial tests I exported all the tables content one by one and imported in the new database and table adding the missing columns and renaming what was different.
I'm not sure this is the best way to deal with this, I'm not a big expert of MySQL databases, so my help is probably not the most expert one you might get here.

jekkos:
Think this approach is probably the best, export the data and then import it again in the new table structure. You'll need to deal with a couple of issues as well here (add data for new columns eg).

I did something similar some time ago and wrote an sql for this. If you want then I can share it here. It literally copied data from the old database into the new one.

jekkos:
I just checked the script and I think it can do more or less what you are looking for, however it requires some modification as it contains a couple of queries for tables that only exist in another branch. It does everything in sql,

So maybe someone can clean it up and the contribute it back so I can add it to the repo?

* query to insert categories has to be adapted
* item_sizes table does not exist in master (remove constraints)

jekkos:
I've added the script to the source and did some baisc cleanup. Maybe test a little futher and update where needed. I made it in february 2015 so any changes done to the db layout since then should also be reapplied to this script.

jekkos:
Think the create tables can be removed as well. In this case only the inserts copying the actual data are what you need. Just create the schema's from the usual database.sql script.

jekkos:
Did an exstra updatre for this script. It's best to create all tables first and add constraints after the data copy is done. MIght be best to split up the initial database.sql script in two parts, one for creating tables and one for adding constraints

* database;sql will then source both script 1 and script 2
* migrate script will source script 1, execute the existing insert/update statements, and then source script 2

lutfiajib:
I'm still got difficulties using scripts. But daN4cat is correct. i already done manually:
1 make ospos database
2. remove all constrain from phppos
3. remove all constrain from ospos
4. compare table one by one, and modify phppos table one by one
5. then copy table one by one
[...]

PR #317, opened 2016-02-04
Fixes to phppos to ospos 2.3.4 migration script #258

daN4cat:
Fixes to phppos to ospos 2.3.4 migration script #258


jekkos:
Ok great thanks for testing, one thing is that the patch should be applied on the original migrate script (dist is the concated result of tables, migrate and constraints). Will try to see if I can solve that quickly.

daN4cat:
Let me fix that, I also forgot to make the change for admin entry.

daN4cat:
OK, fixed.

jekkos:
Ok great, thanks for finalizing this.

Issue #336, opened 2016-02-17
Fix sales stock level to work off item reorder level instead of  0 or below.


odiea:
I don't know what procedure  this error is coming from but it would be nice to have this fixed also.

daN4cat:
Is this a bug or new feature? Not sure I get.

odiea:
Currently the warning does not show unless the items are 0 or below. It would be nice to have the warning show when the items qty is at or below the reorder level. I currently have a way around this that shows the warning when the qty is 2 items below the reorder level.

daN4cat:
Ok, got your point

odiea:
Ok. I got it figured out. The following will show the warning when it reaches the reorder level.
function out_of_stock($item_id,$item_location)
	
	{
		//make sure item exists
		if($this->validate_item($item_id) == false)
        {
            return false;
        }
		$item_info = $this->CI->Item->get_info($item_id);		
		$data['reorder_level'] = $item_info->reorder_level;
		$item_quantity = $this->CI->Item_quantity->get_item_quantity($item_id,$item_location)->quantity;
		$quantity_added = $this->get_quantity_already_added($item_id,$item_location);
		
		if ($item_quantity - $quantity_added <= $data['reorder_level'])
		{
			return true;
		}
		
		return false;
	}

jekkos:
Would be nice to integrate this change in 2.4

odiea:
Feel free to integrate any of my ideas. That's why I post them so others can try or integrate them into OSPOS

daN4cat:
@odiea no chance you can use github and go for a PR so it's easy to integrate?

odiea:
Just let me know how to do this.

jekkos:
If you edit the files directly on your github repo, you should be fine. Then you can make a pr just like that. It's quite easy.

jekkos:
Eg the snippets you paste here, just click edit on the file in your cloned github repo, ofc after choosing the correct branch first. github wil automatically add those changes to a new branch so you can make pr more easily with this repo.

odiea:
https://github.com/odiea/tabbed-dev/tree/tabbed-dev is the branch that contains my changes. If you can use it great.I have to go to work now.

odiea:
So what do I need to do. Start with a virgin copy of ospos then add my snippets to that because all mine are hacked versions.

odiea:
[...]

PR #365, opened 2016-02-27
Fix sales stock level to work off item reorder level instead of 0 or below #336

daN4cat:
Fix sales stock level to work off item reorder level instead of 0 or below #336

### DIFF
--- a/WHATS_NEW.txt
+++ b/WHATS_NEW.txt
@@ -1,3 +1,15 @@
+Version 2.3.4
+-------------
++ Migration script fixes
++ Improved continuous integration setup
++ More integration tests
++ Virtualized container setup (docker install)
++ Live clock functionality + favicon
++ Improved PHP 7 compatbility
++ Added de_CH (German) as language 
++ Minor code cleanup
++ Removal of annoying backup prompt on logout
+
 Version 2.3.3
 -------------
 + Item kit fixes (search, list, ..)


application/libraries/Sale_lib.php
```diff
@@ -336,19 +336,19 @@ function add_item($item_id,$quantity=1,$item_location,$discount=0,$price=null,$d
...
 		if($this->validate_item($item_id) == false)
         {
             return false;
         }
-
-		
-		//$item = $this->CI->Item->get_info($item_id);
+		$item_info = $this->CI->Item->get_info($item_id);		
+		$data['reorder_level'] = $item_info->reorder_level;
 		$item_quantity = $this->CI->Item_quantity->get_item_quantity($item_id,$item_location)->quantity;
```
jekkos: Why save this value in a associative array here? Can't we declare it directly?

odiea: I'm not a programmer so that's what I came up with that worked. Please clean it up if you want to.

jekkos: ok will do anytime soon

jekkos:
I had a closer  look at this patch. Maybe we should show another warning then the quantity is insufficient? At least there are items left to sell so stock wise there's nothing really wrong as you're at least not having negative stock counts for the product at hand..

odiea:
Yes that would be great. I just didn't want to play with the language files. The warning  will still work when it is 0 or below though.

odiea:
Should 2 warnings be placed 1 for low inventory and 1 for 0 inventory ?

jekkos:
Good point. Maybe just one or the other? Maybe the reorder level warning is only meaningful is the  level itself is higher than 0?

odiea:
Yes. If it has both then someone can see that the level is low before going completely out.

odiea:
Or change the wording to Your inventory level is low or out

odiea:
So I duplicated the out_of_stock and set the new one to low inventory. I changed the warnings by duplicating them also. Does this sound like too much work or can it be set to work in the current low inventory function without duplicating these things.

jekkos:
I've added a patch for this issue in latest master. An extra change in the code was needed to show the warnings when editing quantities directly.

Issue #377, opened 2016-03-05
feature BootstrapUI cannot create or update item


ais-one:
Cannot create or update item because does not recognize the "none" dropdown option for supplier, fix is to remove the  required class on the input

ais-one:
Pull request sent

[https://github.com/jekkos/opensourcepos/pull/378](url)

daN4cat:
Merged

PR #378, opened 2016-03-05
Fix for issue #377

ais-one:
Fix bug #377 where item cannot be edited or created due to not recognizing none as a supplier.

### DIFF
--- a/application/views/items/form.php
+++ b/application/views/items/form.php
@@ -47,7 +47,7 @@
 		</div>
 
 		<div class="form-group form-group-sm">
-			<?php echo form_label($this->lang->line('items_supplier'), 'supplier', array('class'=>'required control-label col-xs-3')); ?>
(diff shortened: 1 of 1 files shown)


application/views/items/form.php
```diff
@@ -49,7 +49,7 @@
 		<div class="form-group form-group-sm">
 			<?php echo form_label($this->lang->line('items_supplier'), 'supplier', array('class'=>'required control-label col-xs-3')); ?>
```
daN4cat: you need to remove it from the label too

ais-one: I have removed it in next commit, I did not remove it in earlier commit as I thought we could still display label as a required field.

daN4cat: It makes sense to do not make it mandatory as some people might just want to create items without bothering about suppliers and etc.

Issue #339, opened 2016-02-20
PHP 7 Compatibility


nairgh:
A PHP Error was encountered
Severity: 8192
Message: Methods with the same name as their class will not be constructors in a future version of PHP; open_flash_chart has a deprecated constructor
Filename: ofc-library/open-flash-chart.php
Line Number: 63
A PHP Error was encountered
Severity: Warning
Message: Declaration of Customer::save(&$person_data, &$customer_data, $customer_id = false) should be compatible with Person::save(&$person_data, $person_id = false)
Filename: models/customer.php
Line Number: 273
A PHP Error was encountered
Severity: Warning
Message: Declaration of Employee::save(&$person_data, &$employee_data, &$grants_data, $employee_id = false) should be compatible with Person::save(&$person_data, $person_id = false)
Filename: models/employee.php
Line Number: 367
A PHP Error was encountered
Severity: Warning
Message: Declaration of Supplier::save(&$person_data, &$supplier_data, $supplier_id = false) should be compatible with Person::save(&$person_data, $person_id = false)
Filename: models/supplier.php
Line Number: 310
Similarly so many errors

jekkos:
The app is not compatible with php7 yet.

daN4cat:
And why should we?

daN4cat:
We will look into this going forwards.

nairgh:
Happy to see this going. Always my support available. By the way it is a  very useful project. I don't want to see the grey side of this ..

daN4cat:
Would be useful to get some help with this once the bootstrapUI step is completed.

daN4cat:
@ais-one & @nairgh could you please work on this using bootstrapUI branch since you seem to have PHP7 installations?

Of course it has to be at least 5.5 backward compatible, so it should not break things.

ais-one:
Hi I will look into this as well, yes I am working using php7 on both win (32 bit) and linux (64 bit).

I do recall seeing those errors and having a fix. It is definitely a php7 thing.

Will report on fix later.

ais-one:
Btw, php7, I had to use mysqli, I read there is some issue on this too, which issue is this happening? I did not read through to see what was failing.

daN4cat:
mysqli is set in the database config since we moved to CI3.

ais-one:
@nairgh 

What version of PHP are you using, I get the same problem for versions below 7.02 on XAMPP (both windows and linux version).

[...]

PR #383, opened 2016-03-07
Fix For #339

ais-one:
https://github.com/jekkos/opensourcepos/issues/339

We should also rollout the fixes to  develop/2.4 if necessary. This fix
was adapted from BootstrapUI which fixed the PHP 7 error regarding
classes.


daN4cat:
No need to merge this because it's solved in 2.4, now in master.
It was a 2.3.4 issue pre CI3.

PR #415, opened 2016-03-21
Fix for 393 - issue with bulk edit

ais-one:
https://github.com/jekkos/opensourcepos/issues/393


application/views/items/form_bulk.php
```diff
@@ -152,12 +152,8 @@
 //validation and submit handling
 $(document).ready(function()
 {	
-	$("#category").autocomplete("<?php echo site_url('items/suggest_category');?>",{max:100,minChars:0,delay:10});
-    $("#category").result(function(event, data, formatted)
-    {
-    });
-	$("#category").search();
-
```
jekkos: The css issue can better be solved by using appenTo in the autocomplete call, check usage in other forms.

jekkos:
Merged in bootstrapUI now. cb26c8cf3307943b7a80547897fb08b17d33e392

PR #436, opened 2016-03-31
Feature/bootstrap ui

joesalty:
Hi!

I've created Hungarian language for bootstrap.
Here is the files.
[translations.zip](https://github.com/jekkos/opensourcepos/files/197149/translations.zip)

EDIT: Something wrong with the language file, because if I switch to hungarian, I can't modify the items.
So don't use this. Yet... Thx


daN4cat:
Sorry but this PR cannot be accepted. You are attempting to merge bootstrapUI into master, which is not what we are supposed to do here.

Therefore I close this PR.

PR #501, opened 2016-04-23
Mannual sms feature 

RamkrishnaMondal:
Mannual sms feature


application/views/messages/sms-sender.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+* Change the  variables 'uid', 'pwd', 'phone' and 'msg' according to your sms-api on line no. 55 & 56 and in the form-field names below.
+* Change the '$res = sendsms' part (on line no. 56) according to your 'sms-api.php'
+* Put your 'sms-api.php' file in views/messages folder and change the file name below (on line no. 8) accordingly.
```
daN4cat: I'm afraid this is not correct.

The sending of the message needs to happen via the controller not the view.
What we discussed seems having not been implemented.

database/database.sql
```diff
@@ -43,9 +43,6 @@ INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
 ('barcode_page_cellspacing', '20'),
 ('barcode_generate_if_empty', '0'),
 ('receipt_show_taxes', '0'),
-('receipt_show_total_discount', '1'),
```
daN4cat: This must stay, it cannot be removed it's in master.

PR #503, opened 2016-04-25
new

RamkrishnaMondal:


application/helpers/table_helper.php
```diff
@@ -125,7 +125,8 @@ function get_people_manage_table($people,$controller)
 	$CI->lang->line('common_first_name'),
 	$CI->lang->line('common_email'),
 	$CI->lang->line('common_phone_number'),
-	'&nbsp');
```
daN4cat: removing this is not a good idea because the icons will not be placed correctly in the table

application/language/en/module_lang.php
```diff
@@ -6,14 +6,16 @@
...
+$lang["module_messages_desc"] = "Send Messages to Customers, Suppliers, employees";
+$lang["module_giftcards"] = "Giftcards";
+$lang["module_giftcards_desc"] = "Manage Giftcards";
 $lang["module_home"] = "Home";
 $lang["module_item_kits"] = "Item Kits";
 $lang["module_item_kits_desc"] = "Add, Update, Delete and Search Item Kits";
 $lang["module_items"] = "Items";
 $lang["module_items_desc"] = "Add, Update, Delete, and Search items";
-$lang["module_receivings"] = "Receivings";
+$lang["module_receivings"] = "Recevings";
```
daN4cat: typo, this should be reverted

application/helpers/table_helper.php
```diff
@@ -207,8 +212,9 @@ function get_supplier_manage_table($suppliers,$controller)
 	$CI->lang->line('common_first_name'),
 	$CI->lang->line('common_email'),
 	$CI->lang->line('common_phone_number'),
+	$CI->lang->line('common_send_msg'),
 	$CI->lang->line('suppliers_supplier_id'),
-	'&nbsp');
+	$CI->lang->line('common_update'),);
```
daN4cat: the last '&nbsp' is for the edit glyphicon, don't remove that

application/controllers/Messages.php
```diff
@@ -0,0 +1,26 @@
...
+	/*
+	Loads the sms sender form
+	*/
+	function view($person_id=-1)
+	{ 
+		$data['person_info']=$this->Person->get_info($person_id);
+		$this->load->view("messages/sms-sender",$data);
+	}	
+}
+?>
```
daN4cat: This would be correct, you need to add a function send as discussed already.

application/views/messages/sms.php
```diff
@@ -0,0 +1,58 @@
...
+	      </div>
+	    </div>
+	  </fieldset>
+	</form>
+       </div>
+
+	<!---------------------------------------------- END OF FORM ------------------------------------------------------>	
+	<?php
+
+	if (isset($_POST['uid']) && isset($_POST['pwd']) && isset($_POST['phone']) && isset($_POST['msg'])) {
```
daN4cat: you don't need this, you should check other views. Take store config as example where there is a submit button on each tab.

See how the view is done and how the data is read from the controller side.

database/database.sql
```diff
@@ -288,10 +293,11 @@ CREATE TABLE `ospos_modules` (
 --
 
 INSERT INTO `ospos_modules` (`name_lang_key`, `desc_lang_key`, `sort`, `module_id`) VALUES
-('module_config', 'module_config_desc', 100, 'config'),
+('module_config', 'module_config_desc', 110, 'config'),
```
daN4cat: why this change?

PR #504, opened 2016-04-26
Change some Thai translations

sharecode:


daN4cat:
Thank you for the translations contribution.

However before merging I would suggest to go through the other .cvs files as some were added recently. It would make sense to get a general fix and merge everything in one go.

What do you think?

PR #517, opened 2016-04-28
Pequenas modificações

arteiro:
Alterados e incluídos palavras nos arquivos .csv


jekkos:
Thanks for the translations. However we currently don't support any translation on the readme (or we'll have to switch to another language) so if you could then maybe omit that part?

arteiro:
Ok, I do not know how to use the right desktop github, put on my copy to anyone who is Brazilian not hit much. if you give problem retreat there.
So I made one with .csv.
Hugs
Thank you

daN4cat:
Merged

arteiro:
OK,

Thanks

Issue #472, opened 2016-04-10
Trying to add manual SMS feature.


RamkrishnaMondal:
I am using  manual sms sending tool written in php to send messages to my customers. The tool is working fine if i use from my localhost root directory ( ie var/www/ ). But showing lot of errors while trying to use it as a' module' from within the OSPOS, as the tool is written to use 'mysql' not 'mysqli' driver.

Can any one help converting the mysql functions to mysqli function? 

![screenshot from 2016-04-10 14 06 54](https://cloud.githubusercontent.com/assets/14151032/14408928/83675892-ff25-11e5-928b-8fb2c7b57ed3.png)

Using 'error_reporting(0); it looks like below : 

![screenshot from 2016-04-10 17 53 58](https://cloud.githubusercontent.com/assets/14151032/14410169/c786eafe-ff45-11e5-8d0a-bf46e201d7ff.png)

jekkos:
This code looks like an ugly hack. You should try to use CI models + controllers, add new entries in modules table and all should be fine.

jekkos:
Just read CI documentation a bit and you'll find out how it works

RamkrishnaMondal:
Thanks to @jekkos, as per his advice using CI models I have done the following:

![screenshot from 2016-04-15 20 43 30](https://cloud.githubusercontent.com/assets/14151032/14570856/fb39b2ec-0363-11e6-88da-8a9813569a6d.png)

![screenshot from 2016-04-15 20 42 57](https://cloud.githubusercontent.com/assets/14151032/14570791/a4336dd0-0363-11e6-8629-8e418821b564.png)
![screenshot from 2016-04-15 20 42 33](https://cloud.githubusercontent.com/assets/14151032/14570794/aaa703fc-0363-11e6-8720-f6e6347c3982.png)
![screenshot from 2016-04-15 20 42 07](https://cloud.githubusercontent.com/assets/14151032/14570803/b16a0dce-0363-11e6-86f2-85546b558f03.png)
![screenshot from 2016-04-15 20 41 39](https://cloud.githubusercontent.com/assets/14151032/14570862/07b017fa-0364-11e6-89fe-74cbcb890d46.png)


Wanted to use the Modal dialogue like this one:

![screenshot from 2016-04-15 23 38 01](https://cloud.githubusercontent.com/assets/14151032/14570906/355e0284-0364-11e6-9d8a-f88aee9d835a.png)

[...]

PR #524, opened 2016-04-30
SMS message sending feature #472

daN4cat:


jekkos:
@daN4cat @RamkrishnaMondal Thanks for this contribution. It might be nice to support implementation of a spi through a comon api for sending sms. A good usecase here would be to attach a cell phone to pc over serial line and send messages using a simcard inside this cell phone. This might be a lot cheaper in comparison with using a sms gateway as most providers give a lot of free sms on prepaid cards.

daN4cat:
The API in library/sms.php should be generic enough. However if you want to dynamically load custom interfaces for extra functionality even at configuration time the design is not there indeed.

RamkrishnaMondal:
@jekkos  I was also thinking of using an 'AT command supported GSM' modem alongwith 'kannel', as I am on Ubuntu based Linux.

Here is a nice how to : https://aacable.wordpress.com/2012/11/26/howto-configure-your-local-http-gateway-using-kannel-on-ubuntu/

RamkrishnaMondal:
@daN4cat  I just read the 'read me' of ospos, was expecting my name there. Perhaps you have forgotten to update that file. Not a big issue though.

daN4cat:
Hi @RamkrishnaMondal you are right, apologies for that. I checked many things but forgot to update the COPYING file. Let me do that now.

RamkrishnaMondal:
Thanks.

daN4cat:
done

PR #671, opened 2016-06-18
Update Customer.php

likozhang:
check this same like before

### DIFF
--- a/application/models/Customer.php
+++ b/application/models/Customer.php
@@ -177,7 +177,7 @@ class Customer extends Person
 		$this->db->group_start();		
 			$this->db->like('first_name', $search);
 			$this->db->or_like('last_name', $search); 
-			$this->db->or_like('CONCAT(first_name, " ", last_name)', $search);
+			$this->db->or_like('CONCAT(first_name,last_name)', $search);
 		$this->db->group_end();
 		$this->db->where('deleted', 0);
 		$this->db->order_by('last_name', 'asc');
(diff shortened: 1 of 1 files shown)


application/models/Customer.php
```diff
@@ -241,7 +241,7 @@ public function get_found_rows($search)
 			$this->db->or_like('email', $search);
 			$this->db->or_like('phone_number', $search);
 			$this->db->or_like('account_number', $search);
-			$this->db->or_like('CONCAT(first_name, " ", last_name)', $search);
+			$this->db->or_like('CONCAT(first_name,last_name)', $search);
```
daN4cat: This change just removes a space between first name and last name.

So instead of Liko Zhang you get LikoZhang.

Anyway you closed, but I wanted to make sure that was clear.

likozhang: oke, thankyou

daN4cat: That said, I'm not sure that CONCAT really makes sense because there is already:

> 			$this->db->like('first_name', $search);
> 			$this->db->or_like('last_name', $search);

I'm trying to read the code to understand what the purpose of that was.

likozhang:
yep but in case when search firstname(ABC) lastname (DEF)
and search with "ABC DEF" ? 
db->like(firstname(ABC),Search(ABC DEF))??
or 
db->like(lastname(DEF),search(ABC DEF))??

may use like '%Search%'

daN4cat:
CI like or_like does that for you already.

PR #673, opened 2016-06-19
Update Sale.php

likozhang:
Validation for calculation


likozhang:
test

for this bug
$this->db->where('DATE(sale_time) BETWEEN ' . $this->db->escape($filters['start_date']) . ' AND ' . $this->db->escape($filters['end_date']));

no filter in search and calculate

likozhang:
@daN4cat how about this ?

Issue #678, opened 2016-06-20
change display for print receipt


likozhang:
![win_20160620_23_47_13_pro](https://cloud.githubusercontent.com/assets/6987941/16202791/73461436-3741-11e6-924b-439e4eea9bee.jpg)
![win_20160620_23_47_25_pro](https://cloud.githubusercontent.com/assets/6987941/16202792/73490eac-3741-11e6-8bf8-60c6139f2d67.jpg)

font to small because fit the width
how about

ITEM NAME
Qty Price           sub_total_price
Item Name 2
qty price            sub_total_price

                          total_price

likozhang:
`<?php $this->load->view("partial/header"); ?>

<?php
if (isset($error_message))
{
	echo "<div class='alert alert-dismissible alert-danger'>".$error_message."</div>";
	exit;
}
?>

<script type='text/javascript'>
function send_email() {
	$.get('<?php echo site_url() . "/sales/send_receipt/" . $sale_id_num; ?>',
		function(response) {
			$.notify(response.message, { type: response.success ? 'success' : 'danger'} );
		}, 'json'
	);
}

$(document).ready(function()
{
	<?php if(isset($email_receipt) && !empty($email_receipt)): ?>
		send_email();
	<?php endif; ?>

	<?php if(isset($customer_email) && !empty($customer_email)): ?>
		$("#show_email_button").click(function(event) {
			send_email();
		});
	<?php endif; ?>
});
</script>

<?php $this->load->view('partial/print_receipt', array('print_after_sale'=>$print_after_sale, 'selected_printer'=>'receipt_printer')); ?>

<div class="print_hide" id="control_buttons" style="text-align:right">
	<a href="javascript:printdoc();"><div class="btn btn-info btn-sm", id="show_print_button"><?php echo '<span class="glyphicon glyphicon-print">&nbsp</span>' . $this->lang->line('common_print'); ?></div></a>
	<?php /* this line will allow to print and go back to sales automatically.... echo anchor("sales", '<span class="glyphicon glyphicon-print">&nbsp</span>' . $this->lang->line('common_print'), array('class'=>'btn btn-info btn-sm', 'id'=>'show_print_button', 'onclick'=>'window.print();')); */ ?>
	<?php if(isset($customer_email) && !empty($customer_email)): ?>
[...]

PR #679, opened 2016-06-20
Update receipt.php

likozhang:
tes


daN4cat:
I would like to make a standard receipt and a shorten version of it selectable from the Store Config.

So I will merge this but creating a different type of receipt.

daN4cat:
See #678 for commit notes.

PR #962, opened 2016-11-07
Translate to Spanish for new config field to filter with datetime format

jlctmaster:
Hello. 

Add the translate to Spanish for new config field to filter with datetime format, please check the changes.

Greetings.


jekkos:
Ok thanks looks alright

PR #979, opened 2016-11-17
Implemented dotenv used for different configurations for different environments

deepshah:
Sample file: .env.example is stored in application/config
Copy it to application/config/.env

Add custom settings as required in the .env file.


application/config/config.php
```diff
@@ -358,7 +358,7 @@
 | https://codeigniter.com/user_guide/libraries/encryption.html
 |
 */
-$config['encryption_key'] = '';
+$config['encryption_key'] = isset($_SERVER['ENCRYPTION_KEY']) ? $_SERVER['ENCRYPTION_KEY'] : '';
```
daN4cat: This clashes with another feature I added to automatically generate an encryption key and add it here.

For this type of change we need to change the encryption support and save the autogenerated key in .env if that's the way we want to go.

application/config/config.php
```diff
@@ -41,7 +41,7 @@
 | Logs are stored in application/logs
 |
 */
-$config['db_log_enabled'] = FALSE;
+$config['db_log_enabled'] = (ENVIRONMENT !== 'production');
```
daN4cat: Not sure I would make this automatically enabled if not production.

Better to treat it independently?

deepshah: Okay, have changed it.

application/config/config.php
deepshah: Which file contains the feature? I will see if I can update it.

application/config/database.php
```diff
@@ -75,10 +75,10 @@
...
 $db['default'] = array(
 	'dsn'	=> '',
-	'hostname' => 'localhost',
-	'username' => '',
-	'password' => '',
-	'database' => '',
+	'hostname' => isset($_SERVER['DB_HOSTNAME']) ? $_SERVER['DB_HOSTNAME'] : '',
+	'username' => isset($_SERVER['DB_USERNAME']) ? $_SERVER['DB_USERNAME'] : '',
+	'password' => isset($_SERVER['DB_PASSWORD']) ? $_SERVER['DB_PASSWORD'] : '',
+	'database' => isset($_SERVER['DB_DATABASE']) ? $_SERVER['DB_DATABASE'] : '',
```
jekkos: I remember using getenv() in the docker environment. Might be better to use that one directly instead

application/config/config.php
deepshah: would it be fine to add it in the upgrade notes?

README.md
```diff
@@ -79,9 +79,9 @@ Local install
 1. Create/locate a new mysql database to install open source point of sale into
 2. Execute the file database/database.sql to create the tables needed
 3. unzip and upload Open Source Point of Sale files to web server
-4. Copy application/config/database.php.tmpl to application/config/database.php
-5. Modify application/config/database.php to connect to your database
-6. Modify application/config/config.php encryption key with your own
+4. Copy application/config/.env.example to application/config/.env
+5. Modify database settings in application/config/.env to connect to your database
+6. Modify encryption key in application/config/.env with your own
 7. Go to your point of sale install public dir via the browser
```
deepshah: I've made these changes.

jekkos: Thanks!

application/config/.env.example
```diff
@@ -3,7 +3,7 @@ CI_ENV="production"
 ENCRYPTION_KEY=""
 
 #Database Settings
-DB_HOSTNAME="localhost"
-DB_USERNAME=""
-DB_PASSWORD=""
-DB_DATABASE=""
\ No newline at end of file
+MYSQL_HOST_NAME="localhost"
```
jekkos: Indeed this was the last required change, will merge this one asap.

deepshah: @daN4cat what do we do about the encryption key? do we keep it as it is in the current master, or do we add it in .env and add instructions in the upgrade file?

daN4cat: To be honest I still have a gutfeeling that this ENV story is opening a backdoor to some attack.
Something to read carefully: http://stackoverflow.com/questions/10090014/can-i-trust-my-environment-variables

Therefore I prefer to leave the encryption key as it is at the moment.

Even here they suggest to encrypt what goes in the ENV vars: https://help.fortrabbit.com/env-vars

Also we should probably sanity check the ENV content from any malicious code.
We need to be careful....

jekkos: Well indeed I guess that in this case phpinfo could leak the environment values in some cases. I had this feeling as well and came across the same stackoverflow post. On the other hand the dotenv thing seems to be widely promoted as a best practice in managing environment variables. That made me think that they must have addressed possible security issues. 

Beside that it might also be a good idea to check the software's manual on how to configure it properly.

daN4cat: So let's leave the encryption key out and the automatic generation as it is at the moment (write directly to config.php file).

The config file can pick up the other ENV var if one wants to, but we don't bind the code to dotenv. That's what I prefer at this stage.
Just a connect dotenv <-> config file and nothing more.

Agree?

jekkos: What do you mean exactly with just a connection? In this case the values are loaded using getenv() call. I just found in dotenv documentation that they don't advise to use it in production. In that case we should just stick with using the config files as before and leave dotenv usage for dev/test environments (overriding the default values in config)

daN4cat: Amended as I meant to say dependency.

So, we can add dotenv in composer, rename config.php to config.dotenv and have a .env. But we keep the config.php clean as it is at the moment.
We write something in the wiki for the ones that want to use this support.

jekkos: Or we could leave code in this pull request in the username / password fields and instruct users to fill in their credentials directly. This saves us the config.php file duplication and leaves the setup intact to use dotenv if desired. This also facilitates docker setup which is done through environment variables anyway.

deepshah: I did not follow the last two comments.  Can you explain them?

Not sure if you'll meant this but this what if we have two config files? One (which is under version control) imports the other (not under version control). So instead of environment variables, we have regular PHP variables.

daN4cat: So you mean to make it default.
But then we need to remove the encryption key from the ENV vars otherwise I wouldn't know how to add from code automatically the key.

deepshah: Sorry not PHP variables, but constants. 
We do not keep any settings under ENV vars, we define all environment related settings as constants. Like you mentioned earlier, encryption key stays in config.php as it is now.

daN4cat: ENV are variable from an environment point of view, from PHP perspective they are constants indeed.

jekkos: That could be an option as well. Using define in a separate file for production values. If they are not filled in then we can fallback and look for a variable in the ENV (in this case filled in by dotenv)

jekkos: Indeed that's what I meant to say.

deepshah: I'll fix it in a day or two

daN4cat: @deepshah just add each singular `if(empty(XYZ))` after the config item so we keep it confined to each atomic part.

jekkos: @daN4cat Where did you mean to add this statement? In the database.php file itself instead of the ternary check for the env variable?

daN4cat:
@deepshah do you plan to complete this work with the latest comments?

jekkos:
As currently most installs have their env set to production, debugging problems for new installs becomes harder. Think this PR would be useful as adding a dotfile can then set it back to development mode which should enable display_errors in PHP.

daN4cat:
@jekkos this needs completion. But you need to get it to a closure.

jekkos:
So I guess we're find then using .env file instead of the regular CI config.php? In any case the latter is adapted to pick up the settings by reading the from an environment variable.

Only thing left to fix is the encryption key generation which is now done directly in config.php, will move this to env file then as well

jekkos:
Might still be better to stick with config.php directly for production environment, otherwise credentials might leak into phpinfo.php which can pose posible issues

daN4cat:
@jekkos indeed, production environment needs to be safe we cannot use this story as a security risk.

jekkos:
PR was integrated in latest master. A default database.php is now filled in with the standard credentials. .env files can be used to override this configruation is desired. In that case credentials are  visible as environment varialbles. Using the old config.php method is sitll recommended  for production settings.

.env file also makes it easy to enable display_errors in php. This is useful to generate error messages when troubleshooting installation problems. This releases the user from seeing a blank page in some situations

PR #989, opened 2016-11-20
Modifications to group by functions which should help with MySQL 5.7 compatability

SteveIreland:
Sorry about the multiple commits on this.  I meant to consolidate them.


odiea:
Can this be pulled into a new branch for testing?

daN4cat:
@SteveIreland great, thank you for the contribution.
I'm afraid there are some conflicts to solve but I do hope they are the last ones as I want to release 3.0.1!

application/models/Receiving.php
```diff
@@ -238,28 +238,28 @@ public function create_temp_table(array $inputs)
 			' (INDEX(receiving_date), INDEX(receiving_id))
 			(
 				SELECT 
-					DATE(receiving_time) AS receiving_date,
-					receiving_time,
+					MAX(DATE(receiving_time)) AS receiving_date,
```
daN4cat: Not sure I get, why MAX function is used here?

application/models/Item.php
```diff
@@ -60,13 +60,61 @@ public function get_found_rows($search, $filters)
...
+
+        $this->db->select('max(item_quantities.item_id) as qty_item_id');
+        $this->db->select('max(item_quantities.location_id) as location_id');
+        $this->db->select('max(item_quantities.quantity) as quantity');
+
+        $this->db->from($this->db->dbprefix('items').' as items');
+        $this->db->join($this->db->dbprefix('suppliers').' as suppliers', 'suppliers.person_id = items.supplier_id', 'left');
+        $this->db->join($this->db->dbprefix('inventory').' as inventory', 'inventory.trans_items = items.item_id');
+
+        $this->db->from($this->db->dbprefix('items').' as items');
```
daN4cat: this section looks a copy & paste of the above. Mistake?

vendor/codeigniter/framework/system/database/drivers/mysql/mysql_driver.php
```diff
@@ -281,7 +281,6 @@ protected function _prep_query($sql)
 		{
 			return trim($sql).' WHERE 1=1';
 		}
```
daN4cat: Please revert this change, we don't touch codeigniter original code unless there is a recognised bug stopping us.

application/models/Item.php
```diff
@@ -60,13 +60,61 @@ public function get_found_rows($search, $filters)
 	*/
 	public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'items.name', $order = 'asc')
 	{
-		$this->db->from('items');
-		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
-		$this->db->join('inventory', 'inventory.trans_items = items.item_id');
+
+        $this->db->select('max(items.name) as name');
```
daN4cat: I think just using `$this->db->select('items.*, suppliers.*, inventory.*, item_quantities.*')` would do.
Why listing all and then state the AS + max?
It looks an overkill to me.

application/models/Item.php
```diff
@@ -60,13 +60,61 @@ public function get_found_rows($search, $filters)
...
+
+        $this->db->select('max(inventory.trans_id) as trans_id');
+        $this->db->select('max(inventory.trans_items) as trans_items');
+        $this->db->select('max(inventory.trans_user) as trans_user');
+        $this->db->select('max(inventory.trans_date) as trans_date');
+        $this->db->select('max(inventory.trans_comment) as trans_comment');
+        $this->db->select('max(inventory.trans_location) as trans_location');
+        $this->db->select('max(inventory.trans_inventory) as trans_inventory');
+
+        $this->db->select('max(item_quantities.item_id) as qty_item_id');
```
daN4cat: This is an optional join, so this could throw errors.

application/models/Receiving.php
jekkos: Group by will now require to use aggregates for every field in the select.

application/models/Item.php
jekkos: Same thing here.. If we want to use these fields in select we need to use aggregate function

application/models/Receiving.php
daN4cat: Aggregate means `AS` ?

daN4cat: And what about this `max(` or `MAX(` ?

jekkos:
Thanks for this pull request. Will we stick with full group by sql mode in this case? I think we could also disable this sql mode and stick with how the code is now?

http://stackoverflow.com/questions/23921117/disable-only-full-group-by

jekkos:
We could just add this statement to database.sql

`SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));`

jekkos:
On the other hand if we leave it on we're SQL compliant.

daN4cat:
The `SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));` could go in 3.0.1 and do for the MySQL5.7 support.
For 3.1 we can be SQL compliant do the proper changes.

jekkos:
Ok sounds fine, thanks.

odiea:
Works great. No issue now with MySQL 5.7.

jekkos:
Ok I hoped that it would be persistent for the db configuration itself. Another way to set it is in the database config file itself. This might not be a solution for all in case of shared hosting one does not often have access to db configuration parameters

daN4cat:
@jekkos you mean that adding in the database scripts is not enough/correct?

odiea:
That would be a good idea while this transition is going on. People could continue or even use 5.7 without issues. Once it is resolved that line could be removed.

application/models/Item.php
daN4cat: OK, I read better about this point as I'm not that expert in SQL stuff and got the point.
No GROUP_BY if there is no aggregate function like SUM, MAX, MIN and similar.

daN4cat: The issue here is to understand first why a GROUP_BY is added if there is no aggregate function.
I'll check that part of the code better to understand.

daN4cat: In this case we need to solve this problem:

```
		// avoid duplicated entries with same name because of inventory reporting multiple changes on the same item in the same date range
		$this->db->group_by('items.item_id');
```

differently.

application/models/Receiving.php
daN4cat: In Receiving the GROUP BY is probably a copy & paste mistake.
I need to check and remove it, so no need for the MAX(

application/models/Item.php
jekkos: Think we could try to use DISTINCT, however I'm not sure if it allows one to check for uniqueness of one column only in the select.

application/models/Item.php
```diff
@@ -60,13 +60,61 @@ public function get_found_rows($search, $filters)
...
+        $this->db->select('max(items.name) as name');
+        $this->db->select('max(items.category) as category');
+        $this->db->select('max(items.supplier_id) as supplier_id');
+        $this->db->select('max(items.item_number) as item_number');
+        $this->db->select('max(items.description) as description');
+        $this->db->select('max(items.cost_price) as cost_price');
+        $this->db->select('max(items.unit_price) as unit_price');
+        $this->db->select('max(items.reorder_level) as reorder_level');
+        $this->db->select('max(items.receiving_quantity) as receiving_quantity');
+        $this->db->select('items.item_id as item_id');
```
daN4cat: If the group_by is by items.item_id I don't understand how this could work without an aggregate.
It's more a question because I'm 100% sure I got my head around this issue.

daN4cat: To test this you need to go to Items and apply a date range to filter the Items selection according to when they where changed.

application/models/reports/Specific_discount.php
```diff
@@ -24,8 +24,10 @@ public function getDataColumns()
 	
 	public function getData(array $inputs)
 	{
-		$this->db->select('sale_id, sale_date, SUM(quantity_purchased) AS items_purchased, customer_name, SUM(subtotal) AS subtotal, SUM(total) AS total, SUM(tax) AS tax, SUM(cost) AS cost, SUM(profit) AS profit, payment_type, comment');
-		$this->db->from('sales_items_temp');
+		$this->db->select('sale_id, MAX(sale_date) AS sales_date, SUM(quantity_purchased) AS items_purchased,
```
daN4cat: This should be AS sale_date and not AS sales_date

application/models/reports/Specific_employee.php
```diff
@@ -24,8 +24,10 @@ public function getDataColumns()
 	
 	public function getData(array $inputs)
 	{
-		$this->db->select('sale_id, sale_date, SUM(quantity_purchased) AS items_purchased, customer_name, SUM(subtotal) AS subtotal, SUM(total) AS total, SUM(tax) AS tax, SUM(cost) AS cost, SUM(profit) AS profit, payment_type, comment');
-		$this->db->from('sales_items_temp');
+		$this->db->select('sale_id, MAX(sale_date) AS sales_date, SUM(quantity_purchased) AS items_purchased,
```
daN4cat: This should be AS sale_date and not AS sales_date

PR #1018, opened 2016-11-30
Support for MySQL 5.7 -  part 1 of ?

SteveIreland:
Change SQL statements involving the GROUP BY clause to include determinant columns as required by the MySQL 5.7 default.


application/models/Sale.php
```diff
@@ -51,23 +51,23 @@ public function get_info($sale_id)
...
+				MAX(customer_p.first_name) AS first_name,
+				MAX(customer_p.last_name) AS last_name,
+				MAX(customer_p.email) AS email,
+				MAX(customer_p.comments) AS comments,
 				' . "
 				IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals)) AS amount_due,
-				payments.sale_payment_amount AS amount_tendered,
-				(payments.sale_payment_amount - IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals))) AS change_due,
+				MAX(payments.sale_payment_amount) AS amount_tendered,
+				(SUM(payments.sale_payment_amount) - IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals))) AS change_due,
```
daN4cat: are you sure this SUM is correct? unless you found a bug and fixed with this I would have expected a MAX

application/models/Sale.php
```diff
@@ -150,23 +150,23 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 's
...
 				' . "
 				ROUND($sale_subtotal, $decimals) AS subtotal,
 				IFNULL(ROUND($sale_tax, $decimals), 0) AS tax,
 				IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals)) AS total,
 				ROUND($sale_cost, $decimals) AS cost,
 				ROUND($sale_total - IFNULL($sale_tax, 0) - $sale_cost, $decimals) AS profit,
 				IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals)) AS amount_due,
-				payments.sale_payment_amount AS amount_tendered,
-				(payments.sale_payment_amount - IFNULL(ROUND($sale_total, $decimals), ROUND($sale_subtotal, $decimals))) AS change_due,
+				SUM(payments.sale_payment_amount) AS amount_tendered,
```
daN4cat: here it's SUM and not MAX as per previous comment?

application/models/Sale.php
```diff
@@ -763,34 +763,34 @@ public function create_temp_table(array $inputs)
...
-					sales_items.serialnumber,
-					sales_items.item_location,
-					sales_items.description,
-					payments.payment_type,
-					payments.sale_payment_amount,
+					MAX(sales_items.serialnumber) AS serialnumber,
+					MAX(sales_items.item_location) AS item_location,
+					MAX(sales_items.description) AS description,
+					MAX(payments.payment_type) AS payment_type,
+					MAX(payments.sale_payment_amount) AS sale_payment_amount,
```
daN4cat: and here it's MAX on both cases. maybe I'm missing something. this needs some testing for sure

application/models/reports/Detailed_receivings.php
```diff
@@ -48,15 +48,18 @@ public function getDataByReceivingId($receiving_id)
...
-		$this->db->from('receivings_items_temp');
-		$this->db->join('people AS employee', 'receivings_items_temp.employee_id = employee.person_id');
-		$this->db->join('suppliers AS supplier', 'receivings_items_temp.supplier_id = supplier.person_id', 'left');
+        return $this->db->get()->row_array();
+    }
+
+    public function getData(array $inputs)
+    {
+        $this->db->select('receiving_id, MAX(receiving_date) AS receiving_date, SUM(quantity_purchased) AS items_purchased, 
+		MAX(CONCAT(employee.first_name," ",employee.last_name)) AS employee_name,
```
daN4cat: it's better to break this on multiple lines to easy readability

application/models/reports/Detailed_receivings.php
```diff
@@ -48,15 +48,18 @@ public function getDataByReceivingId($receiving_id)
...
+    }
+
+    public function getData(array $inputs)
+    {
+        $this->db->select('receiving_id, MAX(receiving_date) AS receiving_date, SUM(quantity_purchased) AS items_purchased, 
+		MAX(CONCAT(employee.first_name," ",employee.last_name)) AS employee_name, 
+		MAX(supplier.company_name) AS supplier_name, SUM(total) AS total, SUM(profit) AS profit, MAX(payment_type) AS payment_type, 
+		MAX(comment) AS comment, MAX(reference) AS referemce');
+        $this->db->from('receivings_items_temp AS receivings_items_temp');
+        $this->db->join('people AS employee', 'receivings_items_temp.employee_id = employee.person_id');
```
daN4cat: some indentation issues

application/models/Item.php
```diff
@@ -60,15 +60,60 @@ public function get_found_rows($search, $filters)
 	*/
 	public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'items.name', $order = 'asc')
 	{
-		$this->db->from('items');
-		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
-		$this->db->join('inventory', 'inventory.trans_items = items.item_id');
+        $this->db->select('max(items.name) as name');
```
daN4cat: Indentation issue

application/models/Sale.php
SteveIreland: Thanks for reviewing this.  To answer the first part of your question I was a little squeamish about changing to SUM from MAX, but since the other parts of line 68 were SUM based I figured the better guess was to use SUM.    For the second part of your question that was a mistake on my part.  It should have been SUM based on the logic I used for the first part.

application/models/reports/Detailed_receivings.php
SteveIreland: I'm happy to do that - some were broken down and some weren't and I think I tried to maintain the coding pattern that was already there.  I'm not saying I maintained that discipline, but I recall deciding that is what I probably should do.

application/models/reports/Detailed_receivings.php
SteveIreland: Interesting ... this isn't what I see in my code.  I'm new to PhpStorm and thought it was handling that for me so I have a bit more to learn here.  I shall fix that. Thanks again for pointing that out.

SteveIreland:
I know the stock report options without any underlying report is still showing on this branch, but I've run out of time to resolve it.  All of the other reports seem to be working now, so even though it needs a bit of tweaking I think this is something that's pretty close to being acceptable.

daN4cat:
@SteveIreland any reason behind the conflict? Have you rebased with the latest code from master?

database/database.sql
```diff
@@ -197,6 +197,8 @@ CREATE TABLE `ospos_items` (
   `item_id` int(10) NOT NULL AUTO_INCREMENT,
   `pic_id` int(10) DEFAULT NULL,
   `allow_alt_description` tinyint(1) NOT NULL,
+  `item_type` TINYINT(2) NOT NULL DEFAULT 0,
```
daN4cat: The correct thing to do here would be to change tables.sql and run grunt to get database.sql and migrate_phppos_dist.sql generate automatically containing the changes from tables.sql.

jekkos:
Hi Steve, hope you're not giving up on this one, if I can be of any help with it let me know

Issue #1026, opened 2016-12-01
pic_id and file names which actually makes sense?


i92guboj:
Hi, this is neither a support request nor a feature request nor a bug report. I tried to locate a mailing list, which I think would be best for this kind of discussion but didn't find any. So, if this isn't the right place for this kind of topic, I apologize and kindly ask you to redirect me to whatever channel I should be using instead.

Right to the topic...

I have some concerns on how images are imported and stored. Right now, images can't be related to products in any meaningful way, which means you are bound to introduce them one by one. It seems ospos just puts the image in uploads/item_pics/, renames it as <number>.jpg, and puts <number> in ospos_items.pic_id

So, a nonsensical number gets stored in a field, and is arbitrarily used as a filename. I haven't looked at that part of the sources yet, I'll come to that later, tomorrow if I have the time. I've only been poking around with mysql-workbench...

Let's imagine, for a second, we changed pic_id to varchar. Let's imagine I have a lot of images for my products stored in such a typical way as <PRODUCT_REFERENCE>.png so that I can identify them quickly, as people normally do. And then, let's imagine I import my products using a CSV file, which, actually, I do.

All I'd need to add images for a thousand products would be to upload them into uploads/item_pics, and then run this SQL sentence:

`update ospos_items set pic_id=<field_containing_reference> `

It would be just too nice if I could also get a pic_id column in the default CSV template, that would even spare me the step of launching workbench.

Is this something that people would like or need or is it something I should be doing and keeping for myself? What would be the OSPOS-way to do it? 

And, again, forgive me if this isn't the right place for this. I really have no idea where else to ask.

Thanks!

i92guboj:
I tried ALTERing that column, and in a sweep I was able to add pics to a good lot of products. It needs testing, it might break in some other place, but for now I can see the images for all my products in the product list.

daN4cat:
Feel free to contribute for something more practical, this part can be changed and your help is welcomed.

daN4cat:
I assume you will contribute for 3.1.

jekkos:
I don't use this feature myself so I don't really care that much about it. If you have ideas to improve then feel free to propose some code changes in a PR to make it better. We will discuss and review before merging it upstream. Thanks.

i92guboj:
Ok. Just wanted to know if this had any chance of getting upstream.

I will try to do something. I will have to think a bit about how to implement this, it would be a nice improvement albeit simple in terms of coding. Suggestions are welcome.

[...]

PR #1041, opened 2016-12-19
Pic filename

i92guboj:
This would convert pic_id to pic_filename, and add the necessary code to handle the fact that this field, from now on, will hold the full name (.ext included). Where necessary, conditional branching to detect legacy file names has been added, and a couple of silent update points have also been introduced, so, the existing installs should be automatically fixed.

Of course, this needs testing, and you should backup if you really want to test this on a production database.

It shouldn't do much harm, but still...

So, to sum up:

- field pic_id was renamed to pic_filename
- data type was changed from int to varchar(255)
- database/3.0.2_to_3.1.sql script was added to ALTER the items table
- import_items.csv was modified to add an example of the newly available field
- php was modified to be in agreement with the changes above, and also to manage the migration of already existing elements in the altered field


i92guboj:
Hi, as per the conversation in #1026 I am proposing this patch.

Also, please, note that I am a rookie to all this nodejs stuff. Bower touched several files, and git has commited that, I just let it do its work. While I know how to work with git, I have no idea what bower is so, please, forgive me if something useless went through the commit process.

application/helpers/table_helper.php
```diff
@@ -268,13 +268,21 @@ function get_item_data_row($item, $controller)
...
+			$images = glob('./uploads/item_pics/' . $item->pic_filename . '.*');
+		}
+		else {
+			// preferred
+			$images = glob('./uploads/item_pics/' . $item->pic_filename);
+		}
 		if (sizeof($images) > 0)
 		{
-			$image .= '<a class="rollover" href="'. base_url($images[0]) .'"><img src="'.site_url('items/pic_thumb/'.$item->pic_id).'"></a>';
+			$image .= '<a class="rollover" href="'. base_url($images[0]) .'"><img src="'.site_url('items/pic_thumb/' . pathinfo($item->pic_filename, PATHINFO_BASENAME)) . '"></a>';
```
jekkos: In fact this function call will always fallback to the filename without extension, if I read this right? So in this case even if the database column has the file extension, we will always be using the legacy glob functionality inside pic_thumb

application/helpers/table_helper.php
```diff
@@ -268,13 +268,21 @@ function get_item_data_row($item, $controller)
...
-		$images = glob('./uploads/item_pics/' . $item->pic_id . '.*');
+		$ext = pathinfo($item->pic_filename, PATHINFO_EXTENSION);
+		if($ext == '')
+		{
+			// legacy
+			$images = glob('./uploads/item_pics/' . $item->pic_filename . '.*');
+		}
+		else {
+			// preferred
+			$images = glob('./uploads/item_pics/' . $item->pic_filename);
```
i92guboj: As said, I am no php master. But, as I see it, glob() still serves as a way not to reference non-existent files. If the pattern doesn't match a file, it will return an empty array, I am not sure if for php trying to open a non-existant file can have any negative implication.

If you think glob() is superfluous here, I have of course remove it and assume the file exists?

application/helpers/table_helper.php
i92guboj: No, you have a point here, but I think it's actually the other way around. AFAIK, PATHINFO_BASENAME is "filename.ext", which is different from PATHINFO_FILENAME, which would hold just "filename" (no .ext). Just like the *nix basename command line tool.

So, pic_thumb() is always called with full filename, and probably the code inside pic_thumb() to handle the "no-extension case" is redundant. So, probably, just having it in one place or the other should be enough. 

I have to double check this, right now I am busy. I think I'll get to that this evening.
ank you for taking the time to review this, it's much appreciated.

application/helpers/table_helper.php
i92guboj: Ok, I've been looking into this. The code in pic_thumb() is useless, as you noticed.

The code in the table helper, needs one modification. Using pathinfo on $item->pic_filename to construct the url is incorrect, since it can be with or without extension. You need to use $images[0], and then PATHINFO_BASENAME takes meaning. Maybe that's what drove you to confusion.

i92guboj:
The last commit simplifies the code a bit and corrects a little bug. pic_thumb() now receives "filename.ext" (the whole basename) as an argument. 

It's unclear to me whether using glob() when the file name is "full" is really necessary or has any advantage (regarding security or whatever) over using the string directly and hoping that the file is there (if it isn't then the server should issue 404, shouldn't it?).

i92guboj:
For example, in pic_thumb() we have
`		$images = glob('./uploads/item_pics/' . $pic_filename);`
and below we have
`		if(sizeof($images) > 0)`
so this is a way (smart or not, I have no idea) to control whether the file exists. 

So, it glob() here has a purpose. Whether that purpose is legit or not, that's open for discussion. I guess there are better ways to infer if a given file exist, that are more legible. If you think it's worth the effort, I can certainly take care of that.

daN4cat:
Thank you for the contribution, I will try to review the code past Christmas together with other contributions.

daN4cat:
@i92guboj could you please rebase this PR with the latest from master? I also see that @jekkos requested changes, have you addressed those?

Thanks for the work by the way.

i92guboj:
Hi!

The requested changes are adressed as far as I can remember. But, as a rebase is due, I'll review everything ASAP. Can't promise anything but I should be able to do so in the next couple of days if nothing gets in the middle.

Thanks!

daN4cat:
@i92guboj any update on the matter?

i92guboj:
Hi and sorry for the delay. This should be fine now.

daN4cat:
@i92guboj sorry one more rebase due to the last commit to master. Thanks

i92guboj:
That should do it :)

hrshadhin:
Hi, I'm using OSPOS latest version and have a issue with item thumb image.
`/items/pic_thumb/Recipe-Chicken.png` 
above url not map on nginx. its show 404 error.

PR #1101, opened 2017-01-31
Fix merge issues related to line_sequence for receipts and invoices

SteveIreland:
This resolves the issue with the line-sequence.

### DIFF
--- a/application/libraries/Sale_lib.php
+++ b/application/libraries/Sale_lib.php
@@ -3,7 +3,6 @@
 class Sale_lib
 {
 	private $CI;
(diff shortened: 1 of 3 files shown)


odiea:
@SteveIreland  would you also look into why there are no more warnings showing for low inventory or out of stock items that use to show in Register.

SteveIreland:
Yes, I'll take a look at that now.  I'm leaning toward changing the validation to only take place for stocked items.  That makes sense now that there is a distinction .... right?

odiea:
Yes. You are correct and thanks for all the time spent on this contribution. Hope you enjoyed it.

PR #1102, opened 2017-01-31
Fix out of stock warning

SteveIreland:
I believe this will resolve the missing warning message.  It is also a change to sale_lib so it should be merged into or after the previous fix.


odiea:
If you come up with extra time. Is there a way to show the same warnings for 0 or at reorder level for  items in an item kit. Currently it does not display any warnings when an Item kit is added.

PR #1103, opened 2017-01-31
Restrict inventory reports to stock items

SteveIreland:
With the introduction of the new stock type field (stocked or not) it makes sense that the inventory reports now filter out non-stock items.  This pull request will change the inventory reports (Low Inventory and Inventory Summary) to include only stocked items.


odiea:
In inventory low did you mean to remove the following line. Doesn't it compare your current inventory to your reorder level?
$this->db->where('item_quantities.quantity <= items.reorder_level');

PR #1107, opened 2017-01-31
Low inventory check for kits

SteveIreland:
This will check each kit item for a low inventory condition.  If one triggers the condition then the warning message is generated.  Subsequent items are not checked for low inventory.


odiea:
Works great. Thanks

PR #1122, opened 2017-02-05
Preliminary fix for dot corruption in pic_filename

i92guboj:
Please, review it. It's only preliminary, I think we can simplify a bit in a couple of places, but first I'd like to see if this, indeed, fixes the issue at hand.

### DIFF
--- a/application/controllers/Items.php
+++ b/application/controllers/Items.php
@@ -75,12 +75,12 @@ class Items extends Secure_Controller
 	
 	public function pic_thumb($pic_filename)
 	{
-	    if($pic_filename=='')
-	    {
-	        // nothing to do here
-	        return;
-	    }
-	    
+		if($pic_filename=='')
+		{
+			// nothing to do here
+			return;
+		}
+
 		$this->load->helper('file');
 		$this->load->library('image_lib');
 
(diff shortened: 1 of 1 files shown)


application/controllers/Items.php
```diff
@@ -64,14 +64,23 @@ public function search()
...
+				$this->_update_pic_filename($item);
+			}
 		}
 
 		echo json_encode(array('total' => $total_rows, 'rows' => $data_rows));
 	}
 	
 	public function pic_thumb($pic_filename)
 	{
+	    if($pic_filename=='')
```
daN4cat: can you please fix this and use tabs

application/controllers/Items.php
```diff
@@ -216,9 +225,9 @@ public function view($item_id = -1)
 			// redundant.
 			if(sizeof($images) > 0)
```
daN4cat: this doesn't seem required anymore, please remove the whole statement

application/controllers/Items.php
```diff
@@ -809,6 +818,8 @@ public function do_excel_import()
 	 */
 	private function _update_pic_filename($item)
 	{
+        $filename = pathinfo($item->pic_filename, PATHINFO_FILENAME);
```
daN4cat: tabs

i92guboj:
That second commit should fix spacing (I corrected a few more spacing issues while I was at it). And removed some code. A third commit will remove one of the checks, that, at this stage, I think it's unnecessary.

I'll triple test everything, please, do your checks as well before merging.

PR #1138, opened 2017-02-12
This will add support for preparing quotes and generating invoices

SteveIreland:
This seems to be working pretty well.  I'm going to continue testing it, but I thought I'd throw it out here just in case anyone who has an interest in the feature has some time to look at it and tell me if I overlooked something.  I know I'm missing a support for the original "$" tokens but those are quick to add and will  catch those up on Sunday.


application/controllers/Sales.php
```diff
@@ -11,40 +18,36 @@ public function __construct()
...
 	}
-	
+
 	public function manage()
 	{
 		$person_id = $this->session->userdata('person_id');
 
-		if(!$this->Employee->has_grant('reports_sales', $person_id))
-		{
+		if (!$this->Employee->has_grant('reports_sales', $person_id)) {
```
daN4cat: please revert all these types of changes, the code style is what it is and we need to stick to it for consistency reasons and to avoid causing merges difficult for no reasons.

application/controllers/Sales.php
```diff
@@ -11,40 +18,36 @@ public function __construct()
...
 			redirect('no_access/sales/reports_sales');
-		}
-		else
-		{
+		} else {
```
daN4cat: same here, this is not required I'm afraid.

application/controllers/Sales.php
```diff
@@ -56,18 +59,18 @@ public function get_row($row_id)
 	public function search()
 	{
 		$search = $this->input->get('search');
-		$limit  = $this->input->get('limit');
+		$limit = $this->input->get('limit');
 		$offset = $this->input->get('offset');
-		$sort   = $this->input->get('sort');
-		$order  = $this->input->get('order');
+		$sort = $this->input->get('sort');
```
daN4cat: no point in changing spacing...

application/controllers/Sales.php
```diff
@@ -1,6 +1,13 @@
-<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
 
 require_once("Secure_Controller.php");
+require_once(APPPATH . 'libraries/tokens/Token.php');
```
daN4cat: not sure why we need this, library should encapsulate with a facade if possible otherwise you explode many dependencies across the code

application/models/Item_kit.php
```diff
@@ -77,7 +77,8 @@ public function get_info($item_kit_id)
 		custom8, 
 		custom9, 
 		custom10,
-		item_type');
+		item_type,
+		stock_type');
```
daN4cat: is this a fix for something else than this new feature?
if yes it should be submitted independently

application/models/Appconfig.php
```diff
@@ -74,5 +74,19 @@ public function delete_all()
 	{
 		return $this->db->empty_table('app_config'); 
 	}
+
+	public function get_next_invoice_sequence()
+	{
+		$last_used = $this->get('last_used_invoice_number') + 1;
+		$this->save('last_used_invoice_number', $last_used);
```
daN4cat: it's a bit unusual to see a get function do a save action.
maybe instead of get should be some other verb/action?

application/controllers/Sales.php
```diff
@@ -665,14 +752,32 @@ private function _load_sale_data($sale_id)
...
+		$data['cart'] = $this->sale_lib->get_cart();
+		if ($this->config->item('invoice_enable') == '0')
+		{
+			$data['modes'] = array(
+				'sale' => $this->lang->line('sales_sale'),
+				'return' => $this->lang->line('sales_return'));
+		} else {
+			$data['modes'] = array(
+				'sale' => $this->lang->line('sales_sale'),
+				'sale_invoice' => $this->lang->line('sales_sale_by_invoice'),
```
daN4cat: is the intention here to have a pro forma or a final invoice?

daN4cat:
Thank you for the contribution, I'm sure this will be useful to many.
In the past people asked support for pro forma invoicing. I assume you want to solve that problem here too.
I can link those issues later.

Also remember that in case somebody disables the invoice support from OSPOS nothing of this should be visible or affect the standard functionality.
So don't forget to check the proper flag.

application/libraries/Sale_lib.php
```diff
@@ -69,19 +83,67 @@ public function set_invoice_number($invoice_number, $keep_custom = FALSE)
...
 	}
-	
-	public function is_invoice_number_enabled() 
+
+	public function clear_quote_number()
+	{
+		$this->CI->session->unset_userdata('sales_quote_number');
+	}
+
+//	public function is_invoice_number_enabled()
```
daN4cat: what's the plan with this function?

application/models/Item_kit.php
```diff
@@ -77,7 +77,8 @@ public function get_info($item_kit_id)
...
```
daN4cat: is this a fix for something else than this new feature?
if yes it should be submitted independently

application/controllers/Sales.php
```diff
@@ -405,58 +370,142 @@ public function complete()
...
+			$token_customer = new Token_customer();
+			$token_invoice_count = new Token_invoice_count();
+			$token_quote_sequence = new Token_quote_sequence();
+			$token_suspended_invoice_count = new Token_suspended_invoice_count();
+			$token_year_invoice_count = new Token_year_invoice_count();
+			$supported_tokens = array($token_invoice_sequence, $token_customer,
+				$token_invoice_count, $token_quote_sequence, $token_suspended_invoice_count,
+				$token_year_invoice_count);
+
+			$this->token_lib->register($supported_tokens);
```
daN4cat: This looks like a dependency injection story, but the question I have here is whether is needed or could be encapsulated in the library itself.
The dependency injection is useful if you want to have a control on the passed object and aggregate parts from a "composer". Not sure this is the case here....

application/controllers/Sales.php
```diff
@@ -405,58 +370,142 @@ public function complete()
...
+
+					$this->suspend_quote($quote_number);
+					$this->load->view('sales/quote', $data);
+					$this->sale_lib->clear_mode();
+					$this->sale_lib->clear_all();
+				}
+			}
 		}
-		else 
+		else
```
daN4cat: wonder if all the above quote/invoice business logic would be better move to sale_lib or maybe an invoice_lib just to keep sales controller code cleaner.

Thoughts?

database/database.sql
```diff
@@ -16,6 +16,7 @@ CREATE TABLE `ospos_app_config` (
 INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
 ('address', '123 Nowhere street'),
 ('company', 'Open Source Point of Sale'),
+('default_register_mode', 'sale'),
```
daN4cat: It would probably make sense to get used to change tables.sql only and run grunt script.

You need to install node.js and etc. I believe I have updated the wiki explaining how to.

odiea:
Current version needs fine tuning.
Quotes suspends a sale just fine but it still shows up in Takings as a Sale.
Invoicing numbering not working at all.
As you can see $co is the only thing showing as far as a number for Invoicing. And yes I set config to 1.
I also believe that Invoicing and Quotes should be related to a Customer and should only show after a Customer has been selected. Personal opinion.

odiea:
I know this is a personal preference. This is how I have my current version adapted.
2017-02-12 06_11_27-wedidit _ powered by ospos 3 1 0

odiea:
I also ran the following upgrade script on an old database. 3.0.2_to_3.1.0.sql. The following config items finally showed but no quote number is showing and same issue with Invoice number. Quote number shows ok on a new db install.
2017-02-12 06_54_41-open source point of sale _ powered by ospos 3 1 0

application/libraries/Token_lib.php
```diff
@@ -0,0 +1,131 @@
...
+	private $CI;
+
+	private $_registered_tokens;
+
+
+	public function __construct()
+	{
+		$this->CI =& get_instance();
+
+		foreach (glob("tokens/*.php") as $filename)
```
jekkos: We should make sure this directory is not writable in this case, otherwise any script added there would be loaded which could lead to a server takeover

application/libraries/tokens/Token.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/**
+ * Created by PhpStorm.
+ * User: steve
+ * Date: 2/4/2017
+ * Time: 4:01 PM
+ */
```
jekkos: Might probably better to add this info in the COPYING file, just to stay consistent with other code.

application/libraries/tokens/Token_quote_sequence.php
```diff
@@ -0,0 +1,30 @@
...
+	public function __construct()
+	{
+		parent::__construct();
+		$this->CI =& get_instance();
+	}
+
+	public static function get_token_code()
+	{
+		return Token_quote_sequence::$token_code;
+	}
```
jekkos: IMHO this logic is actually common to all Token implementations. If $token_code was an instance field then one could use the [`tell, don't ask`](https://martinfowler.com/bliki/TellDontAsk.html) clean code pattern that tells each token to do it's magic on a given string. In that case you'd need to have the token implementations instantiated in a list at first

application/controllers/Sales.php
SteveIreland: Sorry about this.  Yes I'll search and destroy them.  I must have accidentally activated a code reformat before I had set the conventions properly.  What a pain.

application/models/Appconfig.php
jekkos: +1 save should at least be included in the method name

SteveIreland: ok

application/models/Appconfig.php
```diff
@@ -74,5 +74,19 @@ public function delete_all()
...
+		return $last_used;
+	}
+
+	public function get_next_quote_sequence()
+	{
+		$last_used = $this->get('last_used_quote_number') + 1;
+		$this->save('last_used_quote_number', $last_used);
```
SteveIreland: ok

application/views/configs/invoice_config.php
```diff
@@ -66,6 +84,34 @@
 				</div>
 			</div>
 
+            <div class="form-group form-group-sm">
+				<?php echo form_label($this->lang->line('config_last_used_invoice_number'), 'last_used_invoice_number', array('class' => 'control-label col-xs-2')); ?>
+                <div class='col-xs-2'>
+					<?php echo form_input(array(
+						'max' => '999999999',
```
jekkos: If there's no real max than we could just omit this value. Unless you want to guard against overflowing the database's datatype's width. But in that case we should do the check in the backend.

application/views/configs/invoice_config.php
```diff
@@ -66,6 +84,34 @@
...
+						'class' => 'form-control input-sm required',
+						'value'=>$this->config->item('last_used_invoice_number'))); ?>
+                </div>
+            </div>
+
+            <div class="form-group form-group-sm">
+				<?php echo form_label($this->lang->line('config_last_used_quote_number'), 'last_used_quote_number', array('class' => 'control-label col-xs-2')); ?>
+                <div class='col-xs-2'>
+					<?php echo form_input(array(
+						'max' => '999999999',
```
jekkos: Same thing here

application/views/sales/quote.php
```diff
@@ -0,0 +1,177 @@
...
+        </tr>
+    </table>
+</div>
+
+<script type="text/javascript">
+    $(window).on("load", function()
+    {
+        // install firefox addon in order to use this plugin
+        if (window.jsPrintSetup)
+        {
```
jekkos: Not sure how it's done in other views but this piece of code is common to all printing functionality, maybe we should move it to a partial view somewhere

application/views/sales/register.php
```diff
@@ -437,6 +448,15 @@
 			<?php echo form_open($controller_name."/cancel", array('id'=>'buttons_form')); ?>
 				<div class="form-group" id="buttons_sale">
 					<div class='btn btn-sm btn-default pull-left' id='suspend_sale_button'><span class="glyphicon glyphicon-align-justify">&nbsp</span><?php echo $this->lang->line('sales_suspend_sale'); ?></div>
+					<?php
+					// Only show this part if the payment cover the total
+					if ($quote_or_invoice_mode)
+					{
+					?>
+                    <div class='btn btn-sm btn-success' id='finish_invoice_quote_button'><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $mode_label; ?></div>
```
jekkos: alignment doesn't seem to be quite alright here

application/controllers/Sales.php
SteveIreland: An invoice in the US is simply an agreement that delivery has been made or services have been rendered and that payment will eventually be made.  Followup is usually handled by an accounts receivable system that generates outstanding balance due statements.  There isn't a requirement for a "final invoice".  However, I believe that this should still print a final invoice and I'll test it today to confirm that it does.

My intention is that a quote would be something that can be generated and retained and the retrieved to convert to a sale.

If a pro forma is simply a quote that is not retained then I have provided that with the "Discard" button when the quote is generated.  After it's printed then the user simply clicks on Discard and the suspended quote should be deleted.

application/libraries/Sale_lib.php
SteveIreland: I'll try to clean up my development commented code a little better.

application/models/Item_kit.php
SteveIreland: Okay, I'll pull this out.  I stumbled across this error while I was testing while in development mode.  I do that now you know.  :-)

daN4cat:
Yes I sense that quote and pro forma are similar. Pro forma looks like an invoice and it's sometimes a means to show what you are about to receive after you paid a deposit or the full amount.  Once the payment is done and the delivery is with you a formal invoice is issued.

application/controllers/Sales.php
odiea: I will retest. But after discarding the quote it still showed up in Takings as a sale.

application/controllers/Sales.php
SteveIreland: The requirement here is that not all tokens should be supported for all text replacements.  The register method is intended to indicate, for a particular use of the token render method, which tokens are to be supported.  I think for now I'll just remove this because it is starting to seem like I over-engineered this a bit.  I'll leave it to the user to decide which tokens are appropriate.

database/database.sql
SteveIreland: I know ... I know .... you keep telling me I have to learn grunt ..... but it's sooo hard.

I'm going into surgery in about a week.  While I'm laid up I'll try to learn to grunt.

daN4cat: No worries, I don't always remember if I said it or not. I cannot always memorise comments between submissions here.

We'll give a run through grunt before merging the code.

jekkos: not that hard as you'd think, maybe just a matter of having it set up.. one thing you need to do just once and then it's just a matter of running the build now and then.

it's even `easier` when using docker but then you'll need to learn that first. Think I have something on it on the wiki. If we'd need to choose again then I'd go with gulp instead which is the better choice these days. Think we have an issue open for that.

daN4cat:
Yes there is an issue

SteveIreland:
I'm still working on it .... @odiea pointed out an issue that resulted in a bit of a rewrite which I just finished and I am just now returning to the list of reported issues and suggestions.  I received a lot more feedback than I expected and it's all good.  :-)  Thanks everyone.

application/controllers/Sales.php
SteveIreland: The next update to the pull request will have the issue with quotes appearing in takings fixed.

application/libraries/tokens/Token.php
SteveIreland: I removed them.

application/libraries/Token_lib.php
SteveIreland: I removed this.  I might need some technical advice on how to include the class files since there isn't a previous example.  The approach used was what I found on google.  The glob actually didn't work anyway.  I just forgot to remove it.

SteveIreland:
@odiea The approach I used for not allowing a quote or invoice to be generated without a customer was to hide the Invoice and Quote button.  I wanted an employee to be able to stay in a particular register mode but you are correct that it isn't really valid without a selected customer.

application/views/configs/invoice_config.php
SteveIreland: Okay.  It was intended to put a limit on what I expected to support, but the way I ended up coding it  didn't introduce a limit.

application/views/configs/invoice_config.php
SteveIreland: Got it.

application/views/sales/register.php
SteveIreland: You are correct.

application/libraries/tokens/Token_quote_sequence.php
SteveIreland: The only book I purchased on the topic of patterns was Java AntiPatterns. That was my best at coming up with a solution to acquire the correct object using a string identifier, but seriously if you can rework it to be more elegant I won't be offended at all.  I couldn't figure out how to apply the example you linked to to in order to make it cleaner,

application/libraries/Sale_lib.php
```diff
@@ -28,6 +37,67 @@ public function get_cart()
...
+
+		$filtered_cart = array();
+
+		foreach($cart as $k=>$v) {
+			if($v['print_option'] == '0')
+			{
+				$filtered_cart[] = $v;
+			}
+		}
+
```
jekkos: Think polymorphism could have been used here yielding less duplicated code..

application/libraries/tokens/Token_customer.php
```diff
@@ -0,0 +1,36 @@
...
+		if($customer_id != -1)
+		{
+			$customer_info = $this->Customer->get_info($customer_id);
+			if($customer_info != '')
+			{
+				return trim($customer_info->first_name . ' ' . $customer_info->last_name);
+			}
+		}
+
+		return 'Customer Unknown';
```
jekkos: Shouldn't this string be translated?

application/libraries/tokens/Token_quote_sequence.php
jekkos: Well actually I would leave the getValue() as only method in the subclasses and add a replace method in a base class that calls the implementations to replace the variable name with the returned value

application/libraries/Sale_lib.php
```diff
@@ -37,6 +37,67 @@ public function get_cart()
...
+		if($this->CI->config->item('line_sequence') == '0')
+		{
+			$sort = array();
+			foreach($filtered_cart as $k=>$v) {
+				$sort['line'][$k] = $v['line'];
+			}
+			array_multisort($sort['line'], SORT_ASC, $filtered_cart);
+		}
+		// Group by Stock Type (nonstock first - type 1, stock next - type 0)
+		elseif($this->CI->config->item('line_sequence') == '1')
```
jekkos: bit of code smell here.. sorting the cart could be done using a OO pattern called strategy

jekkos:
Wow about +1200 extra lines of code quite a nice job you did there.

application/libraries/tokens/Token_customer.php
SteveIreland: I think I might as well just return an empty string.  It should never happen anyway.

application/libraries/Sale_lib.php
SteveIreland: The finest wines can have a touch of bloody tar gustatory transducers so I am sure that is what you are smelling.

SteveIreland: And yes I know the sense of taste is not the same as the sense of smell, but good wine was what was on my mind,

SteveIreland:
@jekkos  Are you waiting on me to address some issues before merging?  If so can you itemize them so that I can make a decision? If it's only the code refactoring then I will probably need to refocus on the payment and customer based taxing jurisdiction sales tax issue so that I can get my client up and running.

Although code refactoring per O-O patterns would actually be a fun and interesting educational exercise for me I couldn't do it within my current window of opportunity - but I would be happy to take it on later - after they are up and running on OSPOS.

odiea:
I just downloaded your current version. Invoicing still does not work. After creating an Invoice the screen goes blank. Also did you fix the Invoicing number issue?

SteveIreland:
Well, that would certainly be a reason.  I generated dozens of invoices last night with no problem with the invoice number generation, so I guess it's time to rebuild my test environment from scratch again.  Thanks for letting me know.

odiea:
This is the branch I have been downloading. "opensourcepos-tokens-quotes.zip" If this is not correct please let me know.

SteveIreland:
I rebuilt my test environment using a fresh download of the same branch you used and it again works as I expect so I have a mystery.  Did  you apply the database change that adds the quote number to the `sales_suspended` table?

odiea:
Yes. Quotes have been working fine from the beginning it is the Invoicing feature that is giving me the issues. Quotes auto number properly. If you look at my image you can see that the First invoice number listed is $CO. I will recreate a new database from the latest and see if the issues persist. I use the database.sql

odiea:
I just downloaded the tokens-quotes branch
Unzipped it. Changed database config. Created a new db using database.sql
I added an item went to register selected invoice after entering the customer. Paid the Sale. Clicked on invoice.
2017-02-14 16_32_40-open source point of sale _ powered by ospos 3 1 0
And this is the page I get. Receipts and quotes work just fine.
2017-02-14 16_32_55-mozilla firefox

odiea:
I will delete my submissions later tonight. I know they take up space.

SteveIreland:
Bingo!  I was able to replicate it.  It must be the token.  Checking it out now.

SteveIreland:
Submitted a fix for the broken tokens.

RamkrishnaMondal:
I downloaded and tested this feature...seems working as desired. But I can see the option of `Sales Receipt `under invoice tab in config. Is not it odd to have `Sales Receipt` selection under `invoice` tab? It should have only 2 options,` Invoice `and `Quotes`. Even if I select `Sales Receipt` here, `Create Invoice `checkbox is still showing in the sales register. 

screenshot at 2017-02-15 11 38 18


Thank you once again for the hard work.

jekkos:
Ok let's say that if this new addition is found to work correctly I'll merge the code. But as @daN4cat I think that we should think about restructuring the code a bit as some controllers are having too much responsabilities which reduces cohesion of the code. But this indeed is something we can tackle afterwards.

I don't know if I'll find the time to test myself tonight but I trust @odiea and @RamkrishnaMondal in verifying if everything works as intended. So once these last issues are fixed I think we're ready to go.

odiea:
It is creating Invoices now but the register is still showing the following.
2017-02-15 05_22_13-open source point of sale _ powered by ospos 3 1 0

SteveIreland:
@RamkrishnaMondal I'll test to confirm that I didn't overlook what should happen when the Invoicing Enabled box is not checked, but the option for Default Register Mode is only valid (and useful) if the Invoicing feature is enabled. 

In fact, after this is in production I was going to follow up with a proposed change to add an override by employee (if Invoicing is enabled) so that some employees default to register mode and others would default to Sales Receipt.  Right now the system level default is really only useful for testing the default receipt mode because I'm sure everyone that's using point of sale will default to sales receipt mode and make those who do service work change the mode to Sales by Invoice for invoicing purposes.

SteveIreland:
@odiea I forgot to mention that I changed the trigger point when the final invoice number is actually derived.  On the screen I'm just showing the template that will be used.  In my past experiences the invoice number is immutable and shouldn't be under the control of the user.

We can discuss this because maybe some users will want to override it.  I'll test this area more thoroughly and add a way for the user to override it.

Also, the last invoice number used only comes into play when you use the `{ISEQ}` token.  That's the token that allows the user to take control over the number used for the invoice.  What I mean by that is that I saw an issue where someone wanted to use OSPOS but had a requirement that they be able to start their invoice numbering at a particular point.  The `{ISEQ}` was designed for that - allowing the user to change the config value to start at a particular number.

odiea:
Ok Thanks for the update. My concern would be previous users that have always seen the number there.

odiea:
Another issue. If for some reason the config is changed in Invoices-remove the check mark this is what happens when you re enable it. Plus quotes do not show a number any longer even if the last used number or any number is put in there.
2017-02-15 07_45_55-open source point of sale _ powered by ospos 3 1 0
And here is what the Quote looks like.
2017-02-15 07_50_12-open source point of sale _ powered by ospos 3 1 0

RamkrishnaMondal:
Personally, I always use the invoice template and Sales Receipt under invoice is misleading, that's why i have proposed the change. Anyway I can live with that. But in case of Invoice, `complete `button is only generating Sales receipt. Can't you add Invoice button's action to Complete button? or at least hide the `complete` button when invoice is enabled. After 4-5 testing, I got that I have to click `Invoice` button to generate an Invoice, as we are habituated with clicking the `compete` button.

RamkrishnaMondal:
Whatever i put as last invoice number, Invoice number is always starting from '0'.

SteveIreland:
@RamkrishnaMondal Are you using the `{ISEQ}` token?  Because only `{ISEQ}` will use the last used invoice number.

RamkrishnaMondal:
I can see {QSEQ} at sales quote format, but what is {ISEQ} ? How to use that?

RamkrishnaMondal:
Got it.  QSEQ = Qoute sequence, ISEQ = Invoice Sequence. 
Yes, with ISEQ as Invoice format, custom invoice numbering works.

SteveIreland:
I'll go ahead and hide the Complete button when in Invoice mode so that it is less confusing.  If they need to print a sales receipt instead of an invoice they simply need to change the register mode.

SteveIreland:
This last commit hides the commit button in sales register mode and I also made sure that when the user is in Sales (by Receipt) Register Mode that if paid in full and they click on "Create Invoice" it will also generate the invoice and will optionally accept an overridden invoice number.

SteveIreland:
Just wanted to note that the descriptions for the supported replaceable tokens can be found at https://github.com/jekkos/opensourcepos/wiki/Sales#4-configuration

jekkos:
Ok great, guess we're getting close to finalization.

daN4cat:
@SteveIreland instead of hiding Complete why not charge the button description to be Invoice when in invoice mode? You just need to use a bit of js. There is a similar scenario as example, when you select giftcard as payment type the input text change from amount to be card number. Of course on commit you need to handle differently angry again with a bit of js.

daN4cat:
@jekkos we need to make sure all of this change switches off if invoicing is not enabled. I need to review the code as this is something standard shops in EU don't need.

SteveIreland:
@daN4cat For those shops where the same station might be used to print both sales receipts and invoices I felt the relocation of the "print" button was a better visual clue to help the user be aware of the register mode that is in place (primarily when a user walks up to the station that was being used by someone else).  I toggle the label between "Invoice" and "Quote", but I figured the same person is going to be doing quotes and invoices so they'll just need to be self aware.

Working on a couple of "developer mode only" bugs I stumbled across.

odiea:
So what I found with the quotes issue is if you disable Invoice. Submit and then exit out of config. Go back into config Invoice the Quotes format Q%m{QSEQ:6} is now gone.

SteveIreland:
Yes, after I fixed the developer mode only bugs I stumbled across that one too.  I've been working on it but I'm high on pain drugs after surgery this afternoon ... so development has been slow .... very slow,

But I'm about to push it up.  I hope you don't mind that I rewrote everything in jRuby - it was more sparkly.

daN4cat:
@SteveIreland take it easy and rest. No need to rush on this, we can all wait a week or more if required.

SteveIreland:
Just wanted to highlight that I did fix the "developer mode only" bugs and resolved the issue reported by @odiea.  It's ready to merge now.  :-)

odiea:
So far so good. Is there going to be a permission check for either of the following if the person does not have permission for Item Kits?
2017-02-17 12_06_11-open source point of sale _ powered by ospos 3 1 0

SteveIreland:
I'll go ahead now and condition the display of those fields.  The Stock Type is more in support of services than it is for Item Kits.  However, as it is right now its easy to argue that the Item Type is exclusively in support of Item Kits (because it is), but future development might actually add additional types of items that require unique processing rules.

daN4cat:
Yes please let's fix the item kit issues before moving to the next.

SteveIreland:
Okay, I've added some changes to the pull request.

SteveIreland:
@daN4cat @jekkos @odiea 

Can we get this merged into master soon?  I have the weekend available to address any issues.  I'm getting a little nervous that someone is going to merge in something that will require me to start over again.

Thanks.

jekkos:
I'm OK to merge this PR now. Once it's in master I guess more users will give feedback and so issues can be tackled. 

@daN4cat @odiea you guys OK with this? There's still @daN4cat's approval pending.

odiea:
I checked out his latest and it appears to be working properly.

daN4cat:
@jekkos merging this in master will cause duplicated history.

daN4cat:
@jekkos you probably need to rebase this branch before merging into master. Have a look

jekkos:
Ok I will

SteveIreland:
Soooo ... is this back in my lap?

SteveIreland:
I'm assuming so.  I didn't realize until now that I could "resolve" conflicts using github so that is what I did.  I hope it was correct.

odiea:
@SteveIreland  I just downloaded your latest and there appears to be issues with Sales. When I select The sales Icon it goes to a blank page. Have you added any new db updates?

odiea:
After I swapped out the current register with the previous Register it works.

odiea:
Around line 405. Should that second else be there? After I removed it now works.

SteveIreland:
@odiea I will check this morning to see what happened.  It's most likely that I mucked up the conflict resolution.  Yesterday @daN4cat introduced changes to the master branch on a couple of functions that I also had several changes to.  I'll fix the pull request and report back to  you on what the issue was

jekkos:
Let me know when you're done and then I'll see how to integrate in master properly

jekkos:
Found quite some duplicated history already in this feature branch

daN4cat:
@jekkos yep, as I told you earlier...

jekkos:
I've tried to rebase this functionality on top of what is in master, but then  I need to go through resolving every conflict since the first commit in this PR. Can be a bit error prone as I haven't followed the progress in that much detail but I'll have a look anyway

SteveIreland:
I've rebased and squashed the branch.  I'm trying to test it, but OSPOS is locking up on me and I'm trying to work through that.  I'll resubmit when it's done.  Please don't spend any more time on it.

rminformaticos:
im doing this changes, and when i click on sales button (register.php) it showing error 500.

rminformaticos:
im doing this enhancement, and on sales section (register.php) when i add a product, not showing add payment button, only complete sale button.

odiea:
It may be that you do not have a customer selected for the function you are trying to use.

rminformaticos:
i re download the master, and files was updated, and works

PR #1149, opened 2017-02-20
Tokens quotes

SteveIreland:
Grab it quick.  I don't know how long it will last and I pray that there are no new conflicts since the squash.


SteveIreland:
@jekkos @daN4cat @odiea 

Is it in yet?  :-)

odiea:
Won't know until tomorrow. It's sleepy time in Europe.

SteveIreland:
Well, I know I won't get any sleep until it is in!   Well ....maybe I will if I double up on my pain pills.  ;-)

odiea:
Ya. A couple shots and an extra pain pill.

application/controllers/Sales.php
```diff
@@ -401,58 +404,128 @@ public function complete()
...
+			else
+			{
+				$invoice_format = $this->config->item('sales_invoice_format');
+			}
+			$invoice_number = $this->token_lib->render($invoice_format);
 
-			$this->_reload($data);
+			$quote_number = null;
+
+			// TODO If duplicate invoice then determine the number of employees and repeat until until success or tried the number of employees (if QSEQ was used).
```
daN4cat: todo means to me code not ready...

daN4cat:
In sale controller there are many indentation or spacing changes that could be avoided. Again the code becomes inconsistent with the rest...

application/controllers/Sales.php
SteveIreland: It can mean that ... or it can suggest that there is a future internal improvement I'd like to make but would like to hear the opinions of others before submitting a pull request for it - which is the case here.

odiea:
Sales.php 806 could you add something to that line so all the drop down choices only show when a customer is selected.
if ($this->config->item('invoice_enable') == '0' or empty($customer_info))
I was having issues with the Invoice mode after returning back to the Register. It would print an Invoice instead of a receipt.
So in Sale_lib I added in function clear all  $this->clear_mode(); and now it appears to be working better.
I am sure you have a better way to do this. I just copy and paste to change the issues that bother me.

jekkos:
Thanks for the history cleanup already, think it's better like this.

SteveIreland:
@odiea Thanks for the feedback.  As soon as it's merged in I'll revisit that question again.  You mentioned it earlier so I took a look at it. What I found was that if I made the change you requested then the workflow for the user who's world is all about invoicing because more painful.  I'd prefer to address this when I'm able to add employee permissions to the workflow.  That way I can do exactly what you want for employee's whose primary mode is sales by receipt because it makes sense there.  What I did for you until then is prevent the user from clicking on Invoice (hopefully it's still there and didn't get lost in the merge).

jekkos:
Ok I'll merge once I find time to do support on issue tracking / review

daN4cat:
@jekkos any plans with this merge?

jekkos:
Will merge today.

jekkos:
Just trying out these new changes. When I set the sale by invoice mode I'm not seeing a complete button after adding a payment. Is this expected behavior?
schermafdruk van 2017-02-26 11-33-15

jekkos:
Complete button only shows when enabling the return mode

jekkos:
Issues seemed to be related to my local setup, things look fine. Let's wait and see if community finds any extra bugs

jekkos:
I will now update weblate to import the translations related to this part

jekkos:
Weblate should have sent out emails about those new translations, so every translator should be notified.

odiea:
I believe his intensions here were If a Customer is not selected and Invoice or Quote mode are enabled you cannot complete a transaction. Add a Customer name and it will work.

SteveIreland:
Thanks @jekkos.   I'll give it whirl too and will submit any corrections ASAP.

SteveIreland:
I never touched returns so I never tested it ... So with my luck the odds are pretty good that there was some collateral damage.  I will work through any issues I stumble across and submit a fix for those.

SteveIreland:
I am much relieved to report that I was able to process a return without it crashing.  So the bones have not been seriously damaged.

Issue #1153, opened 2017-02-20
Question: Why qty, price calculation is not updated in real-time.


nick9822:
Hi there,

Thanks for such a nice product. While evaluating I came across one issue or question you can say.
Why qty, price is not updated real-time? Please see following screen, Price is 500 and 20 qty so amount should be 10000 but instead its 5000. Amount is getting updated after pressing update icon.

![capture 55](https://cloud.githubusercontent.com/assets/10139291/23138866/c3cf20b0-f7cf-11e6-804a-d1c11e777e1e.png)

If its default behavior, can anyone point where I can automate it by JS or PHP, so every focusout can update the amount?

Thanks,
Nick

daN4cat:
Have type enter after you updated the field?

nick9822:
@daN4cat 
Oh thanks, it worked. So sorry for that.

But how can we make on focusout or TAB press, any idea? It would be great.

Thanks,
Nick

daN4cat:
I would need to check....

RamkrishnaMondal:
@daN4cat   I have seen commercial POS system build on PHP, updates values automatically, without pressing enter or clicking on update button, can't we modify OSPOS that way ?

jekkos:
Sure but this is not a commercial system but a community driven one. Technically it is certainly possible as long as someone has the motivation to pick up the issue and fix it

nick9822:
@jekkos 
Thank you for reply. You have done the great work creating this product. 👍 
I would love to do this feature, if you guide me somewhat.

daN4cat:
@nick9822 many thanks for the contribution, right approach to contribute. Somebody here will pick up and test, I'll see if I can find the time.

jekkos:
See latest commit, should work for both receiving and sales

nick9822:
Thank you for your efforts. Now we have the needed feature running.

PR #1154, opened 2017-02-21
Realtime line updation

nick9822:
Hi Team,
This is with reference to #1153 

Please advice whether is it a good way to achieve so I can apply same changes to Sales.

Thanks,
Nick

### DIFF
--- a/application/views/receivings/receiving.php
+++ b/application/views/receivings/receiving.php
@@ -142,7 +142,7 @@ if (isset($success))
 							if ($items_module_allowed && $mode !='requisition')
 							{
 							?>
-								<td><?php echo form_input(array('name'=>'price', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['price'])));?></td>
+								<td><?php echo form_input(array('name'=>'price', 'id'=>'cart_'.$line, 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['price'])));?></td>
 							<?php
 							}
 							else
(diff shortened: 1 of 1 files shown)


RamkrishnaMondal:
WOW! this is very simple yet very effective solution.
Thank you very much. 
Hope @jekkos & @daN4cat  will love this too.

application/views/receivings/receiving.php
```diff
@@ -516,6 +516,17 @@
 			}
 		}
 	}
+	$('[name="price"]').focusout( function(e) {
+		document.getElementById(this.id).submit();
```
jekkos: Why not use `$(this).submit()` instead of combining jQuery with regular JS?

jekkos: Also the selectors can be combined which would reduce the change to only three lines, as follows

`$('[name="price"],[name="quantity"]).focusout($(this).submit)`

jekkos:
Also you are duplicating id's among elements in the DOM which isn't allowed in any way. They should be unique at all times

application/views/receivings/receiving.php
nick9822: Nice review. I will check and update pull request.

nick9822:
@jekkos
Yes its not permitted. As classes and names were occupied, I thought to give it a try.

Anyways latest changes will not require ids, so I will remove those.

jekkos:
Thanks for making the PR, I've added this bit in the code myself without using the input field id's. Too bad it does not have you're name on it

nick9822:
@jekkos 
Its okay that I do not have name on it. I am happy that the new feature added to the pos. 🖒😊

OSPOS is good POS, lets build it more perfect. ☺

PR #1157, opened 2017-02-22
Fix on latest commit for (#1154)

nick9822:
@jekkos 
With reference to https://github.com/jekkos/opensourcepos/commit/e55b5a743a2b8de59f56a217721314c93f04ce22
Rows were not updating, So added some changes to make it work.

### DIFF
--- a/application/views/receivings/receiving.php
+++ b/application/views/receivings/receiving.php
@@ -131,7 +131,7 @@ if (isset($success))
 				{
 			?>
 					<?php echo form_open($controller_name."/edit_item/$line", array('class'=>'form-horizontal', 'id'=>'cart_'.$line)); ?>
-						<tr>
+						<tr id='<?php echo $line?>'>
 							<td><?php echo anchor($controller_name."/delete_item/$line", '<span class="glyphicon glyphicon-trash"></span>');?></td>
 							<td style="align:center;">
 								<?php echo $item['name']; ?><br /> <?php echo '[' . to_quantity_decimals($item['in_stock']) . ' in ' . $item['stock_name'] . ']'; ?>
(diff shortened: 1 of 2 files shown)


application/views/receivings/receiving.php
```diff
@@ -131,7 +131,7 @@
 				{
 			?>
 					<?php echo form_open($controller_name."/edit_item/$line", array('class'=>'form-horizontal', 'id'=>'cart_'.$line)); ?>
-						<tr>
+						<tr id='<?php echo $line?>'>
```
jekkos: I would think your intent was to fix this for the description and serialnumber fields that are in the table row below the row that contains price, quantity and discount. In that case I think you forgot to add the extra selectors?

So this PR doesn't really add any extra working functionality. Also I would rather avoid adding just plain numbers as id's in the DOM to fix this issue. Best solution would be to get rid of these tables wrapping the register lines and use plain responsive CSS, so that we can use a relative jQuery Dom traversal method to get ahold of the form element

jekkos: Or if you don't feel up to this CSS rework, then it might be nicer to use some kind of remainder division hack to get the correct cart index number in this case. Something along the lines of 

`var index = $(this).parents("tr").index(); var final_index = index - 1 - (index % 2)`
and then use final_index to fetch `$("#cart_" + final_index)`

`

jekkos:
Thanks for this PR, appreciate it but I think register needs a bit of a DOM rework to fix this in a proper way.

application/views/receivings/receiving.php
jekkos: It's hacky after all but still I think one should be able to just fetch the form elements coming from the inputs in this case

nick9822:
@jekko
Thanks for reviewing.

These changes were done, because latest commit was only able to update single row. And when we add more than one items (rows) it was not working due to wrong index.

I would love to do CSS rework, but it will take time.
For this instance I find this useful to put table row id as cart id. We can make table row id as cart1, cart2 and so on, so it will not be simple number. 😜

jekkos:
is fixed now in latest master

Issue #1166, opened 2017-02-26
Undefined property in Sale_lib.php


SteveIreland:
Version: Open Source Point Of Sale Version 3.1.0 - 4f5ad57 (after commit 032bb34)

The following error occurred when I attempted to unsuspend a quote.  Working on a fix for it now,

```
A PHP Error was encountered
Severity: Notice
Message: Undefined property: stdClass::$stock_type
Filename: libraries/Sale_lib.php
Line Number: 736
```

SteveIreland:
I tried to cherry pick a commit but it didn't work the way I wanted (which was I just wanted the commit which contains the change, but it seems to have picked up a few empty commits).

If you want me to redo this just let me know.  My feelings won't be hurt ... much.

odiea:
@SteveIreland  I believe @daN4cat already fixed this in the master.

SteveIreland:
No problem.  "Practice makes perfect."  I think the last thing I read was "try again" - that's why it was resubmitted.

daN4cat:
I part fixed, @SteveIreland had more to add and therefore his was the proper fix!
Shall we consider this closed now?

SteveIreland:
I think it can be closed now.

PR #1167, opened 2017-02-26
Fix undefined property error in Sale_lib.php

SteveIreland:
Fixes undefined property in Sale_lib.php #1166


daN4cat:
@SteveIreland do you think you can squash to just one commit?

PR #1169, opened 2017-02-26
Support for Configurable Dinner Tables Feature

jlctmaster:
This feature was initiated by @deepshah where the support was given to the sales engine to take into account the billing for the restaurant tables, what I did was give it support to be activated through the configuration of the business, This is a functionality that does not apply to 100% of OSPOS users, as the translation is supported only for the English and Spanish languages.


jlctmaster:
Hello @daN4cat, please review the changes.

Greetings

application/models/Sale.php
```diff
@@ -493,7 +493,8 @@ public function save($items, $customer_id, $employee_id, $comment, $invoice_numb
 			'customer_id'	 => $this->Customer->exists($customer_id) ? $customer_id : null,
 			'employee_id'	 => $employee_id,
 			'comment'		 => $comment,
-			'invoice_number' => $invoice_number
+			'invoice_number' => $invoice_number,
+			'dinner_table_id'=> $dinner_table
```
jekkos: hm think the data model is being polluted with all kinds of specific functionality.. not sure if there's a more generic way to tackle these additional links

jlctmaster: I think these are specific events that end up being part of the billing, which is the fact made in the process, all of the before the invoice generated could be considered optional and will depend on the business where you are using

daN4cat:
@jekkos are you ok with adding this functionality?

The review seems ok to me.

jekkos:
As long as we can disable it in config then ok

odiea:
Congrats to all. This merge went well. Thanks

SteveIreland:
Yeah!

jlctmaster:
Thanks, the next step is to support the ospos to connect through a Rest Full API and get a mobile app for taking orders in a restaurant. :)

odiea:
Pretty soon we will be calling this the SAK of POS (Swiss Army Knife)

PR #1178, opened 2017-03-04
Feature/reward points customer

joshua1234511:


application/config/config.php
```diff
@@ -358,7 +358,7 @@
 | https://codeigniter.com/user_guide/libraries/encryption.html
 |
 */
-$config['encryption_key'] = '';
+$config['encryption_key'] = 'ec58a27d5854523044e3283b673a97c1deb59fe1e6f2ca7f77ed5e0f4c5aa02a';
```
daN4cat: please remove your key

jekkos:
Cool to see those PR's coming in. On the other hand I think we should consider doing a small release first as we have quite some functionality already?

daN4cat:
We can pull this in and wait for the tax rework by @SteveIreland. Also I need to test and polish a bit things before we can baseline and release.

jekkos:
With polish you mean refactor? Maybe we might want to make an epic for that

daN4cat:
I mean adjust few things or inconsistencies. You can call it small refactoring too.

application/config/config.php
joshua1234511: Done Removed key and pushed

joshua1234511:
I agree on refactor as the language files as i see are in proper order  i did not add properly .

I am working now on adding a front end for customers where they can add id or email  with date birth and be able to check their sale and Reward points available.

daN4cat:
And how do you plan to interface OSPOS from the mobile app?

joshua1234511:
@daN4cat 
Did not understand interface OSPOS from the mobile app?
There is a mobile app also

daN4cat:
How your mobile app will call OSPOS service?

jekkos:
@joshua1234511 would be interested to see whether you're gonna open source it or not?

joshua1234511:
Yes it would be open source

joshua1234511:
Oh by  mobile I did not mean the app 
It was the login using e-mail and mobile number for validation of customer

daN4cat:
Do you envision a solution where the customer could login to a different OSPOS page? Is that what you mean?

If yes I would be very careful as you open a backdoor for potential attacks. There are security implications to consider and talk about.

joshua1234511:
Yes something like that 
Yes have added a new controller customer_login and managing through that. 
For customer added a new user-data customer_id for handling the security side
Pushed the new changes to new commit 

Attached a new Screen cast with new features
: https://www.screencast.com/t/lZbtEJidV

Admin:         /login
Customer:   /customer_login

joshua1234511:
Please review the code and suggest if any security issues

daN4cat:
Thank for the extra contribution but please revert to the previous version without the customer login.
I prefer to keep that separate and not have everything in one big pot.

joshua1234511:
@daN4cat I have reverted back changes to the last commit and pushed
ie: the previous version without the customer login Changes

daN4cat:
@joshua1234511 thanks for reverting the commit.

Please add all the *.sql content to the upgrade script 3.0.2_to_3.1.0.sql, also you need to make sure tables.sql contains the new tables as a new installation needs to be fully correct from an SQL schema point of view.

joshua1234511:
Added the sql  to upgrade script 3.0.2_to_3.1.0.sql
And Also modified the tables sql to incorporate these table changes

jlctmaster:
Hello @joshua1234511 , thanks for this improvement, @daN4cat  who should apply the translations? For example, I use the module translated into Spanish, can I do it?

daN4cat:
@jlctmaster you can using weblate, check the wiki for instructions.

PR #1181, opened 2017-03-05
Update for Receiving quantity. Multiplier now works.

odiea:


daN4cat:
You need to add your name to LICENSE file 😊

application/views/receivings/receipt.php
```diff
@@ -75,7 +75,8 @@
 				<td><?php echo to_currency($item['price']); ?></td>
 				<td><?php echo to_quantity_decimals($item['quantity']) . " " . ($show_stock_locations ? " [" . $item['stock_name'] . "]" : ""); 
 				?>&nbsp;&nbsp;&nbsp;x <?php echo $item['receiving_quantity'] != 0 ? to_quantity_decimals($item['receiving_quantity']) : 1; ?></td>
-				<td><div class="total-value"><?php echo to_currency($item['total']); ?></div></td>
+								
+              <td><div class="total-value"><?php echo to_currency($item['total'] * $item['receiving_quantity'] ); ?></div></td>
```
daN4cat: could you please adjust the indentation?

odiea:
One of these days I will add my name. But not for this.

application/models/Receiving.php
```diff
@@ -264,7 +264,7 @@ public function create_temp_table(array $inputs)
 					MAX(serialnumber) AS serialnumber,
 					MAX(receivings_items.description) AS description,
 					MAX(item_unit_price * quantity_purchased - item_unit_price * quantity_purchased * discount_percent / 100) AS subtotal,
-					MAX(item_unit_price * quantity_purchased - item_unit_price * quantity_purchased * discount_percent / 100) AS total,
+					MAX(item_unit_price * quantity_purchased * receivings_items.receiving_quantity - item_unit_price * quantity_purchased * discount_percent / 100) AS total,
```
jekkos: What about discount? It's only applied on the original quantity and not on the total of items?

application/views/receivings/receipt.php
```diff
@@ -75,7 +75,7 @@
 				<td><?php echo to_currency($item['price']); ?></td>
 				<td><?php echo to_quantity_decimals($item['quantity']) . " " . ($show_stock_locations ? " [" . $item['stock_name'] . "]" : ""); 
 				?>&nbsp;&nbsp;&nbsp;x <?php echo $item['receiving_quantity'] != 0 ? to_quantity_decimals($item['receiving_quantity']) : 1; ?></td>
-				<td><div class="total-value"><?php echo to_currency($item['total']); ?></div></td>
+				<td><div class="total-value"><?php echo to_currency($item['total'] * $item['receiving_quantity'] ); ?></div></td>
```
jekkos: Not sure but in receiving_lib the total does already take receiving_quantity into account. Do we need to multiply again then here?

odiea: When I leave the receipt the original way no Total shows on the item line. I will see where I screwed up with receiving lib.

odiea:
I hope this works better. Appears to be working on discounts now.

odiea:
I would squash the commits but I don't know how. I would also like someone to download it and test it. I did download from my branch and it appeared to work ok.

odiea:
I have this down to 2 commits now. I changed the following line and added the receiving quantity multiplier.
MAX(item_cost_price * quantity_purchased * receivings_items.receiving_quantity ) AS cost.
Now my question is on the following line should I add the receivings multiplier to the last part. I have tried it with it and without it and I do not see much difference.
(item_cost_price * quantity_purchased)) AS profit,
I want to keep it at 2 commits the db update separate from the program update so there is no confusion that the db part is necessary. If needed I will cancel this pull and do another one.

odiea:
I am going to remove this until I can work on it some more.

PR #1188, opened 2017-03-15
Langlais115 item by article

Langlais115:
Hi There,

Here is an update for the French translation.

I hope that will help.

Best regards,


daN4cat:
@jekkos can I merge this and then you upstream in Weblate? I believe that we need to upstream a bit of new additions to Weblate related to the latest additions anyway.

jekkos:
OK would have been cleaner to have it one commit, but as it's already merged I guess too late now. I've pulled these new translations and synced with weblate. Should be clear in stats now there as well.

daN4cat:
Sorry about that, seeing no reply I thought you were to busy and pushed the button.
Hopefully the upstreaming was not too much of a pain, but in any case we had other instances requiring a resync (e.g. Dinner Tables and Rewards contributions).

jekkos:
Hmm some issues with weblate, from the docs 

```
Following things are known to be broken:

    Adding new units to translation, every translation has to contain all strings (even if empty).
```

jekkos:
So newly added strings to english should be added to other languages as well, to have them exported correctly. Might want to fix that maybe in weblate itself.

PR #1257, opened 2017-04-21
Support for change user password

jlctmaster:
This support allows to change the password of the authenticated user from any part of the system, simply by clicking on the link of the user name.


application/views/configs/manage.php
```diff
@@ -1,5 +1,9 @@
 <?php $this->load->view("partial/header"); ?>
 
+<script type="text/javascript">
```
daN4cat: Wouldn't be easier to add it to partial header?

application/models/Employee.php
```diff
@@ -411,5 +411,24 @@ public function get_employee_grants($person_id)
 
 		return $this->db->get()->result_array();
 	}
+	/*
+	Change password for the employee
+	*/
+	public function change_password(&$employee_data, $employee_id = FALSE)
```
daN4cat: I understand that in PHP there is no optimisation in using & (alias)

daN4cat:
Shouldn't we add a current password before inputting twice the new one so we make sure only the correct logged in user can change their own password?
We need to be careful because employees might leave OSPOS open to other to use and those could play silly and change the password.

Please add that feature before I can merge this.

jekkos:
@daN4cat Correct, at least the current password should be known before changing it.

application/views/configs/manage.php
jlctmaster: Hello, I did not add that script in the partial header file because only in those windows does not call that function and call it there would make it in the windows where already it is calling separately, could create some conflict in the js.

jlctmaster:
Okay, thanks for the suggestions, I'll apply it.

application/views/configs/manage.php
daN4cat: it's ok

daN4cat:
Good, thanks for the contribution.

application/views/configs/manage.php
daN4cat: it's ok

jlctmaster:
Hi, I already applied the suggestions, check it out in what they can.

Greetings.

PR #1264, opened 2017-04-24
fixes for giftcard

joshua1234511:
While I was checking the working, I noticed that gift card can add customer, but for sale it is never checked so I added code check for customer if gift card is assigned. 

Attached scrrencast: https://www.screencast.com/t/nMfwmylKFYO

The code I added checks if gift card has customer assigned to it, if it does that the customer should match, else gift card will give error invalid Customer.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -227,8 +227,12 @@ class Sales extends Secure_Controller
 				$payment_type = $payment_type . ':' . $giftcard_num;
 				$current_payments_with_giftcard = isset($payments[$payment_type]) ? $payments[$payment_type]['payment_amount'] : 0;
 				$cur_giftcard_value = $this->Giftcard->get_giftcard_value($giftcard_num);
-
-				if(($cur_giftcard_value - $current_payments_with_giftcard) <= 0)
+				$cur_giftcard_customer = $this->Giftcard->get_giftcard_customer($giftcard_num);
+				$customer_id = $this->sale_lib->get_customer();
+				if(isset($cur_giftcard_customer) && $cur_giftcard_customer != $customer_id){
+					$data['error'] = $this->lang->line('giftcards_cannot_use', $giftcard_num);
+				}
+				elseif(($cur_giftcard_value - $current_payments_with_giftcard) <= 0)
 				{
 					$data['error'] = $this->lang->line('giftcards_remaining_balance', $giftcard_num, to_currency($cur_giftcard_value));
 				}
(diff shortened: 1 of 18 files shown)


daN4cat:
Hi, the reason why giftcard are anonymous is because you may give as a gift to somebody and not be named.
It was discussed in the past and so I would not enforce any rules at sales time between customer and giftcard.
As long as the code is fine it should be OK. Unless you have some other scenario in mind where it could be wrong?

joshua1234511:
Usually  when we give birthday cards to customers, or anniversary we provide gift cards fro them. this bonus is only for them. and not transferable. 
The other cards can still be there. 
This is also helpful in case of misplacing of cards

joshua1234511:
Doing the above you can have both 
1: gift to somebody and not be named.
2: gift to somebody and restrict.

daN4cat:
I want to hear an opinion from other people too.

joshua1234511:
Yes that is there
It must be agreeable to all. 
This was just my opinion,

daN4cat:
@jekkos @odiea @SteveIreland any view?

odiea:
I will download his branch and see how it does.

odiea:
The only issue I see is when you start with generate in series they auto number correctly but when you switch over to a Random number say 1234-56 and then go back to series the series then uses the 1234 and not the previous number of say 45. Other then that it appears to be working. Eventually I hope he will add the ability to add a customer to the gift card form instead of add new customer then add new giftcard. Giftcards on the fly

daN4cat:
You mean when creating a new giftcard for a customer not in the system? That would be handy but it's probably difficult to manage as it implies two modals

odiea:
Yes but we can dream can't we.

daN4cat:
Indeed

daN4cat:
So @odiea you are fine with this change. @jekkos any opinion?

odiea:
It is another nice option. So yes.

daN4cat:
Ok if no more feedback I'll merge this next days.

jekkos:
I agree to merge as long as one can still make giftcards without having to add a customer.

odiea:
Yes you can still use blank customer giftcards. If the customer name is selected he can also use a blank name giftcard.

daN4cat:
Merged and submitted to master.

daN4cat:
I was wondering, in the sale register it would be nice to suggest the gifcard number as it is entered, like the search suggestion we give for customer or items.
In case a customer is selected the suggestion should be narrowed down to the giftcard allocated to the customer.

@joshua1234511 is that something you can look into as a nice user experience addition?

joshua1234511:
https://github.com/jekkos/opensourcepos/pull/1273

PR #1273, opened 2017-04-27
auto suggest gift card

joshua1234511:
Added Auto complete for gift card

Screen cast: https://www.screencast.com/t/xKjIRJjUXzb

@daN4cat 
In case a customer is selected the suggestion should be narrowed down to the gift card allocated to the customer.
Did not narrow down,
As customer might have anonymous gift card also if you still want me to do that ill add it

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1,0 +1,3 @@
+input#amount_tendered:disabled {
+    display:none;
+}
--- a/application/controllers/Giftcards.php
+++ b/application/controllers/Giftcards.php
@@ -44,6 +44,14 @@ class Giftcards extends Secure_Controller
 	/*
 	Gives search suggestions based on what is being searched for
 	*/
+
+	public function suggest()
+	{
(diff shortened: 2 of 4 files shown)


public/css/style.css
```diff
@@ -0,0 +1,3 @@
+input#amount_tendered:disabled {
```
daN4cat: this probably needs to move to the correct stylesheet.
I will look into that when I'll merge this code, there are also few other minor things to adjust but I prefer to do rather than writing comments here,

daN4cat:
@joshua1234511 looks good, thanks. I'll see if I can merge it in this evening.

With regards to the filtering point, maybe we can sort the suggestion putting the giftcards allocated to the customer first. That will be a good trade off between the too narrow and nothing. 
What do you think?

daN4cat:
Thought about what I said earlier with regards to sorting the list and it doesn't make complete sense.
I'll merge this PR now.

daN4cat:
Merged

PR #1281, opened 2017-04-30
Sales tax history migration

SteveIreland:
This pull request provides support for updating the `sales_items_taxes` table with a calculated tax amount and populating the `sales_taxes` table with an appropriately rounded tax entries for the sale.  These updates will be the future source for tax amount reporting.

The migration action takes place under the general configuration panel.  You will need to enable migration and enter the number of sales transactions to be upgraded each time you press the "Save" button. 

When you press the Save button (or in many cases simply press Enter) and "Migrate sales history on Save" has been checked then the system will check to see if there are any sales that need to be migrated.   If there are then it will start migrating sales.  The number of sales it will migrate is specified in the configuration value labeled "Sales to migrate per Save"  This is a value that you specify.  I like a value of around 100 and will sit there and press Save until there are no more sales to be migrated.  You can specify a larger number but the response time will be corresponding longer.  If the response time is too long then your session might exceed the allowed timeout value and crash before returning to you - so be conservative in the number of sales to migrate at a time.

Each time you press Save, after the system has migrated the selected number of sales, the system will respond with a message telling you how many sales it migrated and will tell you the number of sales remaining to be migrated.

At this point, the only way to check the result of the migration is to use an SQL query tool to check to see if new records have been added to the `sales_taxes` table. 

A side effect of this migration effort is that now records will be added to the `sales_taxes` table even if the amount is zero.  The migration will create the records even if the sales amount is zero.  The existence of these records is how I know that migration does not need to be repeated for these sales.


application/models/Sale.php
```diff
@@ -752,11 +752,14 @@ public function save_sales_item_tax(&$customer, &$sale_id, &$item, $tax_basis, &
 		}
 	}
 
-	private function save_sales_tax(&$sales_taxes)
+	public function save_sales_tax(&$sales_taxes)
 	{
+		error_log('>>>Sale.save_sales_tax started'.print_r($sales_taxes, TRUE));
```
daN4cat: there is an error_log that was left in

application/models/Sale.php
```diff
@@ -752,11 +752,14 @@ public function save_sales_item_tax(&$customer, &$sale_id, &$item, $tax_basis, &
...
 		foreach($sales_taxes as $line=>$sales_tax)
 		{
+			error_log('>>>Sale.save_sales_tax $sales_tax - '.print_r($sales_tax, TRUE));
```
daN4cat: here other error_logs

application/models/Sale.php
```diff
@@ -1205,5 +1208,63 @@ public function get_suspended_sale_info($sale_id)
...
+		$this->db->limit($block_count);
+		$query = $this->db->get();
+
+		return $query;
+
+	}
+
+	public function update_sales_items_taxes_amount($sale_id, $line, $name, $percent, $tax_type, $item_tax_amount)
+	{
+		error_log('>>>Sale.update_sales_items_taxes_amount '.$sale_id.', '.$line.', '.$name.', '.$percent.', '.$tax_type.', '.$item_tax_amount);
```
daN4cat: another error_log

daN4cat:
I decided to merge it in even with the error_logs so some people can test, but please @SteveIreland fix those comments and I'll merge the next PR.

SteveIreland:
Done, thanks.

jekkos:
Hmm actually I don't really like the idea of doing a migration when saving something in the config screen. Think it'd be better to isolate this migration code in it's own place if it isn't possible to do it with plain SQL. There's some support in CI to detect whether migration is necessary or not. We don't we just leverage that ?

jekkos:
https://www.codeigniter.com/user_guide/libraries/migration.html

In fact this tax story would just be one case of such a migration, which can then be tracked by CI itself whether  it needs to be done or not.. This system is something I had on my list for adding later on to automate the updating process further

daN4cat:
@jekkos would you prefer to revert these two commit?

jekkos:
Also this Migration class would make this an async process. we shouldn't bother the user batching up the number of records to avoid session session timeouts,

daN4cat:
OK, I revert the two commits and we can talk about...

jekkos:
I think this is just a nice opportunity to have this Mirgation thing introduced as this time we're relying on actual code instead of SQL. Should just be a matter of moving this migration code out of the config section imho.

daN4cat:
reverted

daN4cat:
This repo has a migration example that would be useful to "copy & paste": https://github.com/bthiban/codeigniter-migration

Also some articles: http://stackoverflow.com/questions/9154065/how-do-i-run-codeigniter-migrations, https://codeinphp.github.io/post/database-migrations-in-codeigniter/, https://www.tutorialsplane.com/codeigniter-migrations-class-library/

daN4cat:
@SteveIreland can you please explain me in detail what you need to migrate with that script?
Migrate from sales_items_taxes to sales_taxes? Is that all or something else?
Also are we supposed to have both tables kept or should we then drop sales_items_taxes?

SteveIreland:
The detail reports are reporting taxes at the item level.  Of course, trying to balance item level tax amounts to the actual tax amount collected will often result in rounding differences.  So the `sales_taxes` table is intended to identify what is actually collected and the `sales_items_taxes` table is intended to approximate the amount of the tax collected that is attributable to that item.

I wasn't planning on removing that table because it might assist in some reporting requirements.  Since we do report tax amounts at a line level I was going to continue to use that table for that level of reporting.

But at the sales level, I was  going to use the `sales_tax` table to accurately report the tax that was collected.

SteveIreland:
So .... if you are going to incorporate CodeIgniter's Migration class, then are you going to use the migration class to upgrade from 3.0.1 to 3.1.0 instead of using the database version to version upgrade script?

Also, I'm a bit confused on what the state of my pull request is and my role now that it's been backed out.

Lost and confused.

daN4cat:
The complexity of this tax story escapes me and at the same time I'm worried that we build some complex that will be not easy to maintain.

The way I understand it is the following:

1. An item can have 0 to N taxes
2. There could be different taxes (e.g. children clothing 0%, adult clothing 20%, wine 30%, city 5%, county 10%, VAT 20%)
3. Item sale price can be tax included (VAT) or tax excluded
4. If a customer is selected with a valid address the Taxation can be based on the customer address (online sale)
5. The sale tax can be rounded

Is that all?

daN4cat:
@SteveIreland I reverted to allow time to discuss.
I wanted to understand if using an SQL script could still do the copy. Is the issue the calculation of the sale_tax_basis the problem why you didn't use it?

daN4cat:
I was checking and the logic implemented in SQL is probably more complicated.

daN4cat:
@jekkos was suggesting to use the migration approach from CodeIgniter. Not sure @SteveIreland you want to use that?
Or do you prefer to leave the code as it's in your PR?

daN4cat:
@jekkos it's better if you look into this. From my side I'm fine with any solution as long as we get to an end of this.

jekkos:
In my case it was mainly a usability issue. i don't have a very good view on this whole tax story and that's maybe because things aren"t that complicated here in Belgium.

@SteveIreland I appreciate all the work you're doing but I was planning to improve the upgrade experience for OSPOS and this solution doesn't really fit in there. AFAIK this migration thing resolves some of the problems you solved ad hoc in that PR. If you don't have the time then I'll look at it eventually, however I won't be able to commit on any time contraints for now.

SteveIreland:
I may be mistaken but I don't see where the Migration class addresses the session timeout issue, but that might just be an old issue from decades ago that I had to be aware of.   With current hardware and processor speeds maybe it's not an issue.

I'll give it a shot.  The core of the migration has been coded up.  It was how to initiate it where I actually spent most of my time since I tried three variation before settling on what was there.  I'll give the Migration class since it offers a "standard" approach.

"That's the nice thing about standards ... there are so many of them" ...  Borrowed quote.

Did you want me to incorporate `3.0.2_to_3.1.0.sql` into it?

@daN4cat Your list of tax features is pretty close to what's there.  Some of the complexity comes from the fact that we needed to continue to support an item based sales tax in addition to the customer based sales tax.  VAT tax remains item based, but I wanted to include the capability to incorporate a VAT tax in the sales tax configuration (in an effort to simplify tax maintenance in future releases.  It's more complicated at the moment than it needs to be, but I sensed that we didn't want to force anyone to change if they weren't ready for it.

jekkos:
Ok great thanks for going that exfra mile here. Would be nice if it's possible to include the SQL script as well.

daN4cat:
With the migration addition I would first start with one example based on this upgrade by @SteveIreland and then see how to port the current SQL script.

PR #1296, opened 2017-05-07
Add support for database migration

SteveIreland:
This pull request will run the sales tax history migration using the Codeigniter Migrate tool.  It's accessed via a new module called Migrate.


jekkos:
Look good. Beside having this migration class for taxes I guess we can have one for update scripts as well. Then it'd be a matter of numbering scripts from the pasts and find a way to execute them against the current db connection. This would then automate the process as a whole.

jekkos:
I'll try later today to see if it works so we can merge.

daN4cat:
I would move all sale and tax_lib migrate functions to migrate class. Please don't merge yet.

daN4cat:
We don't want to pollute normal classes with one off migration.

SteveIreland:
The way it is program now is more "generic" than just the data migration for sales tax history.  It is coded to "catch up" to whatever version is specified in the migration.php configuration file which is "20170502221500" at this time.

@daN4cat I would disagree with the term "pollution".  Since those functions are going to be there forever, from a maintenance perspective it seems it would be preferable to be associated with the class on which they act (shouldn't an object know how to upgrade itself?).  And from an ongoing maintenance perspective, if there is an improvement or change to the tables or logic then they would also need to be correspondingly changed and improved.   It doesn't seem like a good idea to having similar logic spread out so that more scripts need to be searched and modified than there already is.  When I was coding it up I appreciated having those methods in the same class.

daN4cat:
@SteveIreland you need to plan upgrades in incremental steps and upgrade always from the previous step. Once an incremental step is working you don't touch it otherwise you need to run regression tests always for all the upgrades.
So if you keep common code behind those incremental upgrade steps you end up running a lot of regressions testing everytime and longer any time you add one new incremental upgrade.
I would replicate the way we have done so far with the SQL script and once done we don't touch.
So I still strongly suggest to have migration code left to each migration class and incremental migrations.
This migration facility would just remove the complication of the sql scripts for the ones that don't want to do that and in such way we can start considering automatic upgrades.

jekkos:
@SteveIreland I must say I appreciate all the effort you put in this part, but I think I agree with @daN4cat . This code is used just once and although it belongs in the sale etc model, I'd rather have it in some migration specific entity so that it's isolated from the 'real' business logic.

I might take a look and fix that if I find some time soon.

SteveIreland:
I started relocating the functions yesterday and I believe I've relocated the relevant functions to the migration file and refreshed the pull request.

If someone is going to take a look at the code, where I would really appreciate a more knowledgeable pair of eyes is a look at WHY I am not getting a successful completion message being displayed when it is in fact a successful migration.  I am reasonably certain this was working before I moved the functions into the migration file.

Of course, the way things have been going for me over the last week I would suspect demonic forces in action ... but the answer is probably simpler.

SteveIreland:
I also changed the migration option to run all migrations in the migrations folder through the latest one in the file instead of the one that is configured as "current".  The reason is that it simplifies the testing of migrations when one doesn't need to go in and change the migration.php file for each test.

SteveIreland:
I added a comment about the migration to the Wiki at https://github.com/jekkos/opensourcepos/wiki/Sales-Tax-by-Customer#migration

SteveIreland:
Now that this is "done?"  I'm going to move on to starting a new branch to change the report SQL statements to use the new sales tax table.  It should also help with checking the migration results.

daN4cat:
I'll check into this over the weekend. I've been slowed down in recent months by a new job and much stuff to sort out at architecture level.

jekkos:
Same for me.. regular job taking up a lot of energy, so less time to work / invest in this project.

daN4cat:
Thanks for the improvement @SteveIreland, I merged the code to master.

I would say that we could have moved the migrate functionality to Store Config but it's fine this solution for the time being. When I find the time I'll move to Store Config.

daN4cat:
Looking forward for the completion of the work so we have a final migration to the new tax module and no old tax anymore (hopefully).

SteveIreland:
@daN4cat Feel free to move it to Store Config.  I started to add it as new tab under Store Config (before using the Migration class), but it was just too different in behavior and I concluded it wasn't  going to be acceptable.  It was also unnerving to me that all configuration tabs are loaded at the same time even though they may not be used for update.    So I just felt the migration behavior needed to be better isolated from the other configuration tabs.

I've started working on the report changes.  I pretty confident that I'll have a pull request ready for it for next weekend.

daN4cat:
@SteveIreland as said for the time it's OK. I just found funny that in order to support a database migration we need to run a SQL script update just for that module addition. But it's fine we can live with that.

PR #1308, opened 2017-05-21
Upgrades sales tax reporting

SteveIreland:
This includes the changes to tax reporting that should reflect the changes made by sales tax migration.

Please be aware that this change will make the sales reports unusable for reporting historical data without first running the sales tax migration. 

I'm waiting on someone to tell me that something isn't right (with an example as to why) so that I can fix it.  Overall this was all too complex and has too many nuances for everything to be correct on the first iteration.   The changes to the reporting weren't all that complex but the impact of the tax changes was because of the permutations possible with the tax styling. 

Thanks and good luck.


application/models/reports/Detailed_sales.php
```diff
@@ -113,8 +113,9 @@ public function getData(array $inputs)
 
 	public function getSummaryData(array $inputs)
 	{
-		$this->db->select('SUM(subtotal) AS subtotal, SUM(tax) AS tax, SUM(total) AS total, SUM(cost) AS cost, SUM(profit) AS profit');
-		$this->db->from('sales_items_temp');
+		$this->db->select('SUM(subtotal) AS subtotal, SUM(sales_taxes.sale_tax_amount) AS tax, SUM(total) AS total, SUM(cost) AS cost, SUM(profit) AS profit');
```
daN4cat: I'm not sure I get. `sales_items_temp` has a newly calculated tax why isn't that used in the sum here?

daN4cat:
What about Summary_report.php? That should be changed too.

application/models/reports/Detailed_sales.php
SteveIreland: The sales item tax amount is calculated according to local tax rules.  And a simple SQL tax calculation isn't up to the task.  So the sales item tax amount is now persisted and used for reporting purposes.  This provides a lot more flexibility at the point of sale.  At least that was my rationalization for doing it.

SteveIreland:
Yeah ..... I recall now looking at `Summary_report.php` and deciding that would be a good one to put off until later.  I did, and then conveniently forgot about it.   Working on it today.

application/models/reports/Detailed_sales.php
daN4cat: Is it because the SUM of SUM(sales_items_taxes.item_tax_amount) is not equal to the SUM of sales_taxes.tax?

Is that a rounding issue or else?

Because there are to concept of taxes one is the sum per item and the other is the total sum and we are calculating differently. So it looks to me....

daN4cat: If I'm not mistaken this looks like two dimensional modes serving two purposes (I mean the temp table and the taxes table).

SteveIreland:
Sorry, but I have to start asking some apparently dumb questions.  For the summary tax report what is the value of the Count field?  For sales tax reporting it makes sense to count the number of invoices.  However, the current report is counting the number of items being charged that tax rate.  Is it important that the count field be the number of items or would it be okay for this to be the number of sales?

daN4cat:
@SteveIreland invoices are optional so you can count that if invoices are used. Count of items is an "obvious" choice. I assume one wants to read the break down by tax % to pay the taxes correctly. So probably given the new model you want to summarise that only? But make sure it's not invoices centric.

daN4cat:
@SteveIreland any progress about this? It would be good to close this tax story....

daN4cat:
@SteveIreland I merged these changes after having applied the same approach to the rest of the reports.
However it doesn't look right because the taxes in detailed reports are not right and not matching the summary ones.

Please look into this as I want to get to an end of this tax story to release 3.1.0

daN4cat:
I fixed the tax summary in all the detailed reports.

There were a number of regressions around the rewards introduction as some reports were broken and not managing the disabled flag.

PR #1342, opened 2017-06-14
Fix undefined table_status error in Sale.php

SteveIreland:
This has resolved the error for me with the undefined table_status which occurs when the dinner table feature is not enabled.  I default the table status to '0' and I'm just assuming that the value is irrelevant when the dinner table feature is not enabled but if not then '0' will be  acceptable.

### DIFF
--- a/application/models/Sale.php
+++ b/application/models/Sale.php
@@ -522,6 +522,10 @@ class Sale extends CI_Model
 				$table_status = 0;
 			}
 		}
+		else
+		{
+			$table_status = 0;
+		}
 
 		$sales_data = array(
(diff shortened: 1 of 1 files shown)


application/models/Sale.php
```diff
@@ -522,6 +522,10 @@ public function save(&$sale_status, &$items, $customer_id, $employee_id, $commen
 				$table_status = 0;
 			}
 		}
+		else
+		{
+			$table_status = 0;
```
jekkos: Ok for this fix, but might be better to use defines to have assign some kind of meaning to these constants

PR #1358, opened 2017-06-24
Update Receiving.php

odiea:


odiea:
Apparently copy and paste did not work well. Lines 265 and 266 are duplicate. Can 1 of them be removed?

daN4cat:
Fixed

PR #1397, opened 2017-07-15
Improve tax category maintenance.

SteveIreland:
This changes tax category maintenance to insure that the tax category is not being referenced in items nor is being referenced. It also prevents the "Standard" tax category from being deleted (tax_category_id = 0);

### DIFF
--- a/application/controllers/Config.php
+++ b/application/controllers/Config.php
@@ -637,6 +637,7 @@ class Config extends Secure_Controller
 		);
 
 		$success = $this->Appconfig->batch_save($batch_save_data) ? TRUE : FALSE;
+		$delete_rejected = FALSE;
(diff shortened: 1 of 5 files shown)


SteveIreland:
This changes tax category maintenance to insure that the tax category is not being referenced in items nor is being referenced.  It also prevents the "Standard" tax category from being deleted (tax_category_id = 0);

daN4cat:
I merged from my mobile but can't test so I hope it's fine.

jekkos:
Hmm so you disallow delete from the application. Think we should then also add a fk constraint to prevent delete if there's a fk present for this table. Otherwise it can still be circumvented by db queries ( and break application logic)

PR #1398, opened 2017-07-15
Add missing language element for rounding code "Unknown" and eliminate duplicate statements in upgrade statement, and fix issue with 0 tax_category_id

SteveIreland:


SteveIreland:
Okay, this pull request isn't perfect but in my preliminary testing it resolves the issues that have been reported.  However, it has .... again ... resulted in the successful popup panel not displaying and has broken - in tax maintenance - the drop down list for tax categories.  I think it is a javascript issue, but the only javascript error I'm seeing is in the minimized css.  I'm not sure what to do with that, but I'll continue to work on it today.  I need to take a little break here but will resume testing and debugging when I get back.

odiea:
I added a new tax code and then changed that to the default code.
2017-07-16 08_03_02-wedidit _ powered by ospos 3 1 0
I then tried to change back to the original code. In config it shows it has changed but it still is charging the new code I put into place earlier. Plus it appears the math is off even after the rounding has been changed. Enjoy your time off.

odiea:
Currently I would suggest to not merge this pull request. The db update appears to break the Items taxes also. Or something went haywire with the taxes somewhere.

SteveIreland:
I've updated this pull request to include a fix for the tax rounding issue - a couple of things have "appeared" that are actually unrelated to the original change, but I've included those fixes here since they are small code changes.

odiea:
This appears to have fixed my previous issues. It now uses the Item tax after I deselected  using the customer tax.

SteveIreland:
I know you know this, but just in case anyone in the future stumbles across this thread I should point out that the concept of switching between VAT tax and Sales tax isn't something that should ever be done without starting from scratch.  Sales history is locked in to either one or the other.  Reports would be totally useless.

odiea:
Thanks. To test Vat Tax I created a different database just for that.

odiea:
I downloaded your branch and had no issues. I copied and pasted this request and I do now get the next page success message and there is no dropdown selection in Add or update new tax {Alcohol,Service}. So you did something to fix this issue. Can't wait for this request so I  can see how it works out.

daN4cat:
The popup doesn't pop up because there is an error in the code, I always ended up fixing it.
Most likely some referenced variable is not existing or assigned with null.
It's not the js for sure.
Now I'm far from any pc or laptop until end of the month, so it is you fix or we wait.
The way to check in Chrome inspector is in the Network tab the Response to the call, you'll see some error dump.

SteveIreland:
Yes, I've been chasing it down for a couple of hours and have found the problem.

I've uploaded a temporary fix that will restore the popup message behavior.  It is actually usable in a production environment for sales tax only because the solution is bound to the English language.

I'm having a bit of a challenge with an an unterminated string JavaScript error.  I've tried about 6 approaches to resolve the issue but I'm running out of time and will need to do additional research.  I'm sure it's an oh-duh solution, but it seems I have a lesson to learn.

odiea:
After replacing the old form with this form all appears to working properly. The form change was the only new submit I saw.

SteveIreland:
I know ... it's pathetic.  So many hours I put into this and ended up changing only a couple of lines of code. 

:scream:

odiea:
It's not a game breaker but I am still unable to delete an exception. The table shows it is being deleted but when I go back and open the form it is still there.

odiea:
If there are 2 rates only 1 of them can be deleted.

SteveIreland:
I bet it's the same issue that @daN4cat  resolved for for the other table maintenance.  I'll clone what he did in order to fix that.  I probably borrowed that logic from the other routines.

jekkos:
@SteveIreland if you encounter this kind of error then it might help to use the debug versions of the javascript to see in which part it breaks.. just add ?debug=true at the end of the URL and it 'll load the sources that you can debug using chrome inspector.

Best to put a breakpoint in manage_tables.js where ajax response for the dialogs are parsed so you can check the return code of the last ajax call (that probably has non valid json in it)

database/3.0.2_to_3.1.0.sql
```diff
@@ -244,17 +239,33 @@ INSERT INTO `ospos_grants` (`permission_id`, `person_id`) VALUES
...
 ALTER TABLE `ospos_tax_categories`
   MODIFY COLUMN `tax_category_id` int(10) NOT NULL AUTO_INCREMENT;
+
+INSERT INTO `ospos_tax_categories` ( `tax_category_id`,`tax_category`, `tax_group_sequence` ) VALUES
+  (1, 'Standard', 10),
+  (2, 'Service', 12),
+  (3, 'Alcohol', 11);
+
+ALTER TABLE `ospos_items`
+  MODIFY COLUMN `tax_category_id` int(10) NOT NULL DEFAULT 1;
```
jekkos: think we're still missing FK constraints here as mentioned in the other issue. Check out the constraints for other tables and maybe add an delete cascade update on this one so that the reference is cleared once category is removed

database/database.sql
```diff
@@ -821,9 +821,9 @@ CREATE TABLE IF NOT EXISTS `ospos_tax_categories` (
 --
 
 INSERT INTO `ospos_tax_categories` ( `tax_category_id`,`tax_category`, `tax_group_sequence` ) VALUES
-  (0, 'Standard', 10),
-  (1, 'Service', 12),
-  (2, 'Alcohol', 11);
+  (1, 'Standard', 10),
```
jekkos: If you have auto_increment then you don't need to explicitly define the ID's, as they should be in line with where the increment start (which should be 1 by default)

database/3.0.2_to_3.1.0.sql
```diff
@@ -244,17 +239,33 @@ INSERT INTO `ospos_grants` (`permission_id`, `person_id`) VALUES
...
+
+UPDATE `ospos_items` SET `tax_category_id` = 1 WHERE `tax_category_id` = 0;
+
+-- If you have added any tax codes, the following will correct the rate_tax_category_id on the tax_code_rates table,
+-- but you might need to add more UPDATE statements depending on how may tax codes and/or tax categories you've added
+
+UPDATE `ospos_tax_code_rates` SET rate_tax_category_id = 4 WHERE rate_tax_category_id = 3;
```
jekkos: I don't really consider this one as complete for users that were already on master, a more generic version that adds 1 to existing categories if a category exists with id = 0 might work better..

database/3.0.2_to_3.1.0.sql
```diff
@@ -244,17 +239,33 @@ INSERT INTO `ospos_grants` (`permission_id`, `person_id`) VALUES
...
-  
 -- long alternate description
   
 ALTER TABLE `ospos_sales_items`
   MODIFY COLUMN `description` varchar(255) DEFAULT NULL;
   
 -- fix tax category maintenance
-  
+
+DELETE FROM `ospos_tax_categories` where tax_category_id in (0, 1, 2, 3);
```
jekkos: this delete will now leave the db in an inconsistent state (as it will break references on entities that have fk's to this one)

database/tables.sql
```diff
@@ -821,9 +821,9 @@ CREATE TABLE IF NOT EXISTS `ospos_tax_categories` (
 --
 
 INSERT INTO `ospos_tax_categories` ( `tax_category_id`,`tax_category`, `tax_group_sequence` ) VALUES
-  (0, 'Standard', 10),
-  (1, 'Service', 12),
-  (2, 'Alcohol', 11);
+  (1, 'Standard', 10),
+  (2, 'Service', 12),
+  (3, 'Alcohol', 11);
```
jekkos: same remark here, defining the id isn't necessary if you have auto_increment enabled

database/3.0.2_to_3.1.0.sql
SteveIreland: This would only be an issue if they are on 3.1.0 master and using Sales Tax.

Surely I am the ONLY person on the planet doing that now.  :-)

database/tables.sql
SteveIreland: Normally I agree and key's 2 and 3 are certainly unnecessary.  I could change them to zero and let the system reassign them.  But I need to be sure that key 1 exists, so that is why that must b provided.

However, in an effort to try to handle those who installed the system but aren't using Sales tax, knowing the keys being used for those allowed me to know what records to migrate using the key value.

SteveIreland:
Do you want me to back out the auto_increment effort?

database/tables.sql
jekkos: Ok just a small detail, but I like things to be consistent with how other tables are initialized. Same is done for stock_locations, in fact, we shouldn't realy on ID's to do specific inserts. In that part I insert subsequent records by using a subquery to look for the default stock location

database/3.0.2_to_3.1.0.sql
jekkos: Well most probably users that are on docker will have latest master if they have auto redeploy enabled..

jekkos: But they might not yet be using it indeed.

SteveIreland:
This last change properly fixes that string literal (in the sense that it does what I needed it to do).  I also tested the add, and removal of the exceptions and they seem to be updating the database correctly.

application/models/enums/Rounding_mode.php
```diff
@@ -40,6 +40,7 @@ public static function get_html_rounding_options()
 	{
 		$CI =& get_instance();
 		$CI->load->helper('language');
+		$x = "";
```
daN4cat: please use '' wherever possible because "" forces the php parser to parse the content of the string.

SteveIreland: Yes.  I know that. :-)  But after coding in Java all day sometimes my fingers don't pay attention to me.  Thanks for catching it.

SteveIreland:
Are youse guys waiting on me to do something else before this merge takes place or just seeing how much stress I can handle? ... or just waiting until you have time to revisit this again ... for the umpteenth time.

daN4cat:
@jekkos?

jekkos:
I guess we need this commit to get a functional whole so let's merge it for now.. I'll check db script later on

PR #1421, opened 2017-07-25
Provide support for updating items using CSV Import

SteveIreland:
This modifies the CSV Items Import to allow it to update existing items.  It will first attempt to update by item_number (a.k.a. UPC or barcode).  If a unique item_number is not found then it will attempt to update record based on a unique item description.  If one or the other is not found (or is not unique) then an add is assumed.  There is the possibility that an item could be added that is a third or more duplicate, but you should insure that is not the case before running this.  If the item_number is unique then the item description is not considered.


application/controllers/Items.php
```diff
@@ -658,42 +658,49 @@ public function do_excel_import()
...
 				while(($data = fgetcsv($handle)) !== FALSE)
 				{
 					// XSS file data sanity check
 					$data = $this->xss_clean($data);
-					
+					$i = $i + 1;
+
 					/* haven't touched this so old templates will work, or so I guess... */
-					if(sizeof($data) >= 23)
+					if(sizeof($data) >= 23 && $data[1] != '')
```
jekkos: what does data[1] contain exactly? Might be good to make a local variable there that denotes the field being checked

jekkos: Ok I see it's item name now by looking at subsequent changes

application/controllers/Items.php
```diff
@@ -708,116 +715,259 @@ public function do_excel_import()
...
+							{
+								$item_update_data['custom2'] = $item_data['custom2'];
+								$diff = TRUE;
+							}
+							if($item_data['custom3'] != '' && $item_data['custom3'] != $item_update_data['custom3'])
+							{
+								$item_update_data['custom3'] = $item_data['custom3'];
+								$diff = TRUE;
+							}
+							if($item_data['custom4'] != '' && $item_data['custom4'] != $item_update_data['custom4'])
```
jekkos: Ok why not iterate over these fields in a simple loop, define them in an array first,  DRY principle here

application/controllers/Items.php
```diff
@@ -708,116 +715,259 @@ public function do_excel_import()
...
+				$message = "Added " . $added . ", changed " . $changed . ', errors ' . count($failCodes);
+
+				if (count($failCodes) > 0)
 				{
-					$message = $this->lang->line('items_excel_import_partially_failed') . ' (' . count($failCodes) . '): ' . implode(', ', $failCodes);
-					
-					echo json_encode(array('success' => FALSE, 'message' => $message));
+					$success = FALSE;
+					$message = $this->lang->line('items_excel_import_partially_failed');
+					$message = str_replace("?1",$added,$message);
```
jekkos: Why not use a online printf function that replaces all positional variables at once?

application/language/en/items_lang.php
```diff
@@ -33,8 +33,8 @@
 $lang["items_error_updating_multiple"] = "Error updating items";
 $lang["items_excel_import_failed"] = "Excel import failed";
 $lang["items_excel_import_nodata_wrongformat"] = "Your uploaded file has no data or wrong format";
-$lang["items_excel_import_partially_failed"] = "Most Items imported. But some were not, here is the list";
-$lang["items_excel_import_success"] = "Import of Items successful";
+$lang["items_excel_import_partially_failed"] = "Most Items imported.   Adds-?1, Changes-?2, Errors-?3.  Records in error are";
```
jekkos: Think there was already a search / substitution mechanism in place for language lines.. check MY_Lang for the code behind it. It will allow you to do this string format magic by adding varargs arguments to the lang->line method

SteveIreland: Thanks ... will do.  I did a quick scan for something but didn't spot it.  Now that I know it exists I'll look more closely.

application/controllers/Items.php
SteveIreland: They were already defined in the item_data array so I refactored to use that.

SteveIreland:
This pull request should not be installed or used at this point in time.  It has a couple of issues that I'm working on (in addition to the recommended cleanup by @jekkos)

application/controllers/Items.php
jekkos: Ok but isn't there a generic way to check for differences between these arrays?

application/models/Item.php
```diff
@@ -42,6 +42,51 @@ public function item_number_exists($item_number, $item_id = '')
 	}
 
 	/*
+	 * Determines if an item exists for the given number and only one item has that number
+	 * If not found or more than one record exists then return false
+	 * else return the item data
+	 */
+	public function get_item_by_number($item_number)
```
jekkos: there is already an item_number_exists(), maybe rework that one to return a truthy value (the item itself) and false if item does not exist

application/models/Item.php
```diff
@@ -42,6 +42,51 @@ public function item_number_exists($item_number, $item_id = '')
...
+	}
+
+	/*
+	 * Determines if an item exists for the given name and only one item has that name
+	 * If not found or more than one record exists then return false
+	 * else return the item DATA
+	 */
+	public function get_item_by_name($item_name)
+	{
+		$this->db->from('items');
```
jekkos: same for this one, exists() is already there

application/models/Item.php
daN4cat: changing behaviour of existing functions risks to introduce regressions unless each function call is evaluated to avoid issues.

application/controllers/Items.php
daN4cat: a simple for loop should do

application/models/Item.php
jekkos: It doesn't really change behavior in fact, if the function is result is just checked for true or false (as it is now), then it should just keep working as it was

SteveIreland:
I used a variation of this in a production environment and there are some problems with it which I will fix as soon as time allows (hopefully next weekend).  The first problem is that it doesn't recognize soft deleted records as not being updatable and the second problem is that if there item doesn't have a barcode then the record falls into add mode regardless of the existence of an item with the same description.

daN4cat:
Ok this feature will not make 3.1.0 as we need to cut here.

SteveIreland:
That's not a problem for me.  In fact I would recommend closing this as I think when I rework it for the new attribute feature that @jekkos is working on it will be a much more useful tool.

PR #1431, opened 2017-07-30
Display tax decimals or currency decimals for tax amounts.

SteveIreland:
Display tax decimals or currency decimals for tax amounts depending on if sales tax is enabled.  This needs to take place for the register, receipt, invoice, and quote.

### DIFF
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -64,7 +64,16 @@ function to_currency_no_money($number)
 
 function to_currency_tax($number)
 {
-    return to_decimals($number, 'tax_decimals', \NumberFormatter::CURRENCY);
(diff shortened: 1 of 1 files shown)


daN4cat:
That generic helper function affects reports too. I'm not sure it's a good idea to have this approach. Maybe it's better to have two functions and call the right one where needed?

SteveIreland:
For a customer sales tax environment the sales tax amount collected should be rendered in the currency decimals, so I think this is a necessary thing.  At an item level on the reports the customer sales tax amount really doesn't mean much so rounding to the currency decimal positions should be okay, but if I learn that's not the case then I would either use the other function or create a new function named to_tax_amount() that renders it in the tax decimal amount.

Issue #1243, opened 2017-04-17
Weighing Scale Barcode


WebShells:
Dears, 
Is there anyway so the software will scan auto generated barcodes by weighing scales ? 
Trying to make it fit for a grocery shop... 
Any ideas ? 
Thanks in advance....

jekkos:
Actually I did some of this work in the past for a client on a private branch. If I'm not mistaken then quantity was an optional field in the barcode  encoding which would then be prefilled in the register once scanned in

WebShells:
Well, after scanning the barcode 3 values should be input, 
1) item name set as PLU/item# ( if they scan yougurt 1 kg  for autogenerated barcode 000023XXXQTYXXPRICE)
2) Weight or qty will be input
3) Price that was automatically generated depending on scale weight to be copied into price field...
Any help ? Is this what we are talking about ?

WebShells:
....

jekkos:
I'd have  to dig it up, it's based on a rather old version and will probably need polishing. i'm willing to do this if you  then can send a PR with the polished up feature.

WebShells:
Sure, no problem about that, ill edit the code to fit the current version and will share it back... Just need to touch the end of the code in order to know where to start from...

WebShells:
Jekkos... Any updates regarding this ?

WebShells:
Updates ?

odiea:
Don't hold your breath waiting for this one.

WebShells:
Hello Odiea, why ? already jekkos said he did it before... maybe sharing it or giving an idea how it can be done so i can proceed on my own...

WebShells:
Updates :/ ? 
Who can i talk to for paid development ?

jekkos:
Actually what I did at the time was make a websocket talk to a serial device (in this case a scale) and readout the byte stream in the register... once something changed on the scale then quantity was updated in the register. Also this quantity in barcode sounds familiar to me.. think it was somewhere in that branch as well.

I could try to integrate this functionality in upstream however, I'll have to take a closer look first and will probably expect some donation in return.. the DO reference hasn't given me any credit this month so I'll need to pay for weblate hosting myself now.

WebShells:
Hello Jekkos,

Thanks for your reply, can i have any private messaging details ? email ? am ready for the donations required in order to accomplish this feature. 
Kindly share a way to contact you...
Regards

jekkos:
[...]

PR #1438, opened 2017-08-03
Add function to parse barcode fields

jekkos:
Added a scale barcode parsing functionality that is configurable in by setting a regex for price and quantity variables. Found a nice bootstrap widget to display multi value select (as array) as labels in an input field. This could be reused for other GUI functionality as it dramatically reduces CRUD logic in javascript. 

Also some smaller GUI fixes for glyphicon labels here and there


odiea:
For this non geek. What is this going to do? An example would be nice.

jekkos:
It's in fact related to #1243. It allows you to read in barcodes printed by a scale, having a specific format. IN this case a scale could encode UPC and quantity fields. You have to configure the barcode formatting once and then ospos will fetch the fields and load them into the register after scanning

application/libraries/Barcode_lib.php
```diff
@@ -45,10 +46,31 @@ public function get_barcode_config()
...
+	function parse_barcode_fields(&$quantity, &$item_id_or_number_or_item_kit_or_receipt)
+	{
+		$barcode_formats = json_decode($this->CI->config->item('barcode_formats'));
+		foreach($barcode_formats as $barcode_format)
+		{
+			if (preg_match("/$barcode_format/", $item_id_or_number_or_item_kit_or_receipt, $matches) && sizeof($matches) > 1)
+			{
+				$qtyfirst = strpos('d', $barcode_format) - strpos('w', $barcode_format) < 0;
+				$quantity = $matches[$qtyfirst ? 1 : 2];
+				if (strstr($barcode_format, '02'))
```
jekkos: kind of specific for a certain type of scale here.. according to it's manual having this prefix the qty would be in grams so we divide by 1000.. actually a more generic solution would be adding quantity units so we can convert between them (g's, kg's, liters, whatever)

odiea:
Thanks for the clarification.

application/libraries/Barcode_lib.php
daN4cat: Are we sure this format will not interfere we any other barcode out there? Is this really a different standard for this purpose only?

jekkos: In this case the scale is set in weighing mode and outputs item number + qty as barcode.. but it depends on how the format looks, if none is defined then the code does nothing at all.

jekkos:
Any opinions on whether to merge / not to merge?

application/libraries/Barcode_lib.php
```diff
@@ -45,10 +46,31 @@ public function get_barcode_config()
...
+				{
+					$quantity = $quantity / 1000;
+				}
+				$item_id_or_number_or_item_kit_or_receipt = $matches[$qtyfirst ? 2  : 1];
+				return TRUE;
+			}
+		}
+		$quantity = 1;
+	}
```
daN4cat: there is just a return TRUE and nothing else. the function looks incomplete to me

application/controllers/Receivings.php
```diff
@@ -90,7 +90,8 @@ public function add()
 
 		$mode = $this->receiving_lib->get_mode();
 		$item_id_or_number_or_item_kit_or_receipt = $this->input->post('item');
-		$quantity = ($mode == 'receive' || $mode == 'requisition') ? 1 : -1;
+		$this->barcode_lib->parse_barcode_fields($quantity, $item_id_or_number_or_item_kit_or_receipt);
```
daN4cat: the return value is not checked. also should quantity a returned value instead of passing it by reference?

application/controllers/Sales.php
```diff
@@ -323,10 +323,11 @@ public function add()
 			$discount = $this->config->item('default_sales_discount');
 		}
 
+		$item_id_or_number_or_item_kit_or_receipt = $this->input->post('item');
+		$this->barcode_lib->parse_barcode_fields($quantity, $item_id_or_number_or_item_kit_or_receipt);
```
daN4cat: same comment hete

database/3.0.2_to_3.1.0.sql
```diff
@@ -69,6 +69,7 @@ INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
   ('last_used_invoice_number', '0'),
   ('last_used_quote_number', '0'),
   ('line_sequence', '0'),
+('barcode_formats' '[]'),
```
daN4cat: indentation

daN4cat:
I'm fine for the merge after the review comments are addressed

application/libraries/Barcode_lib.php
jekkos: Once a format match is found, it shoudln't look any furhter. So it expects a user to input his most specific formats first. Once a match is found the function returns

application/controllers/Receivings.php
jekkos: Issue here is that there are two return values, quantity and item reference. I could indeed just return quantity and only set item number by reference, don't know what you prefer? It might be a little cleaner as quantity is always returned

application/libraries/Barcode_lib.php
daN4cat: but after you set quantity = 1 there is no return FALSE nor TRUE or else... this function looks odd to me.

application/controllers/Receivings.php
daN4cat: ok, so at least return a clear boolean if you have two values to return

daN4cat: I removed the boolean return as it is of no use

application/controllers/Sales.php
daN4cat: fixed

daN4cat:
I rebased and fixed various issues:
- sql upgrade script had a mistake
- bootstrap-tags style was not in line with the rest
- jquery version clash so now upgraded to ~2.1.1
- dist files not up to date
- barcode lib function not explicitly public

jekkos:
Regarding jquery upgrade, hope everything still works fine ?

database/3.0.2_to_3.1.0.sql
```diff
@@ -69,8 +69,7 @@ INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
   ('last_used_invoice_number', '0'),
   ('last_used_quote_number', '0'),
   ('line_sequence', '0'),
-('barcode_formats' '[]'),
```
jekkos: You removed this from the upgrade script and then added an empty check in the code?

daN4cat: No no, it's in the script at the end so people that keep up with master just applies the last line of the script.

daN4cat:
so far so good with jquery update to 2

PR #1441, opened 2017-08-04
Cleanup Sales controller

jekkos:
* Some refactoring in sales controller
* Removed quite some duplicate code
* Removed old substitution logic in favor for Token_lib
* Added upgrade logic to get rid of the old suspended sales tables


application/models/Sale.php
```diff
@@ -532,27 +532,14 @@ public function save(&$sale_status, &$items, $customer_id, $employee_id, $commen
 			return -1;
 		}
 
-		if($sale_status == '1')	//suspend sales
```
jekkos: @SteveIreland we might better use constants here instead of denoting meaning of these digits in comments.. any suggestion as where to define these?

application/models/Sale.php
```diff
@@ -1317,5 +1284,96 @@ public function get_suspended_sale_info($sale_id)
...
+	}
+
+	/**
+	 * @param $sale_status
+	 * @param $dinner_table
+	 * @return int
+	 */
+	private function determine_sale_status(&$sale_status, $dinner_table)
+	{
+		if ($sale_status == '1' && $dinner_table > 2)    //not delivery or take away
```
jekkos: @SteveIreland hope this is logically equivalent to what was written there before ;)
Comparing numbers doesn't really tell alot

application/controllers/Sales.php
```diff
@@ -1236,13 +1095,6 @@ public function unsuspend()
 			$this->sale_lib->copy_entire_sale($sale_id);
 			$this->Sale->delete_suspended_sale($sale_id);
 		}
-		else
-		{
-			// This will unsuspended older suspended sales
```
jekkos: @SteveIreland any idea how quote suspend is supposed to work? I found in the Sales controller that it has sale_status = 2. After  suspend it is not showing up in the popup box. Is this intended behavior or a bug in my rework?

daN4cat:
If I'm getting correctly things here the suspended state is now a state of a sale (or potential sale) entry in ospos_sale, indicated by sale_status.

It would look a neat solution, however the drawback I can see is that a suspended sale that is not converted becomes a deleted (or abandoned) sale and therefore the sequence numbering will have a gap. Not sure that is acceptable or not, just concerned for cases where fiscal control might ask justification for numbering gaps.

Thoughts?

application/models/Sale.php
```diff
@@ -1313,9 +1280,49 @@ public function get_suspended_sale_info($sale_id)
...
+
+	/**
+	 * @param $customer_id
+	 * @param $sale_id
+	 * @param $total_amount
+	 * @param $total_amount_used
+	 */
+	private function save_customer_rewards($customer_id, $sale_id, $total_amount, $total_amount_used)
+	{
+		if (!empty($customer_id) && $this->config->item('customer_reward_enable') == TRUE) {
```
daN4cat: if we want to be consistent with the placement of this { it should be in the next line....

application/models/Sale.php
```diff
@@ -1313,9 +1280,49 @@ public function get_suspended_sale_info($sale_id)
...
+			$package_id = $this->Customer->get_info($customer_id)->package_id;
+
+			if (!empty($package_id)) {
```
daN4cat: same comment

application/models/Sale.php
daN4cat: maybe we should use descriptive (in English) strings like we have for "sale", "receiving" and etc., so I read of the database table gives you a clear idea of the status of the entry?

daN4cat:
also I guess the status could be abandoned or deleted too... so we have implemented a sale soft delete

SteveIreland:
I'll need to review all of this over the weekend.

application/models/Sale.php
jekkos: I would just add some define's in dinner_table model here..

application/controllers/Sales.php
SteveIreland: This is working fine (at least according to my expectations). I created a quote.  Clicking on Quote automatically suspends it and the displays the quote document for printing.  It is now available for retrieval via the Suspended sales view.

There are a few changes I'd like to make such as identifying the document type in the suspended sales panel.  But since there would rarely be more than one suspended document per customer and we now have the ability to filter the list by customer I really think it is usable.  Another change I'd like to make is to NOT have the sales data removed from the table when it is "unsuspended" and the document type is a quote.  I carried forward that behavior but I regret it now.  It's good for temporary suspended sales, but needing to Click on Quote again to insure that it is persisted again as a quote is really dicey because it could be lost if the user gets distracted by a customer (those pesky customers) and doesn't return to it.

SteveIreland: Sorry, I was referring to master.   I'll test your changes now.

SteveIreland: As far as the workflow goes your changes seem to be working as expected for me.  While testing your changes I noticed that I lost some of my support for long alternate descriptions on the quote and invoice. But I don't think that has anything to do with your changes so I'm going to revert back to the master branch and check it out.

I'll submit a pull request for any corrections I make.

application/models/Sale.php
SteveIreland: I've always considered attribute values to be part of the data model.  So does it make sense to include those constant definitions in the `Sale.php` model?

application/models/Sale.php
SteveIreland: Just a note regarding the removal of the old suspended sale logic.  The reason I left that in there was to handle a "slow migration" behind the scenes.  Suspended sales could be in the old suspended tables or the existing sales tables.  But new suspended sales were always written to the existing sales tables.  So the next release is where I was going to propose dropping the suspended sales tables.

SteveIreland: Just a little concern here.  Prior to your later changes the key to the dinner_table was fixed for the first few keys (similar to what I was doing with the Tax Category.  Now that you've removed the key from the table preloads do you think it's still safe for us to check for key values when controlling program logic?

application/models/Sale.php
jekkos: Yes agree that would be  the best place to put them.

application/models/Sale.php
jekkos: OK I see what you mean, in this case there's no real reason to do the change gradually as we can migrate the old records right away. I also thought about the tax upgrade and for this one we'll probably clean up old tables in favor of the new ones in the next release?

jekkos: I suppose you're referring to the id removal in the db script? Think it should create the entries with the same IDs as before. 
Actually I don't really like the idea to rely on technical database ID's in application logic. Functionally they don't have any meaning and maybe we should then just disable the functionality if there's no tables defined?

daN4cat:
@jekkos I fixed the barcode sql script regression and I also rebased the branch to latest master

application/models/Sale.php
SteveIreland: Sounds good to me

jekkos:
I've changed magic numbers with defined constants in the models now.. for my part the cleanup is complete, might require some additional testing first. I'll be on hiking holiday coming week so won't have time to edit any code until next weekend..

application/models/reports/Specific_customer.php
```diff
@@ -61,19 +61,19 @@ public function getData(array $inputs)
...
 		}
 		elseif($inputs['sale_type'] == 'all')
 		{
-			$this->db->where('sale_status = 0');
+			$this->db->where('sale_status = ' . COMPLETED);
 		}
 		elseif($inputs['sale_type'] == 'quotes')
 		{
-			$this->db->where('sale_status = 1 and quote_number IS NOT NULL');
+			$this->db->where('sale_status = ' . SUSPENDED . ' and quote_number IS NOT NULL');
```
jekkos: @SteveIreland I came across this situations a couple of times while find / replacing the magic numbers. We check if  we want to see quotes and then query on sale_status = SUSPENDED. Is this correct ? Shouldn't it check for sale_status = 2 (which is QUOTE in the constants)?

SteveIreland: Oops.  I think I failed to remove a non functioning function.  I believe that function that has that comment in it is not used.  I believe I was exploring what it would take to make it work before realizing that I could just leave a quote in a status of 1 or "Suspended".

SteveIreland: What I really want to do is add another field named `sale_type` that identifies it as standard, credit_only, return, quote, work_order, etc.  But that's a proposal for another release. :smile:

daN4cat: @SteveIreland so is this fine like this? Do you think we can merge these changes? Not sure how extensively you tested the branch.

SteveIreland: Yes, it looked good to me.  It wasn't extensively tested by me but I went through all of the basic transactions and didn't find a problem.

daN4cat: @SteveIreland I'll merge and if any issue you'll fix as @jekkos is away for a week (at least). So we keep moving and close this one too.

Issue #1457, opened 2017-08-08
Can't modify qty field when receiving (on row > 1)


i92guboj:
### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [x] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information

- OSPOS version is: 3.1.0
- OSPOS git commit hash is: 4f5ad57
- PHP version is:   5.6.31
- MySQL or MariaDB version is: mysql 5.6.36
- OS and version is:  gentoo 
- WebServer is:  2.4.27

### Expected behaviour

When entering new lines on reception/devolution/requisition, I expect to be able to manually modify the quantities on the quantity field.

### Actual behaviour

I can modify the quantity in the first row. But, after that, in the following rows, whatever quantity I enter is reset to 1 when I press enter.

**Some funny facts:**

- I can edit the first line I entered as many times as I wish, even before entering more rows.
- This affect all three receiving modes, but it doesn't affect the selling department. When selling, I can edit quantities without problems in any row.
- Funniest of all: when I erase the first line I entered, then I can freely edit all quantities in the rest of lines. Erasing any other line won't do. (This is what I've been doing to be able to work with this bug during the last few weeks).

I should truly take some kind of php debug course, some day. Suggestions welcome ;)

I've seen some commits in the past two months that are related to receiving quantities, maybe that's where it all started. Can't be sure.

But, I am fairly sure this worked some months ago, since there's a lot of stuff I deal with in batches of 6, 10, 12, etc. and I used a lot that qty-edit feature.

### Steps to reproduce the issue

1. Go into receiving mode
2. Receive something, modify qty, make sure it works
3. Receive some other line, modify qty, it should go back to 1
[...]

PR #1464, opened 2017-08-12
Fix receiving issue #1457

SteveIreland:
I believe this resolves the issue with updating receivings with the same solution that was used for sales.   I did a quick test and it does the job. 

The fact that it issues a submit and reload of the form for every change isn't a problem for me and has the advantage of showing the result of the change.

I understand that might be a problem for those with limited internet resources and I'd like to propose a configuration option that either provides immediate update or delayed.


odiea:
Works great on this end.

PR #1465, opened 2017-08-12
Update config_lang.php

Altoubi:
trying to add translate


daN4cat:
@Altoubi thank you for the contribution, which I will merge.
Once @jekkos has resynchronised all the languages, please use the instructions here: https://github.com/jekkos/opensourcepos/wiki/Adding-translations

Don't change right now on Weblate, wait that we sync these changes to Weblate and then you can use that tool for the translations.

PR #1508, opened 2017-08-24
Increase acceptable length of custom fields and allow null values

objecttothis:
Currently the maximum length of these custom fields is 25 characters.  It's not too difficult to imagine a case in which this is far too little.  Also, currently none of the fields are allowed to remain null, which for most uses doesn't make sense as every item wont necessarily have all the fields populated.


daN4cat:
What would be the storage space impact of this change?

objecttothis:
Good question.  According to [this article](https://stackoverflow.com/questions/27671572/whats-the-impact-of-changing-the-length-of-a-varchar-in-mysql)

> The storage of the varchar is the bytes of data with the addition of one or two bytes to declare the length of the string. If the maximum length is 255 or less then only 1 byte of length will be added.

This means that if there are 1000 rows of data and every one is filled to the max you are looking at the following storage:
- Current: 25.39MB
- Proposed: 250MB

Now, that's worst case scenario.  Since varchars only actually take the space that data in them contains.  For example, my 2.0.2 database has over 13,000 rows in the items table and the entire items table is only 4.1MB with varchar at 255 length.

daN4cat:
Ok

daN4cat:
@jekkos are you OK with this change? Anyway there will be a migration of custom fields to attributes in the near future

jekkos:
Yes ..so this will only affect 3.1. I have a script ready to remove them /

daN4cat:
Ok, merged

PR #1520, opened 2017-08-24
add missing keys delete dropped key

objecttothis:
added several missing keys to app_config
deleted one key that was removed from the app at some point


objecttothis:
please look at line 323.  I don't know if client_id is added at first login or what, but I added it with an empty set because it looked as if this was populated with a specific number used for reporting statistics.  perhaps I should have populated the value as NULL?

daN4cat:
It's not required to have one added to the list.

daN4cat:
It's better to remove it.

PR #1525, opened 2017-08-25
Adding line to drop location after migrate

objecttothis:


database/database.sql
```diff
@@ -497,7 +497,7 @@ CREATE TABLE `ospos_receivings` (
   `receiving_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `supplier_id` int(10) DEFAULT NULL,
   `employee_id` int(10) NOT NULL DEFAULT '0',
-  `comment` text NOT NULL,
+  `comment` text DEFAULT NULL,
```
daN4cat: please apply these changes to database.sql to tables.sql as that's the seed file used to regenerate database.sql and others.

objecttothis: changes applied there.

PR #1544, opened 2017-09-03
Add Office Menu

SteveIreland:
This pull request will add an office menu behavior to the module navigation menu.

When installed and the database upgrade is applied, by default the following modules will be shoved back to the office menu.

* Employees
* Taxes
* Store Configuration
* Migration

After installation they can be shoved back to the home page or can be flagged to appear on both (accomplished via Employee grants configuration).


daN4cat:
Merged into feature/backoffice and marked 3.2.0. If no 3.1.1 is required I will merge that branch into master   at a later date.

PR #1548, opened 2017-09-03
Work Order Feature

SteveIreland:
This is the work order feature.  It is based on the office menu (a.k.a. feature/backoffice).

The details for the work order system are documented in the wiki.

It's was also installed in a production environment today, so it will be given a pretty good test over the next week.  Any fixes will result in an update to the pull request.


daN4cat:
Merged

PR #1549, opened 2017-09-04
Fix database and table script.

SteveIreland:
This fixes a typo on the database.sql and tables.sql on populating the grants table.  This same issue was also fixed on the work order pull request ... I hope.


SteveIreland:
The fix was trivial.  Is this failure because of how I applied the fix?  I checked out the feature branch, created a new branch from it and submitted the corrected new branch.

PR #1551, opened 2017-09-04
Fix for 3.1.0 migration issue

SteveIreland:
This resolves the migration issue that was reported plus a couple of other issues that were spotted but not yet reported.

### DIFF
--- a/application/migrations/20170502221506_sales_tax_data.php
+++ b/application/migrations/20170502221506_sales_tax_data.php
@@ -8,7 +8,7 @@ class Migration_Sales_Tax_Data extends CI_Migration
 
 		$this->load->library('tax_lib');
 		$this->load->library('sale_lib');
-		$CI =& get_instance();
+
 	}
 
 	public function up()
@@ -37,8 +37,10 @@ class Migration_Sales_Tax_Data extends CI_Migration
 
 	private function upgrade_tax_history_for_sale($sale_id)
 	{
-		$tax_decimals = $this->CI->config->config['tax_decimals'];
-		$tax_included = $this->CI->config->config['tax_included'];
+		$CI =& get_instance();
+		$tax_decimals = $CI->config->config['tax_decimals'];
+		$tax_included = $CI->config->config['tax_included'];
+		$customer_sales_tax_support = $CI->config->config['customer_sales_tax_support'];
 
 		if($tax_included)
 		{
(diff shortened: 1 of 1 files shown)


application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -65,15 +67,20 @@ private function upgrade_tax_history_for_sale($sale_id)
...
 			}
 			$this->update_sales_items_taxes_amount($sale_id, $item['line'], $item['name'], $item['percent'], $tax_type, $item_tax_amount);
-			$this->tax_lib->update_sales_taxes($sales_taxes, $tax_type, $tax_group, $item['percent'], $tax_basis, $item_tax_amount, $tax_group_sequence, '0', $sale_id, $item['name']);
+			$this->tax_lib->update_sales_taxes($sales_taxes, $tax_type, $tax_group, $item['percent'], $tax_basis, $item_tax_amount, $tax_group_sequence, PHP_ROUND_HALF_UP, $sale_id, $item['name']);
 			$tax_group_sequence += 1;
 		}
 
-		$this->tax_lib->apply_invoice_taxing($sales_taxes);
+		// Not sure when this would ever kick in, but this is technically the correct logic.
+		if($customer_sales_tax_support)
```
jekkos: Hm I don't really like that comment, in which situation is this needed / not needed?
Shouldn't we just upgrade the invoices if they are there or otherwise leave the whole alone?

SteveIreland: My head was into making the necessary changes so that the sales_taxes table will be loaded properly.  That statement wasn't being conditioned properly so I added the condition and added the rounding statement and corrected the oversight of this particular migration file when the rounding codes were changed to to be named using the standard PHP codes as the starting point.

After I had done the work and tested it I then realized that this migration isn't going to be run by anyone that already had customer sales tax enabled because no one on 3.0.2 has customer sales tax enabled.  If they want to convert to it then they should do so AFTER the migration.

I added the comment so that someone reviewing the code would ask about it ... leading to some thoughtful discussion.  :-)

But if someone DOES do the migration after converting to customer sales tax then it will at least try to do what it can and do it correctly.  Assuming that someone set up the tax code tables correctly.  Honestly I haven't given that scenario a lot of thought but if someone decides that would be something they would like to do then I'm happy to spend more time on it - think it through - and run some tests for that scenario.

daN4cat: I'm going to test with my scenario. No invoices involved I'm afraid.

daN4cat: Yes, works fine. Many thanks @SteveIreland I merge this fix now.

Issue #1491, opened 2017-08-21
/public/ folder discussion


objecttothis:
### New Feature / Enhancement
Hi everyone.  Being new to the current version of ospos, there are all kinds of new things to get familiar with.  One of the first things is needing to go to the public folder to access the app.  From seeing the sheer number of repeat questions on it, it appears I'm not the only one that is thrown by that one.  From searching the closed issues, this seems to be a feature introduced with composer.  I have a suggestion that would fix this problem and eliminate all the "I get HTTP 403 when I go to http://url_to_ospos/" questions with one line of code.

An index.html with `<meta http-equiv="refresh" content="0; url=./public/" />` in the root would redirect to public and everybody is happy!  Thoughts?

### Installation information

- OSPOS version is: 3.1.0
- OSPOS git commit hash is: 
- PHP version is:   (e.g. 5.5, 5.6, 7.0, 7.1) 5.6.31
- MySQL or MariaDB version is:   (e.g. MySQL 5.5, MySQL 5.6, MySQL 5.7, MariaDB 10.0, MariaDB 10.1, MariaDB 10.2) 5.7
- OS and version is:   (e.g. CentOS 6.9, Ubuntu 16.4, Windows 10) FreeBSD 11.0
- WebServer is:   (e.g. Apache 2.2, Apache 2.4, Nginx 1.12, Nginx 1.13) Apache 2.4
- (If applicable) Installation package for the LAMP/LEMP stack is:   (e.g. WAMP, XAMPP) FAMP

jekkos:
Might be better to handle this on webserver level.. like adding a rule in apache .htaccess

objecttothis:
@jekkos Yes, I looked into using an apache rule but there's a problem.  We would need to add the standard index.html into the root directory then add the redirect rule below to the .htaccess in the root directory.

`Redirect 301 /index.html /public/`

the above code works, but if the user is using a subdirectory, they will need to change the line to 

`Redirect 301 /{subdirectoryname}/index.html  /{subdirectoryname}/public/`

Users needing to change their .htaccess files already creates more issues being created.  That too would be resolved by an installer that could detect and customize based on install location.

A solution that should be more secure than the HTML redirect but not contain any of the potential to need to modify .htaccess files is to do the redirect with php. An index.php file in the root directory with the following two lines of code would do the trick

`header("Location: ./public/",TRUE,301);`
`die();`

Thoughts?

jekkos:
[...]

PR #1552, opened 2017-09-05
Update to #1491

objecttothis:
Two changes for #1491
Check for mod rewrite to be enabled
Correct the code which before was not redirecting on Apache 2.4 and nginx.

Please test.


odiea:
This appears to be working properly. Thanks

Issue #1546, opened 2017-09-03
Release 3.1.1


daN4cat:
I'm afraid we will need to issue a Release 3.1.1.

I found a bug with Giftcard.
When you create a new one and save it, there is an error thrown and the dialog box doesn't close.

daN4cat:
I need to check if there are more similar issues, somehow something went wrong in recent time.
Time to retest everything before seal a 3.1.1.

daN4cat:
Also found out that some en_GB translations for Giftcard are not correct, fixing those too in file and on Weblate to keep them in sync.

odiea:
Which Giftcard setting are you using?  It looks ok using generate in series.

daN4cat:
@jekkos there is something strange with Weblate, it messed up quite some strings and replaces all the "You have successfully deleted {Gifcard, Employee, Customer, etc.}" with the same string all over the different modules. Basically I've got all the came sentence with Giftcard. I'm changing in GitHub but you need to sync back to Weblate and check why this link between strings.

daN4cat:
Create a new card with random generation. I got an error and there is a bug in the code.

odiea:
And you have Gift Card number set to generate random. I do not get this issue.

daN4cat:
Also I just found out that tax calculation with tax included (VAT 20%) doesn't appear to be right in reports, I mean the total at least. Oh well.... this tax story, how much I love it!

daN4cat:
I'm puzzled by the VAT mistake I saw. It might be an issue moving the old database to the new version. So something wrong with my migration?

objecttothis:
Not sure if this is new to 3.1.0 but when changing the currency symbol manually it throws a HTTP 400 error after hitting tab "http://domain/subfolder/public/[object Object]" The type of request is post.  I have tested this on both my server and an external server to eliminate my server as being the culprit.

objecttothis:
I'm also noticing on two different servers that the .htaccess mod_rewrite is not working to forward to the /public/ folder.  In both cases I am using a subdirectory.  I will look into why it's not working and submit a pull request.

daN4cat:
The migrate function is not working correctly with Tax included (VAT) case.

daN4cat:
@objecttothis I've got a problem with an Apache option to control syblink race condition exploits.
Make sure you don't any such thing enabled when testing. I disabled even because that Apache patch is not recommended in any case.

SteveIreland:
[...]

PR #1560, opened 2017-09-07
en-US Translation Corrections #1546

objecttothis:
This translation is complete. Hopefully no errors.


application/language/en-US/common_lang.php
```diff
@@ -62,6 +62,6 @@
 $lang["common_view_recent_sales"] = "View Recent Sales";
 $lang["common_website"] = "website";
 $lang["common_welcome"] = "Welcome";
-$lang["common_welcome_message"] = "Welcome to OSPOS, click a module below to get started!";
-$lang["common_you_are_using_ospos"] = "You are using Open Source Point Of Sale Version";
-$lang["common_zip"] = "Zip";
+$lang["common_welcome_message"] = "Welcome to OSPOS, click a module below to get started.";
+$lang["common_you_are_using_ospos"] = "You are using Open Source Point Of Sale version";
+$lang["common_zip"] = "Postal Code";
```
daN4cat: I was under the impression that US is ZIP not Postal Code. You sure this change is right?

application/language/en-US/customers_lang.php
```diff
@@ -28,10 +28,10 @@
 $lang["customers_mailchimp_member_rating"] = "Rating";
 $lang["customers_mailchimp_status"] = "Status";
 $lang["customers_mailchimp_vip"] = "VIP";
-$lang["customers_max"] = "Max spent";
-$lang["customers_min"] = "Min spent";
+$lang["customers_max"] = "Max. spent";
```
daN4cat: not sure this is correct, I never seen Min and Max with a full stop. Uh?

application/language/en-US/common_lang.php
objecttothis: it's either "ZIP Code" which is specific to the United States Post Office or the more generic term "Postal Code" used also.  We can stick with "ZIP Code" but "Zip" is slang.  The reason I suggested Postal Code is non-native US English speakers would better understand it.  I don't feel strongly about ZIP Code vs Postal Code.

application/language/en-US/customers_lang.php
objecttothis: US English abbreviation rules state that if you abbreviate a word it should be followed by a period.  International should be abbreviated Intl. Address should be abbreviated Addr. Street should be abbreviated St.

SteveIreland:
Obviously an English teacher in a former life.  However min and max have become common words in their own right and no longer require a period.

application/language/en-US/common_lang.php
daN4cat: Not an issue for me, as long as @odiea and @SteveIreland are fine with it I'm fine with it.

application/language/en-US/customers_lang.php
daN4cat: Yes, theoretically bu practically Min and Max are without dot. Anyway I left it like that, but on en-GB there is no dot.

PR #1561, opened 2017-09-08
Add Work Order and Cancellation Reporting

SteveIreland:
This pull request does several things.

1. It adds a filter for work orders and cancellations to reports.
2. It fixes an oversight that return order types weren't being set.
3. It adds the document type to reports so that it's quicker to distinguish between the document types..


SteveIreland:
I tried.  I'm punting on this. 

I was just beginning to feel like I knew what I was doing in keeping my pull-merge-rebase-push under control on the master branch.  This addition of a feature branch seems to have thrown me for a loop.

My preference would be to swing things back to master.

jekkos:
I guess once we have 3.1.1 release this development can go to master. Another option would be moving to more of a git flow way of working in which we could have a develop and release branch.. but that's another topic

application/models/reports/Detailed_sales.php
```diff
@@ -66,7 +67,15 @@ public function getDataBySaleId($sale_id)
 
 	public function getData(array $inputs)
 	{
-		$this->db->select('sale_id,
+		$this->db->select('sale_id, 
+			MAX(CASE 
+			WHEN sale_type = 1 && sale_status = 0 THEN \'' . $this->lang->line('reports_code_invoice') . '\'
```
jekkos: not using constants here?

application/models/reports/Specific_customer.php
```diff
@@ -46,6 +47,14 @@ public function getDataColumns()
 	public function getData(array $inputs)
 	{
 		$this->db->select('sale_id,
+			MAX(CASE 
+			WHEN sale_type = 1 && sale_status = 0 THEN \'' . $this->lang->line('reports_code_invoice') . '\' 
+			WHEN sale_type = 2 && sale_status = 1 THEN \'' . $this->lang->line('reports_code_work_order') . '\' 
+			WHEN sale_type = 3 && sale_status = 1 THEN \'' . $this->lang->line('reports_code_quote') . '\' 
+			WHEN sale_type = 4 && sale_status = 0 THEN \'' . $this->lang->line('reports_code_return') . '\'
```
jekkos: same here

application/models/reports/Detailed_sales.php
daN4cat: Agree wtih @jekkos, please use the already defined constants

daN4cat:
I would release 3.1.1 tomorrow and then merge everything to master and start 3.2.0. Hopefully the few wrong bits are sorted.

application/models/reports/Detailed_sales.php
SteveIreland: Oops.  Yes ... I agree too.  Oversight on my part.  I worked it out in SQL and was focusing on the language remapping the language definitions for the sale type coding and totally blew by the sale_type and sale_status codes.

daN4cat:
Code is merged in feature/backoffice

PR #1567, opened 2017-09-08
Adding Azerbaijani language for 3.2.0

objecttothis:


jekkos:
Ok thanks for the patch, this is indeed the last thing needed to make it work.
I suppose translations is still a work to be done?
In that case I'll wait the merge until 3.1.1 is ready.

daN4cat:
It's better to have the translations together with the change of language.
If this comes by end of the weekend I can merge it in 3.1.1. If not we skip to 3.2.0

objecttothis:
that's why I marked it as being for 3.2.0. it won't be translated by the weekend and I didn't want to hold up 3.1.1

daN4cat:
ok, clear

PR #1571, opened 2017-09-09
Individual languages

objecttothis:
These are the changes necessary for adding the ability to select language at the employee level.  If employee language is not selected it will default to the system language in the App Config.


objecttothis:
Yes, I agree that varchar looks large.  I selected that because that's the key->value size in app_config.  I'm guessing that varchar 48 would be plenty.  What do you think?

objecttothis:
@daN4cat I will look into building a separation between OSPOS user and store customer.  It may be as simple as always looking to the system language to print receipts.

I agree that the language list should be retrieved from one source.  Currently it's hard-coded into the view in AppConfig.php, however even if we weren't implementing multiple languages I would recommend migrating this to a database driven model.  The biggest reason is that when the language is not English, those languages in the list still appear in English.  It really should be done one of two ways (IMO): Either have the list of languages display in the current display language, or have each of the languages display in their own form.  For example American English would be English (United States) and Russian would be Русский. The first option could potentially be a lot of work whereas the second option is one table storing language code, language name and country locale.

objecttothis:
I corrected the name down to 48 (most will take up 24 or less) and the code to 8 (most will take up 2-5).  It's a varchar so it only takes up what is used.

application/helpers/locale_helper.php
```diff
@@ -6,12 +6,34 @@
 
 function current_language_code()
 {
-    return get_instance()->config->item('language_code');
+//Returns the language code of the employee if set or system language code if not
+	if(get_instance()->Employee->is_logged_in())
+	{
+		//$employee_language_code = 'ru';
```
daN4cat: please remove commented out code

application/helpers/locale_helper.php
```diff
@@ -6,12 +6,34 @@
...
+		$employee_language_code = get_instance()->Employee->get_logged_in_employee_info()->language_code;
+		if($employee_language_code != NULL && $employee_language_code != '')
+			return $employee_language_code;
+		return get_instance()->config->item('language_code');
+	}
+	else
+		return get_instance()->config->item('language_code');
```
daN4cat: we tend to use {} in all cases even if one line

application/helpers/locale_helper.php
```diff
@@ -6,12 +6,34 @@
...
 
 function current_language()
 {
-    return get_instance()->config->item('language');
+//Returns the language of the employee if set or system language if not
+	if(get_instance()->Employee->is_logged_in())
+	{
+		//$employee_language = 'russian';
+		$employee_language = get_instance()->Employee->get_logged_in_employee_info()->language;
+		if($employee_language != NULL && $employee_language != '')
```
daN4cat: same comment here, {} always

application/helpers/locale_helper.php
```diff
@@ -6,12 +6,34 @@
...
+
 }
 
 function current_language()
 {
-    return get_instance()->config->item('language');
+//Returns the language of the employee if set or system language if not
+	if(get_instance()->Employee->is_logged_in())
+	{
+		//$employee_language = 'russian';
```
daN4cat: please remove commented lines from production code

application/helpers/locale_helper.php
```diff
@@ -6,12 +6,34 @@
...
+		$employee_language = get_instance()->Employee->get_logged_in_employee_info()->language;
+		if($employee_language != NULL && $employee_language != '')
+			return $employee_language;
+		return get_instance()->config->item('language');
+	}
+	else
+		return get_instance()->config->item('language');
```
daN4cat: same here

application/views/employees/form.php
```diff
@@ -68,6 +68,37 @@
 						</div>
 					</div>
 				</div>
+				<div class="form-group form-group-sm">
+					<?php echo form_label($this->lang->line('employees_language'), 'language', array('class' => 'control-label col-xs-3')); ?>
+					<div class='col-xs-8'>
+						<div class="input-group">
+							<?php 
+								echo form_dropdown('language', array(
+									':' => 'System Language',
```
daN4cat: We need to have a means to have just one list of languages otherwise we end up maintaining two in two places.

Probably this could be part of some model as it's data that one day could be even added to a config database.

daN4cat:
Please rebase as there is a conflict with the database upgrade script.

I read your comments after my review, agreed. But we need to make sure the separation between the languages is done properly before merging this change.

objecttothis:
@daN4cat is there a way to rebase on the github website?  In essence I created the 3.1.1_to_3.2.0.sql file before it was created in the master, so this is why my version won't automatically merge.

daN4cat:
@objecttothis ok, don't worry I will sort it out when it comes to merge this change. No big deal.

objecttothis:
nevermind, I think I figured it out.

objecttothis:
So, weblate is giving me 404 errors when I try to go in and add the language line needed for this commit.  How do you want me to proceed then?  There is just the line `$lang["employees_language"] = "Language";` in employees_lang.php for English languages and then `$lang["employees_language"] = "";` for all others.  I don't want to submit a pull request for them in github if that is going to mess up weblate.

daN4cat:
No idea, @jekkos should tell us. But at least submit to en-US and I assume Weblate will do the rest.

objecttothis:
submit to en-US on weblate or on github?

daN4cat:
Github

objecttothis:
yes, I confirmed that the invoice and emails get generated in the display language and not the system language.  I will commit changes to separate the display language and system language for emails and printing so that those still get generated in the system language.

objecttothis:
lets also come to some agreement about how we should maintain the language list.  I think everyone is in agreement that we should not maintain two lists.  We could either add it to a model or add it to the database.  I'm in favor of creating a languages table that stores the language name in original language, language code and locale information, but I don't feel incredibly strongly about it.

daN4cat:
Fine with the database table, I assume we add new languages as part of the upgrade script.
Also please adjust the language description as per your suggestion.

jekkos:
Hmm I'm still in the process of finishing the weblate upgrade, let's see how far it is now

jekkos:
Ok think I move all the source translations, @objecttothis can you check if it's working now?

daN4cat:
@jekkos 

Access denied

You're seeing this error because you have DEBUG = True in your Django settings file. Change that to False, and Django will display a standard 404 page.

jekkos:
On which page exactly? Seems to work fine here

objecttothis:
http://weblate.jpeelaer.net/projects/ospos/ I get a 404 error.

objecttothis:
Request Method: | GET
-- | --
http://weblate.jpeelaer.net/projects/ospos/
weblate.trans.views.basic.show_project

Looks like a permissions issue.  When I go to manage watched projects it doesn't show me any to watch or unwatch.

application/views/configs/locale_config.php
```diff
@@ -160,7 +160,7 @@
 						'ar-EG:arabic' => 'Arabic (Egypt)',
 						'az-AZ:azerbaijani' => 'Azerbaijani (Azerbaijan)'	
 					),
-					current_language_code() . ':' . current_language(), array('class' => 'form-control input-sm'));
+					$this->config->item('language_code') . ':' . $this->config->item('langugage'), array('class' => 'form-control input-sm'));
```
daN4cat: wouldn't be better to have common functions for system_language... and employee_language..?

daN4cat: Also the list of languages could be put in the same locale_helper file, if we don't go for the database table in the short term.

application/views/employees/form.php
```diff
@@ -68,6 +68,37 @@
...
+									'th:thai' => 'Thai',
+									'tr:turkish' => 'Turkish',
+									'ru:russian' => 'Russian',
+									'hu-HU:hungarian' => 'Hungarian',
+									'pt-BR:portuguese-brazilian' => 'Portuguese (Brazil)',
+									'hr-HR' => 'Croatian (Croatia)',
+									'ar-EG:arabic' => 'Arabic (Egypt)',
+									'az-AZ:azerbaijani' => 'Azerbaijani (Azerbaijan)'
+									),
+									$person_info->language_code . ':' . $person_info->language, array('class' => 'form-control input-sm'));
```
daN4cat: here I would make employee_current_language()

application/views/configs/locale_config.php
objecttothis: I had thought about that.  I could refactor current_language() and current_language_code() in locale_helper.php to include an optional boolean that when false returns the display language but when someone calls it and passes TRUE to the function it gives the system language rather than the display language.

objecttothis: For now I will put the language list in locale_helper and if that proves to be insufficient then later we can move to a database driven method.

daN4cat: That will do

daN4cat:
@objecttothis you are doing a lot "push & pop" of languages. Isn't there any way to do something at hooks/load_config.php level? That loads the language files, and if I'm right everytime you switch view. So you may want to check if there is a generic way to deal with the problem using that?
For instance if receipts, invoice and quotes have a common base class you can detect that and so switch to system language automatically. It's a run time type identification story. Just a thought...

objecttothis:
Yeah, let me look into it.  I'm probably going to create a function to refactor and consolidate code at the very least.  Something like 
`function load_language($system_language = FALSE, $lang_files array())
{
//if not system language
    for each $lang_files in array
        load employee display _lang
//else
    for each $lang_files in array
        load system display _lang
}`

This will at least reduce things down to one line of code to load system and one line to load employee in each page where we have a receipt, invoice or quote.

I will look into your suggestion of hooks to see if we can automatically load it as well.

application/helpers/locale_helper.php
```diff
@@ -40,6 +40,29 @@ function current_language($load_system_language = FALSE)
 	}
 }
 
+function get_languages()
+{
+	return array(
+		'en-US:english' => 'American English',
```
daN4cat: do you want to change these to English (US) and English (UK)?

application/views/employees/form.php
objecttothis: This is done.  I added a method in locale_helper.php called get_languages() and that is called from both places.

daN4cat:
@objecttothis let me know when this is ready for a merge

application/helpers/locale_helper.php
```diff
@@ -4,14 +4,63 @@
...
+		$employee_language_code = get_instance()->Employee->get_logged_in_employee_info()->language_code;
+		if($employee_language_code != NULL && $employee_language_code != '')
+		{
+			return $employee_language_code;
+		}
+		return get_instance()->config->item('language_code');
+	}
+	else
+	{	
+		return get_instance()->config->item('language_code');
```
daN4cat: these two last returns are the same, it looks to me like the else is superfluous and just a final return at the end of the function will do for the last two returns

application/helpers/locale_helper.php
```diff
@@ -4,14 +4,63 @@
...
+		$employee_language = get_instance()->Employee->get_logged_in_employee_info()->language;
+		if($employee_language != NULL && $employee_language != '')
+		{
+			return $employee_language;
+		}
+		return get_instance()->config->item('language');
+	}
+	else
+	{	
+		return get_instance()->config->item('language');
```
daN4cat: same comment here

objecttothis:
OK, everything is done except your thoughts on the pushing/popping.  Basically, all the pushing/popping is being done at the view level.  Those lang files are pretty much limited to sales, employees, common, customers, receivings, suppliers and items.  The problem is that I can't see a way to do this at the hook level and not have it affect all the other places where those languages are used.  The associated controllers and models do not seem to be specific to receipts, quotes and invoices.  With that said, at the very least I'm going to refactor my code into a single common function to push and pop so that at least we are just introducing one line of code instead of several that are similar.  I'm definitely open to ideas.

application/helpers/locale_helper.php
objecttothis: good point.  i'll modify it

daN4cat:
The "push & pop" for sure gets better having a common function to reduce the maintenance burden of this story.
Let me check the code and see if I can find a way to manage that in the hook.

objecttothis:
ok, gotta be done for the night.  I'll pick it back up in the morning.  let me know if you find anything in the code that might be a good way to automatically load appropriate languages for invoices, quotes and receipts.

daN4cat:
Here what my brain came up with.....

in hook load_config.php you can detect the controller/subdir/function in this way:

```
    $controller = $CI->uri->segment(1);
    $subdir = $CI->uri->segment(2);
    $function = $CI->uri->segment(3);
    
    error_log('hook: ' . $controller . '/' . $subdir . '/' . $function);
```
The hook is called every time you navigate, so you'll see that when in sales/register and you go to receipt or invoice you will get always sales/<something> (what set in controllers/Sales.php).

Now, in order to have a generic story we need to have a subdir concept to sales so you detect that (and sales) and apply the system language always.
That means that all the views related to receipts, quote, invoices, work_order need to move to a subdir in views/sales. I don't know what name we can give, but I'm sure you'll come up with a generic name for that.
Once you have moved to the subdir all those views you need to change controllers/Sales.php to have the subdir in the `$this->load->view` call.
After that in the hook you check segment(1) and segment(2) to be sales and subdir and job done.

All the rest of the cases you can apply the employee languages, so in that hook you can do all the job just calling the function you created with TRUE or FALSE.

objecttothis:
ok I follow. two things. 1- receivings views are also involved in this. That is surmountable. 2- what do you do then about parts of the view which should still be displayed in the employee display language. off the top of my head I only think of the print, email and takings buttons but there may be other. if I understand the method you explained then any views which appear in that controller->subdir->function will entirely be in the system language.

daN4cat:
yep you are right, header, footer and the three buttons will be system language.
Those are removed in the stylesheets for the printing and etc.

On the other hand you can avoid the "pop" functions because the hook will load the correct language.
So you only need to change the "push" functions and leave it as it is without touching the hook.

objecttothis:
OK, we should be good to go for merging this.  I just finished the last of the changes.  Let me know if there's anything else I need to change.

application/views/employees/form.php
objecttothis: done.  It's now stored in locale_helper.php

daN4cat:
@objecttothis I merged the code but taking it, squashing it to one commit, fixed few other bits and rebased.

Please don't use your branch anymore, create a new one taking latest master otherwise history will be messed up and you will get crazy rebasing the current branch.

jekkos:
Ok seems like new weblate has some issues on permissions with the users registered. @daN4cat I have added your account again to the project. Canyou check if it's visible for you?

objecttothis:
@daN4cat Sorry, I want to make sure that I understand correctly how I need to make code modifications.  I don't have write permissions on opensourcepos/opensourcepos so when I go to the master on opensourcepos and go to edit a file which creates a branch, the option it gives me is to edit the file on my fork of the project.  If I rebase my fork of the project before starting a new branch (which I also cannot create from opensourcepos/opensourcepos) by creating a pull request on my fork::master to pull in all the changes from opensourcepos is that enough, or do I need to do something different?

daN4cat:
Don't you have the rights to create a branch in opensourcepos? @jekkos is that correct?

objecttothis:
Nope, not at the moment I don't.  The restrictions don't bother me as trust must be earned, not demanded.  At the same time, if it's creating problems with the workflow, then do as you see fit. :)

daN4cat:
The idea would be to allow you and Steve to create and work on branches at the moment. Jekkos and I will control master, just because we are more familiar with the overall code after few years of reworking it. Still not ideal though.

objecttothis:
Sounds good to me.

PR #1573, opened 2017-09-09
Semi-final touches on Work Orders

SteveIreland:
This pull requests adds the missing language elements.

It also adds a secured ability to delete and reinstate sales.

The security is provided by a new permission which I've added to the end of the database upgrade script. 

It its current manifestation the ability to reinstate a canceled sale is accomplished from the sales reports where the user can list canceled sales.  They select the canceled sale to reinstate and then click on the "Delete" button (think of it as deleting it from the canceled sales report .... and yes, I know that's a stretch).

The Delete button effectively serves as a toggle button.  Ultimately I would like to change the label on the button to Reactivate but in this scenario the button instantiation is embedded deep in the bowels of the JavaScript realm and it will take me a lot longer to accomplish this otherwise trivial task.  I tried it but the water was getting deep and I didn't know how much deeper it was going to get .... if you know what I mean.


daN4cat:
I merged. However I have the following observations:

- I don't see a means to get deleted sales listed. Details sales doesn't show anything even because the drop down menu has Completed Sales & Returns, Completed Sales and Returns (are the Returns completed or no?)
- The delete button on the pop up dialog box should be removed if the user has no rights to cancel the sale

Issue #1531, opened 2017-08-26
Barcode Number required to be unique


objecttothis:
### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [x] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information

- OSPOS version is: 3.1.0
- OSPOS git commit hash is: 
- PHP version is:   (e.g. 5.5, 5.6, 7.0, 7.1) 5.6.31
- MySQL or MariaDB version is:   (e.g. MySQL 5.5, MySQL 5.6, MySQL 5.7, MariaDB 10.0, MariaDB 10.1, MariaDB 10.2) MySQL 5.7
- OS and version is:   (e.g. CentOS 6.9, Ubuntu 16.4, Windows 10) FreeBSD 11.0
- WebServer is:   (e.g. Apache 2.2, Apache 2.4, Nginx 1.12, Nginx 1.13) Apache 2.4
- (If applicable) Installation package for the LAMP/LEMP stack is:   (e.g. WAMP, XAMPP)

### Expected behaviour
Be able to add multiple items with the same barcode number.  For example if we have two of the same books in the store.  One is new and the other is used, so they are at different prices and two different wholesale amounts.

### Actual behaviour
I can't add an item if it has the same barcode number

### Steps to reproduce the issue
[...]

PR #1583, opened 2017-09-12
Allow duplicate barcode

objecttothis:
- Adds a checkbox to config->barcodes to optionally allow duplicate barcodes (default is disabled).
- disables the unique item check when option is enabled.


objecttothis:
sorry @daN4cat, I know this isn't the ideal.  I pulled the latest master from the base before creating this branch so hopefully there are no conflicts.

objecttothis:
This pull request goes with #1531

daN4cat:
Looks fine, but does the database support Barcode duplications? Asking as I can't check this until this eve.

objecttothis:
It's a good thing you asked, because my database upgraded from 2.0.2 is not unique, but the straight install has item_number as a unique, so that means altering the item_table as well.  Since I am running the latest master with a database that has duplicate barcodes, I don't see any major problems created by changing item_number to be a non-unique key. The Primary key in the items table is the item_id so while item_number is a key, it isn't primary.

jekkos:
Hm there's quite some validation logic on top of that unique constraint. I guess we have it for a reason.. it might be better to add a new field as UPC fields are not likely to have duplicates..

You would eg expect to have only one result when scanning a barcode in the sales register. Or did you change the logic in this case to just add everything you find from the query?

objecttothis:
I'm not sure at what point the change was introduced.  I suppose that isn't all that important.  The reason for the change is that not all businesses want only one item-per barcode.  In our case we have used and new books.  Two different books will have the same barcode but be at different retail prices, conditions (a custom field), different quantities and have different wholesale values.  Further some publishers (especially children's chapter books) issue the same barcode on different books which means different title as well, but same barcode.  We have our barcode scanners programmed to not submit after scanning, which means that the barcode will scan and then search suggestions will show all the items with that barcode.  Coupled with my search suggestions format customization that I am working on now, search suggestions will show three customizable fields which will allow the user to pick the one that they want.

I assume that the validation checks are to make sure the barcode is correctly formatted and unique?  My code is keeping the check for unique barcodes in place when the option is not enabled, so it should not break anyone's inventory that doesn't want duplicate barcode numbers.  Are there other validation checks that break the data in a way we don't want when the barcode numbers are duplicated?

objecttothis:
As far as the logic at sale time, I didn't change any of that.  The way the current logic works is that if there is more than one barcode it will not allow you to add it to the register unless you choose one of the items.  If there is only one then pressing enter will automatically accept it to the register.  This is what you'd want anyway unless it was more desireable to have pressing enter just take the first item.

In short, anyone who does not want duplicate barcodes will have their scanners programmed to automatically submit after scan.  Those that want duplicate barcodes will re-program their scanners to not automatically submit (you just scan a specialized barcode and it switches that option in the scanner).

jekkos:
Ok but on the other hand I think it's still useful for other usecases to have a warning in case one tries to enter an item_number that already exists. This check is present in items modal and prevents a user from inputting an item twice. I understand from your usecase that this might not always be required..

objecttothis:
Yeah, I can add a warning.  

Are you thinking of something like this:
warning

only that it would allow you to proceed instead of forcing you to change the barcode number?

daN4cat:
I'm a bit wary about this story. For my case I don't want to allow duplications. The point of the Barcode is to be unique. If you want to differentiate you create a different Barcode of you create an alt to adjust the item at the time of sale. I believe most stories are like that so I'm for a default behaviour to be unique always.  Different story is for extension as an optional.

objecttothis:
@daN4cat The barcode being unique is and will remain the default option.  With my changes, the only way that non-unique barcodes would be allowed is if the user goes into the store config and check's the box to "allow duplicate barcodes".  This changes absolutely nothing for users who want unique barcodes.

objecttothis:
While I agree that most businesses will want unique barcodes, not all do.  Anyone in the book industry won't.  As well as at least some businesses that carry used and new versions of the same product.  The whole reason that I made the extra effort to make this an optional setting with the default requiring that the barcode be unique is because I knew that it currently requires unique barcodes and that many would want to make sure that it stays that way.  All I'm trying to do is to move that decision into the hands of the user to accommodate the businesses (like ours) that depend on the system allowing duplicate barcodes.

odiea:
It sounds to me like a good use case for barcodes. When we had the thrift store we just let the program create the barcodes for us so we never had to face this issue.

objecttothis:
most ean13 barcodes are pre-printed on the product, so it doesnt make a lot of sense to use all of those resources to print new barcodes when they already exist.

daN4cat:
@odiea indeed. But I guess @objecttothis wants to save money/time on printing of barcodes.

So @objecttothis is the duplication of barcodes prevent at entry form? What about csv imports?

odiea:
@objecttothis  I assume you have never been to a thrift store where all items have been donated. When you get home start pulling random stuff out of your kitchen cabinets and clothes drawers and see how many of them have barcodes.

objecttothis:
haha, good point odiea

objecttothis:
@daN4cat if duplicate barcodes are not enabled (default) everything stays the same. duplicate barcodes are not allowed in the entry form or in CSV import. 

If allow duplicate barcodes is enabled then in the entry form it will allow them and in CSV import. @jekkos asked for a warning to be displayed on the entry if the entered barcode is duplicate but just allow it to continue. in CSV import it would be the same thing. I can add a warning for that too. Again, all of this is only if the check box allowing duplicates is enabled.

daN4cat:
Yes, it's all implemented in `item_number_exists`. It seems fine, I questioned a lot because I'm tired this evening and have no time to read all the code.

daN4cat:
I'll merge both PRs tomorrow because I need to squash the long list of changes. It's better to reduce to one line to compress history.

objecttothis:
no worries. it's better to ask questions and be sure than introduce a feature that takes the project away from what we are going for.

no rush o. the merge. I need to add the warnings that @jekkos requested.  Good night.

jekkos:
Ok fine for me if default behavior stays as it is..

daN4cat:
Merged

PR #1584, opened 2017-09-12
Search suggestions formatting

objecttothis:
Allows for the customization of the way that search suggestions appear in the sales register and receivings.  For those that want it, they are able to have up to three columns of information appear.  Currently I have only placed three options in there because it is unclear to me what other people are going to want.  Feel free to request more in the list based on what may be wanted by people.


objecttothis:
This pull request is needed for the allow duplicate barcodes pull request.  Without it, the search suggestions will show multiple barcodes in the register but no further information to know which one you want.

objecttothis:
#1583

daN4cat:
Let me understand this story. Can somebody add any key without any control?
If that's the case we need to sanitise the search key from SQL injection.
Alternative you provide a drop down of possible keys that are the fields in the item structure.
The latter is easier to implement and more under control. However there is the point of custom fields, are those supposed to be part of the story?

objecttothis:
currently it's a drop-down with pre existing options that we control, so I believe it fits your second scenario

objecttothis:
it's modeled after the barcode formatting in config->barcodes

objecttothis:
I can add custom fields to be an option if you think that will be useful.

daN4cat:
Apologies, I didn't read the code properly, it's indeed a drop down. Let's leave it as it is and we can extend it later. I will also check for past issues where people asked to add fields to scan suggestion. Maybe you can look up by yourself, it could be also some close past issue where people customised by themselves (and never contributed of course!).

objecttothis:
yeah, I will do a search and see if there were fields people were asking for

jekkos:
hmm to me this is a dangerous one. I'd rather have attribute integrated soon and then have this search based on that code.. it would make it a lot more generic as sales can have also attributes bound to them which are immutable in time... but I'll need to finish first.
@daN4cat what you think on this part?

daN4cat:
@jekkos yes the attribute story will change this. We should give a good kick and submit to master otherwise we build technical debt here. However we need to clarify category and attribute as different concepts: part of (category), contains (attribute).

objecttothis:
right not the attributes changes are a little abstract so I'll take your word for it. I don't know if this helps at all but this code change doesn't actually change what search suggestions appear or how they are searched. it only changes the label that is used to display the results

jekkos:
But I suppose you'd want to link attribute to a category as well, no? So then you have some kind of self reflection and recursive way or how do you see that

jekkos:
@objecttothis the idea is to have custom attributes linked to an item.. in that case you'd just enable search for attribute values and then resolve a sale through the link table..

objecttothis:
@jekkos if I understand correctly that doesn't make this code obsolete, but if a user wants search suggestions labels to show a customer. attribute then this code would need to be modified... or am I missing something?

jekkos:
Yes indeed it would have been developed on different ground.. queries would be different etc.. so not sure if we should integrate it now and then adapt afterwards

objecttothis:
I'm ok with either. I'd prefer to merge now and adapt but you know better what that will entail. either way I don't mind doing the adaptation work since I have a vested interest in the feature

daN4cat:
As it stands it's fine. Let's not touch the custom fields at the moment so we don't have a problem with attributes.

@jekkos WRT category and attributes, I would say attributes could be associated to a category or not. In my case size, colour and season are disjoint from category (shoes, trousers, shirts). Only size ranges could differ but it's so variable that I would not put any restriction. Also sizes can be US, UK, French, Italian and they differ: 0,1,2,3 or 36,38,40 or XS, S, M and etc.

jekkos:
Ok I agree think I did foresee that usecase but need to check again to make sure..

daN4cat:
Currently the item form doesn't show the attributes if not part of a category.

jekkos:
Hm ok there is no way to add attributes beside through a category, maybe that's what you mean?
Agree I was thinking about adding some neat UI trick to do that in the modal itself.. any suggestion on an add / remove for these fields? Maybe a plus button that then shows a dropdown (that allows to select attribute name).. once that is selected the associated values can be chosen in the input field next to it.
of course deleting attribute links should work from the modal as well

daN4cat:
@jekkos that's the issue. Maybe you need to make a distinction between category (part of) and group of attributes (aggregate group of). If the latter is to simplify the association.
But as it stands it's not right. Also you may want to have category (womenswear) and subcategory (dresses). So the form needs to be dynamic in being recursive with category nest and allow picking up a group of attributes (to simplify the matter). For the first round you can skip the subcategory if too much work on the form.

daN4cat:
Merged

odiea:
Any idea when this is going to be a zip. Nothing new in bintray as far as a stable or unstable and the search suggestions do not appear to be working after doing a merge.

odiea:
application/controllers/config.php public function save_general() does not appear to be saving the search suggestion information.

objecttothis:
let me look into it. it's possible I forgot to commit that change. I'll be home in a bit

odiea:
I will hold off submitting a bug until I hear back from you. I also tried filling in the app config manually and still only the name  shows as an extra search feature..

objecttothis:
Yup, I just forgot to do that commit.  I think I really need to get github setup in eclipse.  @daN4cat @jekkos did we come to some kind of consensus about giving rights for me to create a branch in the master to reduce problems caused by me creating a branch in my fork and then submitting a pull request from that to the master?

PR #1602, opened 2017-09-16
Add support for item kit costing

SteveIreland:
This provides support for being able to specific a cost at an item kit level.  It includes a major refactoring of the logic in order to simplify the code.  Hopefully it is now a little easier to incorporate changes.

A bonus feature is that I added a stock item total to the sales register.  In a sale where there are service charges as well as stock sales, the count only includes stock items.  The information can easily be made available for custom sales receipts and/or invoices if anyone needs it.  But I'm guessing that most sites don't need to filter out service items.

### DIFF
--- a/application/models/Item.php
+++ b/application/models/Item.php
@@ -10,9 +10,18 @@ define('PRINT_ALL', 0);
 define('PRINT_PRICED', 1);
 define('PRINT_KIT', 2);
 
+define('PRINT_YES', 0);
+define('PRINT_NO', 1);
+
 define('PRICE_ALL', 0);
 define('PRICE_KIT', 1);
 define('PRICE_KIT_ITEMS', 2);
+
+define('PRICE_OPTION_ALL', 0);
+define('PRICE_OPTION_KIT', 1);
+define('PRICE_OPTION_KIT_STOCK', 2);
+
+
 /**
  * Item class
  */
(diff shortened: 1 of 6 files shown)


application/models/Item.php
```diff
@@ -393,7 +402,7 @@ public function delete($item_id)
 
 		$this->db->trans_complete();
 
-		$success &= $this->db->trans_status();
+		$success &= $this->db->trans_();
```
daN4cat: not sure I understand why this was changed. trans_status reports the status of the close transaction: https://www.codeigniter.com/userguide3/database/transactions.html

application/models/Item.php
```diff
@@ -428,7 +437,7 @@ public function delete_list($item_ids)
 
 		$this->db->trans_complete();
 
-		$success &= $this->db->trans_status();
```
daN4cat: same question here

SteveIreland:
Good question. It wasn't intentional.  Maybe a mangled rebase. I'll review and correct when I get home. I'm out taking my wife shopping at the moment.

daN4cat:
@SteveIreland ok, so I can merge reverting those two changes. I'll do after dinner.

daN4cat:
Merged. Thanks @SteveIreland for the work and contribution

Issue #1614, opened 2017-09-18
Error on running upgrade script


esbyrt:
### IMPORTANT IMPORTANT IMPORTANT

1. Chose what you want to report between the sections "New Feature / Enhancement" and "Issue / Question / Bug".
2. Remove just the one of the two sections you don't need 
3. Remove this section as it's just a guideline
4. Fill in all the details as requested by the section you chose

If you remove the complete template when submitting a Bug your issue will be closed as we cannot help without basic information.

### New Feature / Enhancement

*For New Features or Enhancements please clearly write your proposal.
It's important to state whether you expect the community to implement it or you will contribute the work.
Please bear in mind that we will implement new features only on the current code, there is no support for older versions.*

### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:
Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

[x] I'm reporting an issue of an unmodified OSPOS installation
[x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
[x] I read the README, WHATS_NEW and UPGRADE
[x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
[x] I read the wiki
[x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
 I'm aware the latest master could be a development version and therefore not stable
 I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report
Installation information

OSPOS version is: 3.1.1
OSPOS git commit hash is:
PHP version is: (e.g. 5.5, 5.6, 7.0, 7.1) 5.5
MySQL or MariaDB version is: (e.g. MySQL 5.5, MySQL 5.6, MySQL 5.7, MariaDB 10.0, MariaDB 10.1, MariaDB 10.2) 10.0.32
OS and version is: (e.g. CentOS 6.9, Ubuntu 16.4, Windows 10) Windows 10
WebServer is: (e.g. Apache 2.2, Apache 2.4, Nginx 1.12, Nginx 1.13) Apache 2.4.27
(If applicable) Installation package for the LAMP/LEMP stack is: (e.g. WAMP, XAMPP)

### Expected behaviour
[...]

PR #1618, opened 2017-09-19
Remove unneeded app_config entry

objecttothis:
this entry is automatically created on first save, so is not needed to manually be entered.


objecttothis:
#1614

PR #1621, opened 2017-09-20
Warning tooltip for allow duplicate barcodes

objecttothis:


application/views/configs/barcode_config.php
```diff
@@ -84,6 +84,11 @@
 						'checked' => $this->config->item('allow_duplicate_barcodes'))); ?>
 				</div>
 			</div>
+			<div class="col-xs-1">
+				<label class="control-label">
+					<a href="http://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes" target="_blank"><span class="glyphicon glyphicon-warning-sign" data-toggle="tooltip" data-placement="right" title="<?php echo $this->lang->line('config_barcode_tooltip'); ?>"></span></a>
```
jekkos: think you need to get rid of the link here

objecttothis: yes, good catch.  I copied the code from another part and forgot to remove the link.

daN4cat:
Guys, I fixed this because I was looking wrong on the UI.

PR #1626, opened 2017-09-20
Office module icon location

objecttothis:


objecttothis:
I'm not sure why all these commits are appearing here.  It should only be 09fb74c and 9f6f7be

PR #1667, opened 2017-10-05
not so secure changes of ceonfig

PudiPudi:
try go to .env later on


PudiPudi:
I am using a new git thing, wasn't supposed to make a request here but on my own branch

Issue #1624, opened 2017-09-20
Office Module Icon Position


objecttothis:
### New Feature / Enhancement

I think the Office Module Icon should appear as the rightmost icon.  I think the reason that it's current position feels a little funny is the following:
- The app config icon formerly appeared on the right end.  
- Since most languages are left to right reading the first thing that people will look at on the top bar is the office icon whereas I don't think that's the first thing we want to draw people's attention to.

Thoughts?

SteveIreland:
That's fine with me.  We only need to change the sort value in the modules table.

SteveIreland:
And there might be a simple code change required in the configuration since I think sort value 1 might be hard-coded.

We would just need to change that to 999

SteveIreland:
I sort value of 0 means that it wlll not display in the menu.  I thought I better point that out.

objecttothis:
ok if no one objects I'll submit the pull request

SteveIreland:
none from me.

daN4cat:
Fine with me

daN4cat:
@objecttothis any plans to work on this?

objecttothis:
Monday morning. My ISP put me behind a NAT so I had to sort that out. it's a quick fix.

objecttothis:
#1671 pull request ready to merge

PR #1671, opened 2017-10-09
Relocate settings module

objecttothis:
My first pull request with added permissions and eGit on eclipse.  Sorry for the messy commit.


objecttothis:
#1624

SteveIreland:
It is a bit messy and hard to follow.  But I don't believe it should be accepted.

It appears to be missing support for enabling and disabling the Office Icon from the Home menu.  If it is then as soon as the Office icon is disabled and re-enabled then it will probably be set back  to a sort value of 1 and would re-appear left aligned.

If I missed the code change then I apologize.  This IS after all, my first code review.  :-)

objecttothis:
Short form is that Github was rejecting the push because it would expose my email address and I had made it private, but eGit wasn't allowing me to resubmit the push, so I reverted the commit and made it again.  In the end it worked but I was expecting it to just push the final commit, and instead it pushed the series of commits and reverts.

With that said, you are correct that I did not take into account what would happen if access is revoked to Office or Messages.  I will fix that as well.

objecttothis:
@SteveIreland Now I'm confused.  If I remove permission for an employee to view the Office Module icon and then re-enable it, the Office Module icon appears in the same place.  Can you clarify what you mean by my changes to the database not offering "support for enabling and disabling the Office Icon from the Home menu"? I might be missing something here, so let me know your thoughts.

SteveIreland:
I'm not referring to permissions.  There is a config option under the general tab that will hid the office icon from the home menu.

The Module.php file is where that action takes place.  Note that $sort is set to 1.  I think you would want to set it to 99.  You might also want to consider using 999 as the value.  That way we don't need to fiddle with the sort values of existing modules and there would always be room to add more modules without needing to adjust the office module sort value since we want it to always appear to the right regardless of what other modules are added.

```
	/**
	 * This method is used to set the show the office navigation icon on the home page
	 * which happens when the sort value is greater than zero
	 */
	public function set_show_office_group($show_office_group)
	{
		if($show_office_group)
		{
			$sort = 1;
		}
		else
		{
			$sort = 0;
		}
		$modules_data = array(
			'sort' => $sort
		);
		$this->db->where('module_id', 'office');
		$this->db->update('modules', $modules_data);
```

objecttothis:
Agreed. I changed the sort order to 999 in Module.php and in the other two files.

application/models/Module.php
```diff
@@ -97,7 +97,7 @@ public function set_show_office_group($show_office_group)
 	{
 		if($show_office_group)
 		{
```
jekkos: If we already set the value through SQL, can't we then get rid of this code?

SteveIreland: I don't understand the question. The code toggles the display of the office icon on and off. It supports a use case where the site doesn't want their store front staff to even be aware that there is a back office menu.   So they can hide it from ever being lncluded in the menu and the only way to access the menu is directly through the URL

Maybe its one feature too far?  It is easy enough to remove. But it was introduced in the original pull request.   I thought I was going to implement it at my site but never got around to it

jekkos: Ok so I see I wasn't aware that we were able to configure this behavior. I guess it's not user specific then.

objecttothis:
So, are these changes approved for merge with the master?

daN4cat:
I need to fix something, let me do the merge once I fixed the database upgrade story.

objecttothis:
sounds good!

daN4cat:
done and in master

Issue #1616, opened 2017-09-19
Uncaught exception


objecttothis:
### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [x] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information

- OSPOS version is: 3.1.1
- OSPOS git commit hash is: 4f5ad57
- PHP version is:   (e.g. 5.5, 5.6, 7.0, 7.1) 5.6.31
- MySQL or MariaDB version is:   (e.g. MySQL 5.5, MySQL 5.6, MySQL 5.7, MariaDB 10.0, MariaDB 10.1, MariaDB 10.2) MySQL 5.7
- OS and version is:   (e.g. CentOS 6.9, Ubuntu 16.4, Windows 10) FreeBSD 11.0
- WebServer is:   (e.g. Apache 2.2, Apache 2.4, Nginx 1.12, Nginx 1.13) Apache 2.4
- (If applicable) Installation package for the LAMP/LEMP stack is:   (e.g. WAMP, XAMPP)

### Expected behaviour
When entering an incorrect format in the price column on the sales register to manually adjust the price of an item ospos would pop up a toast notification explaining that the format of the number entered is incorrect.

### Actual behaviour
When entering a number with incorrect format it gives me a blank screen and the log produces the line:
`ERROR - 2017-09-19 15:47:05 --> Severity: error --> Exception: Number parsing failed /usr/local/www/public_html/pos_alpha/application/helpers/locale_helper.php 192`

### Steps to reproduce the issue
This can very easily happen if your localization settings are expecting a comma but you give it a period.

jekkos:
[...]

PR #1672, opened 2017-10-10
Insert Try->Catch clause (#1616)

objecttothis:
Force an error message on mis-formatted numbers rather than the blank
screen caused by an uncaught exception.

#1616

### DIFF
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -195,7 +195,6 @@ function parse_decimals($number)
     }
     catch (Exception $e)
     {
-    	echo 'Caught exception: ', $e->getMessage(), "\n";
     }
 }
 
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -189,7 +189,14 @@ function parse_decimals($number)
         $fmt->setAttribute(\NumberFormatter::GROUPING_SEPARATOR_SYMBOL, '');
     }
 
-    return $fmt->parse($number);
+    try
+    {
+    	return $fmt->parse($number);
+    }
+    catch (Exception $e)
+    {
+    	echo 'Caught exception: ', $e->getMessage(), "\n";
+    }
 }
 
 /*
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -195,6 +195,7 @@ function parse_decimals($number)
     }
     catch (Exception $e)
     {
+		return FALSE;
     }
 }
 


objecttothis:
Here is an example of the new output when changes are incorrectly formatted.  You can see the red error bar.

capture

application/helpers/locale_helper.php
```diff
@@ -189,7 +189,13 @@ function parse_decimals($number)
...
     }
 
-    return $fmt->parse($number);
+    try
+    {
+    	return $fmt->parse($number);
+    }
+    catch (Exception $e)
+    {
+    }
```
daN4cat: I would say that in the catch statement you should `return $number;` right?

objecttothis: I can do that, but when the exception is thrown, it stops the altering of the field and sticks the red error bar on the screen, so adding `return $number;` doesn't change the outcome unless I'm missing something.

objecttothis: actually, I take that back.  It has a different effect.  Making that modification causes the error to no longer to be displayed and instead it modifies the value and curiously formats in correctly... in essense fixing the formatting, but I don't think that's what we want it to do.  Here's why.  Let's say I take an item that is 0.50 USD but I modify it to 0,45 with a comma.  Adding return $number, corrects the problem, but lets say I type in O.45 with the letter 'O' instead of a zero.  Then it changes the amount to 0.00 USD because it can't parse it.  I think we should leave out the `return $number` so that you get the error message and forces correction of the amount.

daN4cat: I need to check the rest of the code, but returning nothing and expect the calling function to behave correctly doesn't sounds completely right to me. If we want to return FALSE for the case then we need to change the code elsewhere.

jekkos: yes leaving catch clauses empty is certainly no clean coding practice. I would expect the system to stick with the previous value if it can't parse the newly inputted one

objecttothis: I think it would be cleaner to `return FALSE;` and modify the calling code to display the same error if it gets a FALSE return.  If that is acceptable, I can make that change.  @jekkos with the catch block left empty (and if we were to return FALSE the behavior would be to display an error (screenshot above) and keep the original value.

objecttothis:
Here is the code in the Sales Controller.

`		$quantity = parse_decimals($this->input->post('quantity'));
		$discount = parse_decimals($this->input->post('discount'));
		$item_location = $this->input->post('location');

		if($this->form_validation->run() != FALSE)
		{
			$this->sale_lib->edit_item($item_id, $description, $serialnumber, $quantity, $discount, $price);
		}
		else
		{
			$data['error'] = $this->lang->line('sales_error_editing_item');
		}`

I think we just need to return FALSE on the parse_decimals call and it will cause $data['error'] to spit out and use the previous value.

objecttothis:
OK, I tested the code with that change and it appears to do what we are wanting.  The only thing we may want to do is customize the error depending on what the failure is.  Currently the error is fairly generic.  We could potentially return the Exception rather than FALSE and then test for the value being of type Exception.  The only problem with that is that the language of the exception would be English.  My vote is to just return FALSE for now... perhaps that's because I'm trying to triage this against other work that is more important to me.

daN4cat:
Return FALSE will do. No need for more at this stage.

application/helpers/locale_helper.php
```diff
@@ -189,7 +189,14 @@ function parse_decimals($number)
...
     }
 
-    return $fmt->parse($number);
+    try
+    {
+    	return $fmt->parse($number);
+    }
+    catch (Exception $e)
+    {
+		return FALSE;
```
jekkos: Just some minor indentation issue here

objecttothis: hmmm. I'm seeing the issue on your code snippet, but the code in Eclipse and in the github website looks correct.

PR #1686, opened 2017-10-20
Updated Giftcards 81 line

ertanyildiz:
it throws error if there is no gift cart in database. 

*check if row is set. if not, set $max_giftnumber to zero


daN4cat:
Thank you for the contribution!

PR #1717, opened 2017-11-13
improve instructions of ISSUE  template, improve marks, explain instr…

mckaygerhard:
improve instructions of ISSUE  template, improve marks, explain instructions

* improve instructions of ISSUE  template,
* explain instructions better


daN4cat:
The text is already clear enough, certain people just keep ignoring it because they think they can get away with it. What can you do about it?

Issue #1720, opened 2017-11-19
Tax included error when paying with gift card


ackuan:
### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information
*system using latest version of Docker to implement OSPOS local install
- OSPOS version is: 3.1.1
- OSPOS git commit hash is: 
- PHP version is:  
- MySQL or MariaDB version is:   
- OS and version is:   Ubuntu 16.4
- WebServer is:  Apache 2
- (If applicable) Installation package for the LAMP/LEMP stack is:   n/a

### Expected behaviour
We expect for the gift card payment to behave like any other payment type even when "tax included" is checked.

### Actual behaviour
"Tax included" is checked in Sales Configuration section. When implementing a sale with a gift card, the sales tax gets added despite the "tax included" setting. The total is calculated properly, but the "payments total" and the "amount charged" are wrong.
![giftcarderror](https://user-images.githubusercontent.com/12613413/32995481-fec5a506-cd42-11e7-9432-8f932703646f.png)

### Steps to reproduce the issue
Make a gift card $10
Make an item $2
Have store configured with tax included
Charge customer $6 worth of $2 item
Use gift card as payment

jekkos:
Weird behavior.. need to check whether I can reproduce

daN4cat:
@SteveIreland maybe you want to look into this?

SteveIreland:
Yes.  I'll take a look at this.

SteveIreland:
A pull request was submitted for this issue.

daN4cat:
@SteveIreland, merged and thanks for the fix!

daN4cat:
[...]

PR #1723, opened 2017-11-22
Fix gift card amount due in a VAT tax environment.

SteveIreland:
This will hopefully resolve issue #1720


odiea:
This appears to have fixed the giftcard Issue. Thanks @SteveIreland

PR #1737, opened 2017-11-30
Allow receipt item name to be wrapped.

SteveIreland:
I believe this should resolve the issue with the item name on the receipt not wrapping and instead moving content off of the page.

### DIFF
--- a/public/css/ospos_print.css
+++ b/public/css/ospos_print.css
@@ -27,11 +27,6 @@
 		text-align:center;
 	}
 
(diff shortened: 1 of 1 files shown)


SteveIreland:
I should also mention that simply dropping the declaration doesn't get the job done.  It will need to go through the compression phase of the build in order to be incorporated into the compressed stylesheet.

daN4cat:
merged

PR #1742, opened 2017-12-03
Fix complete button not showing when paid in full.

### LINKED ISSUE #1437: GST / VAT inclusive can't complete sale
### IMPORTANT IMPORTANT IMPORTANT

1. Chose what you want to report between the sections "New Feature / Enhacement" and "Issue / Question / Bug".
2. Remove just the one of the two sections you don't need 
3. Remove this section as it's just a guideline
4. Fill in all the details as requested by the section you chose

If you remove the complete template when submitting a Bug your issue will be closed as we cannot help without basic information.

### New Feature / Enhacement

*For New Features or Enhacements please clearly write your proposal.
It's important to state whether you expect the community to implement it or you will contribute the work.
Please bear in mind that we will implement new features only on the current code, there is no support for older versions.*

### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [x] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information

- OSPOS version is: latest master
- PHP version is:   5.6
- MySQL or MariaDB version is:   MySQL 5.5
- OS and version is:   Ubuntu 16.4
- WebServer is:  Apache 2.4

### Expected behaviour
Would be able to complete the sale as there's no amount due (tax included  scenario)

### Actual behaviour
Sale can't be completed for some reason, probably a regression

### Steps to reproduce the issue
schermafdruk van 2017-08-03 12-06-29

SteveIreland:
This should resolve #1437.  I left the code in there that applies a threshold (as a backup to the more simplistic "is the amount paid more than or equal to the amount due"), although I honestly don't understand when it would kick in.  There shouldn't be a rounding issue involved that requires the threshold logic but if there is I'd like to have an example of when it is triggered.

### DIFF
--- a/application/libraries/Sale_lib.php
+++ b/application/libraries/Sale_lib.php
@@ -478,43 +478,29 @@ class Sale_lib
 			$current_due = $amount_due;
 		}
 
+		// 0 decimal -> 1 / 2 = 0.5, 1 decimals -> 0.1 / 2 = 0.05, 2 decimals -> 0.01 / 2 = 0.005
+		$threshold = bcpow(10, -totals_decimals()) / 2;
+
 		if($this->get_mode() == 'return')
 		{
-			if($current_due >= 0)
+			if($current_due >= 0 || $current_due > -$threshold)
 			{
 				$totals['payments_cover_total'] = TRUE;
 			}
 			else
 			{
-				// 0 decimal -> 1 / 2 = 0.5, 1 decimals -> 0.1 / 2 = 0.05, 2 decimals -> 0.01 / 2 = 0.005
-				$threshold = bcpow(10, -totals_decimals()) / 2;
-				if($current_due > -$threshold)
-				{
-					$totals['payments_cover_total'] = TRUE;
-				}
-				else
-				{
-					$totals['payments_cover_total'] = FALSE;
-				}
+				$totals['payments_cover_total'] = FALSE;
 			}
 		}
 		else
 		{
-			if($current_due <= 0)
+			if($current_due <= 0 || $current_due < $threshold)
 			{
 				$totals['payments_cover_total'] = TRUE;
 			}
 			else
 			{
-				$threshold = bcpow(10, -totals_decimals()) / 2;
-				if($current_due < $threshold)
-				{
-					$totals['payments_cover_total'] = TRUE;
-				}
-				else
-				{
-					$totals['payments_cover_total'] = FALSE;
-				}
+				$totals['payments_cover_total'] = FALSE;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


application/libraries/Sale_lib.php
```diff
@@ -478,16 +478,44 @@ public function get_totals()
...
+				else
+				{
+					$totals['payments_cover_total'] = FALSE;
+				}
+			}
 		}
 		else
 		{
-			$totals['payments_cover_total'] = $current_due < $threshold;
+			if($current_due <= 0)
```
daN4cat: I'm afraid I don't see the better logic here unless I'm missing something due to be a sleepy Sunday morning, because what the previous conditional statement said before was:

`if($current_due < $threshold)` -> where `$threshold = {0.5, 0.05, 0.005}` according to the configured `totals_decimals = {0, 1, 2}`.

changing it to `if($current_due < 0)` makes it in my opinion even less true.

The issue here is that `$current_due` is `= 0.01` in the GST defect scenario. So no matter one solution or another the source of the problem seems to me the due having +1 penny due to some rounded sum on one side and sum of rounding on another (my wild guess here).

SteveIreland:
I think we might have to agree to disagree on this.  There should NEVER be a .01 difference.  If that is the case then the tax calculation is defective.  Remember, I changed the logic to calculate the amount of tax due and that is now stored at the sale level and the amount due is the rounded tax amount plus the actual sales.  So a rounding issue isn't possible. 

When I debugged it before I made a change, the threshold calculation was what was causing the problem.  When I simply compared the amounts then the Complete button started appearing again.

Personally I would prefer to remove the threshold calculation so that if there is an upstream error in calculating the difference between the amount due and the amount paid then I can focus my attention on what is going wrong with calculating the amount due.

SteveIreland:
But regardless of whether it is "better logic" or not, the error that was reported is fixed with this change.

When debugging it the threshold logic is what was causing the problem.

PR #1744, opened 2017-12-07
Fix issue with discount not being rounded at the appropriate point.

SteveIreland:
This should hopefully resolve the final problem reported on issue #1437  with the discount causing the Complete button to not appear.


objecttothis:
@SteveIreland yeah, this was not code that I modified during my adjustments to rounding.  I only looked at the method responsible for always rounding up rather than half up or down.  Am I correct in saying that code was not the source of the problem here?

PR #1748, opened 2017-12-11
Incorrect header in spanish invoice template

i92guboj:
The paper header says "tarjeta de Crédito" which means "credit Card", and it should read "Factura de venta", which means "Sales invoice".


daN4cat:
Thank you for the contribution

Issue #1647, opened 2017-09-27
SalePrice / ProductPrice


crls12opazo:
New Feature / Enhancement
 
                        Dear community,

I propose that when updating the price in the sales chart, this makes the calculation for the quantity. In order to improve the sales module and also that the community can do it.
As is done now with quantity which changes the price according to the result. This can also be implemented in the event that the price of the sale changes.

This would help a lot for those who make small sales and have to type the weight of the products according to the balance. It is easier to enter the price of the sale instead of the grams.
The modification must be only in the sale and this should not modify the selling price of the article.

The formula would be
(SalePrice / ProductPrice) this would result the quantity as a fraction.

Greetings.

SteveIreland:
I'm a bit confused by the request and it's probably due to terminology/language mapping challenges.  So to clarify, do you want to be able to enter the "Total" and have the quantity derived from Total divided by Price?

It sounds simple enough, but I don't think I would want the total amount to be input capable for most of my items.  So maybe this feature is driven by item type or a system configuration option.  I would lean toward an item type of "Weight-based".

crls12opazo:
your suggestion is better

daN4cat:
That requirement doesn't really make sense to me. How do you monitor your inventory if you go by extrapolation? You sell by price instead by real number of pieces?
Unless you mean like a candy shop where you buy $10 and instead of counting you check how many using OSPOS, but I would say use a scale and work by weight which is probably better.

nathanzky:
crls12opazo suggestion is the same as my requirement I would like to test the OSPOS running a petrol gas station let say the customer need to load USD 50. Instead of entering how many liters you enter the amount and it automatically calculates how many liters it is.. Would this be possible?

mckaygerhard:
same for the stores that sale fruits by exampel a client need "20 $ for stranberry" event a "1.2 kg of stranberry"

SteveIreland:
I'd like to take a stab at this to see if a practical solution can be put together for this.

crls12opazo:
thank you!!

SteveIreland:
It's in progress.  

I believe this should be an optional feature called "Price Based Sales".   

If enabled then a new item type of "Price Based" will be available for selection.  

A "Price Based" item will be treated like a "Standard" item except that the option of entering an extended total will be available.  The extended line item total on the register will be input capable (which is currently output only).  

If the extended amount is entered then the system will recalculate the quantity of the item sold -
[...]

PR #1766, opened 2017-12-31
Provide a way to enter a line total and have the quantity sold automatically calculated

SteveIreland:
This pull request addresses the issue reported in #1647


daN4cat:
I fixed few things directly to avoid the review turnaround.

Thanks @SteveIreland for the work and contribution and Happy New Year!

odiea:
@SteveIreland  What happened to this line of code. It appears to be missing in the Register.
```							
<td><?php echo to_currency($item['price']*$item['quantity']-$item['price']*$item['quantity']*$item['discount']/100); ?></td>

**Comment by daN4cat** [comment]
Indeed just noticed that there is an issue in the register now. Just applying a discount the total doesn't change.

**Comment by daN4cat** [comment]
Fixing

**Comment by odiea** [comment]
I think this if should be updated to include if this app config is enabled. Currently the item still allows the change when just this is encluded.
```
if($item['item_type'] == 2)
									{
									echo form_input(array('name'=>'total', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['total']), 'tabindex'=>++$tabindex));
									}

daN4cat:
Fixed and in master

odiea:
@SteveIreland If you get a spare minute can you explain the following. Occasionally I get an undefined error.  ITEM_PRICE_BASED. Models/item.php
```
	public function get_stock_search_suggestions($search, $filters = array('is_deleted' => FALSE, 'search_custom' => FALSE), $unique = FALSE, $limit = 25)
	{
		$suggestions = array();
		$non_kit = array(ITEM, ITEM_PRICE_BASED);

**Comment by daN4cat** [comment]
Is ITEM_PRICE_BASED defined? How can you rrproduce this error?

**Comment by odiea** [comment]
I only get this when I use Receivings. It's possible I had a bad merg but I can't see where this is defined anywhere.
ERROR - 2018-01-31 06:12:56 --> Severity: Notice --> Use of undefined constant ITEM_PRICE_BASED - assumed 'ITEM_PRICE_BASED' C:\xampp\htdocs\featuremaster3\application\models\Item.php 628

**Comment by daN4cat** [comment]
I see it's used in `get_stock_search_suggestions` and `get_kit_search_suggestions`.

@SteveIreland what was the aim there? I can't find it defined....

File: application/models/Item.php [hunk 1]
```diff
@@ -598,11 +600,12 @@ public function get_search_suggestions($search, $filters = array('is_deleted' =>
 	public function get_stock_search_suggestions($search, $filters = array('is_deleted' => FALSE, 'search_custom' => FALSE), $unique = FALSE, $limit = 25)
 	{
 		$suggestions = array();
+		$non_kit = array(ITEM, ITEM_PRICE_BASED);
```
daN4cat: @SteveIreland ITEM_PRICE_BASED is not defined...

daN4cat:
I'm not sure @SteveIreland is still watching this space....

PR #1767, opened 2018-01-02
Provide a way to enter a line total and have the quantity sold automatically calculated.

SteveIreland:
This should result in the behavior last discussed.

### DIFF
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -179,11 +179,11 @@ if(isset($success))
 								<?php
 								if($item['item_type'] == 2)
 								{
-									echo form_input(array('name'=>'total', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['total']), 'tabindex'=>++$tabindex));
+									echo form_input(array('name'=>'discounted_total', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['discounted_total']), 'tabindex'=>++$tabindex));
 								}
 								else
 								{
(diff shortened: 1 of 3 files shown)


application/views/sales/register.php
```diff
@@ -179,11 +179,11 @@
 								<?php
 								if($item['item_type'] == 2)
 								{
-									echo form_input(array('name'=>'total', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['total']), 'tabindex'=>++$tabindex));
+									echo form_input(array('name'=>'discounted_total', 'class'=>'form-control input-sm', 'value'=>to_currency_no_money($item['discounted_total']), 'tabindex'=>++$tabindex));
 								}
 								else
 								{
-									echo to_currency($item['price']*$item['quantity']-$item['price']*$item['quantity']*$item['discount']/100);
```
daN4cat: I'm afraid I cannot get where this is now calculated. Your changes don't show this calculation being moved somewhere else. Or am I missing something?

SteveIreland: Sale_lib.php line 903

$line['discounted_total'] = $this->get_item_total($quantity, $price, $discount, TRUE);

daN4cat: Ok, got the point

SteveIreland:
Okay, I've removed the statements that were attempting to check for a required field on the extended total.  To have properly corrected it I would have needed to add an item info retrieval to the method.  I didn't think it was worth the overhead so I just removed the code.  If the user inexplicably enters a zero amount then the quantity will be set to zero. I think it will be okay.

PR #1777, opened 2018-01-07
More missing language elements and one bad reference to a config value.

SteveIreland:


mckaygerhard:
there's any other working with spanish related languaje files?

PR #1827, opened 2018-02-12
This adds the comment back to quote and corrects a couple of bad DEFINE references

SteveIreland:
I'm a bit "out of practice" so my description on the commit was residual from an earlier commit.  If that's important please let me know and I'll redo it all tomorrow night.


daN4cat:
Thanks

PR #1834, opened 2018-02-13
Update detailed reports

odiea:


application/models/reports/Specific_customer.php
```diff
@@ -67,15 +67,46 @@ public function getData(array $inputs)
 			MAX(payment_type) AS payment_type,
 			MAX(comment) AS comment');
 		$this->db->from('sales_items_temp');
-		$this->db->where('customer_id', $inputs['customer_id']);
+		$this->db->where('customer_id', $inputs['customer_id']);  
+		
+		if($inputs['payment_type'] == 'cash')
+		{
+			$this->db->group_start();
+				$this->db->like('payment_type', $this->lang->line('sales_cash'));
```
jekkos: why exactly do we need the parentheses here and not in the following statements?

application/controllers/Reports.php
```diff
@@ -865,10 +884,20 @@ public function specific_customer($start_date, $end_date, $customer_id, $sale_ty
 		}
 
 		$customer_info = $this->Customer->get_info($customer_id);
+		if(!empty($customer_info->company_name))
+		{
+			$customer_name ='[ '.$customer_info->company_name.' ]';
+		}
+		else
+		{
+		$customer_name = $customer_info->company_name;
```
jekkos: minor indentation issue here

application/models/reports/Specific_customer.php
```diff
@@ -134,7 +166,35 @@ public function getSummaryData(array $inputs)
 		$this->db->select('SUM(subtotal) AS subtotal, SUM(tax) AS tax, SUM(total) AS total, SUM(cost) AS cost, SUM(profit) AS profit');
 		$this->db->from('sales_items_temp');
 		$this->db->where('customer_id', $inputs['customer_id']);
+		if($inputs['payment_type'] == 'cash')
```
jekkos: why not just write as follows
```
if (isset($input['payment_type'])) 
{
  $this->db->like('payment_type', $this->lang->line('sales_'.$input['payment_type']));
}
```

application/models/reports/Specific_customer.php
```diff
@@ -67,15 +67,46 @@ public function getData(array $inputs)
...
+		
+		if($inputs['payment_type'] == 'cash')
+		{
+			$this->db->group_start();
+				$this->db->like('payment_type', $this->lang->line('sales_cash'));
+			$this->db->group_end();
+		}
+		if($inputs['payment_type'] == 'due')
+		{
+			$this->db->like('payment_type', $this->lang->line('sales_due'));
```
jekkos: you could easiliy reuse the key - value pairs from get_payment method here to make this a lot shorter? As follows

```

Comment (jekkos) [reply application/models/reports/Specific_customer.php]: see comment below..

File: application/models/reports/Specific_customer.php (hunk 1 above)
Comment (jekkos) [reply application/models/reports/Specific_customer.php]: Ok I guess it will only be needed if we're able to select multiple payment_types at once, but not sure if that's the case in the input screen?

File: application/models/reports/Specific_customer.php (hunk 4 above)
Comment (odiea) [reply application/models/reports/Specific_customer.php]: Thanks. Great idea. I am still in the learning phase. Good to see you back. My computer is not available now (grand daughter) I will look at this later

**Comment by odiea** [comment]
I have it to use just 1 payment type input.

File: application/models/reports/Specific_customer.php (hunk 3 above)
Comment (odiea) [reply application/models/reports/Specific_customer.php]: I could not get the isset to work on all or invoices so I used the following.
```
if ($inputs['payment_type'] != 'all') 
		{
			$this->db->like('payment_type', $this->lang->line('sales_'.$inputs['payment_type']));
		}
		elseif($inputs['payment_type'] == 'invoices')
		{
			$this->db->where('sale_type', SALE_TYPE_INVOICE);
		}

application/models/reports/Specific_customer.php
odiea: Is there an issue with removing the edit from Discounts?

Issue #1863, opened 2018-02-26
[feat] focus and pre-select ready to change amount item/tendered for easy of use


mckaygerhard:
### Issue / Bug  / Question / New Feature

in the amount_tendered the employee must quickly put the amount, but due currently first need to erase the predefined  and input the client's amount.. that little action to erase the predefined amount take a precious time (must move the mouse, hit the erase key)

please @jekkos , @odiea , @objecttothis  or @daN4cat its a very usefully change, if a pull request will be acepted i'll made it, but at the moment if not, it's here!

this improvement are based on the experience of usage in employees.. 

### 1) at the items amount number in  `register.php` of `application/views/sales`

``` diff
172c172
-echo form_input(array('name'=>'quantity', 'class'=>'form-control input-sm', 'value'=>to_quantity_decimals($item['quantity']), 'tabindex'=>++$tabindex));
436c436
---
+echo form_input(array('name'=>'quantity', 'class'=>'form-control input-sm', 'value'=>to_quantity_decimals($item['quantity']), 'tabindex'=>++$tabindex, 'onClick'=>"this.select();"));
```

### 2 at the amount tendered of the payment

``` diff
436c436
- 									<?php echo form_input(array('name'=>'amount_tendered', 'id'=>'amount_tendered', 'class'=>'form-control input-sm disabled', 'disabled'=>'disabled', 'value'=>'0', 'size'=>'5', 'tabindex'=>++$tabindex)); ?>
---
+ 									<?php echo form_input(array('name'=>'amount_tendered', 'id'=>'amount_tendered', 'class'=>'form-control input-sm disabled', 'disabled'=>'disabled', 'value'=>'0', 'size'=>'5', 'tabindex'=>++$tabindex, 'onClick'=>"this.select();")); ?>
480c480
- 									<?php echo form_input(array('name'=>'amount_tendered', 'id'=>'amount_tendered', 'class'=>'form-control input-sm non-giftcard-input', 'value'=>to_currency_no_money($amount_due), 'size'=>'5', 'tabindex'=>++$tabindex)); ?>
+ 									<?php echo form_input(array('name'=>'amount_tendered', 'id'=>'amount_tendered', 'class'=>'form-control input-sm giftcard-input', 'disabled' => true, 'value'=>to_currency_no_money($amount_due), 'size'=>'5', 'tabindex'=>++$tabindex, 'onClick'=>"this.select()")); ?>

```

dvidic:
[...]

PR #1864, opened 2018-02-26
improve autoselect amounts featured

mckaygerhard:
* improve autoselect amounts as per discuted in #1863 in amount_tendered and item count
* TODO: this can be a option configuration in office, but for now made directly in the code

### DIFF
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -169,7 +169,7 @@ if(isset($success))
 								}
 								else
 								{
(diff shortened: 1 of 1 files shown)


jekkos:
so this change is not usable then after all?

odiea:
It works for me.

mckaygerhard:
hi @jekkos  and @odiea this was already done by @daN4cat recently.. and seems his commit was better

WebShells:
Well, it is actually in some cases better than back spacing couple of times...

daN4cat:
It's in master already

PR #1865, opened 2018-02-26
Total Inv. Retail Price & Quantity

WebShells:
Inventory Summary: Added support for total inventory retail price and quantity for square meters average stock calculation.


application/models/reports/Inventory_summary.php
```diff
@@ -54,11 +54,13 @@ public function getData(array $inputs)
 	 */
 	public function getSummaryData(array $inputs)
 	{
-		$return = array('total_inventory_value' => 0);
+		$return = array('total_inventory_value' => 0, 'total_quantity' => 0, 'total_retail' => 0,);
```
daN4cat: There is a mistake in this line, an extra comma at the end.

WebShells: I missed that :x

odiea: The only issue I see is in the summary amounts. I don't think quantity should show a dollar sign.

Total Inventory Value: $12900.89
Total Quantity: $1148.00
Total Inv. Retail Value: $25168.67

WebShells: @odiea i guess i should add an "if statement" to that part since the summary part in the report is all set to_currency...

jekkos:
@WebShells the PR went into conflict, will need to resolve this first

application/views/reports/tabular.php
```diff
@@ -15,11 +15,17 @@
 <div id="report_summary">
 	<?php
 	foreach($summary_data as $name=>$value)
-	{
+    {
```
daN4cat: could you please fix the indentation?

Use tabs no spaces and align all the code accordingly.

thanks.

application/language/ar-EG/sales_lang.php
```diff
@@ -34,19 +34,19 @@
 $lang["sales_date_required"] = "يجب إدخال تاريخ صحيح.";
 $lang["sales_date_type"] = "التاريخ مطلوب.";
 $lang["sales_debit"] = "بطاقة خصم";
-$lang["sales_delete"] = "اسمح بالمسح";
-$lang["sales_delete_confirmation"] = "هل أنت متأكد أنك تريد حذف عملية البيع هذه؟ لايمكن التراجع بعد الحذف.";
+$lang["sales_delete"] = "مسح";
```
daN4cat: sorry these translations have nothing to do with your changes.
Please just use Weblate for the translations and add here only new strings.

thanks.

application/views/reports/tabular.php
WebShells: Done...

application/language/ar-EG/sales_lang.php
WebShells: Yeah that was just before using weblate... i guess some conflicts will show up due to language...

daN4cat:
In master but please next time don't commit language changes that are not required.
I reverted all of them.

daN4cat:
@WebShells you would need to translate the strings in reports_lang.php using Weblate.

PR #1883, opened 2018-03-14
Multiple attachments

amora-paralleldevs:
I added support for sending several files as attachments.


application/libraries/Email_lib.php
```diff
@@ -45,10 +45,17 @@ public function sendEmail($to, $subject, $message, $attachment = NULL)
 		$email->to($to);
 		$email->subject($subject);
 		$email->message($message);
-		if(!empty($attachment))
-		{
-			$email->attach($attachment);
-		}
+		if(is_array($attachment)){
```
mckaygerhard: this need code identation and some guidelines, good patch but must read wiki for coding> https://github.com/opensourcepos/opensourcepos/wiki/OSPOS-development-index#tech-architecture

daN4cat:
Thank you for the contribution.
What would be the use case to send more than one email  attachment?
Your code just changes a library but does not add any business logic to do more.

application/libraries/Email_lib.php
amora-paralleldevs: Please check the code again. I applied the code indentation according to wiki. Please let me know if I need to change something else.

WebShells:
Missing php tag though ? if am not mistaken...

application/libraries/Email_lib.php
amora-paralleldevs: Please check the code again. I applied the code indentation according to wiki. Please let me know if I need to change something else.

daN4cat:
Thank you for the contribution but I still don't understand what feature this brings in outside changing a library for something more generic.

jekkos:
Hope this is not about installing a backdoor in ospos?

amora-paralleldevs:
The use case for this contribution is:
As an user able to email invoices, I want to send another files along with the invoice document.

In the project I'm working on there is a requirement for sending the invoice as a pdf file and another file. I added this change in the library in the project, which is a fork of opensourcepos. 
If you have a better approach for this kind of change, let me know.

jekkos:
Ok so why not contribute the whole feature instead of just an unusable part of it?

mckaygerhard:
umm @jekkos has a good point! it's true!

amora-paralleldevs:
Yes, you are right. 
Ok, my reason for not including the whole feature is because this feature is custom and private. This feature generates an additional file with same data of an invoice. Also there is a contract I must honor.
With that said, I can not include the whole thing, only the change for allowing multiple attachments.

I should have said this first. I am sorry for not stating this earlier, and I apologize for any inconvenience that I made.

mckaygerhard:
you should be know that this software are open and licence and derivative works must also be?

amora-paralleldevs:
Well, I guess I must close this. Sorry for any inconvenience

PR #1926, opened 2018-04-05
Improve a INSTALL file with more specific info and not vage

mckaygerhard:
WIP ill put later two commit after some correction to the wiki page

but @jekkos as you noted this file are more explained to users, still neeed some links to wiki but i must revise the spell checks


mckaygerhard:
umm well as wish

Issue #1940, opened 2018-04-12
get_found_rows queries causing page load slowdown


Ancient123:
### Background information
- [X] Reporting an issue of an unmodified OSPOS installation
- [X] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [X] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [X] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [X] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [X] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [X] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is:  3.1.1
- PHP version is: 7.1
- MySQL or MariaDB version is: 10.1
- OS and version is: Ubuntu 17.10
- WebServer is: nginx 1.12.1

### Bug 
 get_found_rows function is not performant and greatly slow down page load times in some instances.

I am helping debug a friends deployment of opensourcepos and their item set is very large (56,000 items). I profiled the code and MySQL query logs and saw that the initial query for items is written with a limit statement, but then the get_found_rows call happens to get '$total_rows' and promptly grabs the entire result set (56,000 items).

opensourcepos/application/controllers/Items.php line 61
`$total_rows = $this->Item->get_found_rows($search, $filters);`

This should probably be replaced by a query with a simple `count` statement which would not return the entire result-set and would execute significantly faster when there are large result sets.

Ancient123:
[...]

PR #1941, opened 2018-04-12
Speed up item list by preventing giant select statement

Ancient123:


mckaygerhard:
ok but seems you repeat the query complete, maybe a better solution its to count in the results, by using the query object and then by codeigniter conver to array, then make count in that array.. so respect to #1940 

``` php
$arrayresults = $queryresultet->result_array();
count($arrayresults)
```

@jekkos , @daN4cat i think it's more cheat rather than repeat the query and make overload to the DB engine.. due its more cheap to the procesor make the count in RAM event make it on the engine that not use the RAM at first!

Ancient123:
The old code re-runs the complete query without the limit flags. 
We can't re-use the initial query object because that has a limit on it. The full count number is grabbed (as far as I can tell) to help generate the page count that shows up at the bottom of the page. 

The other option would be grabbing the full result-set every time and iterating over that array. I am pretty sure that is worse performance too. :(

mckaygerhard:
the second its better, get the full array and cut that array, with array_slice($arrayrow, 0, <LIMIT>);, this avoid the limit, and avoid the double query to teh database that still are slow

its the best solution and can be merged into 3.2.0 due its critcal performance @jekkos  and @daN4CAT

Issue #1866, opened 2018-02-28
Mobile VIEW


simonaogstino:
Hello Team,

is a mobile view available ?
if NO , is it possible to make one for those who want to use the system in an easy way on tablets and mobiles.

REgards
Simon

WebShells:
Hello @simon,
Actually the code so far can run on mobile/tablets yet it's not with the best responsiveness maybe it needs some time to release a full responsive theme now...

simonaogstino:
Hello,

Yes , it shows the same as the disktop view.

daN4cat:
It's in the TODO list since a long time waiting a brave developer to take, work on it and contribute.

This is how far we could get with the energy we had.

mckaygerhard:
just modification to the sales view css, no great "TODO" .. just a smart modification:

where the paid part are just below of the items list, making more responsive and usable the interface in the moviles (very few space and the inputs for paid are outside of the screen mobile)

mckaygerhard:
@simonaogstino  can be possible, as i explained, @daN4cat just modification to the sales view css, no great "TODO" .. just a smart modification, the problem jajaj its the very complicate developer tools choosen right? note that i can made it, in the css, and the view can be as:

HERE THE RESULTS IN DESKTOP AND MOBILE, my employees are good and happy:

![osposbelow-mobile](https://user-images.githubusercontent.com/1571724/36792901-45feacaa-1c72-11e8-82a5-cc29102b1c3e.png)

![osposbelow-desktop](https://user-images.githubusercontent.com/1571724/36792907-4a3eb634-1c72-11e8-90e6-cdbcdc68cc8d.png)

of course i not work with that complicated bowser npm nodejs etc etc i just use directly the files more easyle and can use directly the repository

daN4cat:
So we can close it's done, right?

WebShells:
I guess so... At least for the time being..

mckaygerhard:
no we cannot close it, the code and the changes are not doe in repository, and i cannot done that, due i dont know how to work tiht that complicated development setup, well @WebShells as your preferer

seems here its more the attention pay to close issues..

mckaygerhard:
the issue must be renamed to "fields sales in mobile view are outside of the box", and the solution its made the changes i proposed in the pictures! the changes does not affect the themes! ok

the code and the changes are not doe in repository, so how tis can be closed! ? and i cannot done that, due i dont know how to work tiht that complicated development setup

[...]

PR #1942, opened 2018-04-12
Remove disables button when SALE COMPLETE button appears

dvbava:
I think it will make sense to not have disables control at 'sale complete' stage in sales register.

### DIFF
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -407,33 +407,13 @@ if(isset($success))
 				// Show Complete sale button instead of Add Payment if there is no amount due left
 				if($payments_cover_total)
 				{
-				?>
-					<?php echo form_open($controller_name."/add_payment", array('id'=>'add_payment_form', 'class'=>'form-horizontal')); ?>
-						<table class="sales_table_100">
-							<tr>
-								<td><?php echo $this->lang->line('sales_payment');?></td>
-								<td>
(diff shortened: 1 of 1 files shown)


mckaygerhard:
can you explain that @dvbava ?

mckaygerhard:
i mean, some users are not so smart and need to see the "behaviour" of the buttons i think!

dvbava:
The button itself changing to "Complete" from "Add payments", which is clear indication that it's good to proceed further.
For me personally, the monitor is 15" (1366x768 resolution), so saving space is equally important.
Once you are use to it, you will have no issue. however you have saved some space on register for long time.

daN4cat:
The problem is that if somebody has a touch screen the UI shifts and the user will tend to push the wrong screen location.
That's why the disabled button is not removed to keep the layout and button position consistent.

daN4cat:
In any case this will not merge in 3.2.0 as I want to release it this weekend.

dvbava:
I see your point regarding touch scree. all right.

mckaygerhard:
hi @dvbava if your space its the issue there's others options, try this: https://github.com/opensourcepos/opensourcepos/issues/1866#issuecomment-370199504 maybe gives you ideas, its a good hack

dvbava:
@mckaygerhard, actually i have done some changes to https://github.com/dvbava/opensourcepos/tree/dev  and built it..works like magic!

mckaygerhard:
please post/share some photos  here for more ideas! glad to see it helps to you @dvbava credits to @nerv55  that makes the hack, i'll try to document that in the wiki later days

Issue #1958, opened 2018-04-25
#1924 Needs to be reopened


SteveIreland:
I was trying to test an invoice and found that the invoice was broken as a result of a change made for #1924.

I have an idea as to how to accomplish the same objective but I'll do that later this week.

daN4cat:
@jekkos to comment

jekkos:
@SteveIreland what exactly is broken then? Only thing I added is show discount whether we applied it or not.

odiea:
Should the columns have some kind of filler or column size change so the displayed columns stay to the right right ?

SteveIreland:
Good question, we don't specify column widths so the table will either be narrower or the other columns will be automatically expanded to fit the width.  The later usually happens when the width is set to or inherited as 100%.  I'm not yet sure what it is, but I'll check on it when I'm testing it later this evening.

What would be your preference?

odiea:
None really . I would like to see the invoice before and after the item[discount] issue is fixed. Like you stated no other adjustments may be necessary

odiea:
@SteveIreland here is the original issue #1824

SteveIreland:
Thanks @odiea I see what's going on now.   Hopefully I can submit a fix for it tonight.

SteveIreland:
If there is no discount on the sale then the discounted unit price column will not be rendered..

daN4cat:
@SteveIreland have you checked if quote and work order have the same issue/behaviour?

SteveIreland:
I have now.  The  revised pull request includes a fix for quotes.  Work orders did not need it.  In most small shops the owner does not want the prices included on the work orders.

daN4cat:
In master

jekkos:
@SteveIreland thanks for ironing this out.

PR #1961, opened 2018-04-27
Fix rendering of discounted unit price on invoice.

### LINKED ISSUE #1924: Discounted item info on invoice / quote
Issue linked to #1824. Yesterday I made an invoice using ospos and found a new column showing discount / item in the item table. This column also shows when there is no discount. I think it should not be the case as it confused me as it seemed that there was a discount.

SteveIreland:
This should close #1924 and #1958.

### DIFF
--- a/application/views/sales/invoice_email.php
+++ b/application/views/sales/invoice_email.php
@@ -72,9 +72,16 @@
 			<th><?php echo $this->lang->line('sales_quantity'); ?></th>
 			<th><?php echo $this->lang->line('sales_price'); ?></th>
 			<th><?php echo $this->lang->line('sales_discount'); ?></th>
-			<?php if ($item['discount'] > 0): ?>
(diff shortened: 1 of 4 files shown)


application/views/sales/invoice_email.php
```diff
@@ -72,9 +72,16 @@
 			<th><?php echo $this->lang->line('sales_quantity'); ?></th>
 			<th><?php echo $this->lang->line('sales_price'); ?></th>
 			<th><?php echo $this->lang->line('sales_discount'); ?></th>
-			<?php if ($item['discount'] > 0): ?>
-			<th><?php echo $this->lang->line('sales_customer_discount');?></th>
-			<?php endif; ?>
+			<?php
+			$invoice_columns = 6;
```
jekkos: It would have been more dynamic to fetch the number of columns in javascript and do adjustments there based on a class or label attached to each column.. but I guess this does the job

Issue #1966, opened 2018-04-28
Checkbox in settings for skipping Sales Complete page


kosimoff:
Please somebody add checkbox in setting for skipping Sales Complete page
![capture](https://user-images.githubusercontent.com/14353101/39392606-4c52dd5e-4ad2-11e8-82a3-42f905c6c686.PNG)

daN4cat:
FIY that's the page you print, not sure how you could skip.

daN4cat:
FIY that's the page you print, not sure how you could skip.

WebShells:
Well I wasn't sure how could I reply to that lol... Jeeeeeesssss!

kosimoff:
Right after print command is sent to printer, go to Sales Register page again, instead of staying in Sales Complete page.

WebShells:
Or set windows header :x

daN4cat:
@kosimoff feel free to contribute the feature.

kosimoff:
If I were a programmer

daN4cat:
If we had all the time of this world :-)

WebShells:
@daN4cat we can simply add header or windows location on receipt page :/

daN4cat:
Not sure I get what you mean... today I'm sleepy and slow so bear with me.

mckaygerhard:
must take in consideration most scenarios, not only those wehre "windows.locations" works..

WebShells:
@mckaygerhard in what cases it might not work :) ? if you can inform me please...

WebShells:
@daN4cat i meant, cant we add windows.location and point it to sales, so once printing page reached if "Print after sale automatically is selected" after reaching print page for Kiosk or non-kiosk users will press enter ( to print ) or escape ( to not ) and the header will redirect the user to sales again.

daN4cat:
I remember having looked in to something like that but I had an issue. I can't remember what was.
But I should look into this option again. Also there is the receipt re-print option to sort out.

mckaygerhard:
umm i doub will work in mostly cases, seems for mobiles will cause some problems, i try to develop a mobile agnostic client app that uses webview and seemms here are not know waht i taking about.. 

webview embebs the client rendering and take output (talking in easy words) and the window location may cause confusions.. for mobile printers..

mobile rpinting are a great fetaure that will be lost if not work, and i not have today a mobile device to test, i must waith until July for that (i not living in a country that today can wastes resources )

daN4cat:
[...]

PR #1968, opened 2018-04-29
Skip receipt windows

WebShells:
Skip receipt windows and redirect to sales if print after sale is selected. 
Can partially solve issue #1966

### DIFF
--- a/application/views/partial/print_receipt.php
+++ b/application/views/partial/print_receipt.php
@@ -70,7 +70,7 @@ if($print_after_sale)
 	{
 	   // executes when complete page is fully loaded, including all frames, objects and images
 	   printdoc(); 
-	   window.location.replace('../public/sales');
+	   window.location.href = "<?php echo site_url('sales'); ?>";
 	}); 
 <?php
 }
--- a/application/views/partial/print_receipt.php
+++ b/application/views/partial/print_receipt.php
@@ -69,7 +69,8 @@ if($print_after_sale)
 	$(window).load(function() 
 	{
 	   // executes when complete page is fully loaded, including all frames, objects and images
-	   printdoc();
+	   printdoc(); 
+	   window.location.replace('../public/sales');
 	}); 
 <?php
 }


mckaygerhard:
do you test in a mobile device? the rule it's at the client side! not a server-made instruction

mckaygerhard:
do you test in a mobile device, in different situations browser? the rule it's at the client side! not a server-made instruction so could be not fully reponsive

WebShells:
@mckaygerhard its for the people who would like to use it without waiting aconfigurable options... It doesnt function unless printer settings were always on...

mckaygerhard:
umm well @webshells i can print in konqueror and also in mobile devices to bluethood printers! will work here?

by the way was another feature that's was not taking in consideration the mobile print

WebShells:
@mckaygerhard you print directly from browser ? through which browser you are using it ?

mckaygerhard:
well not directly, i use a webview.. its like a browser, the application runs embebed!

WebShells:
So it prints as an image (screenshot), not through JS... right ?

mckaygerhard:
noT, from where do you make that idea?

application/views/partial/print_receipt.php
```diff
@@ -69,7 +69,8 @@ function printdoc()
 	$(window).load(function() 
 	{
 	   // executes when complete page is fully loaded, including all frames, objects and images
-	   printdoc();
+	   printdoc(); 
+	   window.location.replace('../public/sales');
```
jekkos: might be better to write the url from PHP here using site_url instead of hardcoding the prefix. Instlalations could be setup in different ways and doing it like this might not always work

WebShells: You're right... fixed

jekkos:
Ok thanks for the fix, but now we will always redirect, which might not be desirable?

WebShells:
@jekkos once print after sale is selected, am just sharing it for who might like to use it

daN4cat:
The problem with this is that if you click a receipt from Transactions or Report Transactions it will switch to Sales, which is not right.

WebShells:
@daN4cat you meant in old-takings or detailed transaction ? well no it doesnt as i tested :x can you confirm please ? "Note: its added only if print after sale is enabled"

odiea:
@daN4cat I tested this and the only time it switches back to Sales is in Sales.

WebShells:
Thanks for your time @odiea ...

mckaygerhard:
i not tested in my app mobile .. so well as you wish guys

odiea:
My personal point of view is I wish the receipt would print in the background and not be seen at all unless a reprint was required.

mckaygerhard:
umm if the user do not see something in screen will sayd "does nothing" well maybe a config option may be need

daN4cat:
@WebShells you are right, I just look at the change itself without expanding the view. It's indeed within the print_after_sale conditional.

Would you mind if I add a small delay between print and jump back? This would allow people to print a second time if required.

Also I need to add the option to remove the button Sale Register Transaction if the receipt is viewed from Reports Transaction.

WebShells:
@daN4cat not at all... Whatever you see more suitable man...

WebShells:
@odiea did you try kiosk option :x? It still shows a pop-up but it might help?

daN4cat:
This is in master now.

PR #1975, opened 2018-05-06
Add support for temporary items.

SteveIreland:
This is the "temporary item" pull request.  I started to work on the ajax update but had a bit of difficulty with getting a proper json encoded response back so I abandoned that effort and will come back and revisit that after I submit a pull request to support multi-package items.

### DIFF
--- a/application/controllers/Items.php
+++ b/application/controllers/Items.php
@@ -24,7 +24,8 @@ class Items extends Secure_Controller
 			'is_serialized' => $this->lang->line('items_serialized_items'),
 			'no_description' => $this->lang->line('items_no_description_items'),
 			'search_custom' => $this->lang->line('items_search_custom_items'),
-			'is_deleted' => $this->lang->line('items_is_deleted'));
+			'is_deleted' => $this->lang->line('items_is_deleted'),
+			'temporary' => $this->lang->line('items_temp'));
 
 		$this->load->view('items/manage', $data);
 	}
@@ -50,7 +51,8 @@ class Items extends Secure_Controller
 						'is_serialized' => FALSE,
 						'no_description' => FALSE,
 						'search_custom' => FALSE,
-						'is_deleted' => FALSE);
+						'is_deleted' => FALSE,
+						'temporary' => FALSE);
 		
 		// check if any filter is set in the multiselect dropdown
 		$filledup = array_fill_keys($this->input->get('filters'), TRUE);
@@ -177,6 +179,8 @@ class Items extends Secure_Controller
 
 	public function view($item_id = -1)
 	{
+		$data['allow_temp_items'] = $this->session->userdata('allow_temp_items');
+
 		$data['item_tax_info'] = $this->xss_clean($this->Item_taxes->get_info($item_id));
 		$data['default_tax_1_rate'] = '';
 		$data['default_tax_2_rate'] = '';
@@ -392,6 +396,13 @@ class Items extends Secure_Controller
 			'custom10' => $this->input->post('custom10') == NULL ? '' : $this->input->post('custom10')
 		);
 
+		if($item_data['item_type'] == ITEM_TEMP)
+		{
+			$item_data['stock_type'] = HAS_NO_STOCK;
+			$item_data['receiving_quantity'] = 0;
+			$item_data['reorder_level'] = 0;
+		}
+
 		$x = $this->input->post('tax_category_id');
 		if(!isset($x))
 		{
@@ -412,8 +423,7 @@ class Items extends Secure_Controller
 		}
 		
 		$employee_id = $this->Employee->get_logged_in_employee_info()->person_id;
-		$cur_item_info = $this->Item->get_info($item_id);
-		
+
 		if($this->Item->save($item_data, $item_id))
 		{
 			$success = TRUE;
@@ -424,7 +434,7 @@ class Items extends Secure_Controller
 				$item_id = $item_data['item_id'];
 				$new_item = TRUE;
 			}
-			
+
 			$items_taxes_data = array();
 			$tax_names = $this->input->post('tax_names');
 			$tax_percents = $this->input->post('tax_percents');
@@ -444,9 +454,15 @@ class Items extends Secure_Controller
 			foreach($stock_locations as $location)
 			{
 				$updated_quantity = parse_decimals($this->input->post('quantity_' . $location['location_id']));
+				if($item_data['item_type'] == ITEM_TEMP)
+				{
+					$updated_quantity = 0;
+				}
 				$location_detail = array('item_id' => $item_id,
 										'location_id' => $location['location_id'],
 										'quantity' => $updated_quantity);
+
+
 				$item_quantity = $this->Item_quantity->get_item_quantity($item_id, $location['location_id']);
 				if($item_quantity->quantity != $updated_quantity || $new_item)
 				{
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -19,6 +19,8 @@ class Sales extends Secure_Controller
 
 	public function index()
 	{
+		$this->session->set_userdata('allow_temp_items', 1);
+
 		$this->_reload();
 	}
 
@@ -1274,6 +1276,10 @@ class Sales extends Secure_Controller
 				$this->session->set_userdata('sale_id', -1);
 			}
 		}
+		else
+		{
+			$this->sale_lib->remove_temp_items();
+		}
 
 		$this->sale_lib->clear_all();
 		$this->_reload();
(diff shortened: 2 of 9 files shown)


daN4cat:
I will not merge this until we release 3.2.1.
This feature is for 3.3.0 and I prefer not to create branches otherwise it gets complicated in constantly rebasing and fixing merging conflicts. It's better to keep one master code line and hold PRs until the time is right.

Hope not to disappoint.

application/controllers/Sales.php
```diff
@@ -1387,6 +1389,47 @@ public function get_filtered($cart)
...
+		$item_id = $this->input->post('item_id');
+		$description = $this->input->post('item_description');
+		$this->Item->update_item_description($item_id, $description);
+		$cart = $this->sale_lib->get_cart();
+		$x = $this->searchCartForItemId($item_id, $cart);
+		$cart[$x]['description'] = $description;
+		$this->sale_lib->set_cart($cart);
+	}
+
+	function searchCartForItemId($id, $array) {
```
daN4cat: Guess what here? :-)

SteveIreland:
The corrections to conform to standards have been completed.

jekkos:
Then let's first get 3.2.1 out of the door.

application/controllers/Sales.php
```diff
@@ -1387,6 +1389,58 @@ public function get_filtered($cart)
 
 		return $filtered_cart;
 	}
-}
 
+	public function changeItemNumber()
```
jekkos: using camelCase here instead of snake_case?

application/controllers/Sales.php
```diff
@@ -1387,6 +1389,58 @@ public function get_filtered($cart)
...
+		if($x != NULL)
+		{
+			$cart[$x]['description'] = $description;
+		}
+		$this->sale_lib->set_cart($cart);
+	}
+
+	function searchCartForItemId($id, $array)
+	{
+		foreach($array as $key => $val)
```
jekkos: isn't there a php builtin to do this kind of array operation? Maybe check [this function](http://php.net/manual/en/function.array-search.php)

application/models/Item.php
```diff
@@ -341,6 +351,26 @@ public function get_info_by_id_or_number($item_id)
...
 	}
 
+	/*
+	Gets information about a particular item by item id or number
+	*/
+	public function get_info_by_id($item_id)
+	{
+		$this->db->from('items');
+
+		$this->db->or_where('items.item_id', (int) $item_id);
```
jekkos: according to the comment you look for item id or number.. but here I  find a logical disjunction for item_id and deleted = 0?

application/views/sales/register.php
```diff
@@ -360,12 +399,12 @@
...
 			<tr>
-				<th style="width: 55%;"><?php echo $this->lang->line('sales_quantity_of_items',$item_count); ?></th>
-				<th style="width: 45%; text-align: right;"><?php echo $total_units; ?></th>
+				<th style="width: 55%;" id=""item_count"><?php echo $this->lang->line('sales_quantity_of_items',$item_count); ?></th>
+				<th style="width: 45%; text-align: right;" id="total_units"><?php echo $total_units; ?></th>
 			</tr>
 			<tr>
 				<th style="width: 55%;"><?php echo $this->lang->line('sales_sub_total'); ?></th>
-				<th style="width: 45%; text-align: right;"><?php echo to_currency($subtotal); ?></th>
+				<th style="width: 45%; text-align: right;" id=""subtotal><?php echo to_currency($subtotal); ?></th>
```
jekkos: misplaced quote here

application/views/sales/register.php
```diff
@@ -360,12 +399,12 @@
 
 		<table class="sales_table_100" id="sale_totals">
 			<tr>
-				<th style="width: 55%;"><?php echo $this->lang->line('sales_quantity_of_items',$item_count); ?></th>
-				<th style="width: 45%; text-align: right;"><?php echo $total_units; ?></th>
+				<th style="width: 55%;" id=""item_count"><?php echo $this->lang->line('sales_quantity_of_items',$item_count); ?></th>
```
jekkos: one quote too much in the id attribute?

application/views/sales/register.php
```diff
@@ -601,7 +640,53 @@
...
+
+	$("input[name='item_number']").change(function(){
+		var item_id = $(this).parents("tr").find("input[name='item_id']").val();
+		var item_number = $(this).val();
+		$.ajax({
+			url: "<?php echo site_url('sales/changeItemNumber');?>",
+			method: 'post',
+			data: $.extend(csrf_form_base(),
+				{
+					"item_id" : item_id,
```
jekkos: OK here we have three endpoint to change each attribute separately. Wasn't there a way to condense those in a single generic one?

application/controllers/Items.php
```diff
@@ -392,6 +396,14 @@ public function save($item_id = -1)
 			'custom10' => $this->input->post('custom10') == NULL ? '' : $this->input->post('custom10')
 		);
 
+		if($item_data['item_type'] == ITEM_TEMP)
+		{
+			$item_data['stock_type'] = HAS_NO_STOCK;
+			$item_data['receiving_quantity'] = 0;
```
jekkos: does it make sense to keep this line if you overwrite in the next?

odiea:
@SteveIreland  I downloaded from your site and after hitting complete the item does not show on the receipt just the totals

dvidic:
Hello everyone. Is this feature usable or needs some work?

application/controllers/Sales.php
SteveIreland: Sorry ... brain fart on the standards switchover.  I have been doing a lot of Java development and apparently my brain simply refused that day to  adapt.

application/controllers/Sales.php
SteveIreland: I actually considered it and I think I've used it a couple of times in the past, but the requirement was so simple here and when I looked at the extensive comments in the PHP documentation for this function I was discouraged from using it.  Anything that needs that much commentary must have some problems either in implementation or ease of use.  The final nail was when I read something about a performance issue when implemented in a particular way.

application/views/sales/register.php
SteveIreland: Oops, I reverted some code that I had started to add in order to explore ajax update  for quantity, price and discount ... this should have been reverted.

application/views/sales/register.php
SteveIreland: It's the choice I made.  I wanted the item and the cart line to be updated as soon as the user left the field.  I also didn't want to lose cursor position every time the field loses focus so I avoided the auto submit on the line.  I could have updated all three fields every time but sometimes smaller sections of code are easier to maintain and debug since I don't need to be concerned about what else might be impacted by a change to a generic function. But I could go either way.

application/controllers/Items.php
SteveIreland: uh ... the overwrite in the following line is the one that needed to be removed.

SteveIreland:
@odiea I'll make sure it's working this weekend.  It was working before I started some unrelated changes that abandoned and attempted to back out.

Thanks to you and @jekkos for the review and feedback!

SteveIreland:
I have it working again and will be resubmitting a rebased pull request later today (with NO last minute code changes). I just want to do a little more testing.

SteveIreland:
Hmmm.  Somebody moved my  cheese.  I rebased over the new commit level and found there was conflict in the sales register.  I'll need to retest.

daN4cat:
Damned little mouse....

SteveIreland:
Corrections have been made, it's been rebased and the pull request has been updated.

Enjoy.  It is a change that is adapted to the existing architecture so there are no structural changes.  It is intended to facilitate the entry of orders where temporary items are required.

There are a couple of other tweaks that I could add but I would like to wait to see what others think about it.

odiea:
Item now shows in receipts.

dvidic:
Implemented these changes and making testing...so far so good.

dvidic:
One situation has occured. When I add temporary item, and later I want to edit it, I get error that item with that barcode already exists...and it's impossible to edit it.

SteveIreland:
I'll check that out tonight.

SteveIreland:
Also, since this is a temporary item when do you think it would be necessary in the real world to use item maintenance edit it? I was thinking "never".  The item should only be used for a particular sale and once the sale is complete then there isn't much value in editing the item ... right?  What are you seeing that I don't yet see?

dvidic:
I had temporary item which I used for quote. Then, after a while i received that item on stock. I needed to edit data for.it and I was unable to do it. I'll try changes you made...

SteveIreland:
Whoa.  I didn't have a use case for converting a temporary item to a permanent item.  That cannot be done because the item id permanently classifies it as a temporary item.  You'll need to create a permanent item for it and add it to the quote and delete the temporary item.

SteveIreland:
However, I'm guessing that you tried to change the item type from temporary to some other type.  Let me think about that this weekend because technically that should not be allowed and I need to think about what the impact would be if that were to be allowed.

daN4cat:
A classical, a customer with never ending requirements.

PR #1981, opened 2018-05-12
temporary "fix" to wrong summary_payment report

truchosky:
I still cant find a solution to this,  summary payment amount in cash is wrong, like you know it considers payment amount and it doesn't discount change, so the cash payment will be always a higher value, and if you try to discount change directly on query then it only discounts sales with 1 item, and not discounts sales with multiple items, the condition sales_items.line = 1 cause that.

This change considers item price * item qty, so at lest summary payment in cash will be acurated no matter tendered amount is, but leaves out "due sales"

### DIFF
--- a/application/models/reports/Summary_payments.php
+++ b/application/models/reports/Summary_payments.php
@@ -14,7 +14,7 @@ class Summary_payments extends Summary_report
 
 	public function getData(array $inputs)
 	{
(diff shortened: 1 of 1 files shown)


daN4cat:
At least this improves the matter, although you say it's not perfect. But I don't understand exactly which use case is not right.
I tested a mixed case part cash and part card and it's reporting the tow parts correctly.

PR #1990, opened 2018-05-14
Csrf simplifications

jekkos:
Some simplifications for CSRF support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 node_modules
 tmp
+database/database.sql
+database/migrate_phppos_dist.sql
 application/config/email.php
 application/sessions/*
 application/logs/*
--- a/application/views/configs/locale_config.php
+++ b/application/views/configs/locale_config.php
@@ -234,7 +234,7 @@ $(document).ready(function()
 		var data = { number_locale: $("#number_locale").val() };
 		data[field] = value;
 		$.post("<?php echo site_url($controller_name . '/ajax_check_number_locale')?>",
-			$.extend(csrf_form_base(), data),
+			data,
 			function(response) {
 				$("#number_locale_example").text(response.number_locale_example);
 			},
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,5 +1,7 @@
 node_modules
 tmp
+database/database.sql
+database/migrate_phppos_dist.sql
```
daN4cat: Why this?

jekkos: Because these files are generated by the build, so they should not be in the repository (merging them does not make sense as the next grunt build will overwrite them anyway)

mckaygerhard: it makes sense, but again more and more complicated and separated things

daN4cat:
It doesn't work. I gave a try and the first thing I tested was Mailchimp integration and ajax call. Added Mailchimp key and the Mailchimp lists don't appear.

jekkos:
OK it could be that there's yet another path to $.ajax that I didn't proxy. Will need to check that thoroughly. I prefer to get rid of this csrf thing scattered in the code. If it's just in one place then it will be easier to remove once we move to a better or modern mitigation.

jekkos:
thanks for testing btw

jekkos:
Ok one step closer, now just need to run the docker from PHP. Pretty nice to quickly test any branch after pushing (so no need to switch locally)

jekkos:
Ok works as I want now.. let's try this out

daN4cat:
I don't see any code change...

jekkos:
Ok works on the local machine now

jekkos:
Seems that I accidentally deleted my deploy script.. so no more demo instance for now

daN4cat:
The issue related to Mailchimp was something else in the end.
However I fixed a couple of things, and done some general tidy up.
Based on the testing I made this seems to work fine and I can merge. We take the risk.

daN4cat:
I tried setting `$config['cookie_httponly'] = TRUE;` but it doesn't work and blocks ajax calls.
Also I noticed there are a couple of setup_csrf calls but if I tried removing them the calls won't work.

jekkos:
Just spotted something weird, it seems that the token is posted in some cases and processed as part of a form (invoice comments on the demo instance eg)

jekkos:
Check https://demo.opensourcepos.org/sales/invoice/45

jekkos:
Ok did a quick test and seems fine. WIll merge then now.

daN4cat:
Noooooo

daN4cat:
I said not in 3.2.1....

PR #2011, opened 2018-06-06
Add support for temporary items.

SteveIreland:
Here is version 2 of support for temporary items.

### DIFF
--- a/application/libraries/Sale_lib.php
+++ b/application/libraries/Sale_lib.php
@@ -66,7 +66,7 @@ class Sale_lib
 
 		$filtered_cart = array();
 
-		foreach($cart as $k=>$v)
+		foreach($cart as $k => $v)
 		{
 			if($v['print_option'] == PRINT_YES)
 			{
@@ -78,7 +78,7 @@ class Sale_lib
 		if($this->CI->config->item('line_sequence') == '0')
 		{
 			$sort = array();
-			foreach($filtered_cart as $k=>$v)
+			foreach($filtered_cart as $k => $v)
 			{
 				$sort['line'][$k] = $v['line'];
 			}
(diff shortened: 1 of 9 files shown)


application/libraries/Sale_lib.php
```diff
@@ -685,7 +698,7 @@ public function clear_rewards_remainder()
 
 	public function add_item(&$item_id, $quantity = 1, $item_location, $discount = 0, $price_mode = PRICE_MODE_STANDARD, $kit_price_option = NULL, $kit_print_option = NULL, $price_override = NULL, $description = NULL, $serialnumber = NULL, $include_deleted = FALSE, $print_option = NULL )
 	{
-		$item_info = $this->CI->Item->get_info_by_id_or_number($item_id);
```
daN4cat: you sure removing this is fine?
I don't see the replacement doing exactly the same

application/models/Item.php
```diff
@@ -65,7 +66,7 @@ public function item_number_exists($item_number, $item_id = '')
 		$this->db->where('item_number', (string) $item_number);
 		// check if $item_id is a number and not a string starting with 0
 		// because cases like 00012345 will be seen as a number where it is a barcode
-		if(ctype_digit($item_id) && substr($item_id, 0, 1) != '0')
+		if(is_numeric($item_id) && substr($item_id, 0, 1) != '0')
```
daN4cat: there was a bug that the ctype protected.... again I'm not sure changing all of this is covering all the use cases after

SteveIreland:
Sorry, no one seem interested in reviewing the second iteration.  It's been sitting there for a bit, but since I didn't change it much from the first iteration I assumed there wasn't much new there to find. 

The c_type function caused a failure when I was doing the negative id but since I've reverted back to positive ids I'll go ahead and include it back as part of the multi-package.

I'll look closer into the change from get_info_by_id_or_number to simply get_info_by_id, but there was a reason I had to change it in order to support temporary items.  I believe when we changed from fully unique item numbers to "possibly" unique item numbers it should have been changed at that time.  But I'll look for any negative impact.

daN4cat:
I already spent some time to fix issues around that part in the past. Could you please look back in GitHub history and read the issues? That should give the test cases.
I will check myself too to remember the problems around this.

WRT review, I didn't spend that much time in reviewing it earlier because I focused on 3.2.1 release.

daN4cat:
BTW I should start adding test cases because otherwise we risk forgetting about the rationale behind certain things.

daN4cat:
@SteveIreland please revert this change or fix it because it breaks sale register functionality.
I simply tested on the demo site, you cannot scan an item using the barcode anymore. You can only scan using the item_id, not convenient.
As suspected the change of `get_info_by_id_or_number` is causing the regression.

application/libraries/Sale_lib.php
daN4cat: this change is not right, you cannot scan an item with the barcode anymore.

application/models/Item.php
daN4cat: This change was made because the passed item_id could be numeric or a string, meaning it could be item_id or number (barcode) - item_id is a number, barcode is a string -.
Last bug I fixed in this area was that a barcode starting with 0 was not considered a string and therefore considered a number = item_id. I fixed that.

However is_numeric does not behave like ctype_digit:

image

Changing all this area is tricky as you could face subtle issues with different combinations of barcodes and etc.

We really need to build a test harness otherwise we will easily introduce regressions.

SteveIreland:
Thanks. I will look in to this today

SteveIreland:
Yes, I believe I clearly see the problem - one that you had warned me about it earlier but it just didn't sink in at the time.  I'll make the change tonight, test it and should have a pull request for it available in the next 24 hours.

jekkos:
For some reason I don't get notifications on new pull requests, so I missed out a bit here..

WebShells:
Guys... Demo returned internal http error... Is this PR related in anyway ? @daN4cat  @SteveIreland

mckaygerhard:
see my last two isues some info are here

Issue #2024, opened 2018-06-12
Submit button not working on add new items


whoami1904:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.2.2
- OSPOS git commit hash is: 
- PHP version is: 5.7
- MySQL or MariaDB version is:  5.7
- OS and version is: Ubuntu 16.04
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

- I am facing issue about add new items. Now i can not add or update new items and it is not show error. I had fill everything and click submit button but it is still show me table of add new item with information that i filled before. But i can add new item by csv file :)
- This error appearance after i change date and time system on window machine that is open ospos website. I had change again already  but error still there. :(

[...]

Issue #2021, opened 2018-06-11
 Release 3.2.2 still with mayor csrf error 403 at register items


mckaygerhard:
seems related with #1990 cannot registyer a items, produces **403 invalid access errors, reproducible until git 9796d6cfc33714ff0b9d554e5c879b9719e573a3 commit only with BARCODE input field..**

i download a 3.2.2 stable build for test some of the old reportes bugs> https://bintray.com/jekkos/opensourcepos/download_file?file_path=opensourcepos.20180606204534.3.2.2.c5632a.zip

installed with defaults and full DB and apache2 access.. THE PICTURE ILUSTRATE TE PROBLEM

![osposerror-csfr-items](https://user-images.githubusercontent.com/1571724/41242545-d6b9b8ce-6d6d-11e8-8a7d-8b60cc8b256e.png)

mckaygerhard:
same as #2024

jekkos:
might not have been reverted properly, will have a look tonight

mckaygerhard:
the PR send by you only solved agains 3.3.0 i reviewed, but 3.2.2 sitll will suffer due the csfr are after commit 9796d6c that are totally different and need other rewriting of the fix

nerv55:
Bug confirmed I have tested yesterday.

SteveIreland:
I'm not an expert by any means in the area of cross domain scripting errors but noticed that you are using an IP address instead of a domain name.  Since I am not encountering these errors in my tests could your use of an IP address be triggering this?

I've created a virtual domain on my local host to more closely simulate a production setup.  Maybe that is the difference?

mckaygerhard:
umm maybne you send comment afeter recents, but @SteveIreland do you read that this only happened in 3.2.2 and today the lasted has a csfr commit unrelated to 3.2.2 and that-s why are unreproducible? also do you read i not the only with this error and other have it? do you read that only happened in 3.2.2 that means before commit 9796d6c ? maybe that's why you write that

mckaygerhard:
but there's something strange here @jekkos and @SteveIreland the fix in PR at #2025 seems does not affect the 3.3.0 due at demo side everything works good! but i review the code and seems are need! 

i dont understand that's must be reproducible the error at athe demo side and at the git recent but seems works.. only reproducible with 3.2.2 ...

that's due i just tested the PR #2025 agains 3.2.2 and seems fixed the problem: see the picture i dont understand!

[...]

PR #2025, opened 2018-06-12
Fix check_item_number bug #2024

jekkos:


mckaygerhard:
this solve part of the 3.3.0 bugs but for 3.2.2 release still have many others bugs.. and this pull does not help in that particular case, i have simillar errors in other functions but i not have now enought time.. 

i recomended check the revertion agains olders commits and apply again! that-s wht in my fork i have 3 branchs like debian does!

mckaygerhard:
seems solved, but please revise comment in https://github.com/opensourcepos/opensourcepos/issues/2021#issuecomment-396588289

jekkos:
Ok will merge then

Issue #2020, opened 2018-06-11
Release 3.2.3 still with mayor error bugs at expenses


mckaygerhard:
seems related with #2018 but error gives now at DB level with related tables, some refers incorrect?

i download a 3.2.2 stable build for test some of the old reportes bugs> https://bintray.com/jekkos/opensourcepos/download_file?file_path=opensourcepos.20180606204534.3.2.2.c5632a.zip

installed with defaults and full DB and apache2 access.. here the error SQL log:

```
ERROR - 2018-06-11 09:58:18 --> Query error: Cannot add or update a child row: 
a foreign key constraint fails (`ospos`.`ospos_expenses`, CONSTRAINT `ospos_expenses_ibfk_1` FOREIGN KEY (`expense_category_id`) 
REFERENCES `ospos_expense_categories` (`expense_category_id`)) - 
Invalid query: 
INSERT INTO `ospos_expenses` (`date`, `supplier_name`, `supplier_tax_code`, `amount`, `tax_amount`, `payment_type`, `expense_category_id`, `description`, `employee_id`, `deleted`) 
VALUES ('2018-06-11 09:58:04', 'ASA', 'AsAS', 10, 10, 'Efectivo', '0', '', '1', 0)
INFO - 2018-06-11 09:58:18 --> Final output sent to browser
DEBUG - 2018-06-11 09:58:18 --> Total execution time: 0.2809
```

mckaygerhard:
i found the problem! expense category are empty on ospos at first install so then i try to register a expense, will fail due category are empty..

this are due the 0015bdffcf67816ff5a136758544200a7c093874 commit that remove the example, this are good but need to left a default category! or at leas alert users that category could not be empty!

umm i noted that if some one does not comment or if a issue does not have more that one person involucrated are not take seriusly!

jekkos:
Ok thanks for pointing that out. If we're allowed to delete the last category, then we should also get a proper error message no?

SteveIreland:
Agreed.

mckaygerhard:
umm @jekkos and @SteveIreland why a release has too many bug and git now are too away of the release? a new release will nedd a complete revert of the last commits or a new branch in the git! pufff

daN4cat:
Works fine for me.

jekkos:
I can reproduce it.. the frontend validation is not flexible enough here and refuses to accept a formatted decimal.

daN4cat:
Fixed and in master.

Issue #2018, opened 2018-06-08
bug: decimals are not interpreted correctly at expenses module


mckaygerhard:
### Background information

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is:  demo
- OSPOS git commit hash is: demo
- PHP version is:  ??? please!
- MySQL or MariaDB version is:  ??? please!
- OS and version is:  ??? please!
- WebServer is:  well must be apache2 for sure
- Selected language is: en-US
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

set localte in en-US and let all with default i demo site, then logout and login, then register a expense and said that amount must be set to number, see picture 1, 

so then in picture 2 i register 30000 but when submit picture 3 show register 0!

![ospos-error-locale-decimals](https://user-images.githubusercontent.com/1571724/41160721-919dbf2c-6afe-11e8-9ad4-f1986478c2bb.png)

[...]

PR #2027, opened 2018-06-13
Use backend for amount validation (#2020, #2018)

jekkos:
Some fixes for CSRF + amount validation in giftcards and expenses

### DIFF
--- a/application/controllers/Expenses.php
+++ b/application/controllers/Expenses.php
@@ -159,6 +159,12 @@ class Expenses extends Secure_Controller
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


jekkos:
* Reason why csrf sometimes didnt work, was probably because the dateFilter callback was not always done (setup of the token is now done in the hook itself, before returning the result)
* Giftcard amount formatting was broken, now works and cleans user input interactively
* Same for expenses form

would be nice if someone has some time to do a thorough validation on demo. @mckaygerhard ?

PR #2028, opened 2018-06-13
Use backend for amount validation (#2020, #2018)

jekkos:
*    Reason why csrf sometimes didnt work, was probably because the dateFilter callback was not always done (setup of the token is now done in the hook itself, before returning the result)
*    Giftcard amount formatting was broken, now works and cleans user input interactively
*    Same for expenses form

Would be nice if someone has some time to do a thorough validation on demo.

### DIFF
--- a/application/controllers/Expenses.php
+++ b/application/controllers/Expenses.php
@@ -161,7 +161,8 @@ class Expenses extends Secure_Controller
 
 	public function ajax_check_amount()
 	{
-		$parsed_value = parse_decimals(array_pop($this->input->post()));
+		$value = $this->input->post();
+		$parsed_value = parse_decimals(array_pop($value));
 		echo json_encode(array('success' => $parsed_value !== FALSE, 'amount' => to_currency_no_money($parsed_value)));
 	}
 
--- a/application/controllers/Expenses.php
+++ b/application/controllers/Expenses.php
@@ -159,6 +159,12 @@ class Expenses extends Secure_Controller
 		}
 	}
 
+	public function ajax_check_amount()
+	{
+		$parsed_value = parse_decimals(array_pop($this->input->post()));
+		echo json_encode(array('success' => $parsed_value !== FALSE, 'amount' => to_currency_no_money($parsed_value)));
+	}
+
 	public function delete()
 	{
 		$expenses_to_delete = $this->input->post('ids');
--- a/application/controllers/Giftcards.php
+++ b/application/controllers/Giftcards.php
@@ -131,7 +131,7 @@ class Giftcards extends Secure_Controller
 	public function ajax_check_number_giftcard()
 	{
 		$parsed_value = parse_decimals($this->input->post('giftcard_amount'));
-		echo json_encode(array('success' => $parsed_value !== FALSE, 'giftcard_amount' => $parsed_value));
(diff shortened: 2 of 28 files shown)


jekkos:
Maybe i should leave invalid values untouched instead of resetting them? Any ideas here?

WebShells:
@jekkos for gift cards and expenses things are running smoothly over the demo...

jekkos:
@Webshells can you also test locale config, MailChimp, barcode validation to see if it works well with csrf?

jekkos:
@daN4cat I checked the double submit issue, it does not occur after adding again a form_setup call. I'll verify again on demo  right away.

daN4cat:
Tested and I still see the issue (BTW I'm using PHP7.1)

jekkos:
For my understanding, what you do is just click a submit button twice in config?

daN4cat:
I click submit once then after the pop up window I click it again and from Inspector I see the error.

jekkos:
Ok I can reproduce it..

daN4cat:
Tricky one to catch...

jekkos:
I'll try to look into it this week, went off for the weekend again..

jekkos:
OK think I have it fixed now. Issue was that I was trying to reinstall the new token in a form that wasn't yet in the DOM (so from the $.ajax call that parsed the DOM response). Instead of this the token refresh is done from the ajaxComplete() method now (seems to be a new function in current jQuery).

Testing/validation on demo would be nice @WebShells

daN4cat:
Sounds good

odiea:
Demo Server Bad Gateway

mckaygerhard:
interesting demo server it's not apache... but ospos must run recomended in apache!

daN4cat:
@mckaygerhard never heard of running a reverse proxy in front of a web server? That's standard practise and usually done using nginx.
BTW OSPOS can also run on nginx and there is a tutorial how to set it up. So no, it's not only Apache.

jekkos:
Ok demo should be running this branch now

jekkos:
@mckaygerhard in this case ospos is still running on apache (in the docker config), but it's just easier to setup ssl using let'sencrypt and nginx so..

daN4cat:
@mckaygerhard I would suggest to tune your expectations to the right level as we are not committing ourselves to support 50 OSPOS stations nor enterprise deployments.
There are professional people to deal with those type of problems and they make sure they are paid for that type of job.

mckaygerhard:
i cannot track correctly either backport this pull due there's too code style changes (innecesary in a pull request) can be **code changes next time made outside/after the pull request! there's too many changes not related with the bugfix!**

jekkos:
So no more known issues here? Ready to merge?

SteveIreland:
I'm ready!

mckaygerhard:
and for 3.2.3 version ? there's no backport? i cannot track due those non-related changes

daN4cat:
@jekkos I didn't have the time to test. I'm maxed out. It's a tricky change and needs to be tested properly.

jekkos:
I tested some regular flow.. as well as most of the config submits, seemed to be all fine. In case something else pops up then I'll take care of it, so I'll pull the trigger here now.

SteveIreland:
Excellent, I'll rebase all of my stuff over it tonight.

PR #2049, opened 2018-07-14
Fix Count Function Error

SteveIreland:


odiea:
Here was the pull request from 2 days ago and I did not see your issue.
2018-07-16 13_09_13-window

Issue #68, opened 2015-05-06
categories and subcategories


asadjaved63:
Hello sir. Hope you will fine with your health and business
Sir, it is possible, make a system like Categories have different sub categories and all items have same item number.....

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=13244847)](https://www.bountysource.com/issues/13244847-categories-and-subcategories?utm_source=13244847&utm_medium=shield&utm_campaign=ISSUE_BADGE)

jekkos:
Yes I have a branch set up that will allow you to define quantities in 'units' for each item. In this case it could be used to assign different kinds of items all having the same item id, while keeping their current quantity stock counts separated.

Currently the implementation allows you to assign units to a specific category, but only by defining the data directly in mysql. Thus the main problem is that this functionality needs a configuration interface (so you can setup the units properly) and I have no funding to develop this functionality yet.

asadjaved63:
i will send you some coins if you develop it for me....

asadjaved63:
reply me .. waiting for your reply

jekkos:
I was thinking of making the UI so that one can create categories and units, and then develop something that allows a user to associate mutliple units with a category of choice.

This means that you will be able to add the different colors here, link them to eg category t-shirt, and then select the t-shirt category for a new item. Then all the associated categories will show up with edtiable quantities for that category. This is still quite a profound change and will take some time and effort so I might want to discuss the reward with you first.

asadjaved63:
This great idea. Many companies required same thing... one item have different categories and sub-categories like color, size etc, and have different quantity of sub-item..  , in my case one item have many sub-items(color and size) and all sub-items have own Quantity,, so I would suggest you Do something like that,, its very helpful to maintain stock  in the end of year and maintain profit and loss statement....if you develop it for me I will donate you some amount  on PayPal
![a02bdfb6-f3b9-11e4-882a-4ca5afff67dd](https://cloud.githubusercontent.com/assets/12225271/7535402/1e3327ee-f59e-11e4-86b2-cabe1f398839.png)
 ....

asadjaved63:
If I want to add one module like opening balance and closing balance,
expense, then what you would charge for it... please tell me?

jekkos:
This might be best implemented as a new report. What do you mean with closing balance? The amount of cash in the till? In this case it might be best to add something in the sales module that can track the amount of money in the till?

[...]

PR #2086, opened 2018-08-17
sales + receivings item attributes support (#68)

jekkos:
After letting this code float for more than a year I finally managed to push through and polish what was there to finish a functional whole. A bit of a sad thing I got demotivated as the scope of the initial design was badly incepted by me and some things needed time to come to fruition. Also I found myself losing a lot of time in designing a good way to manage the attributes in the item form.

I wanted to get this out as I'm leaving on a hike trip for a week to northern Sweden (lapland). I hope that some of you can find the time to review and test the branch before I return so I can do the necessary changes and fixes.

Let me describe briefly what is in here.

### Functionally
* 'Custom' fields and all involved ugliness was removed from the config section
* Item attributes can be created / edited  through the attributes module in the backoffice
* 'Groups' of attributes can be defined. This serves as an easy way to add multiple attributes to an item when creating or updating it. For eg. in my case for running shoes we have both color and size that can be linked to the 'Shoes' Group attribute definition
* Two levels of attribute visibility, being 'sales' and 'receivings'. This means that the values for those attribute definitions will be visible in the respective receipts
* Three 'types' of attribute definitions
 - **DATE**: Renders the field in the item dialogue using a datepicker, and saves the values accordingly. Useful for saving 'expiry dates' on items or the like
 - **DROPDOWN**: Choose from a fixed set of values which you define in the attributes module. Rendered using a select box
 - **TEXT**: Just add some free text as an attribute value. This control is rendered using autocomplete widget and will optimally reuse existing database entries
* Attribute 'values' are searchable through the existing attribute search flag in items screen. This should work as before with custom fields

### Refactoring
* Did some minor refactoring / cleanup here and there, mostly around the datepicker widget and table management

### Extra focus for testing
* Migration script is foreseen to move existing custom fields to the new attribute module system. I might **need some help from others** testing the specific SQL here as I don't have a database that used custom fields, so most testing was done on a clean situation

### Design considerations
I found that I lost a lot of time trying to think of an easy and concise way doing CRUD for attribute values for both the insert and update situation. In the end I found a nice way around by leveraging a server side ajax view refresh cycle that sends out form controls using javascript input. Personally I think this is  cleaner than trying to build the DOM form javascript itself, which is done in quite a lot of other places in the app. 

### Open ends
* One thing that is still missing is a way to show attribute values in the items table by default. Initially I added a visibility flag for this but there's no code yet to handle showing attribute values in the items table.

And now I gotta go make my backpack..


jekkos:
@objecttothis thanks for commenting. 
* The branch will be deployed on our demo server again in a couple of minutes, so you can have a quick try there
* We could indeed consider moving some fields to this section, however, will need to go through them to see which ones would be worth moving
* A similar mechanism might be useful adding form controls to customers module (which was asked for some time ago)
* And indeed the DATETIME attribute type uses the user's locale setting for formatting / parsing

odiea:
Migration does not work.
Configuration Error 500.
Cannot delete an or modify attribute.
All custom tables even empty tables (updated to attribute tables) show in Items after upgrade.
Let the games begin.

jekkos:
@odiea these errors occur after migration then?

odiea:
I had to manually add the new attributes. Then I tried to update or delete the current attributes and nothing happened. I don't know if Config worked before adding the attributes. I also did not enable error reporting so I will do that and let you know if any errors show when trying to enter the config module.

odiea:
Missing a comma on line 301 and 302 has a comma at the end of the line. Not anything major.
ERROR - 2018-08-17 16:42:48 --> Severity: error --> Exception: syntax error, unexpected ''multi_pack_enabled'' (T_CONSTANT_ENCAPSED_STRING), expecting ')' C:\wamp64\www\attributes\application\controllers\Config.php 302

odiea:
Has receipt issues also.
2018-08-17 11_13_00-wedidit _ powered by ospos 3 3 0
Receipt quantity changed to 9 after printing receipt.
2018-08-17 11_13_17-wedidit _ powered by ospos 3 3 0
Another odd receipt. Quantity sold was 1.
2018-08-17 11_08_33-wedidit _ powered by ospos 3 3 0

mckaygerhard:
seems there's a re-sending on a part of the screen, by 8x

odiea:
Here is what is shown after all the attributes are removed from the Item.
2018-08-17 12_31_12-wedidit _ powered by ospos 3 3 0

jekkos:
Config section is fixed. Indeed I also notice some strange behavior in sales register (adding an item multiple times when linking attributes), but will check that further somewhere next week alas.

jekkos:
thanks for the quick feedback though!

jekkos:
@odiea Ok no sorry I couldn't stop so I fixed the receipt bug as well. Probably the other issues you're facing are due to the migration doing something wrong. I'll have to check that (if I can move from an old database using custom fields to a new one, to reproduce this). Will take a little more time to set up an old version first and then migrate (as I develop  using docker and thus recreate db everytime)

jekkos:
But so that one I will postpone until earliest 27/08 (when I get back from my trip)

odiea:
The migration will probably be different for everyone depending on how many Categories or how the custom fields were used.

SteveIreland:
I noticed that the field category was dropped from the items table, but it wasn't dropped from the form.  When I go into item maintenance it crashes on the missing field.  Was it meant to be replaced by attributes? If so I'll remove it from my copy of form.php and continue.  Or is there some magic that's supposed to  happen via javascript that didn't happen for me. 

I'm testing from scratch and not running any migrations.

daN4cat:
Is categories an attribute now?

odiea:
Here is what I see in database.sql. category is still there.
2018-08-18 04_47_54-c__wamp64_www_attributes_database_database sql - notepad

SteveIreland:
Okay, I must have built it incorrectly.  I'm also encountering a javascript issue that I'm sure no one else has.  I'll begin anew.

daN4cat:
Ok, because I remember we discussed that point and my position was to leave category and have separate concept of attributes group.

SteveIreland:
Much better.  Sorry for the extraneous noise.

SteveIreland:
The attributes need to be selected for each individual item added.  This is nice because now we can have attributes that are specific to a type of item.  

It would be nice, however, to be able to flag an attribute as auto-selected for all items.  

It would also be nice to set a flag that determines whether or not it is required.

I played around with this and used as my test case the harmonized system code that is needed for India GST.  It is an item attribute, and only has real value in an India GST configuration.  Legally it should be entered for all items since the type of item is used to establish tax rates and the government is probably validating the HSN code against the tax rates used.

The HSN code has to appear on the invoice but since it is simply a numeric value it will need a field label of some type.  It doesn't work if it is just appended to the item name.  So in this case I would need to deselect it as "visible" for sales and receiving and then modify the invoice and receiving document to add it manually so that I can assign a  field label to it.  Right?

For fields that drive program logic there isn't a way to make them "system fields" so that they are automatically included and cannot be removed by the user.  I prefer to avoid user defined fields when program logic is dependent on them but I vaguely recall asking about this a year ago and I thought it was going to be included.  I only raise it for discussion because my intent is to continue with the tax category id as part of the item table and not as part of the dynamic item attributes and I wanted to make sure that was okay with the team.

SteveIreland:
A 10 day vacation ... geesh ... I start feeling "work deprived" after 3 days.  Surely they have internet access in Lapland.   :-)

daN4cat:
I had two weeks of holidays and I didn't think about work at all.

application/views/attributes/item.php
```diff
@@ -50,27 +49,27 @@
...
+                $(this).parents('.form-group').remove();
             });
         };
 
         enable_delete();
 
         $("input[name*='attribute_links']").change(function() {
             var definition_id = $(this).data('definition-id');
-            $("input[name='attribute_ids[" + definition_id + "]").val("");
+            $("input[name='attribute_ids[" + definition_id + "]']").val('');
```
daN4cat: I changed this line because it didn't look right to me (you would expect [ [ ] ], but I had not the chance to test it yet.

application/views/attributes/form.php
```diff
@@ -0,0 +1,174 @@
...
+	$.each(definition_values, function(index, element) {
+		add_attribute_value(element);
+	});
+
+	$('#attribute_form').validate($.extend({
+		submitHandler: function(form)
+		{
+			$(form).ajaxSubmit({
+				beforeSerialize: function($form, options) {
+				    $("select[disabled='disabled'").removeAttr('disabled');
```
daN4cat: I'm not sure this line is correct because there is a [ but no ]. Probably it should be ']".
If somebody could change and test this and the other change I made we could get confidence it's fine.

daN4cat:
I made a couple of changes to js in views/forms related to attribute (item.php and form.php). I believe they were missing a closing ].
I don't have time to test, also because I would need to change my upgrade to an old version database script to fit my custom fields to attributes.
Somehow I can't get the Docker to work properly on my laptop, something I need to investigate.

So if somebody could give test please.

odiea:
"The URI you submitted has disallowed characters". is the error I get when I try to delete an attribute from the attribute table.

SteveIreland:
It's building a URL  of `POST /public/[object%20Object]/delete`

odiea:
I am getting that error when ever I try to delete anything. Items,Sales,Expenses and so on.

daN4cat:
Was it like that before my change?

odiea:
I will re-download the original and let you know. I did note that I could not delete an attribute from the table but I did not test other things to delete.

odiea:
https://bintray.com/jekkos/opensourcepos/download_file?file_path=opensourcepos.20180817002410.item-attributes.ac44d9.zip. This was the 1st one I tested and yes the issue is in there.

odiea:
I just installed the current master and it has no such issues.

SteveIreland:
Oh no ... a change in the deep forests of OSPOS where only the wizards dare tread. 

I think the problem may have been introduced in a change to manage_tables.js

WebShells:
Hola... is there anyone else having problem loading detailed sales report on demo server ? And i think we should display attributes in Summary items report and detailed sales report O.o ! Any thoughts ?

WebShells:
Actually all specific reports ( Employees/Discounts/Customers ) and Transactions... Unable to handle request...

jekkos:
Ok so a couple of things for me to check

(-) attribute delete. Did some changes in manage tables that could be the culprit
(-) detailed sales / receivings reports. I agree that we should see the attribute value as columns if they are flagged visible.

objecttothis:
@jrkkos, one thing to consider is adding a attribute type of number with customizable precision (whole, 1, 2, 3 places, etc). For example our bookstore will need to add an attribute for item weight which should be restricted to a number, but the number attribute type could later be used to convert many different already existing database fields to attributes.

daN4cat:
Welcome back @objecttothis you disappeared for a while. End of the summer?

objecttothis:
The building our bookstore owned was being torn down so I needed to focus my attention of not getting ripped off, moving to a new building and getting the government to give us the necessary documents to reopen. As it stands we are having to take the government to court just to force them to give us documents that they want bribes for. At least I can start giving some attention back to OSPOS changes and writing the connector for the CMS we are using and OSPOS to talk to it.

jekkos:
Ok fixed some more bugs, attributes added to sales / receivings reports. If anyone can find more bugs, let me know, otherwise it'll be rebase time soon

daN4cat:
@jekkos CI is failing because of an issue with cert altnames, maybe you want to check that issue.

daN4cat:
If I find the time I'll test, I fixed a couple of small editorial things not worth commenting.
I guess Items view should have a means to display the attributes in the table, what do you think?

daN4cat:
BTW, I cannot search for an attribute content in the Items table. Let say I want to search for all the colour attribute white, it doesn't work

daN4cat:
Also it might nice to have the possibility to set a default group so you don't need to always select group when you create an Item.

WebShells:
@daN4cat you can search attributes by selecting it in search fields, like we used to do for custom fields... Search by attributes, then it will show.. If i got you right...
Loading the attributes in items list is good idea too...

daN4cat:
@WebShells indeed. Thanks

WebShells:
My pleasure man...

daN4cat:
However, the search by Category doesn't seem to work. Even if I select Search Attribute doesn't work (but I would have not expect to work).

daN4cat:
I tried deleting an item from the table but got a js error: _Uncaught TypeError: b.fn.bootstrapTable.defaults.formatConfirmDelete is not a function_

daN4cat:
Just noticed that search seems broken in general, even by Supplier or simply Item title.

erastus:
@jekkos  it would be good to add quantity to the group of attributes

WebShells:
I guess it will be better if we add stock number next to multiple attribute to each item.
For example we want to receive 3 blue shirts and 3 red ones small in size we can make them as one item with several Attributes...

erastus:
yes, I want to have inventory stock by color and size

jekkos:
@daN4cat Ok thanks for review and testing. Attribute search should indeed work if you select it in the dropdown first. I did test that but maybe didn't check regular search. Will have a look tonight.

Also I siltl need to verify how the migration behaves as @odiea reported some issues there. 
And yes I agree on the fact that we might need a new visibility flag to show an attribute in the items table. But maybe i prefer to add it in followup PR and finish up the rough edges here first, as it's quite a big one already.

Also regarding item creation, you can easily keep an attribute selected at creation time if you use the new button instead of regular submit. In that case the popup will stay open if you're entering multiple new items into the system.

jekkos:
- Codefactor checks seem pretty useless (all checks beside one stating that Item->save() has become  too complex. Otherwise it's detecting duplicate code in the language files
- Fixed the item search
- No more attributes created after migration if they weren't used before
- Enabled the visibility flag for attributes column in Items table. However I'm not very happy with the result, as the table is easily oversized when adding one attribute (a horizontal scrollbar is added). If anyone has some UI ideas here let me know

daN4cat:
I'll disable code factor over the weekend

jekkos:
Ok I have removed the webhook now. It could still be useful but then we need to take the time to configure it properly.

jekkos:
Latest build of this branch is on DEMO now.

jekkos:
Ok one last call for some tests.. then we're ready for merge?

erastus:
I'm going to do a test

SteveIreland:
I also downloaded a fresh copy and will be testing it this evening

SteveIreland:
*Date Attribute Type*
 
The date type doesn't default to the current date and instead defaults to something in the year 2009.

It also appears to be a date-time stamp instead of a simple date field.

Nor am I able to directly type into it.  The only way to change it is to scroll to and select a date ... and that is a long way to scroll.

jekkos:
@daN4cat I'm in the process of upgrading my store's install to this version, it has a considerable big database, sale data for > 5 years and > 20k products. So finally moving away from this custom version I  did for storing categories + item sizes, great opportunity to validate performance on the attributes

@SteveIreland Indeed noticed this issue as well, will have a look today or tomorrow.

daN4cat:
@jekkos sounds good. I wanted to see if I can spend some time today on cashup to test and see what else I need to add to that.

daN4cat:
sorry @jekkos , there is a small conflict to fix since I committed the cashup module.
It's in the autoload.php, just readd attributes at the end. I'm not touching your branch as you may be working on it.

application/migrations/sqlscripts/attributes.sql
```diff
@@ -0,0 +1,185 @@
...
+
+ALTER TABLE `ospos_attribute_links`
+ ADD CONSTRAINT `ospos_attribute_links_ibfk_1` FOREIGN KEY (`definition_id`) REFERENCES `ospos_attribute_definitions` (`definition_id`) ON DELETE CASCADE,
+ ADD CONSTRAINT `ospos_attribute_links_ibfk_2` FOREIGN KEY (`attribute_id`) REFERENCES `ospos_attribute_values` (`attribute_id`) ON DELETE CASCADE,
+ ADD CONSTRAINT `ospos_attribute_links_ibfk_3` FOREIGN KEY (`item_id`)  REFERENCES `ospos_items` (`item_id`),
+ ADD CONSTRAINT `ospos_attribute_links_ibfk_4` FOREIGN KEY (`receiving_id`) REFERENCES `ospos_receivings` (`receiving_id`),
+ ADD CONSTRAINT `ospos_attribute_links_ibfk_5` FOREIGN KEY (`sale_id`) REFERENCES `ospos_sales` (`sale_id`);
+
+
+UPDATE `ospos_modules` SET `sort` = 120 WHERE `name_lang_key` = 'module_config';
```
daN4cat: I set this to be 900 so it will be always at the end before the office one.
You can remove this.

application/migrations/sqlscripts/attributes.sql
```diff
@@ -0,0 +1,185 @@
...
+
+INSERT INTO `ospos_modules` (`name_lang_key`, `desc_lang_key`, `sort`, `module_id`) VALUES
+ ('module_attributes', 'module_attributes_desc', 110, 'attributes');
```
daN4cat: Better to set it to 107, before the expenses and cashup modules as absolute order.

application/migrations/20180905100000_attributes.php
```diff
@@ -0,0 +1,20 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
```
daN4cat: change the date of the model to be 20180910100000, although it would be better to add all of this to the migration to 3.3.0 to be consistent with the previous versions.

application/migrations/sqlscripts/attributes.sql
```diff
@@ -0,0 +1,185 @@
...
+CREATE TABLE IF NOT EXISTS `ospos_attribute_definitions` (
+ `definition_id` INT(10) NOT NULL AUTO_INCREMENT,
+ `definition_name` VARCHAR(255) NOT NULL,
+ `definition_type` VARCHAR(45) NOT NULL,
+ `definition_flags` TINYINT(4) NOT NULL,
+ `definition_fk` INT(10) NULL,
+ `deleted` TINYINT(1) NOT NULL DEFAULT 0,
+ PRIMARY KEY (`definition_id`),
+ KEY `definition_fk` (`definition_fk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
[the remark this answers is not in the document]
daN4cat: The collate, sorry.

application/migrations/sqlscripts/attributes.sql
```diff
@@ -0,0 +1,185 @@
...
+
+
+CREATE TABLE IF NOT EXISTS `ospos_attribute_values` (
+ `attribute_id` INT NOT NULL AUTO_INCREMENT,
+ `attribute_value` VARCHAR(45) NULL,
+ `attribute_datetime` DATETIME NULL,
+ PRIMARY KEY (`attribute_id`)
+) ENGINE = InnoDB DEFAULT CHARSET=utf8;
```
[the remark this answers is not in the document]
daN4cat: The collate, sorry.

application/migrations/sqlscripts/attributes.sql
```diff
@@ -0,0 +1,185 @@
...
+ `item_id` INT NULL,
+ `sale_id` INT NULL,
+ `receiving_id` INT NULL,
+ KEY `attribute_id` (`attribute_id`),
+ KEY `definition_id` (`definition_id`),
+ KEY `item_id` (`item_id`),
+ KEY `sale_id` (`sale_id`),
+ KEY `receiving_id` (`receiving_id`),
+ UNIQUE `attribute_links_uq1` (`attribute_id`, `definition_id`, `item_id`, `sale_id`, `receiving_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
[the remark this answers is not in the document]
daN4cat: The collate, sorry.

daN4cat:
My bad, @jekkos I meant the collate not the charset itself.

odiea:
It appears that the en-us and en-gb Attribute module lang is missing.

application/language/en/attributes_lang.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
daN4cat: This module is not required

application/language/fr/module_lang.php
```diff
@@ -1,5 +1,7 @@
 <?php 
 
+$lang["module_attributes"] = "";
```
daN4cat: This is missing in en_GB and en_US

odiea:
$lang["module_attributes"] = "Attributes";
$lang["module_attributes_desc"] = "Add, Update, Delete, and Search attributes.";

daN4cat:
Fixed `en_GB` and `en_US` and also removed the `en` subfolder.

odiea:
I just downloaded this from the Repository about 20 minutes ago.
Odd looking tables in the attributes edition. In the Items table the new columns are not accurate as per the update form


In the transactions report the tables are also not valid. You can the the Color has both of the attributes.
2018-09-10 17_34_15-window

daN4cat:
@jekkos have you checked the bug reported above?
Also wondering if you managed to do some load testing and checked the performance.

jekkos:
Yes I have it on my todo list, but got blocked early this week due to a broken laptop (someone on the train spilled his coffee over it, and I've been trying to get it back to work since a week, after motherboard replace, etc)

SteveIreland:
Ack ack ack!!!

Did they ever find his body after you threw him off the train?

erastus:
I can not add, the attributes overlap and you can not click
screenshot_18

erastus:
show attributes in the search
screenshot_19

erastus:
problems when trying to add 2 attributes, when it's a new item.
when you selected one, the other disappears.
screenshot_20
screenshot_21

jekkos:
@erastus thanks for testing, I'll go over those cases this weekend.

jekkos:
@erastus I fixed the css issue and the refresh issue as well. However adding the attribute names in the search suggestion is not an easy change. If I understood well a user can configure the field in the suggestion dropdown. So in this case we'd need to query attributes to enable them to show in this config. Feels more like a new feature in one way or another.

In the end it's rather esthetic as you already know on which terms you were querying. The only differentiator could be another attribute added to the suggestion ..

jekkos:
@odiea also the refresh for items screen should be ok now. I need to do some kind of pivot operation on the attribute result set (as rows have to become columns, resulset then needs to be appended to the one of the items) and along the way some extra records can be present (because were joining inventory table as well). This should be fixed now. Also the attribute value positions should be correct wrt their column names (when using multiple visible ones)

erastus:
thank you, I'll compare it tonight

erastus:
does not work, when the items has no attributes or when it only has one attribute.
screenshot_24

jekkos:
> does not work, when the items has no attributes or when it only has one attribute.
> screenshot_24

Just tried locally and couldn't reproduce this case. You mean nothing shows if you have no attrbutes at all? Seems to work fine here.

odiea:
This pertains to the Item add or update form.
1. The table shows it being updated but the new Attributes don't show until the rows are refreshed.
2. Now when updating a form the Attributes default to the way they were entered into the Attributes list and not the way that they were entered into the Item form.

erastus:
@jekkos. add elements without attributes, does not show results.

jekkos:
@erastus , @odiea think I have both fixed now

erastus:
@jekkos there are still problems.
screenshot_27

odiea:
With error logging and development implemented the following shows.
The Modal on the one does not show any error.
2018-09-25 09_51_55-window

jekkos:
@odiea @erastus I can't reproduce the error you were talking about.

jekkos:
Could you try to test again on an empty install? perhaps on the demo server?

jekkos:
installed latest there so test away :dagger:

odiea:
With your last commits in Production mode all works great now. But in Development mode the Items table wont load with this error showing. Message also says Undefined offset:7

A PHP Error was encountered

Severity: Notice

Message: Undefined offset: 6

Filename: helpers/tabular_helper.php

Line Number: 423

Backtrace:

C:\wamp64\www\itematta\application\helpers\tabular_helper.php
Line: 423
Function: _error_handler

C:\wamp64\www\itematta\application\controllers\Items.php
Line: 73
Function: get_item_data_row

C:\wamp64\www\itematta\public\index.php
Line: 315
Function: require_once

odiea:
That is fixed in development mode but when I try to search for an attribute I get that response and this is also what is showing under parameters.
2018-09-29 09_13_46-window

odiea:
I think you nailed it with that change. Great job.

odiea:
Like eating an Elephant. A bite at a time. This shows in development mode but appears ok in Production Mode when adding a new item. This happens when selecting an attribute for the Item. 
2018-09-29 09_41_27-window

odiea:
Fixed. I will download the bintray version in a bit and do more testing. Looks usable now.

odiea:
Still has issues in development mode. After you add a new Item in Production mode.
2018-09-29 10_36_44-window

This is Development mode,
2018-09-29 10_37_43-window

jekkos:
Ok should be fixed now

odiea:
Detailed Receivings and Transactions show the following. It appears it is missing the first attribute but not the attribute column.
2018-09-29 13_28_49-window
Here is what is set on the Item.
2018-09-29 13_27_51-window

erastus:
now it looks good thanks. @jekkos 
screenshot_6

I do not fount by attributes. this happens when you do a search, includes search by attributes.
screenshot_7

jekkos:
Ok so still two open points for detailed reports. @erastus searching for attributes only works when you select the flag in items table

jekkos:
Refactored the attribute parsing code a bit (reused in both reports and items table) and did some more testing. They should show fine now. There are a couple of unallowed chars in attribute values for which I added validators in the forms. (| and :)

erastus:
when you have more than 20 attributes? will all be shown in the table of details items?

jekkos:
Just rebased again with master + aligend the code for attribute expansion in reports + items table. @odiea the category field being absent was probably due to a migration that didn't run (just had this error here as well,logout and login resolved it)

jekkos:
@odiea seemed to be a introduced bug after the supplier branch was merged. Should be fixed now.

@erastus yes currently there is no limit on number of attributes in the table. it's up to the user.

odiea:
Here is your new code. If I comment out $this->db->select('category'); the update works just fine. Is there a reason some of this was added other then the needed attribute stuff? Yes and now there is no longer an issue with the attributes in Transactions showing properly.
```
public function get_multiple_info($item_ids, $location_id)
	{
		$this->db->select('items.*');
		$this->db->select('company_name');
		$this->db->select('GROUP_CONCAT(DISTINCT CONCAT_WS(\':\', definition_id, attribute_value) ORDER BY definition_id SEPARATOR \'|\') AS attribute_values');
		$this->db->select('category');
		$this->db->select('quantity');
		$this->db->from('items');
		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
		$this->db->join('item_quantities', 'item_quantities.item_id = items.item_id', 'left');
		$this->db->join('attribute_links', 'attribute_links.item_id = items.item_id AND sale_id IS NULL AND receiving_id IS NULL', 'left');
		$this->db->join('attribute_values', 'attribute_links.attribute_id = attribute_values.attribute_id', 'left');
		$this->db->where('location_id', $location_id);
		$this->db->where_in('items.item_id', $item_ids);
		$this->db->group_by('items.item_id');

		return $this->db->get();
	}
```
and here is the existing code.
```
public function get_multiple_info($item_ids, $location_id)
	{
		$this->db->from('items');
		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
		$this->db->join('item_quantities', 'item_quantities.item_id = items.item_id', 'left');
		$this->db->where('location_id', $location_id);
		$this->db->where_in('items.item_id', $item_ids);

		return $this->db->get();
	}
,,`

odiea:
@jekkos If you rebased this would you please recheck the commit that showed the attributes on the Items form it is not showing the stored attributes properly.

jekkos:
Ok hope this solves those two last ones..

odiea:
It appears to have fixed the last 2 issues. Could use more testing by different users now.

jekkos:
@odiea in that case I would suggest to proceed and merge to master. As far as we can see it's stable enough and merging it will make it available to more users that then can report back if needed.
@daN4cat waiting your last approval here

jekkos:
Ok it's been sitting here for long enough. Time to pull the trigger.

daN4cat:
And what about the performance with big dataset, have you tested it before merging?
I'm really worried about that because when I tested in the past it was slow.

odiea:
When an Attributes is set to text the modal does not drop down far enough to delete Visibility attributes.
The right hand scroll bar does nothing.
2018-10-15 14_47_59-window

jekkos:
Also item search uses server side pagination which should be pretty fast no?

jekkos:
And if there are issues then I'll tackle them as they are reported.

jekkos:
@odiea I'll check soon. Might be better to crete new issues now if there are bugs as this thread has become too long

odiea:
@jekkos  This may not be an issue after all. I tried it in another browser and it works fine.

PR #2119, opened 2018-09-03
add fields to table.sql discount on sales fixed

erastus:


erastus:
@SteveIreland  you mean tables.sql

erastus:
@SteveIreland  I have made a new installation and does not generate errors, I have a complete database and it does not generate any errors

SteveIreland:
Yes you are correct. Sorry, I was responding via my iPhone at a Mexican restaurant with a margarita in front of me. Thanks for the fix!

erastus:
@SteveIreland  Thanks, I'm Mexican.

daN4cat:
Sorry, but I don't see how this could fix the problem. It changes nothing and the problem it's still there.

I'm using the docker container on the local laptop and it creates everything from scratch.

erastus:
The problem occurs because you do not create the database.sql file with the new fields in the tables.sql file

erastus:
@daN4cat  try to generate the database.sql file with these changes

daN4cat:
The migration script runs the first time and updates the database. It's not that issue either.

erastus:
you can check if in your table ospos_app_config you have the field default_sales_discount_type

daN4cat:
it's probably related to the docker container and database upgrade during migration.
The migration seems executed but something is not right.

daN4cat:
Right, during yesterday's discussion I copied and pasted the following:

INSERT INTO ospos_app_config (key, value) VALUES
('default_sales_discount_type', '0');

and that in latest mariadb is not allowed, where this is correct:

('default_sales_discount_type', '0');

spot the missing `

erastus:
@daN4cat what is the solution, is this pull-requests necessary?

daN4cat:
It's fixed.

PR #2123, opened 2018-09-05
Add support for discount on sales

erastus:
add the new fields to the tables


daN4cat:
@erastus this is not required, because the migration script upgrades the database. Please don't spend more time on this specific issue because it's fine.
Thanks for the help and work so far, don't get me wrong.

erastus:
@daN4cat  Thank you! No problem

PR #2124, opened 2018-09-05
Update issue templates

jekkos:


jekkos:
It seems that there is a way to choose the template based on whether users post a bug report or feature request, might be interesting but I'll only be able to test this once merged to the repo.

odiea:
All you can do is try it and see what happens.

mckaygerhard:
but the bug report template still ask for too many information not necesary, 

1) it better to point to a official files to read 
2) and ask if user follow each step,wich step of the instrucction are failed
3) obviiously each step generates result for following step.. user obviously posted the step wich fail, 
4) if not posted browser resiult

a better solution, due seems this issue are like window: problems->try->reset;problems->try->reseet

jekkos:
Templates can be added through github interface instead. Would be nice to distinguish between bug reports and feature requests

PR #2142, opened 2018-09-16
Fix tax migration script

jekkos:
Still seeing this error when migrating

Severity: Notice

Message: Undefined index: customer_sales_tax_support

Filename: migrations/20170502221506_sales_tax_data.php

I'm not sure whether this value is inserted in the app_config table, as I can't find it anywhere in the migration scripts


SteveIreland:
I’m not sure that this is one is one of those I stumbled across yet. So good catch

jekkos:
Yes correct. Partly because of that, but also found some data in my 2.4 database that stopped some scripts from running.

PR #2146, opened 2018-09-19
update invoice_email.css typo

josh-tf:
_#eeeee_ not valid as requires hexadecimal (i.e _#eeeeee_ or _#eee_ short representation)

**Reference**: https://developer.mozilla.org/en-US/docs/Web/CSS/background-color


josh-tf:
No probs!

PR #2155, opened 2018-09-23
Add support for India GST

SteveIreland:
Here is a small change to the tax component for your review. Since non-trivial pull requests tend to be sitting in the queue for awhile I thought I'd go ahead and get this change in early.  I just finished updating an existing database with this change and it seems to be reasonably clean.

I'm going to continue to test it in as many permutations as possible but I think this iteration is in the ball park.  However, it is one of those changes that a fix in one area might create a problem in another area.  So I do expect to rebuild this pull request a few times and will include any feedback that anyone has time to offer.  Thanks.


application/config/autoload.php
```diff
@@ -132,4 +132,4 @@
 |
 |	$autoload['model'] = array('first_model' => 'first');
 */
-$autoload['model'] = array('Appconfig', 'Person', 'Customer', 'Employee', 'Module', 'Item', 'Item_taxes', 'Sale', 'Supplier', 'Inventory', 'Receiving', 'Giftcard', 'Item_kit', 'Item_kit_items', 'Stock_location', 'Item_quantity', 'Dinner_table', 'Customer_rewards', 'Rewards', 'Tax', 'Expense_category', 'Expense', 'Cashup');
+$autoload['model'] = array('Appconfig', 'Person', 'Customer', 'Employee', 'Module', 'Item', 'Item_taxes', 'Sale', 'Supplier', 'Inventory', 'Receiving', 'Giftcard', 'Item_kit', 'Item_kit_items', 'Stock_location', 'Item_quantity', 'Dinner_table', 'Customer_rewards', 'Rewards', 'Expense_category', 'Expense', 'Cashup', 'Tax', 'Tax_category', 'Tax_code', 'Tax_jurisdiction' );
```
daN4cat: Probably to easy the merge issue we should consider have the array on multiple lines so any automerge will work more efficiently.

application/migrations/sqlscripts/3.0.2_to_3.1.1.sql
```diff
@@ -260,15 +260,8 @@ DELETE FROM `ospos_tax_categories` where tax_category_id in (0, 1, 2, 3);
 ALTER TABLE `ospos_tax_categories`
   MODIFY COLUMN `tax_category_id` int(10) NOT NULL AUTO_INCREMENT;
 
-INSERT INTO `ospos_tax_categories` ( `tax_category_id`, `tax_category`, `tax_group_sequence` ) VALUES
-(1, 'Standard', 10),
```
daN4cat: Please don't touch old upgrades otherwise it becomes a maintainance nightmare. We just need to keep upgrading from the previous to the next without touching history.

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
```diff
@@ -6,9 +6,9 @@ INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
 ('multi_pack_enabled', '0');
 
 ALTER TABLE `ospos_items`
-  ADD COLUMN `qty_per_pack` decimal(15,3) NOT NULL DEFAULT 1,
-  ADD COLUMN `pack_name` varchar(8) DEFAULT 'Each',
-  ADD COLUMN `low_sell_item_id` int(10) DEFAULT 0;
+  ADD COLUMN `qty_per_pack` decimal(15,3) NOT NULL DEFAULT 1 AFTER `tax_category_id`,
```
daN4cat: What happens now to people who upgraded to 3.3.0 dev?
This part will be ignored as already existing.

daN4cat: Anyway not big deal as it's just an ordering story.

application/models/Appconfig.php
```diff
@@ -67,6 +67,7 @@ public function batch_save($data)
 
 		$success &= $this->db->trans_status();
 
+
```
daN4cat: ???

application/migrations/sqlscripts/3.0.2_to_3.1.1.sql
SteveIreland: I wouldn't touch it unless there was a very good reason to do so.  In this particular case we are replacing the notion of customer sales tax with user destination tax.  So "priming the pump" with a tax category of "Standard" is bad ... very bad ... and makes no sense for India GST.  So on a new install I definitely do not want to corrupt the database with a bogus record because then I'd have to delete it as part of migration and as we know, deleting records as part of migration is fraught with peril.

For customer's that haven't run this particular migration script there isn't a problem because after all of the scripts are complete the table should be empty and the user is responsible for manually populating it.  That is a good thing.

For customers who have already run this script then the record now exists and it doesn't need to be rerun again because it would fail.  If they are now going to start using the user destination tax  feature for India GST then they will need to manually delete it and define the categories (tax rate slabs) that they want to use.

One of the things I HAVE done is try to undo some of my changes to the tables. sql file since in those earlier days I wasn't yet in sync with your goals for the migration scripts.  On a new install there shouldn't be any migration script errors but at this time there are because our "starting point" for tables.sql wasn't in sync with the starting point for the migration scripts.  (Not all of the errors are mine and I was going to ask you if you wanted me to address those too.)

This pull requests contains addresses the errors that I introduced.

For those sites that have chosen to upgrade to a development branch then don't they have the burden of tweaking the migration scripts so that they don't rerun them?  I just had to do that with my test upgrade to an existing site.   I had to delete the first part of the t_3.2.0 migration script because it had already run.

If we don't want them to need to do that then it seems we need a separate migration script for each new feature.  Right?

daN4cat: Guys this story sounds a bit unclear to me.
We are splitting paths becoming a maintainability nightmare. In version controlled system you just keep one line and move forward.
If an error was introduced at some point it's dealt later with a fix. So everyone starts from one point and everyone ends on the same other point following one path.

SteveIreland: I understand your point and totally agree with you that maintainability is a goal.  But from my perspective the "fix" to improve maintainability is to remove migration steps that no longer make sense for the 3.3 design.

Rather than add more complexity by creating migration steps that exist solely to undo earlier migration steps (and in which case which should only be done in a new clean install) it would keep things cleaner to just remove the migration steps that make no sense for a 3.3 install.

Existing sites that were configured for customer sales tax will not want a migration script that undoes anything because it not anything that needs to be fixed.  The earlier migration script that needs to be removed was intended to facilitate a new site install.  Since this feature is now more generic the migration script that "primed the pump" for a conventional customer sales tax configuration should not be run.

daN4cat: Agree with you, those tables.sql changes should not have happened, but those errors are not fatal and can be ignored because are just a change redundancy.
Probably we should spend time and clean up table.sql of what is in the migration script, but it's a painful process. I personally don't have the time for that.

My only concern about what you said in the previous post is that people need to manually delete tax rate slabs.
I must admit that in the rush I might have misread that as I interpreted like a manual database operation. Either way it sounds like you need a manual operation process defined somewhere and expect the number of people posting issues and complaints.

SteveIreland: It will be an interesting upgrade for those who have configured their system for customer based sales tax in an earlier release.  I don't believe it will be more than a handful since the basic sales tax system should be handling 95% of all sites.  I'm happy to answer any questions regarding migration and will bear the burden of any fussing.  Hopefully no one has used OSPOS in a truly complex tax scenario but that's what the Issue system is for.

The manual cleanup I was referring to is simply using OSPOS to redefine their tax rates since after migration the `tax_rates` table will be empty. The major new factor in this release process is tracking taxes by tax jurisdiction.  You might remember that there was one U.S. site using OSPOS that insisted that jurisdictions had to be included separately on the invoice.  I disagreed yet I was going to go ahead and do it anyway but simply ran out of time (it was part of the original design).  With this change he now has exactly what he originally wanted. I definitely hope to hear a "I told you so" from him.

application/views/taxes/tax_jurisdictions.php
```diff
@@ -0,0 +1,144 @@
...
+            rules:
+                {
+		<?php
+		$i = 0;
+
+		foreach($tax_jurisdictions as $tax_jurisdiction=>$tax_jurisdiction_data)
+		{
+		?>
+		<?php echo 'tax_jurisdiction_' . ++$i ?>:
+        {
```
daN4cat: there are spacing / indentation issues in this section

application/views/taxes/tax_categories.php
```diff
@@ -0,0 +1,136 @@
...
+                        $("#tax_categories").load('<?php echo site_url("taxes/ajax_tax_categories"); ?>', init_add_remove_tax_categories);
+                    },
+                    dataType: 'json'
+                });
+            },
+
+            errorLabelContainer: "#tax_category_error_message_box",
+
+            rules:
+                {
```
daN4cat: spacing / indentation issues

SteveIreland:
I guess you don't appreciate my Salvador Dali code style.  So I cleaned those two files up a bit.

daN4cat:
I'm a minimalist.

jekkos:
Didn't have time yet to go through this one, but I think that the testing or validation of this story will probably be the most difficult part. There might be quite a couple of scenarios we need to cover. For this of course some unit tests would be great.. but setting up this infrastructure might be a bit out of scope for your PR.

However I think this effort would certainly pay off in the long run (as my accountant frequently finds wrong calculations in the reports..

However we might want to install this one on demo by default once attrbutes is merged

jekkos:
@SteveIreland if you push the branch to a branch in opensourcepos it will get it's own bintray artifact + deployment on demo. Lets do that after attributes is merged.

odiea:
This is what error displays when completing a Sale. Where is the print sequence?

A Database Error Occurred
Error Number: 1048
Column 'print_sequence' cannot be null
INSERT INTO `ospos_sales_taxes` (`sale_id`, `tax_type`, `tax_group`, `sale_tax_basis`, `sale_tax_amount`, `print_sequence`, `name`, `tax_rate`, `sales_tax_code_id`, `jurisdiction_id`, `tax_category_id`, `rounding_code`) VALUES (8, '', '7.5% ', '20.0000', 1.4, NULL, 'Sales Tax', '7.5000', '1', '0', '1', '1')
Filename: C:/wamp64/www/india/vendor/codeigniter/framework/system/database/DB_driver.php
Line Number: 691

And here is a Sale. It appears that the tax is not being included in the total.
2018-10-14 08_37_52-window

odiea:
It also displays 3 to 4 errors when trying to suspend a Sale.

odiea:
It does add it to the Total if I use the Standard tax 1 but the receipt displays errors.

SteveIreland:
@odiea Thanks, I finally rebased it over the latest master and just now spotted your comments.  I'll see if I can replicate your results tomorrow, but it sounds like it is similar to the MariaDB issue that is mentioned elsewhere since that was also related to building the print sequence - apparently I have a fatal flaw there.   Do you recall which flavor of taxing you were testing?

PR #2166, opened 2018-09-27
Fix issue with deleted items not being included in reprinted receipts.

SteveIreland:
Here is a permanent fix to the "deleted item missing from reprinted receipt".  

The other fix allowed deleted items to be added to a sale if the items were added by item id or by number. But it provided a simple manual code fix that allowed receipts to be correctly reprinted.

This more comprehensive fix will not allow deleted items to be added to a new sale while allowing deleted items to be included in a reprinted sale.


SteveIreland:
This should be a low risk merge.  I've already run it through the basic test cycle.

SteveIreland:
I am unclear on what is being asked.  Since what I did is a common pattern used in the system are you asking that I change add_item to avoid Boolean or just add another variation of the retrieve so that there is a retrieve_all and a retrieve_active?

jekkos:
Ah no it just reminded me of some good old clean code practice. But don't know if there is any cleaner or better way to handle this case.

jekkos:
I would suggest use a separate method name for the deleted items only, but then quite a lot of other code would need to be duplicated

SteveIreland:
I will keep this in mind for future development, but in the change here I don't have a need to retrieve deleted items only.

Issue #2088, opened 2018-08-19
[Epic] Credit Receivings


emi-silva:
## Context
In the organic store where I'm currently working, receivings aren't usually paid for immediately. Instead, they're paid on a _factura vencida_ basis, which means that, when the supplier brings new products to the store, she only gets paid for the products she had brought last time. So we need a way to pay for receivings on credit. Or, put in accounting words, we need an Accounts Payable functionality.

## Design
Given that, for this use case, the receiving of the products takes place in a different moment than the transfer of money, we need a way to differentiate these two actions. In order to do that, I propose the following sequentiality from the user's point of view:
1. The supplier brings new products to the store. The store owner goes to the Receivings module, registers the products, and chooses "Due" as a payment type. The supplier's balance increases.
2. The store owner wants to know how much she owes the supplier. She goes to the Reports module, choses "Accounting" and sees a list of suppliers and their balances. The bigger the balance, the more the supplier is owed.
3. When the payment is due, the store owner makes a bank transfer to the supplier. She then goes to the Expenses module and registers the payment she just made. The supplier's balance decreases.

## Implementation Steps
- [x] 1. In the Receivings module, add a new payment type called "Due", just like in the Sales module. This was implemented in #2080.
- [ ] 2. Modify the Expenses module so that its Supplier field is associated with actual suppliers in the database, i.e., add a foreign key from `ospos_expenses` to `ospos_suppliers` and display a dropdown list with the suppliers, just like in the Receivings module.
- [ ] 3. Create a new report called Accounting, where you can see a list of suppliers and their balances. Balances are calculated by subtracting expenses paid to each supplier, from the "Due" receivings associated to them. The more you owe them, the higher their balance.
- [ ] 4. Make it impossible to modify most fields of past expenses so that they become more legally binding, and generate expense receipts. This should be very similar to the "Daily Sales" functionality.

## Discussion
An early proposal and discussion of this improvement can be found in Pull Request #2080. Here are some highlights:

### Accounting in OSPOS
@jekkos [points out](https://github.com/opensourcepos/opensourcepos/pull/2080#discussion_r210785171) that there has been a lot of discussion on whether OSPOS should have accounting functionalities. Although I'm not familiar with that discussion, I have the belief that it should. Accounting is a basic need for any business and, at least in my experience, small businesses who benefit from OSPOS also need basic accounting functionalities very badly. I regard this improvement as a first step toward that goal.

### Purpose of the Expenses module
[...]

PR #2178, opened 2018-10-12
Linked Expenses with Suppliers

emi-silva:
This pull request corresponds to the third implementation step from Epic #2088.


application/views/expenses/form.php
```diff
@@ -201,13 +251,16 @@
...
+		ignore: '',
+
 		rules:
 		{
 			category: 'required',
 			date:
 			{
 				required: true
 			},
+			supplier_id: 'required',
```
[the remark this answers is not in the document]
emi-silva: I would say so, what do you think?

application/migrations/20180819195000_expenses_supplier_id.php
```diff
@@ -0,0 +1,97 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
```
daN4cat: Please use the SQL upgrade script, I moved your script to SQL only from the previous commit as a courtesy but I would like to avoid to do it every time.

At least for vs 3.3.0 we keep that way, for future versions we will do differently and probably using a feature based as long as we don't have inter dependencies.

The rationale for this change is that if we touch the SQL date we need to align this too to avoid this upgrade to start before the other and fail.

application/views/expenses/form.php
```diff
@@ -187,6 +197,46 @@
...
+
+	$('#remove_supplier_button').click(function() {
+		$('#supplier_id').val('');
+		$('#supplier_name').removeAttr('readonly');
+		$('#supplier_name').val('');
+		$(this).css('display', 'none');
+	});
+
+	<?php
+		if(!empty($expenses_info->expense_id)) {
```
daN4cat: no reason for this extra TAB in the whole section.

application/views/expenses/form.php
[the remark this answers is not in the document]
daN4cat: Ehm... I would prefer not having it mandatory because some people might just want to capture some petty cash transaction without being a supplier.
Let say you refund an employee of an extra expense or similar.

public/css/bootstrap.autocomplete.css
```diff
@@ -3,7 +3,7 @@
     position: absolute;
     top: 100%;
     left: 0;
-    z-index: 1000;
+    z-index: 1050;
```
daN4cat: what's the purpose of this change?

application/views/expenses/form.php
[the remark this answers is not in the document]
emi-silva: OK

application/migrations/20180819195000_expenses_supplier_id.php
emi-silva: OK

application/views/expenses/form.php
emi-silva: OK

public/css/bootstrap.autocomplete.css
emi-silva: This is necessary so that the dropdown that lets you choose the supplier appears in front of the New Expense form and not behind.

emi-silva:
@daN4cat and @jekkos: Please check my new commit. Thanks!

application/controllers/Expenses.php
```diff
@@ -139,6 +139,10 @@ public function save($expense_id = -1)
 			'deleted' => $this->input->post('deleted') != NULL
 		);
 
+        if($expense_data['supplier_id'] == '') {
```
jekkos: check indentation, it's not perfectly aligned

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
```diff
@@ -111,3 +111,36 @@ ALTER TABLE `ospos_suppliers`
...
+
+
+--
+-- Link Expenses with Suppliers
+--
+
+-- Add supplier id
+
+ALTER TABLE `ospos_expenses`
+  ADD COLUMN `supplier_id` int(10) NULL;
```
jekkos: do we have the same datatype as the id column in suppliers table?

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
```diff
@@ -111,3 +111,36 @@ ALTER TABLE `ospos_suppliers`
...
+
+-- Link suppliers
+
+UPDATE `ospos_expenses`
+  INNER JOIN `ospos_suppliers`
+    ON `ospos_expenses`.`supplier_name` = `ospos_suppliers`.`company_name`
+SET `ospos_expenses`.`supplier_id` = `ospos_suppliers`.`person_id`;
+
+-- Save name in description for those expenses whose supplier isn't registered
+
```
jekkos: @daN4cat was the idea to stick to using 3.2.0 to 3.3.0 sql for this version only?

application/views/expenses/form.php
```diff
@@ -228,14 +228,14 @@
...
 	<?php
-		if(!empty($expenses_info->expense_id)) {
+	if(!empty($expenses_info->expense_id)) {
 	?>
-			$('#supplier_id').val('<?php echo $expenses_info->supplier_id ?>');
-			$('#supplier_name').val('<?php echo $expenses_info->supplier_name ?>');
-			$('#supplier_name').attr('readonly', 'readonly');
-			$('#remove_supplier_button').css('display', 'inline-block');
+		$('#supplier_id').val('<?php echo $expenses_info->supplier_id ?>');
+		$('#supplier_name').val('<?php echo $expenses_info->supplier_name ?>');
```
jekkos: I think you can chain both calls like attr().val()

jekkos:
Looks good, just some minor remakrs

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
emi-silva: Yes. You can check this with the following SQL command:

    > show columns from ospos_suppliers where Field = 'person_id';
    +-----------+---------+------+-----+---------+-------+
    | Field     | Type    | Null | Key | Default | Extra |
    +-----------+---------+------+-----+---------+-------+
    | person_id | int(10) | NO   | MUL | NULL    |       |
    +-----------+---------+------+-----+---------+-------+
    1 row in set (0.00 sec)

application/views/expenses/form.php
emi-silva: OK

emi-silva:
@jekkos I just added a commit with your requested changes. Thank you very much.

application/controllers/Expenses.php
```diff
@@ -139,6 +139,10 @@ public function save($expense_id = -1)
 			'deleted' => $this->input->post('deleted') != NULL
 		);
 
+		if($expense_data['supplier_id'] == '') {
```
daN4cat: We try to be consistent with the rest of the code, even if it's an annoying detail...

emi-silva: OK, sorry about that...

emi-silva:
@daN4cat @jekkos please review my last commit. I think we're ready to merge? I can squash the commits if necessary

daN4cat:
Looks fine, I will merge the code in the next days. First I need to test and then merge.

Thank you for the contribution.

emi-silva:
You're very welcome. Thank you for your reviews and for testing and merging and everything you do to make OSPOS work.

emi-silva:
@daN4cat I can't reproduce the error you mention. I pulled your commits and tested it, and it works fine for me. Then I cloned a new repository, created a new database, then created a supplier, and then in Expenses the suggestion dropdown showed it correctly. Please show me the steps to reproduce this error so I can correct it. On Sunday I'm leaving for a long trip through South America and won't be able to do any coding for a few months.

SteveIreland:
Should this pull request be closed?

PR #2194, opened 2018-10-20
Add support for India GST

SteveIreland:
Made a few corrections and I'm hoping that this path to a pull request will create a bintray entry (if I'm doing it correctly).


SteveIreland:
Yeah!  It created a bintray entry which I will be using to continue my tests this weekened.

Issue #2221, opened 2018-11-09
Default Attributes Option


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 465a2c
- PHP version is: 5.6
- MySQL or MariaDB version is: MySQL 5.7.2
- OS and version is: FreeBSD 11.0
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Bug
[...]

PR #2237, opened 2018-11-23
Keep attributes selected when creating new items (#2221)

jekkos:
Did some minor change that helps when entering batches of new items in receivings or sales module. The items form will stay open when a user keeps hitting 'new' or when he uses a barcode scanner to finalize an item entry every time. The barcode field will be cleared out while attribute fields remain filled.

When finished entering new items, submit has to be clicked, which will then refresh the sale register or orders view that will then contain all the new entered items with a quantity equal to 1.


objecttothis:
I'm back from vacation today and will look at these code changes and the others you requested either today or tomorrow.  Sorry for the delay.  If it needs to get merged before, I completely understand.

Issue #1461, opened 2017-08-11
Customer Display


lorenzadam:
### New Feature / Enhacement

*For New Features or Enhacements please clearly write your proposal.
It's important to state whether you expect the community to implement it or you will contribute the work.
Please bear in mind that we will implement new features only on the current code, there is no support for older versions.*

i think it would be great if ospos have customer display features. The computer are using two monitors, so customer can see the items (and prices) they have bough while in the chasier.

daN4cat:
Once somebody suggested a solution: use a Raspberry Pi with a screen and a node.js server listening to a SocketIO connection. Add to OSPOS the capability to post over SocketIO the values connecting to the Raspberry PI. SocketIO client is a javascript running on your web browser, in that way it can connect to the locally installed RaspberryPI. Of course you would need to add the IP configuration of your RaspberryPI to the OSPOS configuration.

Quite some work, so if you want to see it done you better learn coding and help here contributing that feature.

lorenzadam:
Do you mean to use one computer and a raspberry pi?
Because not every one in their business are using easpberry pi

daN4cat:
It's an option that was discussed a while back, not suggesting is the solution.

lorenzadam:
@daN4cat Is there a big risk if using auto refresh every n seconds with pure php and html?

SteveIreland:
Someone can correct me on this, but I believe the real problem is session control.  The active order is stored in a session and not persisted to disk until it's saved.  Another browser would instantiate another session  To enable a feature such as this then I believe it would be required to persist each change to disk and then the auto-refresh option might have a chance of working.

daN4cat:
Indeed, the data is stored in a session associated with the browser. So everything needs to be controlled from the browser client dealing with the transaction.

lorenzadam:
@daN4cat I try to make a sale and enter the items then open a new browser window and open the sales module, the list of items that have been included still appear.

Maybe an auto refresh solution with html can be done, by creating another user interface but retrieving data from the same source.

![cuplikan layar dari 2017-08-16 09-28-23](https://user-images.githubusercontent.com/5461309/29344913-51d12d7a-8265-11e7-9281-3fee9d26a84c.png)

daN4cat:
Yes the some browser can duplicate the view and retain the same session.
So what you could do is to add some java script in the register view that allows open a second window automatically and via ajax call read the total from the sale register controller.
It should not be too difficult.
[...]

PR #2238, opened 2018-11-23
Websocket display support

jekkos:
Putting this out here, still needs simple integration with the config section and sale / receiving module

### DIFF
--- a/js/serial.config.js
+++ b/js/serial.config.js
@@ -1,0 +1,206 @@
+(function($) {
+    var serial_ports = [], baud_rates = [];
+    var settings = (localStorage['serial_settings'] && JSON.parse(localStorage['serial_settings'])) || {};
+    var default_baud_rate = 9600;
+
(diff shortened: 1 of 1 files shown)


objecttothis:
@jekkos, how about this one?  Is this something that you plan to merge later?

jekkos:
Well basically I put this here to add support for #1461.  The code here works in conjunction with a [serial-to-http server written in go](https://github.com/johnlauer/serial-port-json-server), which allows you to hook up devices like scales or scanners on the serial port and read them out using a websocket.

Using it is rather straightforward: You need to call bind_keys() in the view that you want to enable it, and then mark the fields you want to have a value read out in with a css class. After hitting the key the value from the scale or device will appear in this field.

The same setup can be used to show text on a display or the like. It could be triggered by the same key mechanism or by some other event.

objecttothis:
OK, that makes sense to keep the branch.

Issue #2245, opened 2018-11-27
attribute_value column in ospos_attribute_values too small


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 
- PHP version is: 5.6.38
- MySQL or MariaDB version is: 5.7.24
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English

### Issue
[...]

Issue #2230, opened 2018-11-15
Attributes Migration Script tweaks


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 465a2c
- PHP version is: 5.6.38
- MySQL or MariaDB version is: 5.7
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### New Feature
[...]

PR #2246, opened 2018-11-27
Attributes migration tweaks

objecttothis:
This is related to #2230 and #2245


objecttothis:
Lets not merge this yet.  Some of my tests are failing.

objecttothis:
Nevermind.  It passes all checks.  If at least one other person wants to try this modified migration script it wouldn't hurt.  What they should see is that no "" (blank but not NULL) custom fields get migrated over and no values larger than 45 get truncated (should be good up to 255).

PR #2289, opened 2018-12-27
Refactor Tax Configuration

SteveIreland:
This is some rework on the tax configuration in order to simplify the code.  This is about all I have time for this year for rework.  I believe it addresses a couple of the requested changes and certainly makes it easier to understand and maintain.


application/views/taxes/tax_categories.php
```diff
@@ -126,7 +120,7 @@
 			foreach($tax_categories as $tax_category=>$tax_category_data)
 			{
 			?>
-			<?php echo 'tax_category_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_category_required'); ?>",
+				<?php echo 'tax_category_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_category_required'); ?>",
```
jekkos: the input field names have changed, so I guess that needs to be reflected here?

application/views/taxes/tax_codes.php
```diff
@@ -129,7 +122,7 @@
 				foreach($tax_codes as $tax_code=>$tax_code_data)
 				{
 				?>
-				<?php echo 'tax_code_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_code_required'); ?>",
+					<?php echo 'tax_code_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_code_required'); ?>",
```
jekkos: same remark, not sure if it will show the correct message now

application/views/taxes/tax_jurisdictions.php
```diff
@@ -137,7 +125,7 @@
 				foreach($tax_jurisdictions as $tax_jurisdiction=>$tax_jurisdiction_data)
 				{
 				?>
-				<?php echo 'tax_jurisdiction_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_jurisdiction_required'); ?>",
+					<?php echo 'tax_jurisdiction_' . ++$i ?>: "<?php echo $this->lang->line('taxes_tax_jurisdiction_required'); ?>",
```
jekkos: double check this one as well

SteveIreland: The name hasn't changed since I still need to generate an id for that field and that appears to be what the JavaScript is using, but I spotted an oddity in the way the rules are being generated because the rules are not being added for rows that are "plus'd in" because they (the rules) are being added only for rows generated from PHP (which includes the first empty row that is used to "prime the pump".  That's what I deserve for copying patterns without taking the time to fully understand them.

SteveIreland: oops ... still learning here.  Apparently I am mistaken again.  The validation plug-in uses name instead of id. Seems wrong to me but I'm adapting.

jekkos: I think it's probably easiest to use a class tag for the input fields that links to a validator. This will then simplify the logic quite a bit (you won't need the php echo anymore in this case).

application/controllers/Taxes.php
```diff
@@ -396,39 +408,19 @@ public function suggest_tax_codes()
...
-		$array_save= array();
+		$tax_code_id = $this->input->post('tax_code_id');
+		$tax_code = $this->input->post('tax_code');
+		$tax_code_name = $this->input->post('tax_code_name');
+		$tax_code_id = $this->input->post('tax_code_id');
+		$city = $this->input->post('city');
+		$state = $this->input->post('state');
 
-		foreach($this->input->post() as $key => $value)
+		$array_save= array();
```
daN4cat: missing space before =

application/controllers/Taxes.php
```diff
@@ -16,8 +16,20 @@ public function __construct()
 	public function index()
```
daN4cat: I noticed that none of the new tax function have an xss_clean call in this function

daN4cat:
@SteveIreland would you address the review feedback so we can merge this over the weekend? Thanks.

SteveIreland:
I've spend numerous hours on this.  It's been educational but frustrating.  The jQuery validation plug-in has some squirreleness that I've been fighting.  It could be due to ignorance but I've about put all I can into this. In my mind it is a lot cleaner and working much better than my last iteration on this.

Give it a shot ... if it doesn't work I won't cry if you decide to pull the plug on it and if  you can spot where I'm doing something wrong I won't hurt my feelings if you want to apply some improvements.

I was "heads down" on the rewrite over the last couple of days so I missed some of the latest comments.  I'll take a look at them when I'm fresher.

application/controllers/Taxes.php
SteveIreland: ok

application/controllers/Taxes.php
SteveIreland: ok

application/views/taxes/tax_categories.php
```diff
@@ -41,22 +41,16 @@
...
 			id = id.replace(/.*?_(\d+)$/g, "$1");
 
-			var previous_tax_category_id_id = 'tax_category_id_' + id;
 			var previous_tax_category_id = 'tax_category_' + id;
-			var previous_default_tax_rate_id = 'default_tax_rate_' + id;
-			var previous_tax_group_sequence_id = 'tax_group_sequence_' + id;
 			var block = $(this).parent().clone(true);
 			var new_block = block.insertAfter($(this).parent());
-			var new_tax_category_id_id = 'tax_category_id_' + ++tax_categories_count;
+            ++tax_categories_count;
```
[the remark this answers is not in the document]
SteveIreland: ok

application/views/taxes/tax_categories.php
```diff
@@ -41,22 +41,16 @@
...
-			$(new_block).find("input[name='"+previous_tax_category_id_id+"']").attr('name', new_tax_category_id_id).val('-1');
-			$(new_block).find("input[id='"+previous_tax_category_id+"']").attr('id', new_tax_category_id).removeAttr('disabled').attr('name', new_tax_category_id).attr('class', 'form-control input-sm').val('');
-			$(new_block).find("input[id='"+previous_default_tax_rate_id+"']").attr('id', new_default_tax_rate_id).removeAttr('disabled').attr('name', new_default_tax_rate_id).attr('class', 'form-control input-sm').val('');
-			$(new_block).find("input[id='"+previous_tax_group_sequence_id+"']").attr('id', new_tax_group_sequence_id).removeAttr('disabled').attr('name', new_tax_group_sequence_id).attr('class', 'form-control input-sm').val('');
+			$(new_block).find("input[name='tax_category[]']").attr('id', new_tax_category_id).removeAttr('disabled').attr('class', 'form-control input-sm').val('');
+			$(new_block).find("input[name='tax_group_sequence[]']").removeAttr('disabled').attr('class', 'form-control input-sm').val('');
+            $(new_block).find("input[name='tax_category_id[]']").val('-1');
```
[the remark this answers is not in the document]
SteveIreland: ok

application/views/taxes/tax_categories.php
SteveIreland: I ended up reworking this entire section of code.  I'll add a comment later to explain why.

application/views/taxes/tax_codes.php
SteveIreland: reworked

SteveIreland:
@daN4cat I believe it can be merged now.

daN4cat:
I'll look into merging it this evening. I need to fix a minor thing before merging it.

jekkos:
Thanks for adressing these comments! code looks quite a bit cleaner now. Just one final remark ..

application/models/Tax_category.php
```diff
@@ -230,6 +230,14 @@ public function get_tax_category_suggestions($search)
 		return $suggestions;
 	}
 
+	public function get_empty_row()
```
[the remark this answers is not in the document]
daN4cat: For an empty object example  look in people model.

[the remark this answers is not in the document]
SteveIreland: The empty object pattern is primarily used in the "get_info" methods where a single object is what is intended to be returned and if the object is not found the method returns an empty object.

That's not what is happening here.  The intent here is to return an array.  It doesn't seem helpful to create an object and then retrieve the same properties from that object to create the array needed by CodeIgniter (the "get_result_array" doesn't work here).   I would be okay with it if there was an elegant solution to convert an object to array but that doesn't seem to be available.

I started to leave this method in the controller but decided at the last minute that even though it doesn't deal with the database, since the structure should mirror the table then it would be better to leave it with the model.  The advantage of being in the model is that it would  be more obvious that the method would need to be changed if the table structure had to be changed.

I went ahead and rebased this over the latest changes to master and pushed it up.

I started to code the changes you requested but since I couldn't find an a way to convert the object that I would be creating to an array without effectively creating twice the amount of code to no advantage I just backed out my changes until I get a clarification from you.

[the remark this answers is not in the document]
daN4cat: It's fine with me to leave it as it is.

daN4cat:
I merged this PR as I believe we were all good with this.

If any bug in master, @SteveIreland will fix it.

Massive thank you for all the work @SteveIreland.

application/models/Tax_category.php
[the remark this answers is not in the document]
jekkos: there is an elegant solution availailable to convert the object to an assoc array. You would then just add the get_info method and cast the result by prepending `(array)` in front of the result.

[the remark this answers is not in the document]
jekkos: Do you sill have part of the changes you were planning to make? Maybe if you can submit it then I'll finish the refactor

application/views/taxes/tax_categories.php
```diff
@@ -41,51 +41,59 @@
...
-		$.validator.addMethod('valid_chars', function(value, element) {
-			return value.indexOf('_') === -1;
+		$.validator.addMethod('validateTaxCategoryCharacters', function(value, element) {
+			if ((value.indexOf('_') != -1)) {
+				return false;
+			}
+			return true;
 		}, "<?php echo $this->lang->line('taxes_tax_category_invalid_chars'); ?>");
 
+		$.validator.addMethod('requireTaxCategory', function(value, element) {
```
jekkos: why not just use the required rule here instead? just trying to adhere to the DRY principle

application/views/taxes/tax_categories.php
```diff
@@ -41,51 +41,59 @@
...
+			if (value_count > 1) {
+				return false;
+			}
+			return true;
 		}, "<?php echo $this->lang->line('taxes_tax_category_duplicate'); ?>");
 
-		$.validator.addMethod('valid_chars', function(value, element) {
-			return value.indexOf('_') === -1;
+		$.validator.addMethod('validateTaxCategoryCharacters', function(value, element) {
+			if ((value.indexOf('_') != -1)) {
```
jekkos: `return value.indexOf('_') == -1`

application/views/taxes/tax_categories.php
```diff
@@ -41,51 +41,59 @@
...
 		// run validator once for all fields
-		$.validator.addMethod('tax_category' , function(value, element) {
+		$.validator.addMethod("check4TaxCategoryDups" , function(value, element) {
 			var value_count = 0;
-			$("input[name*='tax_category']").each(function() {
+			$('input[name="tax_category[]"]').each(function() {
 				value_count = $(this).val() == value ? value_count + 1 : value_count;
 			});
-			return value_count < 2;
+			if (value_count > 1) {
```
jekkos: `return value_count =< 1`

application/views/taxes/tax_categories.php
```diff
@@ -96,41 +104,25 @@
...
-			messages:
-			{
-			<?php
-			$i = 0;
+		<?php
+		$i = 0;
+		foreach($tax_categories as $tax_category=>$tax_category_data)
+		{
+		?>
+		$('<?php echo '#tax_category_' . ++$i ?>').rules( "add", {
```
jekkos: I don't think you explicitly need to enable the rules when you add a class with the validator name to the input fields?

application/views/taxes/tax_codes.php
```diff
@@ -38,59 +38,64 @@
...
 		// run validator once for all fields
-		$.validator.addMethod('tax_code' , function(value, element) {
+		$.validator.addMethod('check4TaxCodeDups' , function(value, element) {
 			var value_count = 0;
-			$("input[name*='tax_code']").each(function() {
+			$("input[name='tax_code[]']").each(function() {
 				value_count = $(this).val() == value ? value_count + 1 : value_count;
 			});
-			return value_count < 2;
+			if (value_count > 1) {
```
jekkos: same remark as above

application/views/taxes/tax_codes.php
```diff
@@ -38,59 +38,64 @@
...
+				return false;
+			}
+			return true;
 		}, "<?php echo $this->lang->line('taxes_tax_code_duplicate'); ?>");
 
-		$.validator.addMethod('valid_chars', function(value, element) {
-			return value.indexOf('_') === -1;
+		$.validator.addMethod('validateTaxCodeCharacters', function(value, element) {
+			if ((value.indexOf('_') != -1)) {
```
jekkos: spotted some code duplication: might be good to extract the validators in a common or partial view here. Eg `view/partial/tax_config.php`. Then include in both views from php this common code.

application/views/taxes/tax_jurisdictions.php
```diff
@@ -42,61 +42,61 @@
...
-			$(new_block).find("select[name='"+previous_tax_type_id+"']").attr('name', new_tax_type_id).removeAttr('disabled').attr('class', 'form-control required input-sm').val('');
-			$(new_block).find("input[id='"+previous_reporting_authority_id+"']").attr('id', new_reporting_authority_id).removeAttr('disabled').attr('name', new_reporting_authority_id).attr('class', 'form-control input-sm').val('');
-			$(new_block).find("input[id='"+previous_tax_group_sequence_id+"']").attr('id', new_tax_group_sequence_id).removeAttr('disabled').attr('name', new_tax_group_sequence_id).attr('class', 'form-control input-sm').val('');
-			$(new_block).find("input[id='"+previous_cascade_sequence_id+"']").attr('id', new_cascade_sequence_id).removeAttr('disabled').attr('name', new_cascade_sequence_id).attr('class', 'form-control input-sm').val('');
-			hide_show_remove();
+			$(new_block).find("input[name='jurisdiction_name[]']").attr('id', new_jurisdiction_name_id).removeAttr('disabled').attr('class', 'form-control required input-sm').val('');
+			$(new_block).find("input[name='tax_group[]']").removeAttr('disabled').attr('class', 'form-control required input-sm').val('');
```
jekkos: are there any fields in the block that you don't want to disable? Otherwise you can combine the calls to add `form-control`and `input-sm` classes as well as the ones to add the `required` class.

application/views/taxes/tax_jurisdictions.php
```diff
@@ -42,61 +42,61 @@
...
 		// run validator once for all fields
-		$.validator.addMethod('tax_jurisdiction' , function(value, element) {
+		$.validator.addMethod('check4TaxJurisdictionDups' , function(value, element) {
 			var value_count = 0;
-			$("input[name*='tax_jurisdiction']").each(function() {
+			$("input[name='jurisdiction_name[]']").each(function() {
 				value_count = $(this).val() == value ? value_count + 1 : value_count;
 			});
-			return value_count < 2;
+			if (value_count > 1) {
```
jekkos: as before, might be better to extract validators (and so no more duplication)

application/views/taxes/tax_jurisdictions.php
```diff
@@ -107,41 +107,24 @@
...
-				?>
-			}
+			errorLabelContainer: "#tax_jurisdiction_error_message_box"
 		}));
+
+		<?php
+		$i = 0;
+		foreach($tax_jurisdictions as $tax_jurisdiction=>$tax_jurisdiction_data)
+		{
+		?>
```
jekkos: also this part seems repetitive and can be extracted, I can do this for you.

jekkos:
@SteveIreland thanks again for the refactor effort! Still have some remarks, Let me know which parts you'e willing to address. I'll pick up my fair share here.

SteveIreland:
Thanks for merging this.  I've been trying to catch up on some personal stuff this week and then I'll be able to get my head back in to it.  I'll review your comments then.

odiea:
@SteveIreland I receive this following error when using a giftcard or rewards  as a payment in the Register.
Message: Undefined index: sales_tax_amount
Filename: libraries/Sale_lib.php
Line Number: 1279
Backtrace:
C:\wamp64\www\my-master\application\libraries\Sale_lib.php
Line: 1279
Function: _error_handler

C:\wamp64\www\my-master\application\libraries\Sale_lib.php
Line: 551
Function: get_total

C:\wamp64\www\my-master\application\controllers\Sales.php
Line: 300
Function: get_amount_due

C:\wamp64\www\my-master\public\index.php
Line: 315
Function: require_once

PR #2307, opened 2019-01-10
Add config item for default order discount

jekkos:
This change makes orders and sales more consistent and adds the 'default discount' feature here as well. Something requested by my users as they tend to get the same discount percentage from many providers. In the end would be nice to match the total receiving price with the invoice for the entered items so the costs can be tracked.


application/language/en-US/common_lang.php
```diff
@@ -42,7 +42,7 @@
 $lang["common_migration_needed"] = "A database migration to %1 will start after login.";
 $lang["common_new"] = "New";
 $lang["common_no_persons_to_display"] = "There are no people to display.";
-$lang["common_none_selected_text"] = "[Select]";
+$lang["common_none_selected_text"] = "Nothing selected";
```
daN4cat: Previous translation was better. I would revert this change.

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
```diff
@@ -132,15 +132,19 @@ SET `ospos_expenses`.`supplier_id` = `ospos_suppliers`.`person_id`;
...
+  ADD CONSTRAINT `ospos_expenses_ibfk_3` FOREIGN KEY (`supplier_id`) REFERENCES `ospos_suppliers` (`person_id`);
 
 -- Delete supplier name
 
 ALTER TABLE `ospos_expenses`
-DROP COLUMN `supplier_name`;
+  DROP COLUMN `supplier_name`;
+
+INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
+  ('default_receivings_discount_type', '0'),
```
daN4cat: This is not execute in current dev installations because the PHP file date didn't change.

daN4cat: Change the date in the filename and it be re-executed

jekkos: but then there might be an ovelrap with the statements that were already in there?

daN4cat: I don't follow. The script is re-executed and just skips on errors.

jekkos: Ok fine then.

odiea:
@SteveIreland  I do not know if this was from a bad merge because the bintray files are not up to date. This is the error I received when trying to pay with a giftcard.

A PHP Error was encountered
Severity: Notice
Message: Undefined index: sales_tax_amount
Filename: libraries/Sale_lib.php
Line Number: 1279
Backtrace:
C:\wamp64\www\my-master\application\libraries\Sale_lib.php
Line: 1279
Function: _error_handler
C:\wamp64\www\my-master\application\libraries\Sale_lib.php
Line: 551
Function: get_total
C:\wamp64\www\my-master\application\controllers\Sales.php
Line: 305
Function: get_amount_due
C:\wamp64\www\my-master\public\index.php
Line: 315
Function: require_once

PR #2310, opened 2019-01-12
Adding expiry date to items

elhadi-kh:
I added the expiry date attribute to the item, it can be very useful for those who has to manage even the expiry dates.


jekkos:
Indeed this functionality is already foreseen with datetime type attributes. However there is still a formatting issue I will need to fix.

jekkos:
Check #2319 for a more final version

SteveIreland:
Isn't expiration date a stock level attribute instead of an item level attribute?  I'm having difficulty imagining a scenario where that would work.

WebShells:
@SteveIreland actually there is no scenario where this could be applicable 100% since during sales nothing could tell to which date this item is related... So somehow the most that can be done is applying a reminder or a new report that could list items where expiry date is close to current month or 2 months later... Any thoughts?

elhadi-kh:
I'm using this software in my own store to manage my products and I'm working on a report to get the product with expiration dates, also I have the idea to use a warning in sales just like the warning for low inventory, the warning will check the expiration date with the system date.

WebShells:
That's not applicable for the same reason I've just mentioned above, unless you are planning to allow duplicate barcodes and assign 2 different stocks of 2 different Expiry date and select which one you'd like to sell upon sale... Too complicated to be accomplished in the right way i think... @SteveIreland agree?

elhadi-kh:
@WebShells The problem with my store is that for the same product I have multi purchase price and multiple selling prices and for each product I have the same name but a different purchase price and also different selling price and different expiry date, So I have to create the product after each purchase ( even if it already exists in the database )

WebShells:
Yes exactly as I said about adding it with allowing duplicate barcodes.... Not very practical...

elhadi-kh:
@WebShells it will not have duplicated barcode, for each product a barcode will be generated. so we will have 2 different products with the same name.

WebShells:
Got you... More and more work for restocking :/

elhadi-kh:
The solution is to create variables for the same product.

SteveIreland:
I believe I understand that @elhadi-kh is trying to satisfy a very specific requirement but it appears that the implementation here seems to trash a lot of rules and features that OSPOS has implemented.  Obviously we all go to a grocery store and check the expiration date and select the item with the latest expiration date, but I've never been to a store where I had to check the prices on the items to make sure I'm selecting the one that has the lowest price.

Software that manages product expiration requires tracking expiration date at a inventory level (not item level). which we don't currently provide support for.

I agree with @WebShells that this is not a candidate for merging.

PR #2319, opened 2019-01-25
Fix attribute datetime formatting

jekkos:
Fix for attribute datetime formatting. Did a quick test on items listing and attribute datetime columns are formatted according to what a users specifies as date / time format. Might also need to check sales and receivings and eventually reports.


odiea:
Ignore the right hand edit columns. I moved them to the left.

jekkos:
Ok thanks for testing, will need to check some other scenarios but couldn't way to put it out here.

odiea:
I also receive the following error when adding an item in the Register.
A Database Error Occurred
Error Number: 1064

You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' GROUP_CONCAT(DATE_FORMAT(attribute_datetime, '%m/%d/%Y :%i:%S ') SEPARATOR ', '' at line 1

SELECT GROUP_CONCAT(attribute_value SEPARATOR, GROUP_CONCAT(DATE_FORMAT(attribute_datetime, '%m/%d/%Y :%i:%S ') SEPARATOR ', ') AS attribute_datetimevalues FROM `ospos_attribute_links` JOIN `ospos_attribute_values` ON `ospos_attribute_values`.`attribute_id` = `ospos_attribute_links`.`attribute_id` JOIN `ospos_attribute_definitions` ON `ospos_attribute_definitions`.`definition_id` = `ospos_attribute_links`.`definition_id` WHERE `definition_type` <> 'GROUP' AND `deleted` = 0 AND `sale_id` IS NULL AND `receiving_id` IS NULL AND `item_id` = 1 AND definition_flags & 2

Filename: C:/wamp64/www/newdate/vendor/codeigniter/framework/system/database/DB_driver.php

Line Number: 691

odiea:
It's getting better. The Items table needs more work. If I just use the Datetime attribute it does not show on the table until another attribute is selected which may be why the table does not refresh after selecting the Datetime attribute.

daN4cat:
what's the status of this?

jekkos:
Still have some pending changes here that I need to test further. Also wonder whether we should keep the time in the datetime type. @objecttothis do you actually use it that way or would a regular datepicker suffice?

objecttothis:
> 
> 
> Still have some pending changes here that I need to test further. Also wonder whether we should keep the time in the datetime type. @objecttothis do you actually use it that way or would a regular datepicker suffice?

@jekkos For our use, we don't need a full timestamp it as we are dealing with the publication date.  So then would you change the database structure to have the attribute_datetime be a DATE in MySQL instead of DATETIME?  Also, does it make sense then to refactor our code and the name of the column to be DATE and attribute_date respectively?  Let me know what you end of deciding on just storing the date and I will adjust the regex accordingly.  The only scenario in which I could forsee the need for a full DATETIME is if we expanded attributes in the future to include Inventory attributes. In that case it could be useful to have a datetime which tells us the hour and minute an item was received, quantity was changed, report was run, etc.

Alternately we could just keep the database as is for future compatibility and ignore the hour, minute and second on the software end.  MySQL will accept YYYY-MM-DD in a datetime without the time portion.  It just stores zeros in the time portion.

jekkos:
As discusses in other issues, usage of this type of attribute is rather limited as an expiry date should be linked to invntory and not to the item `class` itself. So for now I might just make this a date instead.
Still have some testing to do here but that won't be before beginning next week, as i'll be off to London for a couple of days ...

application/models/Attribute.php
```diff
@@ -395,16 +381,13 @@ public function save_value($attribute_value, $definition_id, $item_id = FALSE, $
 	{
 		$this->db->trans_start();
```
jekkos: @objecttothis this it the fix I was talking abouit. You will need to add in another case for decimals then.

daN4cat:
@jekkos what happened to this?

jekkos:
Yes I have to find the time to look into this, it's still on my list. I hope somewhere next week.

application/helpers/locale_helper.php
```diff
@@ -456,6 +456,50 @@ function dateformat_momentjs($php_format)
...
+		'o' => '',
+		'Y' => '%Y',
+		'y' => '%y',
+		// Time
+		'a' => '',
+		'A' => '%p',
+		'B' => '',
+		'g' => '%l',
+		'G' => '%k',
+		'h' => '',
```
skaman123: fill-in this locking associative-array 'h'=>'%H'
in mysql_db -> ospos_app_config -> time_format = 'h:i:s A'

application/helpers/tabular_helper.php
```diff
@@ -529,15 +530,22 @@ function expand_attribute_values($definition_names, $attribute_values)
 	$indexed_values = array();
 	foreach($values as $attribute_value)
 	{
-		$exploded_value = explode(':', $attribute_value);
-		$indexed_values[$exploded_value[0]] = isset($exploded_value[1]) ? $exploded_value[1] : '-';
+		$exploded_value = explode('_', $attribute_value);
+		if (sizeof($exploded_value) > 1)
+		{
+			$indexed_values[$exploded_value[0]] = $exploded_value[1];
+		}
```
skaman123: it works on mine tiny fix/change on a special case if DATETIME attribute_value occur

           //special case for DATETIME attribute_value only.  eg: from this '2120/12/31 11'  turns to -->  '12/31/2120' $dt_value = to_datetime(strtotime(explode(' ', $exploded_value[1])[0]), FALSE, 'm/d/Y');
$exploded_value[1] = preg_match('/[0-2][0-1][0-9][0-9]\/[0-1][0-9]\/[0-3][0-9]/i', $exploded_value[1])? $dt_value : $exploded_value[1];
$indexed_values[$exploded_value[0]] = isset($exploded_value[1]) ? $exploded_value[1] : '-';

skaman123: i change a little on to_datetime() function.

function to_datetime($datetime, $flag = TRUE, $dt_format="")
{
              $config = get_instance()->config;
              if($flag == TRUE) return date($config->item('dateformat') . ' ' . $config->item('timeformat'), $datetime);	
	
     if(!empty($dt_format)) return date($dt_format, $datetime);
     return date($config->item('dateformat'), $datetime);
}

daN4cat:
@jekkos please give a kick to this so we can wrap up and release 3.3.0 soon.

application/helpers/tabular_helper.php
UncleBong: appreciated sir seems working i just put this little fix code it runs on me!
this is just for sample! i used all types of attributes in one records to test
image

in sales register when i add items i've notice that in table field 'Item Name'  w/ lots of attribute values have been setup particularly on this field was bloated with long description
before: 
image

after: i add some tooltip on this table field & view the attribute_values there everytime we hover it
image
correct me if i wrong sir i put some little modifications "views/sales/register.php" and i add some little helper functions

UncleBong: these are code functions that i add&modify! i just print screen it
for clear reference!. :)

at tabular_helper.php
image

image

at locale_helper.php
image

at register.php & receiving.php 
image

objecttothis:
@jekkos it appears that @UncleBong has proposed the changes that you will need to make this work.

daN4cat:
Yep, we need to get to an end of this release. It's been a very long time since we started and it doesn't look like we are getting near to close it.

jekkos:
Ok ill try to give a kick this weekend

daN4cat:
@jekkos managed to get it done?

I really want to close this 3.3.0!

jekkos:
I did give a kick but wasn't hard enough yet.. converted datetime to be a date but there's some bugs (values not showing after a call to search()). Quite a busy week so hope to pick it up later again.

application/helpers/tabular_helper.php
jekkos: Actually I like the tooltip idea, but it might be good to use it only when the content is too long.

jekkos:
Ok had another look and I found no more issues for now, dates are showing and refreshing as it should in items listing now + fixed another bug in the attribute form.

daN4cat:
@objecttothis @UncleBong do you have time to review and test the latest code from @jekkos and see if we are fine now?

objecttothis:
> @objecttothis @UncleBong do you have time to review and test the latest code from @jekkos and see if we are fine now?

I'll install the branch and test it.

objecttothis:
@daN4cat I went to bintray but there doesn't seem to be any builds for this pull request.  If there is another way for me to download a build and test it, please let me know.

jekkos:
Ok ic, account is rate limited again, alas

daN4cat:
Probably you need to clean up from old builds?

objecttothis:
> Ok ic, account is rate limited again, alas
That explains why I'm getting "There are no files yet for this package.
You can browse files after they are published by the package owner.
Check back soon."

Is there a way to keep fewer builds in the account so that we don't hit the size limit?

jekkos:
I removed all the builds and mailed back to jfrog, hope they respond soon.

jekkos:
Ok rate limitation was removed, I'm rerunning the build now.

objecttothis:
> Ok rate limitation was removed, I'm rerunning the build now.

I'll see if I can download and test it later tonight.

odiea:
Nothing was showing in bintray so I had to  download it from the master. 
It is better now. I had to remove the old date attribute and add date back and it shows fine in the Register with only 1 Date showing.
Items table still loads very slowly with attributes .
When I add an Item id at the register I get the following error.
A Database Error Occurred
Error Number: 1054
Unknown column 'company_name' in 'field list'
SELECT `item_id`, `name`, `pack_name`, `company_name`, `name`, `cost_price` FROM `ospos_items` WHERE `deleted` = 0 AND `item_type` IN(0, 2) AND `name` LIKE '%1%' ESCAPE '!' ORDER BY `name` ASC
Filename: C:/wamp64/www/datetime/vendor/codeigniter/framework/system/database/DB_driver.php

odiea:
Please ignore the db error. I changed the db and it the register works without errors.

jekkos:
@odiea how many items do you have in the items table? I might give the performance another look then and try on a dataset of my own first to see how it compares. I mean you could easily benchmark attributes by deselecting the items flag in the definition. It will omit the JOIN in the query and run it as before. Curious if that is where the perf hit is coming from.

jekkos:
And indeed bintray is still disabled.. quite annoying that their software doesnt remove the limitation after the account was wiped, so I'll have to wait for their reply again.

odiea:
I only have 188 items and with the attributes enabled in Items it take 3.6 seconds to show the table. With no attributes in items table it shows almost immediately.

objecttothis:
I have over 14800 items in my database, but I can't test until bintray gets re-enabled.  It's too bad JFrog is a little slow to re-enable the account.  @jekkos you said that the rate-limit had been exceeded.  Do you mean storage size limit or rate-limit as in number of downloads, or bandwidth used?  I ask because if this keep happening we might want to consider alternate solutions to reduce the risk.

application/models/Attribute.php
```diff
@@ -270,11 +270,11 @@ private function check_data_validity($definition, $from, $to)
 			$this->db->where('definition_id',$definition);
 			$success = TRUE;
 
-			if($to === DATETIME)
+			if($to === DATE)
 			{
 				foreach($this->db->get()->result_array() as $row)
 				{
-					if(valid_datetime($row['attribute_value']) === FALSE)
+					if(valid_date($row['attribute_value']) === FALSE)
```
UncleBong: We forgot this function valid_date
function valid_date($date){
return preg_match('/^(([0-9]{2,4})(\\/|-)([0-1][0-9])(\\/|-)([0-3][0-9])|([0-1][0-9])(\\/|-)([0-3][0-9])(\\/|-)([0-9]{2,4}))?$/', $date);
}

seems ok sir for now but for me in the future use its better to stick to DATETIME format cuz in this function we can choose either DATE or DATETIME format if we refactor this function nicely! And also what if, if we change the config to another 'dateformat' type that we rely on?
 sample valid dates: 2001-01-01, 2001/01/01, 12-31-2001, 12/31/2001 01:01:01

application/helpers/tabular_helper.php
UncleBong: yah ic SIR exactly! little bit refactor on this

jekkos:
@UncleBong if you change dateformat then formatting in items table should follow along. After some discussion with @objecttothis we found that the usecase for datetime might be rather limited, that's why I changed it. Where exactly do I need the valid_date method?

@odiea thanks for the quick benchmark. I'll have to look further into this and possibly look into optimizing the query in that case (will run an explain first to see what is causing the slowness)

@objecttothis you're right about bintray. We should maybe set something up that deletes the oldest file once we upload a new build, so that we don't go over this limit again. Or other option would be to use the weblate VPS for this (but will need to check how much download volume these builds account for first).

UncleBong:
> 
> 
> @UncleBong if you change dateformat then formatting in items table should follow along. After some discussion with @objecttothis we found that the usecase for datetime might be rather limited, that's why I changed it. Where exactly do I need the valid_date method?
> 
> @odiea thanks for the quick benchmark. I'll have to look further into this and possibly look into optimizing the query in that case (will run an explain first to see what is causing the slowness)
> 
> @objecttothis you're right about bintray. We should maybe set something up that deletes the oldest file once we upload a new build, so that we don't go over this limit again. Or other option would be to use the weblate VPS for this (but will need to check how much download volume these builds account for first).

ok sir @jekkos ic thanks nothing to worry! we can use  valid_datetime format if we use to_datetime method instead of to_date method but it seems ok to us ready to merge :)

application/models/Item.php
```diff
@@ -192,7 +192,9 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'i
```
UncleBong: little refactor/fix to enable to search items particular fields in attribute_values

$attributes_enabled = count($filters['definition_ids']) > 0;

if(!empty($search))
{
$this->db->group_start();
	$this->db->like('name', $search);
	$this->db->or_like('item_number', $search);
	$this->db->or_like('items.item_id', $search);
	$this->db->or_like('company_name', $search);
	$this->db->or_like('items.category', $search);

      //if ($filters['search_custom'] && $attributes_enabled)
	if($attributes_enabled)
	{
		$this->db->or_like('attribute_values.attribute_value', $search);
	}
$this->db->group_end();
}

if($attributes_enabled)
{
  //$format = $this->db->escape(dateformat_mysql(TRUE));  
  this->db->select('GROUP_CONCAT(DISTINCT CONCAT_WS(\':\', definition_id, attribute_value) ORDER BY definition_id SEPARATOR \'|\') AS attribute_values');
  //$this->db->select("GROUP_CONCAT(DISTINCT CONCAT_WS(':', definition_id, DATE_FORMAT(attribute_date, $format)) SEPARATOR '|') AS attribute_dtvalues");
  $this->db->join('attribute_links AS attribute_links', 'attribute_links.item_id = items.item_id AND attribute_links.receiving_id IS NULL AND attribute_links.sale_id IS NULL AND definition_id IN (' . implode(',', $filters['definition_ids']) . ')', 'left');
  $this->db->join('attribute_values AS attribute_values', 'attribute_values.attribute_id = attribute_links.attribute_id', 'left');
}

objecttothis:
@jekkos let's definitely setup bintray to delete the oldest build when a new build goes up.  Do you think you could reach out to JFrog again? Bintray still isn't up.

jekkos:
OK I see, I have sent them two messages but no response yet.

daN4cat:
Might not be a popular decision but I'm going to merge this to master and if any issue we take it as collective pain and fix it.

As stated a couple of times, I want 3.3.0 to be released end of the month, 6 months later than initially intended (end of 2018).

Past this point we need to agree on a different approach because any release takes half or a whole year.
I can't even remember when we released last version!

daN4cat:
Pushed the button and done.

Now let's all test and fix any issue!

Issue #2321, opened 2019-01-26
Change time zone


erastus:
### Bug
application/helpers/locale_helper.php

> 'America/Belize' => '(GMT-06:00) Saskatchewan, Central America',
'America/Cancun' => '(GMT-06:00) Guadalajara, Mexico City, Monterrey',
'Chile/EasterIsland' => '(GMT-06:00) Easter Island',

change time zone America / cancun by America / Mexico_City

> 'America/Belize' => '(GMT-06:00) Saskatchewan, Central America',
		'America/Mexico_City' => '(GMT-06:00) Guadalajara, Mexico City, Monterrey',
		'Chile/EasterIsland' => '(GMT-06:00) Easter Island',

daN4cat:
Merged

PR #2322, opened 2019-01-26
fixed timezone (#2321)

erastus:
#2321 fixed time zone.

Correction of the time zone of Mexico City.


application/helpers/locale_helper.php
```diff
@@ -112,7 +112,8 @@ function get_timezones()
 		'America/Chihuahua' => '(GMT-07:00) Chihuahua, La Paz, Mazatlan',
 		'America/Dawson_Creek' => '(GMT-07:00) Arizona',
 		'America/Belize' => '(GMT-06:00) Saskatchewan, Central America',
-		'America/Cancun' => '(GMT-06:00) Guadalajara, Mexico City, Monterrey',
+		'America/Cancun' => '(GMT-05:00) Cancun',
```
daN4cat: This needs to be moved down by Cuba's entry because the timezone is the same GMT-05:00

Issue #2317, opened 2019-01-24
Change Attribute Type


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 4f5ad57
- PHP version is: 5.6.38
- MySQL or MariaDB version is: 5.7.24
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: N/A
- (If applicable) Installation package for the LAMP/LEMP stack is: N/A

### New Feature

Currently after the migration and after creating an attribute the option to change the type of attribute is unavailable.  The ability to "migrate" to a different attribute type is especially important after the migration to 3.3.0 but could also be helpful for user created attributes.

[...]

PR #2325, opened 2019-01-28
Check to see if attribute_value already exists

objecttothis:
#2317 Tested with multiple attribute definitions and multiple attribute types.  Result should be that the attribute_value is never duplicated in attribute_values

### DIFF
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -57,6 +57,17 @@ class Attribute extends CI_Model
 	}
 
 	/*
+	 Determines if a given attribute_value exists in the attribute_values table
+	 */
+	public function value_exists($attribute_value)
+	{
+		$this->db->from('attribute_values');
+		$this->db->where('attribute_value', $attribute_value);
+		
+		return ($this->db->get()->num_rows() > 0);
+	}
+		
+	/*
 	Gets information about a particular attribute definition
 	*/
 	public function get_info($definition_id)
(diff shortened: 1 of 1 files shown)


application/models/Attribute.php
```diff
@@ -394,19 +417,27 @@ public function get_suggestions($definition_id, $term)
...
 			if($definition_type != DATETIME)
 			{
-				$this->db->insert('attribute_values', array('attribute_value' => $attribute_value));
+				if($this->value_exists($attribute_value))
+				{
+					$attribute_id = $this->get_id_by_value($attribute_value);
+				}
+				else
+				{
+					$this->db->insert('attribute_values', array('attribute_value' => $attribute_value));
```
jekkos: there's another bug in this part, found that out yesterday. so let's see what can be merged first.

application/models/Attribute.php
```diff
@@ -394,19 +417,27 @@ public function get_suggestions($definition_id, $term)
...
-
+		
 		if(empty($attribute_id) || empty($item_id))
 		{
 			if($definition_type != DATETIME)
 			{
-				$this->db->insert('attribute_values', array('attribute_value' => $attribute_value));
+				if($this->value_exists($attribute_value))
+				{
+					$attribute_id = $this->get_id_by_value($attribute_value);
```
jekkos: more efficent would be to store the result of this get in a variable instead of querying twice? Then you can use `empty($attribute_id)` instead of the separate values_exists?

application/models/Attribute.php
objecttothis: Sounds good.

application/models/Attribute.php
objecttothis: Agreed.  I'll work on it tomorrow barring interruptions.

objecttothis: OK, how do the changes look?  I got rid of the get_value_by_id function and just had the value_exists function return the attribute_id if it exists or empty if it doesn't exists and I'm storing that in a variable so that we have just one db query and one function call instead of two of each.

objecttothis: The reason I had opted for a separate value_exists and get_id_by_value is to follow the behavior of the other functions in the code that perform similar tasks and because I didn't know if we would have need in other places for separate functions.  Now that we've decided to add the UNIQUE constraint to the attribute_value, I don't see how we could need a separate get_value_by_id function, so yes this is approximately twice as efficient.

jekkos: yes indeed I think the model differs in that aspect (having the unique constraint on it)

PR #2330, opened 2019-01-29
Third party integrations

objecttothis:
This is the code side of the changes to pave the way for third party integrations.  The database changes will be in a separate pull request.  The effect of these changes is simply to change the Mailchimp tab in config to be called Integrations and refactoring the appropriate calls to the view.


odiea:
I downloaded and installed your opensourcepos.20190129143149.third_party_integrations.f238bc.zip. It appears that the Integration tab and Licence tab no longer work.

objecttothis:
OK, I'll look into it tomorrow.

odiea:
Applications/views/configs/manage.php I had to change the mail_chimp to the following and it appears to be working for me now. Hope this helps.
~~~
</div>
	<div class="tab-pane" id="message_tab">
		<?php $this->load->view("configs/message_config"); ?>
	</div>
	<div class="tab-pane" id="integrations_tab">
		<?php $this->load->view("configs/integrations_config"); ?>
	</div>
	<div class="tab-pane" id="license_tab">
		<?php $this->load->view("configs/license_config"); ?>
	</div>

objecttothis:
OK.  Not sure how that didn't get transferred over from my dev deployment, but I've fixed it now.  It should be working.

objecttothis:
@jekkos, I added to the en_US config_lang.  Will that automatically get propogated to the other languages?  Also, I used the "header" div id for the new mailchimp section of the Integrations tab.  The idea is that each integration will have a header above it.  However, I don't know if the CSS for the "header" id is too in-your-face.  Is there documentation about what CSS classes are already defined and how they look, or do I need to dig through the code and try them out to find one that works better?

objecttothis:
> 
> 
> I might need to do a resync using some scripts I have to add the empty language string in other languages, but let's handle that after merge.
> 
> Regarding the css classes, we do have a italic text on top of most sections. Will need to have a look at it first to see if it's appropriate, but I would like to try and adhere to a consistent markup in all sections at least (so if you add a header here maybe be consistent and check if other sections have it as well)

@jekkos Are you referring to the italics in the "Fields in red are required" at the top of each configuration tab?  The problem is that the italics fit the warning well, but don't fit a heading very well.  I looked through the other tabs and it seems that we aren't separating vertical sections in any of the other tabs.  In the General tab we could be doing vertical separations, but we only have a little more space between one section of settings and the next.  We could do that here with the integrations, but we really need some sort of clear delineation between integrations so that settings don't get confused.  I think if we had a smaller header class defined in the CSS that is left aligned, larger than other text (perhaps 2em) and bold without a background bar that could look nicer and accomplish the separation we are looking for.

odiea:
You got part of it. I now assume you do not want mail chimp to show in integrations.

objecttothis:
> 
> 
> You got part of it. I now assume you do not want mail chimp to show in integrations.

Mailchimp should show in integrations because it is a 3rd party integration.  The idea for this tab is that all the settings for each third party integration would be found here.  If it should become unruly with too many integrations we could play around with expanding and contracting sections.

odiea:
You forgot the tab-pane part then.
~~~
<div class="tab-pane" id="integrations_tab">
	<?php $this->load->view("configs/integrations_config"); ?>
</div>
~~~

odiea:
I guess if you can find the mailchimp tab then I am missing something.
Does not work.
~~~
<div class="tab-pane" id="mailchimp_tab">
		<?php $this->load->view("configs/integrations_config"); ?>
</div>

objecttothis:
Sorry, you are correct.  Just a minute.  Let me fix that.

objecttothis:
OK, commit hash 00a16ac  should have that all fixed.  The only thing remaining is to decide on how to want to vertically divide the integrations.  The current "header" css class id I am using might be too out of character, but we don't really have divisions in any of the other tabs.  The problem is that we need to clearly divide and identify each third party integration.  I've done so with the name of the integration and maybe we should do that but have it smaller, with no background and left aligned.  Thoughts?

odiea:
The simpler the better. Thanks for all your work.

objecttothis:
OK, so I would like to alter the #header class in /public/dist/opensourcepos.min.css but that file must be generated by the build process.  I'm finding that the following css works well for the header of each 3rd party integration, but don't know if I should just be overwriting the CSS at the file level or modifying the code at the CSS level

`#header `
`{`
` margin:10px 0;`
` text-align:left;`
` color:#000;`
` font-size:1.4em;`
` letter-spacing:2px;`
` padding:10px 0`
`}`

Thoughts?

odiea:
If it does not effect any other header I do not see an issue. I will test it in my style.css and see if effects anything else.

odiea:
I would use the CSS level. I do not see any bad effects on the main program using that. I see the standard header  is .header.

objecttothis:
@odiea sorry, I'm a little daft today.  What css file will I find it in? the dist folder gets generated at build time, so I need to modify it in the css file used to generate the css in the dist folder.

objecttothis:
I can do that, but style.css in /public/css/ or mention of it in /application/views/partial/header.php don't seem to exist so do I need to create it or am I looking in the wrong folder?

objecttothis:
So the two options that I like the best are:
option1

or 

option2

objecttothis:
@opensourcepos/org can you guys weigh in as well?

jekkos:
Also the second one. Maybe make the font a bit less bold?

jekkos:
Yes maybe just add in an existing file like ospos.css, that should be fine. I see no direct need to add a new file just for this?

objecttothis:
So it turns out that if I use the #header CSS id, that causes a bug where darkly and other such themes don't change the color of the font to white.  Instead I created a #integrations_header CSS id in ospos.css and assigned that the proper attributes which seems to work fine.  The only catch is that if I didn't modify the /views/partial/header.php to drop ospos.css down below opensourcepos.min.css the changes are ignored.  I did some regression testing and was unable to find any problems created by moving the insert for ospos.css down, but I think it would be helpful if a few people tested this version before I merge the pull request.

objecttothis:
Super.  I'll merge the pull request then.

application/views/partial/header.php
```diff
@@ -25,7 +25,6 @@
 		<!-- start css template tags -->
 		<link rel="stylesheet" type="text/css" href="css/bootstrap.autocomplete.css"/>
 		<link rel="stylesheet" type="text/css" href="css/invoice.css"/>
-		<link rel="stylesheet" type="text/css" href="css/ospos.css"/>
```
jekkos: @objecttothis I guess you did this for debug purposes you have to leave the source in here so it gets minified.

jekkos: you need to run the build in order to see changes made to the source. If you want a quick preview you need to run the page with `?debug=true`

objecttothis:
Ok that's why. So ospos.css doesn't need to be down below opensourcepos.min.css?

PR #2332, opened 2019-01-30
India GST script is dependent on the 3.3.0 script.

SteveIreland:
The India GST Upgrade script is dependent on the 3.3.0 upgrade script so I changed the name so that it would run after.


SteveIreland:
This was important and was having a negative impact on other branches, so I proceeded with the merge.
The India GST upgrade script is adding new columns after specific columns that were added in the 3.3.0 script.

PR #2333, opened 2019-01-30
Add IGNORE to insert and UNIQUE constraint to attribute_value

objecttothis:
This is to resolve #2327


jekkos:
This changeset actually also requires to have the duplicate check in the code (as we'll get status 500 when adding duplicates now otherwise in the forms)

Issue #2338, opened 2019-02-02
Add migration for tax_code field in Suppliers table


jekkos:
As reported in #2309 there was an issue in creating or updating suppliers as no null values were allowed. In the last india migration script the `tax_code` column in customers table is migrated to `sales_tax_code_id`. 

- [ ] This column seems to refer to the `sales_tax_code` table without using any constraint
- [ ] The new `sales_tax_code_id` in `ospos_customers` does not have any constraint either
- [ ]  The `migrate_customer_tax_codes` uses the `fa.sales_tax_code` in it's where clause. I don't know if this column was ever present or whether it was mistaken with `tax_id` that is present next to `sales_tax_code_id`.

So I think this one was forgotten in the GST migration script. Next to `migrate_customer_tax_codes` we also need a `migrate_supplier_tax_codes`.

daN4cat:
@SteveIreland is this something you plan to fix?

@jekkos shall we move this to 3.4.0?

daN4cat:
Ping again, @jekkos @SteveIreland  see above comment.

SteveIreland:
I assumed @jekkos fixed this with #2343, but since he hasn't yet responded here I'll go ahead and take a look to confirm. If this wasn't included in that pull request I'll do something about this.

daN4cat:
Thanks, I really want to give a kick and close 3.3.0 end of the month.

SteveIreland:
I've done a preliminary review and will follow up more closely when I have a larger slice of time.

But I believe there is some confusion.   The tax code (nor tax_code_id) is not the same as the tax_id.

There is not a foreign key to the tax code on the supplier table.  There is a tax id which is just reference information for those sites that need the information printed on a purchase order or some other document.

On the customer table there is both a tax_id and tax_code (the foreign key which is named tax_code_id)   Prior to this upgrade the id used to be the tax code itself, but post upgrade the key is the tax_code_id (again, not the same as tax_id which is simply a reference number to be used for the customer's tax license when the customer is not supposed to be charged taxes.   I presume that is a U.S. thing.

SteveIreland:
I admit to not including constraints where applicable so I apologize for that..  Having worked for decades with a relational database that initially didn't support constraints (and not experiencing any crisis as a result) I tend to think that constraints are a solution in search of a problem.  Regarding constraints I'm definitely OLD school.  But I didn't exclude them deliberately ... I just overlooked it.

[...]

PR #2343, opened 2019-02-05
Fixes a couple of problems with sales tax migration.

SteveIreland:
It also fixes a minor error with tax_id definition that was causing a problem with the upgrade.

### DIFF
--- a/application/migrations/20190129212600_indiagst.php
+++ b/application/migrations/20190129212600_indiagst.php
@@ -36,17 +36,19 @@ class Migration_IndiaGST extends CI_Migration
 		}
 
 		$count_of_sales_taxes_entries = $this->get_count_of_sales_taxes_entries();
+
 		if($count_of_sales_taxes_entries > 0)
 		{
 			$this->migrate_sales_taxes_data();
 		}
+
+		$this->drop_backups();
 
 		error_log('Migrating tax configuration completed');
 	}
 
 	public function down()
 	{
-		$this->drop_backups();
 	}
 
 	private function get_count_of_tax_code_entries()
@@ -60,7 +62,7 @@ class Migration_IndiaGST extends CI_Migration
 	private function get_count_of_sales_taxes_entries()
 	{
 		$this->db->select('COUNT(*) as count');
-		$this->db->from('sales_taxes');
+		$this->db->from('sales_taxes_backup');
 
 		return $this->db->get()->row()->count;
 	}
@@ -143,9 +145,9 @@ class Migration_IndiaGST extends CI_Migration
 
 	private function drop_backups()
 	{
-		$this->db->query('DROP TABLE IF EXISTS' . $this->db->dbprefix('tax_codes_backup'));
-		$this->db->query('DROP TABLE IF EXISTS' . $this->db->dbprefix('sales_taxes_backup'));
-		$this->db->query('DROP TABLE IF EXISTS' . $this->db->dbprefix('tax_code_rates_backup'));
+		$this->db->query('DROP TABLE IF EXISTS ' . $this->db->dbprefix('tax_codes_backup'));
+		$this->db->query('DROP TABLE IF EXISTS ' . $this->db->dbprefix('sales_taxes_backup'));
+		$this->db->query('DROP TABLE IF EXISTS ' . $this->db->dbprefix('tax_code_rates_backup'));
 	}
 }
 ?>
--- a/application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
+++ b/application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
@@ -148,6 +148,3 @@ ALTER TABLE `ospos_expenses`
 INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
   ('default_receivings_discount_type', '0'),
   ('default_receivings_discount', '0');
-
-ALTER TABLE `ospos_suppliers`
-  CHANGE COLUMN `tax_id` `tax_id` VARCHAR(32) NULL;
--- a/application/migrations/sqlscripts/3.3.0_indiagst.sql
+++ b/application/migrations/sqlscripts/3.3.0_indiagst.sql
@@ -74,7 +74,7 @@ CREATE TABLE IF NOT EXISTS `ospos_tax_jurisdictions` (
 	) ENGINE=InnoDB DEFAULT CHARSET=utf8 AUTO_INCREMENT=1;
 
 ALTER TABLE `ospos_suppliers`
-	ADD COLUMN `tax_id` varchar(32) NOT NULL DEFAULT '' AFTER `account_number`;
+	ADD COLUMN `tax_id` varchar(32) DEFAULT NULL AFTER `account_number`;
 
 ALTER TABLE `ospos_tax_categories`
 	ADD COLUMN `deleted` int(1) NOT NULL DEFAULT 0 AFTER `tax_group_sequence`;
--- a/application/controllers/Suppliers.php
+++ b/application/controllers/Suppliers.php
@@ -117,7 +117,7 @@ class Suppliers extends Persons
 			'agency_name' => $this->input->post('agency_name'),
 			'category' => $this->input->post('category'),
 			'account_number' => $this->input->post('account_number') == '' ? NULL : $this->input->post('account_number'),
-			'tax_id' => $this->input->post('tax_id') == '' ? NULL : $this->input->post('tax_id')
+			'tax_id' => $this->input->post('tax_id')
 		);
 
 		if($this->Supplier->save_supplier($person_data, $supplier_data, $supplier_id))
--- a/application/migrations/20190213210000_indiagst1.php
+++ b/application/migrations/20190213210000_indiagst1.php
@@ -1,0 +1,24 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_IndiaGST1 extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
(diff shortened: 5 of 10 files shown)


application/migrations/20190129212600_indiagst.php
```diff
@@ -36,17 +36,19 @@ public function up()
...
 		}
 
+		$this->drop_backups();
+
 		error_log('Migrating tax configuration completed');
 	}
 
 	public function down()
 	{
-		$this->drop_backups();
```
jekkos: Any reason for moving this line back up? If something goes wrong those tables won't be cleaned up.

application/migrations/sqlscripts/3.3.0_indiagst.sql
```diff
@@ -74,7 +74,7 @@ CREATE TABLE IF NOT EXISTS `ospos_tax_jurisdictions` (
 	) ENGINE=InnoDB DEFAULT CHARSET=utf8 AUTO_INCREMENT=1;
 
 ALTER TABLE `ospos_suppliers`
-	ADD COLUMN `tax_id` varchar(32) NOT NULL DEFAULT '' AFTER `account_number`;
+	ADD COLUMN `tax_id` varchar(32) DEFAULT NULL AFTER `account_number`;
```
jekkos: It would be good to add a foreign key constraint for these columns. So then this is consistent with the other tables in the datamodel and we can assure referential integrity between tables.

application/migrations/20190129212600_indiagst.php
SteveIreland: The theory was that if the migration is successful then the backups (which are actually used as part of the migration) will be removed from the system and the user doesn't need to worry about orphaned tables.

If the migration itself fails then one of the interesting issues with migration is that it doesn't stop ... it just keeps going ... and going ... and going.  So eventually it deletes the backup tables anyway.

So if they revert to an earlier version of the database the down method doesn't need to be attempting to delete something that doesn't exist.

Actually for the down method to be useful wouldn't we need to add a LOT more SQL script to revert all of the previous changes (if they were successful)?

application/migrations/sqlscripts/3.3.0_indiagst.sql
SteveIreland: Actually tax_id is not a foreign key.  The tax_id is just information that is added to the customer and the supplier.  While I actually use it from the customer I don't do anything with the supplier.  I was asked to add it to the supplier table.  However, I did that a long time ago and I subsequently changed the customer tax_id value to be NOT nullable.  I just never got back to working with the tax_id on the supplier table.  After checking how it was being maintained I made it a nullable value on the database rather than changing supplier maintenance.

In my mind I was planning on coming back at a later time to make it a non-nullable value when I made changes to create and manage purchase orders.  But if you want me to go ahead and do that now I am perfectly okay with doing that.

jekkos: Ok so how is the tax_id in customer then used exactly? indeed it's just free text, so then it probably makes sense that it's not an fk. Just wonder how it is being used then.

SteveIreland: It's should be printed on the tax invoice document.

jekkos: Ok just wanted to make sure that we're not duplicating anything in the app, is this something unique and only linked to a customer, I guess it's fine to keep it like that. If we expect a user to define the same input elsewhere, then we should model that properly in our database.

application/migrations/20190129212600_indiagst.php
jekkos: Ok but the issue might be that it's unable to get to the point of dropping the tables. So that's why I think that it does make sense to move this to the down() function, as it will prevent a user's database to remain in an inconsistent intermediate state in case one of the queries fail.

In such a situation you will need to be fixing the db manually anyway, which probably also involves dropping these intermediate tables as one of the inputs for them might have been wrong?

Also reverting the model to a previous version is currently not supported in how we have setup the migrations, and most frameworks don't really provide support for it either. So I would consider a user to be 'on his own' in that situation.

jekkos: Ok my bad, I thought the down() method was used for migration teardown, but that's not the case. It is effectively able to go back to previous versions.

So indeed you're right, it should be in at the end of the up() method. On the other hand I feel there should be some kind of `cleanup` clause that is executed on success and on failure (which is not the case now), but that might be out of scope of this PR.

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
```diff
@@ -148,6 +148,3 @@ ALTER TABLE `ospos_expenses`
 INSERT INTO `ospos_app_config` (`key`, `value`) VALUES
   ('default_receivings_discount_type', '0'),
   ('default_receivings_discount', '0');
```
jekkos: what is a user already has run this alter table, but not the new statement? Will he have the column being nullable in the end? Might have been better to leave this statement here and then alter the column type in the gst script, so that changes remain incremental.

application/migrations/sqlscripts/3.3.0_indiagst.sql
jekkos: Maybe also check my remarks in #2338 and let me know if this applies or not.

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
SteveIreland: Do you mean if they create a production environment from a development branch?  Surely no one would be crazy enough to do that.  It just doesn't make sense.

(Unless one is a developer and accepts the responsibility to build scripts that correct an out of sync database.)

I'll go with the flow, but I really don't like leaving SQL in master that should never have been there in the first place.

SteveIreland: Since we don't get a lot of participation in testing until the code is in master it just doesn't feel right to NOT clean up the database.

However, if "the testing" is in reality users installing the changes into their production environment and then reporting issues then I would just need to grin and deal with it. Then your suggestion that every SQL script has to be built assuming that all previous SQL is immutable (mistakes included) makes sense.

This also means that if I "clean up" the sales_taxes table (for the tax_group fix) I will need to assume that the 3.3.3 development branch has been "published" and my changes will need to be created as if they were mini-release.  In effect the user does a sales tax migration then does another sales tax migration  for the same release.  Cringing here.

jekkos: I would say that it might be acceptable as long as you change scripts in a feature branch, so before it is merged to master. And as we currently have no different stable and development branches I think we should consider master as a release branch (or stable branch) on it's own.

You're right in the sense that the amount of users doing this might be limited, but I think this way of working is very common, at least I've done so in more than half of the projects I worked in. 

We're lucky that the migration system allows us to change old statements afterwards, as in other frameworks (eg liquibase) changing history would just block you from doing any further migration. They enforce you to work that way by hashing the contents of your statements, and then detecting if history was changed by comparing the hash with the executed hashes in the past.

jekkos: This way we ensure people will never get stuck running db migrations once they are merged. So this means that we should try to have any new migration issue free before merge. So that might indeed go at the cost of not having the cleanest SQL code but I think that's acceptable in that case.

daN4cat:
We can't assume people doing development to do not have a test environment ported from a production one.
Starting from scratch to rerun db migration scripts properly is not a good practice. It's always better to be incremental and fix the previous mistake with a new change.
In recent time we retrospectively fixed issues and that's a source of future pain, based on my past professional experience.
We are lucky that we don't need to deploy something not completed in a pre-production environment, because in that case you would not be allowed to fix retrospectively stuff. The only way is forward.

application/migrations/sqlscripts/3.2.1_to_3.3.0.sql
SteveIreland: I applied the fix for the supplier tax_id.  

Notice the "technique" that I used for the migration.  Instead of fixing the earlier migration scripts I added a supplemental migration to clean it up.  The naming convention associates the fix with the original.  I believe this is in the spirit of what is expected for migrations.

SteveIreland:
@jekkos @daN4cat 

Did the last change make it ready to merge?

jekkos:
Ok I guess this is indeed the only way to make it work properly, so we won't be executing SQL that was executed before. Thanks for getting this fixed.

application/migrations/sqlscripts/3.3.0_indiagst_a.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE `ospos_suppliers`
```
jekkos: shouldn't the column be nullable instead? Also can you add a not nullable constraint if the column still has nulls? If not shouldn't the update happen before this change column?

SteveIreland:
Hold off on merging this (not that you were, but just in case you had a change of opinion).

I was testing my changes for multi-currency invoices and and something popped up related to taxes that I need to explore further.

application/migrations/sqlscripts/3.3.0_indiagst_a.sql
SteveIreland: You are correct.  I changed it to apply the update before the alter.  I think it worked anyway but this is better.

SteveIreland:
I also added a supplemental migration step to remove the temporary backups for those who installed a pseudo-release that left them stranded.

SteveIreland:
@daN4cat @jekkos 

I'd like to see this merged this weekend.  All of the pull requests that are based on master are being tested with the tax bugs that are currently in master and if anyone is installing the development branch into production then they too are picking them up.  As always, I'm happy to jump on anything that pops up.

PR #2344, opened 2019-02-05
This fixes the improper display of the tax group

SteveIreland:
The error is triggered by the percent sign being followed by a particular sequence of characters.

### DIFF
--- a/application/libraries/Tax_lib.php
+++ b/application/libraries/Tax_lib.php
@@ -81,7 +81,6 @@ class Tax_lib
 					foreach($tax_info as $tax)
 					{
 						// This computes tax for each line item and adds it to the tax type total
-						$tax_group = (float)$tax['percent'] . '% ' . $tax['name'];
 						$tax_basis = $this->CI->sale_lib->get_item_total($item['quantity'], $item['price'], $item['discount'], TRUE);
 						$tax_amount = 0.0;
 
@@ -99,7 +98,7 @@ class Tax_lib
 						if($tax_amount <> 0)
 						{
 							$tax_group_sequence++;
-							$this->update_taxes($taxes, $tax_type, $tax_group, $tax['percent'], $tax_basis, $tax_amount, $tax_group_sequence, Rounding_mode::HALF_UP, -1, $tax['name']);
+							$this->update_taxes($taxes, $tax_type, $tax['name'], $tax['percent'], $tax_basis, $tax_amount, $tax_group_sequence, Rounding_mode::HALF_UP, -1, $tax['name']);
 							$tax_group_sequence += 1;
 						}
 						$items_taxes_detail = array();
@@ -161,7 +160,7 @@ class Tax_lib
 	*/
 	public function update_taxes(&$taxes, $tax_type, $tax_group, $tax_rate, $tax_basis, $item_tax_amount, $tax_group_sequence, $rounding_code, $sale_id, $name = '', $tax_code_id = NULL, $jurisdiction_id = NULL,$tax_category_id = NULL  )
 	{
-		$tax_group_index = $this->clean('X' . $tax_group);
+		$tax_group_index = $this->clean('X' . (float)$tax_rate . '% ' . $tax_group);
 
 		if(!array_key_exists($tax_group_index, $taxes))
 		{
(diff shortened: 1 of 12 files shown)


application/libraries/Tax_lib.php
```diff
@@ -171,6 +171,7 @@ public function update_taxes(&$taxes, $tax_type, $tax_group, $tax_rate, $tax_bas
 				'sale_id' => $sale_id,
 				'tax_type' => $tax_type,
 				'tax_group' => $tax_group,
+				'tax_group_display' => str_replace('%', '&#37;', $tax_group),
```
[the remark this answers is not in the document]
SteveIreland: I read that htmlspecialchars was redundant since the data is filtered through xss_clean.  The problem as I interpreted the issue was that xss_clean was translating the reported character stream to an unprintable character.  So I had to do something about that percent sign BEFORE xss_clean got it's grimy little hands on it.  The most efficient approach seemed to be the one I used.  I tried a couple of other approaches that I thought were cleaner but xss_clean kept mucking it up.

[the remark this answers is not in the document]
jekkos: this sounds to me like there is another issue, the character should come out fine. Maybe the page encoding is not set properly then or sth, but at least then it would be xss_clean's responsability to handle this conversion correctly.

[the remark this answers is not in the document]
SteveIreland: I think the issue is with xss_clean.  If I go to item name and enter a % symbol it works fine.  If I enter a % C it works fine.   If I enter a % CE it crashes.

If I use SQL to to insert a % CE into the item name then try to display it, the row simply will not display.

If I remove the call to xss_clean then % CE is rendered properly.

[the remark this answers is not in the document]
jekkos: Ok sounds really weird. Doesn't xss_clean escape the esacaped string again then?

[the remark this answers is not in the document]
SteveIreland: It doesn't appear to. I believe it is looking for some embedded ASCII encoding that starts with "%" and "% CE" is triggering something .  Since the percent is now "`&#37;`" it doesn't appear to to applying any escaping to it.

application/views/sales/invoice.php
```diff
@@ -180,7 +180,7 @@ function(response)
 		?>
 			<tr>
 				<td colspan="<?php echo $invoice_columns-3; ?>" class="blank"> </td>
-				<td colspan="2" class="total-line"><textarea rows="5" cols="6"><?php echo $tax['tax_group']; ?></textarea></td>
+				<td colspan="2" class="total-line"><textarea rows="5" cols="6"><?php echo $tax['tax_group_display']; ?></textarea></td>
```
jekkos: might be better to just to the replace here then, or is there a good reason for using another field ?

SteveIreland: I tried that first, but xss_clean had already replaced th data with bad characters by view time

jekkos: Can't we then get rid of the original tax_group field? or is it still needed?

jekkos: Did a quick search and I didn't find any references to it beside in most of the views. So then we can do the replace on the original tax_group and leave the views as-is?

SteveIreland: There is a function in Tax_lib.php where the tax_group is used to consolidate taxes (ultimately for the purpose of tracking and reporting non-included taxes).

That value is also written to the sales_tax table.  So a change to the value would make it very difficult to generate usable reports.

That was the reason I was jumping through hoops to avoid dropping HTML entities into the tax_group - it just doesn't feel right and I'd need to write it to the database which would then screw up tax reporting.

However, having stated all of my excuses I'd actually like to tweak the approach I'm using.  So if you can tolerate another week (or two if something unexpected happens) before the next release then I'll clean this up.

jekkos: To me it feels that this fix is something that belongs in the view and, replacing the entity right before it's displayed. 

Issue is probably that most xss_clean() code is not in the correct place either (should also be done in the views rather than in the controllers). 

I would like to check for myself what the implication of doing it there would be. No hurry or whatsoever here, might have a look myself if I find the time.

SteveIreland:
This last iteration bypasses the issue introduced by xss_clean and doesn't require a change the way sales tax reporting is handled.

application/views/sales/invoice.php
daN4cat: at the time, and to make it secure, it was quicker to add xss_clean in the controllers right before calling a view.
The same way it's checked for any SQL injection before getting to the database in the models.

Move everything to view makes the views even more polluted with PHP code, and it's a massive job. Not sure we should spend some time for that at the moment, or at least for 3.3.0.

jekkos: Agree that it might be safer to escape it before it's saved, but on the other hand it could cause you a lot of other issues as you lose the input as it was originally submitted. 

The way you escape output largely depends on the context in which you're showing it. Eg escaping a string in javascript code should be done differently than escaping something that is shown in a textarea. But agree that this might not be worth the time fixing.

PR #2345, opened 2019-02-05
Correcting positioning of the ospos.css file

objecttothis:
I wasn't aware of the need to build the project before css changes would take effect nor that I could just add `?debug=true` to the URL to see changes.  I've put the line back in its original position and it works without other changes.


daN4cat:
If my memory is not mistaken the car section in the header is automatically generated by the grunt script.

jekkos:
Correct. Might be even more convenient to set the debug flag as a cookie so it is remembered throughout the session

PR #2357, opened 2019-02-09
Cash rounding for Chile requires rounding to the nearest 10 pesos.

SteveIreland:
This pull request apparently will need to be rebuilt to include a fix for the cash rounding feature.

Please don't merge.


jekkos:
I didn't merge this one yet as the reporter in the ticket raised another issue. But I guess we'll need to tackle that in another PR then.

PR #2362, opened 2019-02-15
Add attribute type conversion, error checking.

objecttothis:
This is the pull request associated with #2317.  This enables conversion in the database for the following:

- TEXT -> DROPDOWN
- TEXT -> DATETIME
- DROPDOWN -> TEXT

Javascript has been implemented in the form to prevent all other conversions.  Currently DATETIME->TEXT is disabled because attribute_datetime does not have a UNIQUE constraint but attribute_value does.  It could be implemented in the future but would require a check for each attribute_datetime to see if the attribute_value already exists and then change the attribute_id in the attribute_links table for that row.  This is not a feature that we need so I'm not implementing it for now.

### DIFF
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -497,4 +497,8 @@ function dateformat_bootstrap($php_format)
 	return strtr($php_format, $SYMBOLS_MATCHING);
 }
 
+function valid_datetime($datetime)
+{
+	return preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $datetime);
+}
 ?>
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -247,7 +247,6 @@ class Attribute extends CI_Model
 	private function check_data_validity($definition, $from, $to)
 	{
 		$success = FALSE;
-		$fail = FALSE;
 
 		if($from === TEXT)
 		{
@@ -257,16 +256,16 @@ class Attribute extends CI_Model
 				$this->db->from('attribute_values');
 				$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
 				$this->db->where('definition_id',$definition);
-
+				$success = TRUE;
+				
 				foreach($this->db->get()->result_array() as $row)
 				{
 					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
 					{
 						log_message('ERROR', 'item_id: ' . $row['item_id'] . ' with attribute_value: ' . $row['attribute_value'] . ' cannot be converted to datetime');
-						$fail = TRUE;
+						$success = FALSE;
 					}
 				}
-				$success = $fail ? FALSE: TRUE;
 			}
 		}
 		return $success;
@@ -274,27 +273,27 @@ class Attribute extends CI_Model
 	
 	private function convert_definition_type($definition_id, $from_type, $to_type)
 	{
+		$success = FALSE;
+		
 		if($from_type === TEXT)
 		{
 			//From TEXT to DATETIME
 			if($to_type === DATETIME)
 			{
-				if(!$this->check_data_validity($definition_id, $from_type, $to_type))
+				if($this->check_data_validity($definition_id, $from_type, $to_type))
 				{
-					return FALSE;
+					$this->db->trans_start();
+
+					$query = 'UPDATE ospos_attribute_values ';
+					$query .= 'INNER JOIN ospos_attribute_links ';
+					$query .= 'ON ospos_attribute_values.attribute_id = ospos_attribute_links.attribute_id ';
+					$query .= 'SET attribute_datetime = attribute_value, ';
+					$query .= 'attribute_value = NULL ';
+					$query .= 'WHERE definition_id = ' . $this->db->escape($definition_id);
+					$success = $this->db->query($query);
+	
+					$this->db->trans_complete();
 				}
-
-				$this->db->trans_start();
-
-				$query = 'UPDATE ospos_attribute_values ';
-				$query .= 'INNER JOIN ospos_attribute_links ';
-				$query .= 'ON ospos_attribute_values.attribute_id = ospos_attribute_links.attribute_id ';
-				$query .= 'SET attribute_datetime = attribute_value, ';
-				$query .= 'attribute_value = NULL ';
-				$query .= 'WHERE definition_id = ' . $this->db->escape($definition_id);
-				$success = $this->db->query($query);
-
-				$this->db->trans_complete();
 			}
 		}
 		else if($from_type === DROPDOWN)
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -244,6 +244,42 @@ class Attribute extends CI_Model
 		return $this->search($search)->num_rows();
 	}
 
+	public function convert_definition_type($definition_id,$from_type,$to_type)
+	{
+		if($from_type === TEXT)
+		{
+			//From TEXT to DATETIME
+			if($to_type === DATETIME)
+			{
+				$this->db->trans_start();
+				
+				$query = 'UPDATE ospos_attribute_values ';
+				$query .= 'INNER JOIN ospos_attribute_links ';
+				$query .= 'ON ospos_attribute_values.attribute_id = ospos_attribute_links.attribute_id ';
+				$query .= 'SET attribute_datetime = attribute_value, ';
+				$query .= 'attribute_value = NULL ';
+				$query .= 'WHERE definition_id = '.$definition_id;
+				$this->db->query($query);
+				
+				$this->db->trans_complete();
+			}
+		}
+		else if($from_type === DROPDOWN)
+		{
+			//From DROPDOWN to TEXT
+			$this->db->trans_start();
+			
+			$this->db->from('ospos_attribute_links');
+			$this->db->where('definition_id',$definition_id);
+			$this->db->where('item_id',NULL);
+			$success = $this->db->delete();
+			
+			$this->db->trans_complete();
+		}
+		
+		return $success;
+	}
+	
 	/*
 	Inserts or updates a definition
 	*/
@@ -251,23 +287,37 @@ class Attribute extends CI_Model
 	{
 		//Run these queries as a transaction, we want to make sure we do all or nothing
 		$this->db->trans_start();
-
+		
+		//Definition doesn't exist
 		if($definition_id === -1 || !$this->exists($definition_id))
 		{
 			$success = $this->db->insert('attribute_definitions', $definition_data);
 			$definition_data['definition_id'] = $this->db->insert_id();
 		}
+		//Definition already exists
 		else
 		{
+			$this->db->select('definition_type');
+			$this->db->from('attribute_definitions');
+			$this->db->where('definition_id',$definition_id);
+			
+			$from_definition_type = $this->db->get()->row()->definition_type;
+			$to_definition_type = $definition_data['definition_type'];
+			
+			if($from_definition_type !== $to_definition_type)
+			{
+				$this->convert_definition_type($definition_id,$from_definition_type,$to_definition_type);
+			}
+			
 			$this->db->where('definition_id', $definition_id);
 			$success = $this->db->update('attribute_definitions', $definition_data);
 			$definition_data['definition_id'] = $definition_id;
 		}
-
+		
 		$this->db->trans_complete();
-
- 		$success &= $this->db->trans_status();
-
+		
+		$success &= $this->db->trans_status();
+		
 		return $success;
 	}
 
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -248,23 +248,20 @@ class Attribute extends CI_Model
 	{
 		$success = FALSE;
 
-		if($from === TEXT)
+		if($from === TEXT && $to === DATETIME)
 		{
-			if($to === DATETIME)
+			$this->db->select('item_id,attribute_value');
+			$this->db->from('attribute_values');
+			$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
+			$this->db->where('definition_id',$definition);
+			$success = TRUE;
+			
+			foreach($this->db->get()->result_array() as $row)
 			{
-				$this->db->select('item_id,attribute_value');
-				$this->db->from('attribute_values');
-				$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
-				$this->db->where('definition_id',$definition);
-				$success = TRUE;
-				
-				foreach($this->db->get()->result_array() as $row)
+				if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
 				{
-					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
-					{
-						log_message('ERROR', 'item_id: ' . $row['item_id'] . ' with attribute_value: ' . $row['attribute_value'] . ' cannot be converted to datetime');
-						$success = FALSE;
-					}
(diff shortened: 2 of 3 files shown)


application/models/Attribute.php
```diff
@@ -244,30 +244,80 @@ public function get_found_rows($search)
 		return $this->search($search)->num_rows();
 	}
 
+	public function convert_definition_type($definition_id,$from_type,$to_type)
```
jekkos: Does this method need to be public?

application/models/Attribute.php
```diff
@@ -244,30 +244,80 @@ public function get_found_rows($search)
...
+				
+				$query = 'UPDATE ospos_attribute_values ';
+				$query .= 'INNER JOIN ospos_attribute_links ';
+				$query .= 'ON ospos_attribute_values.attribute_id = ospos_attribute_links.attribute_id ';
+				$query .= 'SET attribute_datetime = attribute_value, ';
+				$query .= 'attribute_value = NULL ';
+				$query .= 'WHERE definition_id = '.$definition_id;
+				$this->db->query($query);
+				
+				$this->db->trans_complete();
```
jekkos: What if the data can't be converted to datetime format? Maybe check how the code behaves in that case, do we get a notification popup to notify the user that the update failed?

You might want to update the $success variable to make that work then here.

application/views/attributes/form.php
```diff
@@ -69,10 +69,31 @@
 	var definition_id = <?php echo $definition_id; ?>;
 	var is_new = definition_id == -1;
 
+	var disable_definition_types = function()
+	{
+		var definition_type = $("#definition_type option:selected").text();
```
jekkos: Kinda like the idea of limiting the dropdown values to the available conversions for each case.

jekkos:
Thanks for this PR! just three minor remarks here.

application/models/Attribute.php
```diff
@@ -244,30 +244,80 @@ public function get_found_rows($search)
...
+			{
+				$this->convert_definition_type($definition_id,$from_definition_type,$to_definition_type);
+			}
+			
 			$this->db->where('definition_id', $definition_id);
 			$success = $this->db->update('attribute_definitions', $definition_data);
 			$definition_data['definition_id'] = $definition_id;
 		}
-
+
```
daN4cat: Please remove the extra tabs, they could cause merge conflicts for no reasons.

application/models/Attribute.php
```diff
@@ -244,30 +244,80 @@ public function get_found_rows($search)
...
 		}
-
+		
 		$this->db->trans_complete();
-
- 		$success &= $this->db->trans_status();
-
+		
+		$success &= $this->db->trans_status();
+
```
daN4cat: Same here

application/models/Attribute.php
daN4cat: Please add a space between the params

application/models/Attribute.php
```diff
@@ -244,30 +244,80 @@ public function get_found_rows($search)
...
+			if($to_type === DATETIME)
+			{
+				$this->db->trans_start();
+				
+				$query = 'UPDATE ospos_attribute_values ';
+				$query .= 'INNER JOIN ospos_attribute_links ';
+				$query .= 'ON ospos_attribute_values.attribute_id = ospos_attribute_links.attribute_id ';
+				$query .= 'SET attribute_datetime = attribute_value, ';
+				$query .= 'attribute_value = NULL ';
+				$query .= 'WHERE definition_id = '.$definition_id;
```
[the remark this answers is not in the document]
daN4cat: There are CI functions to build queries that can be used and those do the escaping for you.
Not sure why we don't use them: https://www.codeigniter.com/userguide3/database/query_builder.html#updating-data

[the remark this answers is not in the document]
daN4cat: ooops, I should have read #2317 first. Disregard my comment here.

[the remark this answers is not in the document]
objecttothis: @daN4cat We can't use Query Builder here because Query Builder does not support db->join() with db->update().  We also can't break the code into two different Query Builder queries because the second query contains a crazy long array and even if it didn't cause Query Builder to error out, the query alone would take over 30 seconds.  Putting the second query in a foreach loop would also take an insane amount of time.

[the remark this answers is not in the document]
objecttothis: @daN4cat @jekkos I'm back in the office tomorrow and will make the formatting and injection prevention changes then.

application/models/Attribute.php
objecttothis: Yeah, we need to hash that out.  Currently with improperly formatted data MySQL puts 0000-00-00 00:00:00.  I posed this question in the issue thread, but didn't get any responses.  We should decide on what the behavior should be.  We can leave it as it, or abort the change and kick back an error.  I'm not sure how we could do the latter unless I do a query prior to this and search the results using php because the query doesn't kick back an error with junk data.  It just quietly inserts the zeros.  Ideas?

application/views/attributes/form.php
objecttothis: It was either that or disabling then so that they are greyed out.  I chose to hide them instead.

application/models/Attribute.php
jekkos: I guess in case we do something like inserting a default value to replace unparseable dates, we should warn the user first. Otherwise I would just throw an error saying that values cannort be converted and that this needs to be fixed in some way.

application/models/Attribute.php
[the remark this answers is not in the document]
objecttothis: $definition_id escaped.

application/models/Attribute.php
objecttothis: Made the function private.  Added space between params

application/models/Attribute.php
objecttothis: added $success to that query, however like I said before, MySQL doesn't kick an error on junk data, so I may need to write a query prior to this using Query Builder to pull the data and examine it.  I'll refactor that into another function since we may want to have data validity checking on other attribute conversions in the future.  I think to simplify it, I will just kick back an array of item_ids with the improperly formatted data and have that output to the error log with a toast notification that the conversion failed and that they should check the error log.

application/models/Attribute.php
objecttothis: Extra tabs removed.

application/models/Attribute.php
objecttothis: Done

application/models/Attribute.php
objecttothis: @jekkos @daN4cat OK, I've added code to check the formatting of the TEXT on conversion to DATETIME and both output each line that isn't either YYYY-MM-DD or YYYY-MM-DD HH:MM:SS and then throw an error in the toast notification.

One problem that we have is that currently the application prevents the ':' symbol in a TEXT attribute but the migration allows it.  This means that if a user needs to correct the formatting on one of the attributes they won't be able to enter a YYYY-MM-DD HH:MM:SS format.  Is there a reason we are blocking ':'?

application/models/Attribute.php
```diff
@@ -244,29 +244,121 @@ public function get_found_rows($search)
...
+				{
+					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
+					{
+						log_message('ERROR', 'item_id: ' . $row['item_id'] . ' with attribute_value: ' . $row['attribute_value'] . ' cannot be converted to datetime');
+						$success = FALSE;
+					}
+				}
+				return $success;
+			}
+			else
```
daN4cat: Just realised that all these elses are unnecessary and it's better to have a return FALSE at the end.
You could also have one if with the && of the two conditionals.

application/models/Attribute.php
```diff
@@ -244,29 +244,121 @@ public function get_found_rows($search)
...
+			
+			$this->db->from('ospos_attribute_links');
+			$this->db->where('definition_id',$definition_id);
+			$this->db->where('item_id',NULL);
+			$success = $this->db->delete();
+			
+			$this->db->trans_complete();
+		}
+		
+		return $success;
```
daN4cat: $success doesn't seem set to any default value outside the two if statements.

Set it to default FALSE at the beginning of the function. Then wrap the query inside the if with check_data_validity so the default will be FALSE at the end of the function without an early return.

daN4cat:
Sorry if I spotted later those two points in my comments.

application/models/Attribute.php
jekkos: Well I think I want to drop the time part for datetime attribute, as I think it's usage is limited. Most of the time people use will use it for expiry date tracking or the like. So not sure if we need a complete timestamp in there?

objecttothis: @jekkos For our use, we don't need a full timestamp it as we are dealing with the publication date.  So then would you change the database structure to have the attribute_datetime be a DATE in MySQL instead of DATETIME?  Also, does it make sense then to refactor our code and the name of the column to be DATE and attribute_date respectively?  Let me know what you end of deciding on just storing the date and I will adjust the regex accordingly.  The only scenario in which I could forsee the need for a full DATETIME is if we expanded attributes in the future to include Inventory attributes. In that case it could be useful to have a datetime which tells us the hour and minute an item was received, quantity was changed, report was run, etc.

Alternately we could just keep the database as is for future compatibility and ignore the hour, minute and second on the software end.  MySQL will accept YYYY-MM-DD in a datetime without the time portion.  It just stores zeros in the time portion.

application/models/Attribute.php
objecttothis: Fixed.

application/models/Attribute.php
objecttothis: Fixed.

daN4cat: Where is the fix?

application/models/Attribute.php
```diff
@@ -244,29 +244,115 @@ public function get_found_rows($search)
 		return $this->search($search)->num_rows();
 	}
 
+	private function check_data_validity($definition, $from, $to)
+	{
+		$success = FALSE;
+		$fail = FALSE;
```
daN4cat: No need of this $fail

application/models/Attribute.php
```diff
@@ -244,29 +244,115 @@ public function get_found_rows($search)
...
+
+		if($from === TEXT)
+		{
+			if($to === DATETIME)
+			{
+				$this->db->select('item_id,attribute_value');
+				$this->db->from('attribute_values');
+				$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
+				$this->db->where('definition_id',$definition);
+
```
daN4cat: Add $success = TRUE;

application/models/Attribute.php
```diff
@@ -244,29 +244,115 @@ public function get_found_rows($search)
...
+				foreach($this->db->get()->result_array() as $row)
+				{
+					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
+					{
+						log_message('ERROR', 'item_id: ' . $row['item_id'] . ' with attribute_value: ' . $row['attribute_value'] . ' cannot be converted to datetime');
+						$fail = TRUE;
```
daN4cat: Add $success = FALSE;

application/models/Attribute.php
```diff
@@ -244,29 +244,115 @@ public function get_found_rows($search)
...
+					}
+				}
+				$success = $fail ? FALSE: TRUE;
```
daN4cat: No need of this $fail

application/models/Attribute.php
objecttothis: OK, sorry, I'm a little slow.  I believe I got it figured out now.  I pushed the changes I think you are asking for.

objecttothis:
The reason for not breaking the loop is so that all failing rows get added to the error log. If we break the loop on the first error then users only know the first problem data to fix and are stuck re-running the conversion until they fix the last one or give up.

objecttothis:
@jekkos @daN4cat OK, I just pushed a minor bug fix in the view which prevents new Attributes from being specified as anything but a GROUP.  I think that's the last fix.  If there is anything else, please let me know, but if it looks good I'll merge the pull request.

application/models/Attribute.php
```diff
@@ -257,44 +256,44 @@ private function check_data_validity($definition, $from, $to)
 				$this->db->from('attribute_values');
 				$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
 				$this->db->where('definition_id',$definition);
-
+				$success = TRUE;
+				
 				foreach($this->db->get()->result_array() as $row)
 				{
 					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
```
jekkos: maybe move this regex to local_helper and put it in a method like `validate_datetime` as it's a more generic thing that can be reused in other classes.

jekkos: On the other hand, why not just use `strtotime` instead? It should just return false if the date is not parseable.

objecttothis: Two reasons why strtotime isn't probably the best solution.  as of PHP 5.1 preg_match is faster than strtotime for parsing YYYY-MM-DD HH:MM:SS format according to the O'Reilly PHP Cookbook.  The second is that strtotime will parse MM-DD-YYYY and other various time formats properly but MySQL is looking for ANSI formatting of YYYY-MM-DD with or without HH:MM:SS so additional checks would have to be made after strtotime() passes to confirm that the date format also passes.

I can move the regex to local_helper to make it more useable by the rest of the code.

objecttothis: OK, moved the regex to the locale_helper.php and tested.

objecttothis:
OK, I've made all the requested changes.  Let me know if there is anything else.  Otherwise we should be able to merge the pull request.

jekkos:
Looks good. Thx for the work.

application/models/Attribute.php
```diff
@@ -244,29 +244,108 @@ public function get_found_rows($search)
...
+			$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
+			$this->db->where('definition_id',$definition);
+			$success = TRUE;
+			
+			foreach($this->db->get()->result_array() as $row)
+			{
+				if(valid_datetime($row['attribute_value']) === FALSE)
+				{
+					log_message('ERROR', 'item_id: ' . $row['item_id'] . ' with attribute_value: ' . $row['attribute_value'] . ' cannot be converted to datetime');
+					$success = FALSE;
```
daN4cat: I still feel you could add a break of the loop here or simply return FALSE as once you have a failure you don't need to iterate the complete list.

objecttothis: @daN4cat So let's say that a user is migrating from TEXT to DATETIME attribute type.  They have 500 items that need to be converted and 150 of them have incorrect format.  If we `return FALSE;` at the first logged error then the user only knows the first of 150 items that they need to update to get their data in compliance.  This means that they are going to have to run the conversion 151 times in order to find all of the incorrectly formatted item_id's instead of 2 times.  I'm guessing most users are going to give up on the conversion in that scenario.  

Correct me if I'm wrong, but I think the concern you have is performance.  I ran this conversion in my own testing on 12,000+ rows and didn't experience any lag between clicking Submit in the view and the toast notification.  It really is snappy.  I even tested it with all 12,000+ being wrong and generating a log line.  The log file took a little more time to download but the transaction itself was just as swift.  My server is running an Intel Core i5 processor, so it's not like it's 8 Xeon Platinum 8176F processors with 256GB of RAM.

application/helpers/locale_helper.php
```diff
@@ -497,4 +497,8 @@ function dateformat_bootstrap($php_format)
 	return strtr($php_format, $SYMBOLS_MATCHING);
 }
 
-?>
+function valid_datetime($datetime)
```
jekkos: Ok thanks for moving this, but still wonder whether we can't use `strtotime()` instead?

application/models/Attribute.php
daN4cat: Good point about the log_message, it's fine like this then.

daN4cat:
Let's move on an merge otherwise it's a never-ending story this 3.3.0 release.

alfatihbaihaqi:
hi, my name is alfatih
i am new in using ospos 3.2.3
i have a trouble in modifying item everytime i login the ospos
help me please

alfatihbaihaqi:
everytime i modify item in ospos 3.2.3 i got forced logout.
i need an explained why

objecttothis:
> 
> 
> everytime i modify item in ospos 3.2.3 i got forced logout.
> i need an explained why

@alfatihbaihaqi please go to the issues section and create an issue.  You'll need to fill in the necessary information there for someone to be able to help you.  Posting your question here clutters it up.

PR #2364, opened 2019-02-16
System & Installation Info

WebShells:
Display system information, user settings, ospos installation version/hash, and files permissions. 
We can add a button to copy the data inside the table and add direct link to git issues, what do you think ?


odiea:
I tried your commit but get the following error loading the system info tab. Is it missing this.
An Error Was Encountered
Unable to load the requested file: configs/system_info.php

WebShells:
Try it now please @odiea sorry renamed the links but didnt change file name... Was working on server stats as name...

odiea:
So far so good.
2019-02-16 17_11_37-window

WebShells:
Shall we move language strings to language files or we keep it that way? 🤔

odiea:
We should let the big boys decide that. Currently en-us is my language so I do not have any complaints. Plus if they copy and paste the install info into issues it would then be English.

application/views/configs/system_info.php
```diff
@@ -0,0 +1,100 @@
...
+		  <th>Extensions & Modules</th>
+			<td>
+			<?php 
+				echo "GD: ", extension_loaded('gd') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+				echo "BC Math: ", extension_loaded('bcmath') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+				echo "INTL: ", extension_loaded('intl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+				echo "OpenSSL: ", extension_loaded('openssl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+				echo "MBString: ", extension_loaded('mbstring') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+				echo "Curl: ", extension_loaded('curl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
+			//	echo print_r(apache_get_modules());
```
daN4cat: Shall we remove all these commented out lines.

daN4cat:
We are missing another .csv, also we shouldn't make .csv templates writable because those are templates not files to be used. For security reasons they must be read only to avoid any injection of code in it.

application/views/configs/system_info.php
WebShells: @daN4cat, sure... The commented lines works on local installations, but for security issues i guess they limit script from running on a shared server... Will remove also cause will load alot of modules and extensions which will make copy/pasting it a big issue...

WebShells:
@daN4cat i didnt force change csv permissions, i just confirmed its writable and i guess you are right it shouldnt be... Its a template to be used, shall i revert to check if its protected or not? Also for customers? And shall we keep labels as it is or add them in language files? To me am not so sure if anything other than main TAB should be translated.. What do you think?

daN4cat:
Yes, security first. Check if protected.
And translations too, where possible of course. We don't translate php into stuff.

application/views/configs/system_info.php
```diff
@@ -0,0 +1,111 @@
+<div id="config_wrapper">
+	<?php echo $this->lang->line('config_server_notice'); ?>
+	<table class="table text-left" >
+		<tbody>
+		<tr>
+		  <th>OSPOS Info</th>
```
jekkos: ok maybe also move this part to a lang->line

application/views/configs/system_info.php
```diff
@@ -0,0 +1,111 @@
...
+				chmod("../.htaccess",0644);
+				chmod("../application/.htaccess",0644);
+				chmod("../public/.htaccess",0644); 
+				chmod("../import_items.csv",0444);
+				chmod("../import_customers.csv",0444); 
+				echo "<br>";
+				echo ".htaccess permissions were reset to 0644 <br>"; 
+				echo "CSV permissions were reset to 0444"; 
+				?><br>
+				<a href="https://github.com/opensourcepos/opensourcepos/issues/new" target="_blank"> Report An issue </a>
```
jekkos: Would there be a way to post the info shown here to the github issue directly?

application/views/configs/system_info.php
```diff
@@ -0,0 +1,111 @@
...
+				<?php echo PHP_VERSION; ?> <br>
+				Server Port: 
+				<?php echo $_SERVER['SERVER_PORT']; ?><br>
+				DB Version: 
+				<?php print mysqli_get_client_info(); ?><br>
+				OS: 
+				<?php echo php_uname();	?></td>
+		</tr>
+		<tr>
+		  <th>File's Permissions</th>
```
jekkos: Better to make these translatable. It's enough if you just add them to the english language files, I can sync the others.

public/system_info.php
```diff
@@ -0,0 +1,104 @@
+<div id="config_wrapper">
```
jekkos: guess we don't need this one as we load the views through a controller instead?

public/system_info.php
```diff
@@ -0,0 +1,104 @@
...
+				<?php 
+					if (is_writable($uploads)) {
+						echo 'Writable &#x2713';
+					} else {
+						echo 'NOT Writable &#x2717 ';
+					} 
+				?><br>
+				public/uploads/item_pics: 	
+				<?php 
+					if (is_writable($images)) {
```
jekkos: might be good to extract a function for this check. like <?php is_writable($variable) ?> that then checks the variable.

objecttothis:
> 
> 
> We are missing another .csv, also we shouldn't make .csv templates writable because those are templates not files to be used. For security reasons they must be read only to avoid any injection of code in it.

To clarify @daN4cat the .csv will need to be writeable by the web server so that ospos can add/remove attribute names.  @WebShells is just checking to see if it's writeable.  It should be no higher than 664 permission for this to be able to happen and no lower than 644

application/views/configs/system_info.php
objecttothis: More specifically en_US since that is the base language

daN4cat:
Right, forgot that with attributes there is a different behaviour. Sounds good to me.

application/views/configs/system_info.php
WebShells: So shall i ? Since en_US is our main lang ? or ?

application/views/configs/system_info.php
```diff
@@ -0,0 +1,111 @@
...
+					} 
+				?><br>
+				import_customers.csv: 
+				<?php 
+					if (is_writable($importcustomers)) {
+						echo 'Writable &#x2717 <br>';
+					} else {
+						echo 'NOT Writable/Read Only &#x2713 <br>';
+					} 
+				chmod("../.htaccess",0644);
```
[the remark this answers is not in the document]
jekkos: Correct you should be cautious here. Also this does not seem to be the right place to correct such a thing. Also might be good to just indicate what the rights are, and leave it up to the user to fix it (as you will never know the exact permissions of the process that tires to correct these btis)

application/views/configs/system_info.php
WebShells: @objecttothis am not sure if the format does any change, copying and pasting first 2 sections will give the needed information, i have no problem in rearranging though...

daN4cat:
@odiea did you have any chances to test this branch?

application/views/configs/system_info.php
jekkos: Yes just english would be great already?

odiea:
@daN4cat No I have been waiting for all the corrections to be made. I copied his original request and it appeared to be just fine. I did post an image from his first set of corrections above.

daN4cat:
I merge this, we'll fix if broken so we keep moving.

Issue #2361, opened 2019-02-15
Design Decision: remove dbprefix?


objecttothis:
### Background information

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 9e2f9f
- PHP version is: 5.6.38
- MySQL or MariaDB version is: 5.7
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### New Feature
I think we should refactor the code and setup a migration script to stop using dbprefix.

dbprefix is a CI optional config and is designed to append a prefix to Query Builder queries so that we could have two or more installations of ospos using the same database but different sets of tables within that database (e.g., ospos1_, ospos2_... inside the same database).

This isn't a configuration that I think we recommend or support regardless of whether dbprefix is being used or not.  Instead we would ask users to install each installation of ospos with its own database.

[...]

PR #2371, opened 2019-02-20
Dbprefix drop

objecttothis:
I'm submitting the pull request now, but it should not get merged into the master after the release of 3.3.0.  This just gives us a chance for some code review.


jekkos:
Ok I guess demo got stuck because it deployed this branch. Will try to reinit the database to get it back up again.

objecttothis:
@jekkos is there a way to get the demo to only deploy the latest master? Or even just the latest release version? The reason it got stuck is that the SQL to migrate the database has to be run manually.

jekkos:
@objecttothis yes you can by either editing .travis yaml file or by making the PR from another repo (so not the organisation's one)

application/models/Attribute.php
```diff
@@ -184,106 +184,192 @@ public function get_definitions_by_flags($definition_flags)
...
 		return $this->search($search)->num_rows();
 	}
-
+	
+	private function check_data_validity($definition, $from, $to)
+	{
+		$success = FALSE;
+		$fail = FALSE;
+		
+		if($from === TEXT)
```
jekkos: why not combine these two if statements?

application/models/Attribute.php
```diff
@@ -184,106 +184,192 @@ public function get_definitions_by_flags($definition_flags)
...
+			if($to === DATETIME)
+			{
+				$this->db->select('item_id,attribute_value');
+				$this->db->from('attribute_values');
+				$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
+				$this->db->where('definition_id',$definition);
+				
+				foreach($this->db->get()->result_array() as $row)
+				{
+					if(!preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $row['attribute_value']))
```
jekkos: think this PR got mixed up with the other one you opened..

application/models/Attribute.php
objecttothis: > 
> 
> I think we need an automatic database migration here (or maybe I missed it or the rationale on why there is none)

Agreed, but I'm not sure how to get it done.  So, with this pull request none of the code contains the dbprefix stuff, but on first login (which is how the automatic migration happens) the database still has all the ospos_ prefixes on the tables so you get a php error.  If you login to phpmyadmin and run the /database/dbprefix_drop.sql then on first login the 3.3.0_to_3.4.0.sql will run like it's supposed to.

application/models/Attribute.php
objecttothis: shoot.  You're right.  I'll fix that later.  It just affects models/Attribute.php

jekkos:
I will force the dbprefix for the time being as the demo server still has an old database.sql in the deploy root that is needed for the docker setup.

application/models/Attribute.php
objecttothis: > 
> 
> why not combine these two if statements?

This also applies to the other PR.  The reason for having them separate was because I was expecting to potentially have other cases in the future where we needed to check formatting before converting.  Perhaps I just combine them now and then we can separate them in the future if we need to.

objecttothis:
> 
> 
> Rework required.

OK, so we agree then that I need to take modifications to migrate_phppos.sql and rename_tables.sql out of this pull request.  No problem.  

Just to confirm though: tables.sql and constraints.sql is used to generate database.sql.  If we don't change those scripts, users installing a new installation will get their tables created with ospos_ in them which we don't want.  In that sense tables.sql and constraints.sql are not migration scripts, but installation scripts.

I've created dbprefix_drop.sql to do the migration, but it can't be an automatic script because at the time that the migration needs to happen the tables will all have "ospos_" in them but the code will be looking for tables that don't have "ospos_" and therefore not even the login screen will appear until the user manually runs dbprefix_drop.sql.  In some ways that isn't any different than before we were using an automatic migration and will be a one-time thing (after, the automatic migration will work flawlessly again).

application/models/Attribute.php
objecttothis: > 
> 
> think this PR got mixed up with the other one you opened..

OK, I rolled back the changes from the other PR.

objecttothis:
@daN4cat one other question.  The manual migration script currently resides in database/dbprefix_drop.sql.  One question is "is that the correct location for it, or should it go in /application/migrations/sqlscripts/ where the automatic sqlscripts are stored?"  We seem to have a legacy and automatic location for location scripts.  The second question is "is dbprefix_drop.sql the best name for the manual script?"  It must be run before any 3.3.0_to_X.Y.Z.sql scripts can work because of above conversation about why it doesn't work as an automated migration script.

daN4cat:
New user creates the table with the old script and the migration scripts since the last time we touched them will update the database to the last version.
That's why you should not touch history but add incremental steps fixing in delta. You replay the whole history incrementally change.

objecttothis:
Please see my post in #2361.  The conversation is happening in two places so it's a little disjointed.  
> New user creates the table with the old script and the migration scripts since the last time we touched them will update the database to the last version.

My understanding of tables.sql and constraints.sql is that every single time we issue code that migrates the structure of the database that these need to reflect those changes so that database.sql always is an accurate representation of the most recent structure of the database.  This is different from migration scripts which are pre-existing database installations whereas database.sql is used for brand new installations.

jekkos:
@objecttothis on your last comment, it used to be like that, but now database.sql is a snapshot of a complete fresh installation before all migration scripts are executed.

so curently we don't have a script that does a one shot fresh database init. In some way it makes sense, if there was then you would be doing every change twice. Once as a migration script and once in the initial init script.

application/models/Attribute.php
jekkos: Ok combine them for now. Doesn't make sense to split it out if it's not needed atm.

jekkos: I see the issue, you will need to keep the ospos_ prefix in the config.php until the last migration script is executed. I guess you'll need to add someting in the php script to clear out the config setting once the tables are renamed to make things work again. There's no other way around.

jekkos: I used following command to replace it in a unix shell

`sed -i 's/'\''dbprefix'\'' => '\'''\''/'\''dbprefix'\'' => '\''ospos_'\''/g' /app/application/config/database.php`

daN4cat:
Basically the migration script is the means to "maintain those old scripts up to date".

daN4cat:
@objecttothis have you tested the upgrade from an old database?
I have one doubt, if you remove ospos_ prefix from database.php due to the code upgrade, how the login page works when this upgrade is required?
Because the login relies on a table starting with ospos_ to check the users. Unless the migration is run before any user check / database interaction?

objecttothis:
> 
> 
> @objecttothis on your last comment, it used to be like that, but now database.sql is a snapshot of a complete fresh installation before all migration scripts are executed.
> 
> so curently we don't have a script that does a one shot fresh database init. if there was then you would be doing every change twice: once as a migration script and once in the initial init script.

OK, now I think I understand why you guys have been referring to database.sql as a migration script.  I was not aware that this was changed.  At what point does database.sql then reflect the removal of ospos_ then?

>@objecttothis have you tested the upgrade from an old database?
>I have one doubt, if you remove ospos_ prefix from database.php due to the code upgrade, how the login page works when this upgrade is required?
>Because the login relies on a table starting with ospos_ to check the users. Unless the migration is run before any user check / database interaction?

Correct, this is what I have been saying and why I have been saying that the dbprefix_drop.sql needs to be run manually the first time.  Login won't work and therefore at some point the changes do need to get reflected in database.sql.

The only way I can possibly imagine this being an automatic update and not touching database.sql is if the code starts with ospos_ in the places needed to access the database (maybe just /application/config/database.php it then migrates the database then modifies itself to remove all the ospos_ references.  I don't know if it can be universally assumed that the software doing a code rewrite of itself on all servers will be allowed though.

objecttothis:
>it used to be like that, but now database.sql is a snapshot of a complete fresh installation before all migration scripts are executed.

I looked at the current master and I see that now.  I wasn't aware this had changed.  So is database.sql perpetually stuck at whatever version number we began the automatic migration scripts?  If that's the case then even a fresh install is running the migration scripts?

If we could figure out how to make the dbprefix_drop.sql an automatic migration, then this wouldn't be a big deal.  

With the system as it is, every old install will have to have this script run one time manually and every new install will have to run this script one time manually.

>have you tested the upgrade from an old database?

yes, I tested it on an install where the database is 3.2.0 and putting the 3.3.0 code in.  It won't get to login until dbprefix_drop.sql is run then it goes to login and runs the rest of the migration scripts automatically.

objecttothis:
@jekkos @daN4cat If database.sql is only used for new installs, what is the benefit of it (and therefore tables.sql and constraints.sql) not being a snapshot of the most recent database structure?  What if we had those always reflect the structure of the most recent database changes and for new installs it just inserted the version in the migrations table reflecting that its "up to date" so that the migrations script didn't try to run any migrations against it.

If we do that then existing installs only need to run dbprefix_drop.sql one time and ALL new installs never need to run the drop script or any other migration for that matter over time that's quite a few migration scripts for a new install.  Without data in the database running even 15-20 migration scripts takes no time at all, so it's not an issue of performance, but without dbprefix_drop.sql being an automatic migration it's the only way I can think of to get new installs not to have to manually run the script.

objecttothis:
An alternate that I can think of that would allow us to keep database.sql frozen but that also would have to be an exception to current policy is the following:
- Alter all migration scripts to remove ospos_
- dbprefix_drop.sql is manually run one time for existing installs regardless of which version they are migrating from.
- New installs don't need dbprefix_drop.sql since database.sql (tables.sql/constraints.sql) have already been modified.
- We continue the policy of database.sql being a reflection of the database when we started doing migrations.

In essence we are simply making it as if dbprefix was never used. This method will work, but requires a one-time exception to the rule of not modifying migration scripts.  However, I am certain that it will not break the migration chain.

daN4cat:
We adopted the automatic migration script approach to solve development, old and new install problems in one shot. You don't need to keep up to date many scripts, only one. You don't need to worry about running manual script but just upgrade the code will do.
Ideally we should go for a continuous upgrade straight from GitHub and reduce the size of releases.

If this drop of dbprefix causes problems with automatic migration then we should reconsider its value. I would not break the upgrade simplicity for normal users with something that has no critical value.

objecttothis:
> 
> 
> We adopted the automatic migration script approach to solve development, old and new install problems in one shot. You don't need to keep up to date many scripts, only one. You don't need to worry about running manual script but just upgrade the code will do.
> Ideally we should go for a continuous upgrade straight from GitHub and reduce the size of releases.
> 
> If this drop of dbprefix causes problems with automatic migration then we should reconsider its value. I would not break the upgrade simplicity for normal users with something that has no critical value.

Understood.  At this point the easiest solution is to drop this change, however the CI devs have said that they don't recommend the use of dbprefix and they plan to deprecate it.  At that point we will be faced with either not using newer versions of CI or dropping dbprefix.

I've been doing a little research and if the /application/config/database.php is writeable by server then I think I may be able to make the dbprefix_drop.sql part of automatic scripting with the following procedure:
1- code has all references to dbprefix removed except in database.php where ospos_ is still set for dbprefix.
2- On first login all the migration scripts run normally including the dbprefix_drop lines.
3- Immediately after the dbprefix_drop lines, migration of the /application/config/database.php needs to run to remove ospos_ from dbprefix.
4- sql scripts continue to run.

This does mean that all future SQL migration scripts (i.e., after the dbprefix_drop) will need to not include ospos_ hardcoded.

If I can get this to work then this satisfies the criteria to keep all migrations automatic and not modify past migration scripts.  Thoughts?

daN4cat:
I'm fine as long as it's not making the current "linear progression" useless, which means let's not broke the machine (or the toy) :-)

jekkos:
Quite a bit risky indeed to have the upgrade path rely on rewriting a source code file at runtime. To make this work we'll need to explicitly state in the installation instructions how the file permissions should be setup. Or we could just wait for people to post their issues here and take action accordingly (after verifying this approach works in a basic setup)

SteveIreland:
I noticed that I can get to the login page with only the existence of `ospos_sessions` and `ospos_app_config`.  No other tables were created, but once the login page was displayed I checked the database and noted that CodeIgniter automatically created `ospos_migrations`.

I wouldn't think it would be out of the question to leave `ospos_sessions`, `ospos_app_config` unchanged and treat those table names as "unprefixed".  I doubt if anyone is using a prefix other than ospos and if they are then they obviously have the necessary skills to adapt.

daN4cat:
I assume the files under the folder database/ need to revert back to the original.

objecttothis:
Yes, don't merge this into the master. It needs work to make it all automatic. The plan is to have this be a post 3.3.0 merge

dtomlins:
Silly question.  What is the "purpose" of this change?  I understand CI wants to drop the prefix, fine.  Is that the only reason?
I really like that all tables are prefixed ospos_ to keep them together and prevent namespace pollution conflicts with any other tables in that database.  It's also a clue to which app is using those tables - don't mess with them unless you know you'll break something (possibly outside of what you are thinking) when there are multiple applications using the same database space.

SteveIreland's comment of a few days ago suggests leaving 2 or 3 of the tables "hardcoded/unprefixed" with the ospos_ still there, how ugly would it be to do the same with all of the tables, and drop the CI prefix?  
That way the namespace stays exactly as it has been, and all the automatic migration issues are gone.
Yes, it's many files changed (anything that does a query), but that is a one-time, application only change that could make life easier.

Just a thought!
Dan

objecttothis:
> 
> 
> Silly question. What is the "purpose" of this change? I understand CI wants to drop the prefix, fine. Is that the only reason?
> I really like that all tables are prefixed ospos_ to keep them together and prevent namespace pollution conflicts with any other tables in that database. It's also a clue to which app is using those tables - don't mess with them unless you know you'll break something (possibly outside of what you are thinking) when there are multiple applications using the same database space.
> 
> SteveIreland's comment of a few days ago suggests leaving 2 or 3 of the tables "hardcoded/unprefixed" with the ospos_ still there, how ugly would it be to do the same with all of the tables, and drop the CI prefix?
> That way the namespace stays exactly as it has been, and all the automatic migration issues are gone.
> Yes, it's many files changed (anything that does a query), but that is a one-time, application only change that could make life easier.
> 
> Just a thought!
> Dan

@dtomlins Good thoughts.  No, CI's preferences are not the only factor here.  #2361 will also give you some background on the situation:
1. CI is planning to deprecate the feature
2. In some places the dbprefix is being hardcoded into the code and in others the dbprefix is being added by CI Query Builder.  This means that the benefit of dbprefix is null since OSPOS was never designed to have multiple instances using the same database space.
3. dbprefix breaks in CI Query Builder in some cases that work without dbprefix. (e.g., join with an update).
4. Handwriting queries (debugging, testing, etc) is more tedious with the ospos_ hard coded

In your case it sounds like you have additional tables you are using for custom code that either have dbprefix hardcoded or are relying on dbprefix.  Without seeing the code I can't say for sure, but it's not all that difficult to remove references to dbprefix and rename additional database tables to whatever you'd like them to be.  I hear you that you like the ospos_ on there to differentiate between your additional tables and the native ones but that's at the expense of the clutter of the rest of the code.  Why not rename your additional tables to make them stand out?

@SteveIreland 's suggestion was trying to find a way to make the mod happen and work around the issue of the OSPOS migration system being incremental.  I believe that this change can be made with having all tables conform (i.e., no dbprefix) AND fit into the migration system so that it doesn't mess up migrations.

daN4cat:
It's not a silly question but a valid one.
However I would argue that it's better to have different databases for better data separation and size management.
If you keep growing with one database having all the data in it, you will face issue one day in sharding the database.
Separate database can be distributed before you enter into a data sharding issue.

dtomlins:
:)  I read that thread too.  
2: The only and huge (external to development) benefit of the prefix is namespace - keeping all tables identified as part of the main project when there are add-ons in the same database.
3: CI breaks on joins anyway, or is at least sub-optimal on many joins - I see too many "using temporary" on this and other projects that depend on it.  There may be other ways that avoid it, but I see it regularly.
4: True - but only done "once"

Yes, I have a lot of other tables in the ospos database space that ospos can run without.  All are "related" to the primary project of ospos however.  Joins don't work well across multiple databases!  QueryBuilder diesn't appear to handle it and it's close to impossible across different machines!
 
My custom code isn't CI based, but hard-coded.  And no, it wouldn't take long to change it to remove the ospos_ prefix either, so that part isn't a real factor for me.
All of my tables in the database however ARE related to the big project, and substantially use the ospos_ tables.  

OSPOS handles the in-store sales well and a lot of stock adjusting and some reporting.

Our zen-cart website is updated periodically (every 30 minutes during business hours) to show proper inventory there.  So there are a number of tables to support that.  All sales from there once completed are recorded in the ospos database for reporting and inventory purposes.  Purchasing info and vendor availability tables, and some specific tables too.  Other project tables are in a different database and server :)

Yes @daN4cat that's certainly a factor, but no-where near needed.  The external tables are relatively small anyway, but are direct support to the function of ospos.  It could all fit in memory on my "next server" :grin:  which is why I try to have any and all user-queries be sub-second response ... and get extremely frustrated when I see a piece of an application have a 2 minute response time!  (like selecting the "items" screen in the current version of ospos.  I know that can't be right!)

Dan

daN4cat:
I wonder if the 2 min Items issue you see it's due to attributes addition.
Did you use custom fields before?

daN4cat:
I wonder if the 2 min Items issue you see it's due to attributes addition.
Did you use custom fields before?

daN4cat:
By the way you can enable the db query logging (check the config for the switch) and read queries and timing.

objecttothis:
> 
> 
> :) I read that thread too.
> 2: The only and huge (external to development) benefit of the prefix is namespace - keeping all tables identified as part of the main project when there are add-ons in the same database.

Or having multiple installations of OSPOS using one database (e.g., ospos1_ and ospos2).  Either case is not a directly supported scenario but can be made to work.  W

> 3: CI breaks on joins anyway, or is at least sub-optimal on many joins - I see too many "using temporary" on this and other projects that depend on it. There may be other ways that avoid it, but I see it regularly.

In my testing Join with update broke with dbprefix but did not break without dbprefix.  With that said the CI dev was surprised that it worked even without dbprefix.  I'm not saying we will be able to use Query Builder for everything, but it should work for most queries.  We have dozens of queries in the code which are not written using Query Builder and eventually I plan on doing some code cleanup to convert those to Query Builder.

> 4: True - but only done "once"

Only done once for coding, but when I and other devs are debugging coding changes we might write dozens of queries in phpmyadmin to make sure that our data is looking the way that we are expecting.

> Yes, I have a lot of other tables in the ospos database space that ospos can run without. All are "related" to the primary project of ospos however. Joins don't work well across multiple databases! QueryBuilder diesn't appear to handle it and it's close to impossible across different machines!

Agreed about Query Builder being limited.  In that case hard coding using db->query() is the route to take, but that's not dependent on dbprefix.  It just means that if dbprefix gets removed you need to take it out of the query.

> My custom code isn't CI based, but hard-coded. And no, it wouldn't take long to change it to remove the ospos_ prefix either, so that part isn't a real factor for me.
> All of my tables in the database however ARE related to the big project, and substantially use the ospos_ tables.

In the end decisions that get made (like this and others) have to primarily be made on whats best for the community but within the context of supported use of the project.  It's perfectly allowed to use your own tables and add custom code as long as the Open Source License is followed but we would be crazy to try to code for every use case outside the stock app.

With that said, we are developing framework for third party integrations which will allow some more flexibility to have third party integrations be written into the code and tables associated with them be used but even that isn't an external plugin system.

> OSPOS handles the in-store sales well and a lot of stock adjusting and some reporting.
> 
> Our zen-cart website is updated periodically (every 30 minutes during business hours) to show proper inventory there. So there are a number of tables to support that. All sales from there once completed are recorded in the ospos database for reporting and inventory purposes. Purchasing info and vendor availability tables, and some specific tables too. Other project tables are in a different database and server :)

Regarding this, what I am doing with our eCommerce site is writing API calls in OSPOS as a third party integration that the eCommerce site receives and processes.  Eventually this will be a two-way integration, but for now it will accomplish the need for communication without intermixing database tables.  It sounds like you have some of this with the "Other project tables" you are referring to.

> 
> Yes @daN4cat that's certainly a factor, but no-where near needed. The external tables are relatively small anyway, but are direct support to the function of ospos. It could all fit in memory on my "next server" 😁 which is why I try to have any and all user-queries be sub-second response ... and get extremely frustrated when I see a piece of an application have a 2 minute response time! (like selecting the "items" screen in the current version of ospos. I know that can't be right!)

In either 3.4.0 or 4.0 I am planning to do some optimization and code cleanup of both the database structure and the code (mostly queries in the code).  Several of the datatypes in the database are too large and not only take up far larger amount of space than they need to, but then when you run queries that are fairly complicated on thousands of rows it adds up in the amount of time that it takes to pull that information.  If done carefully and correctly, those changes will be seamless to the PHP side.  Additionally optimizing queries should speed things up significantly.

objecttothis:
> 
> 
> I wonder if the 2 min Items issue you see it's due to attributes addition.
> Did you use custom fields before?

Could be but I'm seeing a max 2 second delay on 15k items with more than a dozen attributes.  Sounds like his database is larger, but 2 minutes is a lot.  I think him enabling db logging will at least show which queries are taking the longest and then we can start with optimizing those queries first.  Use cases like @dtomlins could prove to be very helpful in query optimization since his is more on the extreme size end.

daN4cat:
I sorted the Reports queries that were a mess and orthogonal using temp tables to build dimensional views.
But I'm sure there is space for improvement.

dtomlins:
Hi.  I don't see an attributes addition (thank you :) - they
can be a nightmare! )   

I do see 10 "custom" fields, but that's nothing.  They are
empty varchars, so are adding maybe 20 bytes to the row.  No
big deal on performance at all.
I have at least 25 additional fields adding probably 200
bytes to each row, and even that is really insignificant in
query times.  It's faster than a join most times.

I've turned on query logging, and I'm seeing a couple of the
issues on that right away - the big one is aggregate
functions, which ALWAYS causes a temporary table build.
I'll look at the code in a while and see if I can find why
it is done that way.

-----Original Message-----

daN4cat:
Good, feel free to suggest query improvements or even better submit PRs so we can review and merge.

Thanks

daN4cat:
@objecttothis I guess this needs a refresh. Not sure where we were left with this. Almost a year old :-~

objecttothis:
@daN4cat yes, it will need a refresh.  It got put on back burner due to needing to implement third party integrations and for me to write the CLCdesq API integration.  We also need to make sure everyone is on board with the change.  If I remember right, there was some hesitation on the implications and the best way to implement the change.  One hurdle that needs to get figured out is how to do the change of removing the dbprefix without changing any pre-existing migrations files.  What I need to wrap my mind around is when we merge the change, the code will no longer look for a dbprefix in the database, but until the update script gets completed, the database will still have the prefix in it.  Unless I'm mistaken, the migration script won't run without logging in, but logging in can't happen if it can't connect to the database tables associated with employees.  Perhaps we need a manual update script that gets deleted upon completion?  I'm open to ideas.

In my opinion, we can punt this change until we are doing code cleanup, but that we should make the change.  The only reason we might not want to make the change is if a multi-site implementation required the use of dbprefix, so that's something that needs to be explored.

PR #2379, opened 2019-02-26
Attribute conversion fixes

objecttothis:
There were a few minor bugs with Attribute and attribute_lang.  They are fixed here.


objecttothis:
@jekkos we will need to propagate the lang_line addition in en_US to the other language files.  I think you said you have a script for that?

application/models/Attribute.php
```diff
@@ -303,6 +303,12 @@ private function convert_definition_type($definition_id, $from_type, $to_type)
 			$this->db->trans_complete();
 		}
 		
+		//Any other allowed conversion does not get checked here
+		else
+		{
+			$success = TRUE;
```
jekkos: isn't this just the start condition? So don't know whether we need to set it to TRUE again if we initialise it as true

jekkos: Ok check maybe in other methods, the regular pattern here is to init the variable as true, and then do a bitwise and for any operation that we need to check the result for.

jekkos:
I'll do the sync once it's merged. Thanks.

application/controllers/Attributes.php
```diff
@@ -105,7 +105,7 @@ public function save_definition($definition_id = -1)
 		}
 		else//failure
 		{
-			echo json_encode(array('success' => FALSE, 'message' => $this->lang->line('attribute_definitions_error_adding_updating').' '.
+			echo json_encode(array('success' => FALSE, 'message' => $this->lang->line('attribute_definition_error_adding_updating').' Attribute: '.
```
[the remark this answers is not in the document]
objecttothis: OK, will do.

application/models/Attribute.php
objecttothis: > 
> 
> isn't this just the start condition? So don't know whether we need to set it to TRUE again if we initialize it as true.

In the original pull request @daN4cat had me set the start condition for $success to FALSE and that's why I did that at the top.

application/controllers/Attributes.php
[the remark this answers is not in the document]
objecttothis: Done

application/models/Attribute.php
objecttothis: If we set the start condition to TRUE for $success then I will have to add an else below `if($this->check_data_validity($definition_id, $from_type, $to_type))` that sets $success to false.  I can do it either way.  Which one do you want?

daN4cat: It's fine like that.

PR #2384, opened 2019-03-01
Searchable Specific Customer Dropdown

WebShells:
Live Data Search for Specific Customer Input Report/Detailed Reports section in order to search for a specific client rather than typing it's name in a fast way.


WebShells:
Any problem in merging this?

PR #2388, opened 2019-03-01
Latest Commits Restore

WebShells:
Solving @jekkos terrible mistakes 🤣


jekkos:
@WebShells are you sure nothing came before these three commits? Actually it would be best to overwrite master with what you have locally as a remote ref to restore it correctly.

WebShells:
Well to what i remember, i was the last to submit these 2 changes... Total of 4002 commits was there... So yep thats all

jekkos:
Ok great then. Will lock the branch for force push then

PR #2391, opened 2019-03-03
Improve the summary payments report.

SteveIreland:
This replaces the existing summary payments report (which has a couple of problems) with a new version that provides a bit more coverage.  I've done all I can do this weekend and if someone has some time I'd appreciate some feedback on this to see if it's a bit closer to what's expected.

The major change is that the report is now segmented into sales and payments.  The sales should now balance to the bottom totals and I tried to distinguish between sales that have a balance due from the customer and those sales that required a cash refund to the customer.  The big problem is that the sale doesn't record the amount of the payment that is given back to the customer in change so it has to be calculated.

Returns are included but I haven't had a chance to closely review how returns are being reflected in the report and whether it's correct or not.


SteveIreland:
Made a change to limit the filter prompt to just the date range.  The order type and stock location is irrelevant for payments.

application/models/reports/Summary_payments.php
```diff
@@ -7,21 +7,121 @@ class Summary_payments extends Summary_report
...
+				. ' FROM ' . $this->db->dbprefix('sales') . ' AS sales '
+				. 'LEFT OUTER JOIN ' . $this->db->dbprefix('sales_items') . ' AS sales_items '
+				. 'ON sales.sale_id = sales_items.sale_id '
+				. 'LEFT OUTER JOIN ' . $this->db->dbprefix('sumpay_taxes_temp') . ' AS sumpay_taxes '
+				. 'ON sales.sale_id = sumpay_taxes.sale_id '
+				. 'WHERE ' . $where . ' GROUP BY sale_id
+			)'
+		);
+
+		$this->db->query('CREATE TEMPORARY TABLE IF NOT EXISTS ' . $this->db->dbprefix('sumpay_payments_temp') .
```
jekkos: Quite a lot of SQL here. Isn't there. I think some of these temporary tables are created in Sale and Receiving models. Maybe we have to move these to sale_payments model to keep code a bit shorter herer?

application/models/reports/Summary_payments.php
```diff
@@ -7,21 +7,121 @@ class Summary_payments extends Summary_report
...
+		$this->db->join('sumpay_items_temp AS sumpay_items', 'sales.sale_id = sumpay_items.sale_id', 'left outer');
+		$this->db->join('sumpay_payments_temp AS sumpay_payments', 'sales.sale_id = sumpay_payments.sale_id', 'left outer');
+		$this->db->where('sales.sale_status = ' . COMPLETED);
+		$this->_where($inputs);
+
+		$this->db->group_by("trans_type");
+
+		$sales = $this->db->get()->result_array();
+
+		$select = '\'' . $this->lang->line('reports_trans_payments') . '\' AS trans_group, ';
```
jekkos: same idea heere, might be better to extract it in a method and give it an appropriate name so it can be reused?

application/models/reports/Summary_payments.php
SteveIreland: Well, I could argue that dispersing code, that isn't going to be shared, across multiple files makes debugging a bit more challenging.  These temporary tables are designed specifically for the summary payments report so they aren't the same as the ones used by 4 of the detail reports  - but I moved their creation to the the Sale.php model. 

I'm not clear on that last statement.  The Sales_payments model that the queries are in is already a report specific model.

application/models/reports/Summary_payments.php
SteveIreland: It's report specific.  If I'm the developer and need a similar query for a different report then I would just copy the code and change it to meet my specific report's requirements.  I haven't touched the graphical summary payments report which has the highest possibility of reuse, but I suspect even that can't use exactly what is here.

Although I have done it in the past, I dislike anticipating possible future uses because then it is doing unnecessary things.  I'm sure you've never done this but I cringe when I think of the number of tables I've added Future Status 1, Future Status 2, etc. decades ago.

SteveIreland:
@daN4cat I believe this last iteration has addressed all of your input.  Thanks.

@jekkos Thank you for your feedback too.

application/models/reports/Summary_payments.php
jekkos: I could argue that it would make the code more testable. If you move the logic for generating the resultset to another object, you can easily swap out the object and it's behavior in case you want to verify how the other parts work. I still feel that this is something I should invest time in as having a project without tests is just horrible.

jekkos:
@daN4cat maybe wait until all conversations are resolved before merging, so we know everyone is aligned on the PR scope and that input has been processed. Reviewing a PR also takes time and even if the remarks look rather minor, I still think it's important to keep the codebase consistent and clean.

application/models/reports/Summary_payments.php
jekkos: Ok that's the idea to have models for the more complex reports, too. At least it will improve readability as you'll directly know from the object and method name what data is fetched and in which order. This saves you some time trying to understand as you don't need to interpret the SQL queries one after the other.

application/models/Sale.php
```diff
@@ -1420,5 +1420,56 @@ private function determine_sale_status(&$sale_status, $dinner_table)
 
 		return COMPLETED;
 	}
+
+	public function create_summary_payments_temp_tables($where)
```
jekkos: I will add this to the 'refactoring' issue as I think this should be in Sale_payments model instead (this class already has 1000+ lines)

application/models/reports/Summary_payments.php
jekkos: Also not really a fan of copying code, might be better to extract out the common parts and reuse as much as possible

daN4cat:
Right, reviewing from my mobile I missed that the function was moved to models/Sale.php
I'll fix it later since I merged this.
Outside that the review feedback was addressed and that's why I merged. Also because we keep going and going without getting to an end of 3.3.0.

daN4cat:
Done

jekkos:
Ok thanks for taking care of this. Maybe what we should try to focus on then is fixing the last known issues for this release... Just have to find some time to take care of this.

PR #2393, opened 2019-03-06
Add decimal attribute type

objecttothis:


application/models/Attribute.php
```diff
@@ -184,84 +185,98 @@ public function get_definitions_by_flags($definition_flags)
...
+		if($from === TEXT)
 		{
 			$this->db->select('item_id,attribute_value');
 			$this->db->from('attribute_values');
 			$this->db->join('attribute_links', 'attribute_values.attribute_id = attribute_links.attribute_id');
 			$this->db->where('definition_id',$definition);
 			$success = TRUE;
 			
-			foreach($this->db->get()->result_array() as $row)
+			if($to === DATETIME)
```
jekkos: I was sort of thinking to use an OO approach for this, so we represent a type of attribute and it's value as an object instance, with a validator and parser. Also the objects themselves would than be able to say if they can be converted to another type or not.

application/views/attributes/item.php
```diff
@@ -41,6 +41,11 @@
                 $value = (empty($attribute_value) || empty($attribute_value->attribute_value)) ? $definition_value['selected_value'] : $attribute_value->attribute_value;
                 echo form_input("attribute_links[$definition_id]", $value, "class='form-control valid_chars' data-definition-id='$definition_id'");
             }
+            else if ($definition_value['definition_type'] == DECIMAL)
+            {
+            	$value = (empty($attribute_value) || empty($attribute_value->attribute_decimal)) ? $definition_value['selected_value'] : $attribute_value->attribute_decimal;
```
jekkos: minor indentation

application/helpers/locale_helper.php
```diff
@@ -494,12 +494,18 @@ function dateformat_bootstrap($php_format)
...
 	return strtr($php_format, $SYMBOLS_MATCHING);
 }
 
 function valid_datetime($datetime)
 {
 	return preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $datetime);
 }
+
+function valid_decimal($decimal)
+{
```
jekkos: this is something that can then be moved to the DecimalAttribute class in case we'd like to take more of an OO approach here.

application/views/attributes/item.php
objecttothis: Fixed.  I'm not sure why this came up like that since it looks fine in eclipse. Nonetheless I deleted the forward tabs and readded them then the subsequent commit looks correct here too.

objecttothis:
> 
> 
> This PR is probably sufficient to make it work on the write side, but we still need the other fix to be able to read the fields properly from a search resultset. as said I might try to refactor that a bit to make the attribute value resultset parsing a bit cleaner (in this case adding a class that then splits up the concatenated GROUP_CONCAT into specific AttributeValue instances. Those will hold their definition and table column ids, so if we pass this as a list to the table_helper he'll be able to choose the correct instance according to the column name.

The issue you are referring to is affecting not just the decimal attribute type but datetime as well. Can we merge this PR then and have another PR that fixes the table display problem for both datetime and decimal attribute types?

objecttothis:
I added to this PR just a bit ago a commit that adds the unit of measurement into the decimal type.  I couldn't think of a reason why other attribute types would need the unit of measure but it can be extended to those in the future if needed.

application/models/Attribute.php
objecttothis: I'm assuming you are talking about an OO rewrite not just for DECIMAL but all the attribute types.  Is that accurate?

jekkos: Yes correct. Might be out of scope for this PR

odiea:
Did something not get passed on with this pull request. When I update a previous attribute I see the following errors?
<?php defined('BASEPATH') OR exit('No direct script access allowed'); ?>
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined variable: attribute_definition_info C:\wamp64\www\my-master\application\controllers\Attributes.php 123
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Trying to get property 'definition_flags' of non-object C:\wamp64\www\my-master\application\controllers\Attributes.php 123
ERROR - 2019-03-07 04:48:13 --> Severity: Warning --> Creating default object from empty value C:\wamp64\www\my-master\application\controllers\Attributes.php 123
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_type C:\wamp64\www\my-master\application\helpers\tabular_helper.php 575
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_id C:\wamp64\www\my-master\application\helpers\tabular_helper.php 585
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_name C:\wamp64\www\my-master\application\helpers\tabular_helper.php 586
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_type C:\wamp64\www\my-master\application\helpers\tabular_helper.php 587
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_group C:\wamp64\www\my-master\application\helpers\tabular_helper.php 588
ERROR - 2019-03-07 04:48:13 --> Severity: Notice --> Undefined property: stdClass::$definition_id C:\wamp64\www\my-master\application\helpers\tabular_helper.php 590

objecttothis:
Let me look into it.

objecttothis:
@odiea can you check your database ospos_migrations table to check to make sure version is 20190301124900

odiea:
20190301124900_decimal_attribute_type.php is what is there. I created a decimal attribute to see if it went away but it did not. I now get the following error when I try to update the decimal attribute
ERROR - 2019-03-07 05:38:27 --> Severity: Notice --> Undefined index: definition_type C:\wamp64\www\my-master\application\models\Attribute.php 359
also the other errors are still there if I try to update 1 of them.

odiea:
What should be placed in the measurement unit?

odiea:
I checked the db and the changes are showing there.

objecttothis:
its optional.  the measurement unit allows you to specify the unit for DECIMAL attributes.  For example, you may have an attribute called Weight.  Your unit might be kg, g, oz, lb, ton, etc.

odiea:
Ok. I was just checking to see if that was blank and creating this new issue.

objecttothis:
OK, I just downloaded the latest master to try to replicate.  It isn't behaving exactly the same for me but something is off.  The only abnormal visual queue I'm getting is that when I change an attribute, the table values in the view disappear for that line until I reload the page.  I did see in the logs:
>ERROR - 2019-03-07 17:06:44 --> Severity: Warning --> Creating default object from empty value /usr/local/www/public_html/pos_delta/application/controllers/Attributes.php 123

Let me look into it and fix it.  My test install wasn't behaving that way so I'll do a diff and see if I can figure out what's going on.

objecttothis:
OK, I found the problem (I think).  On line 122 of controllers/Attributes.php, somehow the variable name that it's using is definition_info, but it should be attribute_definition_info.  I'm not sure how that didn't get in the commit. Once I changed that on my local copy of the latest master it worked fine.  @odiea just to confirm, can you change line 122 on controllers/Attributes.php to read`$attribute_definition_info = $this->Attribute->get_info($row_id);` and confirm for me that the problems go away?

odiea:
Yes. that fixed the update issues for my other attributes but I still get the following error when I try to update the decimal attribute.
ERROR - 2019-03-07 06:25:19 --> Severity: Notice --> Undefined index: definition_type C:\wamp64\www\my-master\application\models\Attribute.php 359

odiea:
Here is what the table looks like in phpmyadmin. It works fine creating a decimal attribute but updating the decimal attribute.
2019-03-07 06_46_16-window

objecttothis:
Ok, you have two rows with the same name. Can you try creating an entirely new decimal and then modify it to see if it does the same thing? I think that may have been a side effect of the variable name problem

odiea:
1 was a deleted attribute so I went and deleted them in the db created a new decimal  attribute and that was ok then I went to update the decimal attribute and I still have the same issue.

odiea:
New info. Works fine with no errors when I take it out of development mode

objecttothis:
OK, I've fixed this line in the master.  I'm not sure what is causing the problem with development mode.

Issue #2016, opened 2018-06-08
Currency symbol not updating properly in locale config


mckaygerhard:
### Background information

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is:  demo
- OSPOS git commit hash is: demo
- PHP version is:  ??? please!
- MySQL or MariaDB version is:  ??? please!
- OS and version is:  ??? please!
- WebServer is:  well must be apache2 for sure
- Selected language is: en-US
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

i setup currentcy symbol and languaje in the demo site to en-US in Localisation, "$" in currency symbol and English in languaje 

but there-s no change currently symbol are same and show a garbage! Same problem with others languajes, when i set the languaje, only english are set!: 

![ospos-currency1](https://user-images.githubusercontent.com/1571724/41157151-a6a324fe-6af2-11e8-8e71-5fd429faed74.png)

jekkos:
[...]

PR #2394, opened 2019-03-08
Add iso currency code

objecttothis:
This adds the ISO 4217 currency code into the app.  At this point it's just needed for a third-party integration but could be used in other parts of the app.  For example, we could add code in that if the currency symbol is blank, the app could instead display the currency code (e.g., 1,234.00USD, when $ is not present)


application/views/configs/locale_config.php
```diff
@@ -229,7 +240,7 @@
 {
 	$('span').tooltip();
 
-	$('#currency_symbol, #thousands_separator').change(function() {
+	$('#currency_symbol, #thousands_separator, #currency_code').change(function() {
 		var field = $(this).attr('id');
 		var value = $(this).is(':checkbox') ? $(this).is(':checked') : $(this).val();
 		var data = { number_locale: $('#number_locale').val() };
```
jekkos: You're adding the field to the change listener but it's not updated after?

objecttothis: whoops, yeah that's a thing.  OK, let I'll fix it.

objecttothis: @jekkos  I'm confused about what modification I need to do here.  I added it to the listener because that listener calls ajax_check_number_locale() in the Config controller which does do checks/assignment.  Do I need to also be adding something in the javascript listener function?

jekkos: Yes the backend will give you the value to fill here. I saw you added the field to the ajax response. So you need to parse that and show it in the correct place. If you enter a locale then intl will figure out the currency code and symbol which we show in the frontend.

objecttothis: on the demo site here is how to replicate what I'm experiencing:

1. Config->Localization
2. Localization erase then type (az-AZ, en-US, en-GB, tr or fr)
3. example number changes format after click or tab to next field
IF currency symbol is empty it populates the currency symbol but with a currency symbol previously used.

If you erase the currency symbol then tab the example currency will change the example number symbol but each new locale repopulates the currency symbol with the wrong symbol.

daN4cat: I confirm that's not working very well. I never found the time to look into the issue and fix it. But agreed, fix the basics first before get the next in.

objecttothis: Glad I'm not taking crazy pills.  My problem is that I am not as proficient with JS and jquery so it takes me longer to figure out the existing code.  To be clear, the functionality of this PR is working in terms of the field itself, database, language, etc.  It's just the locale change that isn't working.  My recommendation is merging it and that way we can fix the jquery together, but I'm OK if you want to wait too.

I am less proficient with js and jquery than I am PHP.  I don't know who implemented the original code, but maybe someone besides me can work out the kinks with the jquery?

daN4cat: I would see like a learning opportunity, something new to learn and have fun with it.
You have done already a great job with your contributions, this is not a too big challenge for you, believe me.

objecttothis:
Thank you for the compliment. I agree that I can learn it and it would be good learning. The problem is that I'm under a huge deadline and have at least 4 more pr.

objecttothis:
As per #2016 I'll implement fixes for the locale not updating properly in this PR.

SteveIreland:
In my changes to the localization configuration the way it works ins that changing the number locale will set the currency symbol to the default for the new number locale.  The user can then change the symbol to the one they want to use.

I believe that is how it is supposed to work.

SteveIreland:
Obviously we are going to have a conflict in this area which is going to be a bit of a pain to resolve.  You might want to not bother "fixing it" at this time since I've already fixed and tested my correction.,

application/views/configs/locale_config.php
jekkos: My guess is that the jquery validation has stopped working after upgrading it. It might be better to move out that code into a separate validator as it doesn't seem to error if you fill in something invalid. That might also be the reason it's not updating?

objecttothis:
> 
> 
> Obviously we are going to have a conflict in this area which is going to be a bit of a pain to resolve. You might want to not bother "fixing it" at this time since I've already fixed and tested my correction.,

That's fine, but your changes don't include the ISO currency code, so the autofill will need to be implemented for the ISO currency code as well.  I will remove changes related to the autofill from my PR so that it doesn't conflict with your fix.  You can either rebase once mine is merged and add in the autofill for the currency code, or I will submit a new PR after you have merged to add in the autofill for the currency code as well.

application/views/configs/locale_config.php
objecttothis: OK, I corrected the code and now the jquery is properly populating the currency_code, currency_symbol and example_number.

objecttothis: @jekkos @daN4cat any reason not to merge now that the jquery is fixed?

application/migrations/20190317102600_add_iso_4217.php
```diff
@@ -0,0 +1,20 @@
...
+class Migration_add_iso_4217 extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		execute_script(APPPATH . 'migrations/sqlscripts/3.4.0_add_iso_4217.sql');
```
jekkos: target release 3.4 ? if we merge in master now it will probably still end up in 3.3 (unless we do a release branch before)

application/language/Language code definition
```diff
@@ -190,57 +132,35 @@ se-NO	Sami (Southern) (Norway)
...
-vi	Vietnamese
 vi-VN	Vietnamese (Viet Nam)
-xh	Xhosa
 xh-ZA	Xhosa (South Africa)
-zh	Chinese
 zh-CN	Chinese (S)
 zh-Hans	Chinese in the simplified scrip
 zh-Hant	Chinese in the traditional script
 zh-HK	Chinese (Hong Kong)
 zh-MO	Chinese (Macau)
```
jekkos: Thanks for cleaning this up, makes sense!

application/migrations/20190317102600_add_iso_4217.php
objecttothis: Ok let's merge now. I'll change the file name and put it in 3.3.0

Issue #2243, opened 2018-11-27
Discount Decimals is NOT included in the Configuration module, localization settings! WHY?


izeblood:
To Whom It May Concern:

As a result of my evaluation/testing of all the features of the OSPOS I found out that the discount decimal is not included in the Configuration,localization settings in order to add decimal point in the discount. As a result during the sales, discount will display like this: if you enter 2.5 the result is only 2, if you enter 2.6 the result is 3, but good to say that the actual percentage of the discount decimal value is CORRECT. 

adding the discount decimal can add simple improvements for the OSPOS.

saanplus:
is there a solution to this yet?

PR #2396, opened 2019-03-10
Fix discount amount formatting.

SteveIreland:
This is a simple resolution to #2243.

I don't believe it would be helpful to have a discount decimals field so i just left it unformatted since it is a multi-use field.

### DIFF
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -197,7 +197,7 @@ if(isset($success))
 
 							<td>
 								<div class="input-group">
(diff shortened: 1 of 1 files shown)


daN4cat:
Not sure we can assume no decimal in discounts.
We don't use as general principle but I remember few cases where it happened. Not sure that was on purpose or a mistake, can't remember to be honest.

daN4cat:
That to say it's fine to remove the no decimal view limitation and ok to leave it free field as long as the rest of the code works fine.

SteveIreland:
I went ahead and merged it since the only change was on the rendering of the field, so it won't impact anything else.

PR #2399, opened 2019-03-13
Add support for alternate currency.

SteveIreland:
Here is another variation of the pull request that supports printing documents in an alternate currency.  It seems to work pretty well, but it affects a lot and I haven't yet tested all permutations.


application/controllers/Sales.php
```diff
@@ -1139,10 +1192,23 @@ public function receipt($sale_id)
 		$this->sale_lib->clear_all();
 	}
 
+	public function receipt_alt($sale_id)
+	{
+		$data = $this->_load_sale_data($sale_id, TRUE);
```
jekkos: the method name is not very descriptive. Also regarding clean code using boolean arguments [is a nono](https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3).

jekkos: maybe we should try to rethink the way most of theses feature toggles works as our views tend to become very complex in the end (also per @daN4cat comment in another PR)

application/helpers/locale_helper.php
```diff
@@ -368,15 +368,35 @@ function to_decimals($number, $decimals, $type=\NumberFormatter::DECIMAL)
...
 	$config = get_instance()->config;
-	$fmt = new \NumberFormatter($config->item('number_locale'), $type);
-	$fmt->setAttribute(\NumberFormatter::MIN_FRACTION_DIGITS, $config->item($decimals));
-	$fmt->setAttribute(\NumberFormatter::MAX_FRACTION_DIGITS, $config->item($decimals));
+
+	$decimals = $config->item($decimal_type);
+
+	$CI =& get_instance();
+
+	$exchange_rate_set = $CI->session->userdata('exchange_rate_set');
```
jekkos: not sure if feasible, but wouldn't we be able to make a generic solution for rendering any view in a different currency by hooking the to_decimals and to_currency only? This could easily be done by setting the exchange rate as a session or flashdata variable at start of the request that is then checked during render after which then the conversion is applied. It would feel more natural as all views can stay as they are and we don't need to pass any variables down the callstack, which might mean that the code would be more concise and the functionality will work for any view we have.

Sorry to keep nitpicking about this as you probably already put in quite some effort (which is definitely appreciated)

jekkos: From that point of view I think you already did the legwork (passing the exchange variables etc). To make it work for any view you could parse the exchange rate variable in a pre_controller hook and copy it as a flashdata CI variable, so that every view can make use of it.

application/controllers/Sales.php
```diff
@@ -223,6 +223,11 @@ public function set_comment()
 		$this->sale_lib->set_comment($this->input->post('comment'));
 	}
 
+	public function set_exchange_rate()
+	{
+		$this->sale_lib->set_exchange_rate($this->input->post('exchange_rate'));
```
jekkos: Imho this should be a generic session or flashdata variable, it's not something bound only to sales but currency conversions might also apply to receiving receipts eg

application/controllers/Sales.php
```diff
@@ -509,6 +519,18 @@ public function complete()
 		$data['dinner_table'] = $this->sale_lib->get_dinner_table();
 		$data['cart'] = $this->sale_lib->get_cart();
 
+		$data['apply_exchange_rate'] = $this->sale_lib->is_apply_exchange_rate();
```
jekkos: this is what you should move to a precontroller hook

application/controllers/Sales.php
```diff
@@ -645,7 +667,9 @@ public function complete()
 				$invoice_view = $this->config->item('invoice_type');
 
 				// Save the data to the sales table
-				$data['sale_id_num'] = $this->Sale->save($sale_id, $data['sale_status'], $data['cart'], $customer_id, $employee_id, $data['comments'], $invoice_number, $work_order_number, $quote_number, $sale_type, $data['payments'], $data['dinner_table'], $tax_details);
+				$data['sale_id_num'] = $this->Sale->save($sale_id, $data['sale_status'], $data['cart'], $customer_id, $employee_id, $data['comments'],
```
jekkos: you can remove the currency as a param here as it will be available for save as a session variable

application/controllers/Sales.php
```diff
@@ -926,6 +956,19 @@ private function _load_sale_data($sale_id)
 		$data['payments'] = $this->sale_lib->get_payments();
 		$data['selected_payment_type'] = $this->sale_lib->get_payment_type();
 
+		$data['alt_currency_enabled'] = $this->config->item('use_alternate_currency') == '1';
+		if($force_exchange_rate)
+		{
+			$this->sale_lib->set_apply_exchange_rate(1);
+		}
+		$data['apply_exchange_rate'] = $this->sale_lib->is_apply_exchange_rate();
```
jekkos: also this code feels like something more generic and could run in a pre controller hook

application/controllers/Sales.php
jekkos: if you read the boolean from a session variable instead (that's set by a controller hook) then you won't need these extra methods

application/controllers/Sales.php
SteveIreland: flashdata suggests that the data has no value beyond the immediate transaction.  In this case, the current exchange rate will have value beyond the immediate transaction.  I expect the value will have value at least during the current day. The user might have a policy of determining the latest exchange rate at the beginning of each morning.   

Persisting it to the disk means that even if the user logs off, goes get lunch, and comes back and then needs to print an invoice in the alternate currency then they don't need to look up the exchange rate again.

"Receiving receipts" weren't part of the requirements and I'm not sure why a site would ever want to print those in the alternate currency - but you would know better than I since you are closer to it.

application/controllers/Sales.php
SteveIreland: ok

application/helpers/locale_helper.php
SteveIreland: The views can't stay as they are because when the document is printed in an alternate currency it is necessary to print on the document the exchange rate that is being used to print the document.  At this point that is the only change to the document views since I'm now using the session to start the set of values needed to render a document in the alternate currency.

I also, for this iteration, undid my changes to to_decimals and to_currency since I thought that was a notable concern of yours because I was originally passing the exchange rate data through those to the to_decimal function.  So at this point there is no "legwork" on those functions.

Also, I am not familiar with the CodeIgniter pre_controller hooks.  But I'll do a bit of research and maybe I can visualize what you are seeing as a possibility. 

CodeIgniter's flash data is new to me too.  I'll make some changes to persist the exchange rate set to flash data instead of the session.

application/controllers/Sales.php
SteveIreland: From the register screen they need to decide whether or not to print the document in an alternate currency.  Only the sales document needs to be generated in an alternate currency.  I'm now going to persist the exchange rate set of data to flashdata I'm not sure where and how a pre-controller hook would add any advantage to this.   Sorry ... it's just not registering.  But if you want to make the change after I update the pull request again my feelings won't be hurt.

application/controllers/Sales.php
SteveIreland: Renamed the methods and dropped the direct use off boolean values.   Still don't have a clue as to what your vision is for the controller hook.

application/controllers/Sales.php
jekkos: Ok you'll need to use a session variable to keep track of the last exchange rate and probably flashdata to store the variables that are needed to render the view in the scope of that request.

application/controllers/Sales.php
jekkos: the idea is to have the alternate currency rendering designed as a 'crosscutting concern'. So I was just thinking that we would probably like to go beyond receipts and render any view that has a currency in it in that 'alternate currency'. 

To make this happen you would normally need to adapt quite lot of controller methods (each one that renders currencies in this case).

jekkos: The idea would be to have a basic convention to render a view in the alternative currency, so that we require not specific controller code to be adapted. In this case I'd just prefix the view with alt_`methodname`. My idea was then to have a precontroller hook to handle this case, but after taking a further look a [custom route (callback)](https://codeigniter.com/userguide3/general/routing.html) might be more suitable. We just need to trigger some common code that sets the flashdata and then proceeds with rendering the view.

Having these flags set will then trigger code that renders the view in the alternative currency. This way we have a generic solution for all views in the application simply by adding a new link with different prefix.

application/controllers/Sales.php
daN4cat: Sorry I'm lost, is flashdata the session data in CI?

jekkos: yes it's just a request scoped set of variables.

jekkos: so in this case it's a perfect fit as the values you set are erased after the request ends

SteveIreland: But I don't want the exchange rate to "disappear" when the request ends OR session ends, I want the last one entered by the user to be displayed for the next document to be printed in the alternate currency so that they can determine whether it is still usable or needs to be changed.

I AM using the flash data which was a perfect solution to pass the alternate exchange rate data to the printed documents because after the document is rendered then the alternate exchange rate data is no longer needed for document rendering.

application/controllers/Sales.php
SteveIreland: I'm having a flashback to someone asking about the system supporting multiple languages by employee.  Before I started this issue I had confirmed that it was only the sales documents that needed to be rendered in the alternate currency.  I probably would not have started this if they had said otherwise because the impact would have been much to large to try to slip into 3.3

application/controllers/Sales.php
jekkos: Ok so that would be a global config option then which you show in the view if the flashdata has the alternate currency render flag set?

application/helpers/locale_helper.php
jekkos: Changing the views is not really an issue I think, it's they way the rendering is triggered that should be generic (convention using routes + setting a flashdata variable). In this case just checking whether the flashdata render flag is set and then show the content you want to show in that view.

SteveIreland:
So what's the plan for this pull request?

daN4cat:
Were all the review comments addressed?

SteveIreland:
I made the changes to the best of my ability.

And just to remind everyone.  I do not have a vested interest in this change other than the time I put in to it.  I volunteered to do this because it seemed like there were 2 or 3 people who were interested in this change and it was something that seemed like a real need, but I haven't heard from an interested third party in a long time, so there might no longer be an interest in it.

odiea:
No good deed goes unpunished. Sorry @SteveIreland

SteveIreland:
I learned a few things out of it so it wasn't wasted time .... even if @jekkos and @daN4cat decide to flush it.

dvidic:
> 
> 
> I learned a few things out of it so it wasn't wasted time .... even if @jekkos and @daN4cat decide to flush it.

Don't flush it, please.

odiea:
@dvidic It is currently in the unstable repo.

SteveIreland:
@jekkos are you still placing a hold on this?

jekkos:
I'm not completely convinced that we should merge this change as it stands here. 
In this case we can show the receipts / invoices in a different currency but I think showing reports or payment overviews in alternate currencies would make sense as well then? that's why I tried to push to think about a more generic approach to solve this problem here.

SteveIreland:
The scope of the project when I agreed to do it was limited to the sales documents and was only intended to include the dollar amounts.

I'm a single language person, so it's hard for me to imagine why reports would need to be given a similar treatment.   In my mind if we go there then that suggests that you might want to include more than just dollar amounts converted and that possibly you'd also want to have labels rendered in the alternate language.

So two instances of OSPOS configured similarly except for the language and currency codes might be a better alternative.  Even then we would need to modify it to support applying a conversion rate to all numeric values and then all reports would need to include a note in the heading that documents what conversion rate is being applied at the time the report is generated.

andriux1990:
Where can I download "Add support for alternate currency" for testing

nukerj:
Es interesante la funcionalidad multoneda como se usa en phppos y A2 Soft y Annual de saint también creo.conveniente se debería agregar el método de pago por transferencias y a crédito a cliente

SteveIreland:
I'm assuming this is a dead pull request. @jekkos made it clear he has no intention of merging it so you can dispose of this as appropriate.

andriux1990:
It's a shame because this feature is super interesting, applicable and makes a lot of sense for countries where the dollar is not the official currency, because they always use the dollar as a means of payment apart from the local currency.

mamleader:
Thanks for all developers worked in this feature and hope we can see future improves

mamleader:
> ompletely convinced that we should merge this change as it stands here.
> In this case we can show the receipts / invoices in a different currency but I think showing reports or payment overviews in alternate currencies would make sense as well. That's w

well , if you are in country that accept multi currency will know it is matter and not special case that need modification . also if you are in Tourist area you will have multi currency for customers it is one example of many

mamleader:
@SteveIreland  I tried multi-currency-sales-2 and I have small problem as in this picture

https://ibb.co/LZvQvRd

After applying exchange rate every cloumn is applied for exchange rate and should be only price not quantity or items count don't know what i done wrong thanks.

dvidic:
> 
> 
> I'm assuming this is a dead pull request. @jekkos made it clear he has no intention of merging it so you can dispose of this as appropriate.

As s person who started this idea/request I'm very sad about this decision. This feature has many good aspects in practice for many people.

odiea:
@SteveIreland This is a really good addition to OSPOS. Just a couple of minor issues after I Tested it. Great work. Thanks

andriux1990:
hey this don't close it yet

SteveIreland:
Do we really have any control?  These bots seem to be running wild and closing tickets faster than I can keep up.  Who let them in the door?

Seriously though, this pull request is is way behind and I'd say it's probably dangerous at this point.  Do I need to put some effort in bringing this current even though it's never going to be merged?

andriux1990:
@SteveIreland ends this ... you almost got it ... and it's very functional and useful.
I like the way you did it

mckaygerhard:
seems alternate currently are based in a fixed exchange rate ! that's are not so kinda usefully but it's a good started point! .. @SteveIreland  .. please rebase the git clone seems a good job that must be merged in main!

objecttothis:
@SteveIreland I just read through the conversation on #2399 and a few things are clear:
- There is demand for the enhancement.
- Some rework is needed on the enhancement design end to minimize the impact of the changes.
- That rework is outside the scope of what you can put into right now.

If it were me I wouldn't delete the branch and one of two things can happen:
1- @jekkos can stub out the design part and @SteveIreland or another dev can fill in the business logic.
2- Another dev can pick it up or be hired to complete the task by those that are interested in the feature.

I'll post this comment in #2399 in case others want to offer to pay for the feature.

andriux1990:
> @SteveIreland I just read through the conversation and a few things are clear:
> 
> * There is demand for the enhancement.
> * Some rework is needed on the enhancement design end to minimize the impact of the changes.
> * That rework is outside the scope of what you can put into right now.
> 
> If it were me I wouldn't delete the branch and one of two things can happen:
> 1- @jekkos can stub out the design part and @SteveIreland or another dev can fill in the business logic.
> 2- Another dev can pick it up or be hired to complete the task by those that are interested in the feature.

@objecttothis You are very smart and you have the ability to finish this ... because you don't give it a push

andriux1990:
> @SteveIreland I tried multi-currency-sales-2 and I have small problem as in this picture
> 
> https://ibb.co/LZvQvRd
> _quantity is only 1 not 10 and the stock is only 46 not 460_
> 
> After applying exchange rate every cloumn is applied for exchange rate and should be only price not quantity or items count don't know what i done wrong and it is shown also in receipt as well .

I made a clone of this and the same thing happens to me

andriux1990:
> > > @SteveIreland I just read through the conversation and a few things are clear:
> > > 
> > > * There is demand for the enhancement.
> > > * Some rework is needed on the enhancement design end to minimize the impact of the changes.
> > > * That rework is outside the scope of what you can put into right now.
> > > 
> > > If it were me I wouldn't delete the branch and one of two things can happen:
> > > 1- @jekkos can stub out the design part and @SteveIreland or another dev can fill in the business logic.
> > > 2- Another dev can pick it up or be hired to complete the task by those that are interested in the feature.
> > 
> > 
> > @objecttothis You are very smart and you have the ability to finish this ... because you don't give it a push
> 
> Thank you for the compliment. Due to the pandemic all of my time is currently focused on the enhancements and bug fixes that my bookstore needs to get data to the eCommerce site we are running. At some point down the road I may have the time to work on it, but I can't make any promises.

ok .. well, we are all affected by covid-19.

I hope you have a free time one day ... I know that in a short time you can solve this. Thank you

andriux1990:
thanks for the patience .. I started using this and I really like it, it's super.
by the time it picks up again, it would be nice to show the change in the two currencies.

The alternative currency a jewel, a diamond in the rough ... it just needs to be polished

andriux1990:
I have been using this for a couple of months and I really like it very much, it is very useful, especially in countries where more than one currency is used.

Guys why don't we put this in the master

daN4cat:
Is it working fine?

WebShells:
I think this is something we really need to add to the project. It needs some editions in order to be fully working as expected including reports etc, but still believe that we can start with adding what @SteveIreland had accomplished and we see where it leads us.

SteveIreland:
This branch hasn't been kept up because it was rejected by @jekkos a long time ago.  The issue was that @jekkos wanted to expand it beyond the scope of the original user request and, to be honest, I wasn't understanding the scope of what was being requested.  So sense I didn't understand the new requirements I felt it best for me to bow out.  However, if requested, I'll be happy to at least bring it up to the current commit level.  I really don't understand the "new reports" issue since the request was to simply generate an invoice or receipt in the customer's preferred currency.  In my mind that shouldn't affect the reporting system at all.  If the requirement is that OSPOS supports two or more currencies at the same time then I'm sure that is beyond the time I have available (and probably beyond my capabilities).

andriux1990:
@jekkos please accept this change .. it is really very useful .. and if it is in the master it is easier to adjust to many needs

jekkos:
I think if we want to add multicurrency support we should go again to the drawing board, as my expectation is that once we add this, bug reports and feature requests will come in for different scenario's we did not think about.

* How will you reconcile payments if you have received a cash amount in a different currency? If you want to do a cashup you need to take the current exchange rate into account
* What if you have an amount due but can't return cash in the currency paid? Shouldn't the receipt then state you have returned dollars instead of rupies? In that case you have two currencies on the receipt.

I just want to clarify that this story goes beyond converting some numbers on the receipt/invoice.

objecttothis:
> 
> 
> I think if we want to add multicurrency support we should go again to the drawing board as my expectation is that once we add this, bug reports and feature requests will come in for different scenario's we did not think about.
> 
> Eg
> 
>     * How will you reconcile payments if you have received a cash amount in a different currency? If you want to do a cashup you need to take the current exchange rate into account and convert the foreign currency amount and sum this up with what you have in your base currency
> 
>     * What if you have an amount due but can't return cash in the currency paid? Shouldn't the receipt then state you have returned dollars instead of rupies? In that case you need two currencies on the receipt.
> 
> 
> I just want to clarify that this story goes beyond converting some numbers on the receipt/invoice. Especially if you want the application to be fiscally correct. At least then we need a full traceability of what was paid in which currency at what time using what exchange rate.

Another situation, although rare, is when a country changes their currency.  They did that here a number of years ago and for a year we had to accept both currencies.  It was a hot mess.

andriux1990:
> I think if we want to add multicurrency support we should go again to the drawing board as my expectation is that once we add this, bug reports and feature requests will come in for different scenario's we did not think about.
> 
> Eg
> 
> * How will you reconcile payments if you have received a cash amount in a different currency? If you want to do a cashup you need to take the current exchange rate into account and convert the foreign currency amount and sum this up with what you have in your base currency
> * What if you have an amount due but can't return cash in the currency paid? Shouldn't the receipt then state you have returned dollars instead of rupies? In that case you need two currencies on the receipt.
> 
> I just want to clarify that this story goes beyond converting some numbers on the receipt/invoice. Especially if you want the application to be fiscally correct. At least then we need a full traceability of what was paid in which currency at what time using what exchange rate.

What can be done is to continue to maintain a main currency .. and use the other currency only when making a sale if the customer pays with the alternative currency ..

what I did on the receipt in register is that when a client pays with an alternative currency, register shows me the total value in both currencies, also on the receipt .. the two exchange rates appear, in the main currency and in the alternative currency

nukerj:
Yo tengo un fork de Open source Pos que le modifique para trabajar con 3
tipos de moneda configurables a la vez adaptado para recibir Bolívares
Venezuela, Dólares y pesos colombianos. En los próximos días trataré de
subir un repositorio a Github para quienes estén interesados puedan ir
trabajando en ello, me ha funcionado y los clientes a quienes se lo
modifique les ha servido de mucho.

Le diseñe un módulo de apertura y cierre de caja.

He dejado de trabajar en ello por falta de tiempo y apoyo económico porque
pues tengo que dedicarme 100% a hacer otros trabajos para generar ingresos.


Interesados dejar un mensaje y al tener listo el repositorio les indico
dónde verlo
Mi WhatsApp +584126447407

jekkos:
> > I think if we want to add multicurrency support we should go again to the drawing board as my expectation is that once we add this, bug reports and feature requests will come in for different scenario's we did not think about.
> > Eg
> > 
> > * How will you reconcile payments if you have received a cash amount in a different currency? If you want to do a cashup you need to take the current exchange rate into account and convert the foreign currency amount and sum this up with what you have in your base currency
> > * What if you have an amount due but can't return cash in the currency paid? Shouldn't the receipt then state you have returned dollars instead of rupies? In that case you need two currencies on the receipt.
> > 
> > I just want to clarify that this story goes beyond converting some numbers on the receipt/invoice. Especially if you want the application to be fiscally correct. At least then we need a full traceability of what was paid in which currency at what time using what exchange rate.
> 
> What can be done is to continue to maintain a main currency .. and use the other currency only when making a sale if the customer pays with the alternative currency ..
> 
> what I did on the receipt in register is that when a client pays with an alternative currency, register shows me the total value in both currencies, also on the receipt .. the two exchange rates appear, in the main currency and in the alternative currency

@andriux1990 fiscally that won't work as when you receive another currency it's value will change wrt to the main currency if you keep it in the cash register (as exchange rate also changes). Not sure how this is typically solved but we call this an exchange rate effect. This actually can make a big difference in profits and statements for companies that have a lot of exposure to foreign markets

andriux1990:
> @andriux1990 fiscally that won't work as when you receive another currency it's value will change wrt to the main currency if you keep it in the cash register (as exchange rate also changes). Not sure how this is typically solved but we call this an exchange rate effect. This actually can make a big difference in profits and statements for companies that have a lot of exposure to foreign markets

Hi @jekkos Thanks for your opinion, you are right ... in fact there are many ways to see how we can take alternate currency ... at least I was using it as it is in this code ... just to simplify it further and made some adjustments

nukerj:
IMG_20210619_222143
IMG_20210619_222111
IMG_20210619_222605
I apply it from the payment methods with 3 types of currencies Dollar, Colombian Pesos and Bolívares and apply formulas to reqlizar the calculations of the operations

nukerj:
Payments Methods on Sales
IMG_20210619_222040
IMG_20210619_222052

ANTOINE1128:
> I'm assuming this is a dead pull request. @jekkos @SteveIreland 
but I would like to set up a quick meeting to discuss this feature further

SteveIreland:
> > I'm assuming this is a dead pull request. @jekkos @SteveIreland
> > but I would like to set up a quick meeting to discuss this feature further

Definitely dead.  After 5 years I believe the corpse of what I started cannot be salvaged.

PR #2400, opened 2019-03-14
Update config_lang.php

objecttothis:
This is to correct a missing language line from integrations.  @jekkos can you run your script to propagate en-US language lines to the rest of the languages and in weblate?

### DIFF
--- a/application/language/en-US/config_lang.php
+++ b/application/language/en-US/config_lang.php
@@ -139,6 +139,7 @@ $lang["config_include_hsn"] = "Include Support for HSN Codes";
 $lang["config_info"] = "Information";
 $lang["config_info_configuration"] = "Store Information";
 $lang["config_integrations"] = "Integrations";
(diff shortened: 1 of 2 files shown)


application/language/en-US/config_lang.php
```diff
@@ -139,6 +139,7 @@
 $lang["config_info"] = "Information";
 $lang["config_info_configuration"] = "Store Information";
 $lang["config_integrations"] = "Integrations";
+$lang["config_integrations_configuration"] = "Third Party Integrations";
```
daN4cat: Could you please add it to en_GB too? Thanks

objecttothis: I can do it but I don't understand why.  @jekkos has a script that he runs that does this.

daN4cat: Thanks

Issue #2397, opened 2019-03-11
Sale Register improvement for Sale Due


odiea:
OSPOS Installation Info | 3.3.0 - deven-US
-- | --
GD: Enabled ✓BC Math: Enabled ✓INTL: Enabled ✓OpenSSL: Enabled ✓MBString: Enabled ✓Curl: Enabled ✓
Browser: ChromeServer Software: Apache/2.4.33 (Win64) OpenSSL/1.1.0g PHP/7.1.16PHP Version: 7.1.16 Server Port: 80DB Version: mysqlnd 5.0.12-dev - 20150407 - $Id: 38fea24f2847fa7519001be390c98ae0acafe387 $OS: Windows NT PAVILION23 10.0 build 17134 (Windows 10) AMD64
Application/logs: Writable ✓Public/uploads: Writable ✓public/uploads/item_pics: Writable ✓import_items.csv: Writable ✓import_customers.csv: Writable ✓ CSV permissions were set to 0664

The Sale due payment currently is not restricted . A customer should be Entered before the Due payment 
is accepted.

odiea:
I have tried this and it appears to work. If someone else could test it that would be great. application/views/sales/register under line 475. This is for the sales due payment needing a customer.
```
<?php
$payment_type = $this->input->post('payment_type');							
/ Only show this part if the payment cover the total and in sale or return mode
											
if($pos_mode == '1' && $payment_type != $this->lang->line('sales_due') && !isset($customer))
	{
	?>
<div class='btn btn-sm btn-success pull-right' id='finish_sale_button' tabindex="<?php echo ++$tabindex; ?>"><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $this->lang->line('sales_complete_sale'); ?></div>
<?php
}
?>
<?php
				
if($pos_mode == '1' && $payment_type = $this->lang->line('sales_due') && isset($customer))
{
?>
<div class='btn btn-sm btn-success pull-right' id='finish_sale_button'  tabindex="<?php echo ++$tabindex; ?>"><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $this->lang->line('sales_complete_sale'); ?></div>
<?php
}
?>

nathanzky:
I test the function and it works, if payment type is due it needs to input customer detail to complete the transaction.. 👍 👍 👍

odiea:
@nathanzky Thanks

odiea:
[...]

PR #2404, opened 2019-03-14
Customer Name for Due Sales & System info changes

WebShells:
Customer Name will be required in order to continue the sales once due payment is selected as per #2397.

### DIFF
--- a/application/language/en-US/sales_lang.php
+++ b/application/language/en-US/sales_lang.php
@@ -174,3 +174,4 @@ $lang["sales_work_order_number"] = "Work Order Number";
 $lang["sales_work_order_number_duplicate"] = "Work Order Number must be unique.";
 $lang["sales_work_order_sent"] = "Work Order sent to";
 $lang["sales_work_order_unsent"] = "Work Order failed to be sent to";
+$lang["sales_customer_optional"] = "(Required for Due Payments)";
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -473,15 +473,25 @@ if(isset($success))
 							</tr>
 						</table>
 					<?php echo form_close(); ?>
-						<?php
-						// Only show this part if the payment cover the total and in sale or return mode
-						if($pos_mode == '1')
+					<?php
+					$payment_type = $this->input->post('payment_type');							
+					// Only show this part if the payment cover the total and in sale or return mode
+																
+					if($pos_mode == '1' && $payment_type != $this->lang->line('sales_due') && !isset($customer))
 						{
-						?>
-						<div class='btn btn-sm btn-success pull-right' id='finish_sale_button' tabindex="<?php echo ++$tabindex; ?>"><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $this->lang->line('sales_complete_sale'); ?></div>
-						<?php
-						}
-						?>
+					?>
+					<div class='btn btn-sm btn-success pull-right' id='finish_sale_button' tabindex="<?php echo ++$tabindex; ?>"><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $this->lang->line('sales_complete_sale'); ?></div>
+					<?php
+					}
+					?>
+					<?php							
+					if($pos_mode == '1' && $payment_type = $this->lang->line('sales_due') && isset($customer))
+					{
+					?>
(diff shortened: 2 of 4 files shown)


application/views/sales/register.php
```diff
@@ -473,15 +473,25 @@
 							</tr>
 						</table>
 					<?php echo form_close(); ?>
-						<?php
-						// Only show this part if the payment cover the total and in sale or return mode
-						if($pos_mode == '1')
+					<?php
```
daN4cat: There are indentation issues, please fix.

application/language/en-US/sales_lang.php
```diff
@@ -174,3 +174,4 @@
 $lang["sales_work_order_number_duplicate"] = "Work Order Number must be unique.";
 $lang["sales_work_order_sent"] = "Work Order sent to";
 $lang["sales_work_order_unsent"] = "Work Order failed to be sent to";
+$lang["sales_customer_optional"] = "(Required for Due Payments)";
```
daN4cat: Could you please add en_GB translation too?

application/views/sales/register.php
```diff
@@ -474,14 +474,24 @@
 						</table>
 					<?php echo form_close(); ?>
 						<?php
+						$payment_type = $this->input->post('payment_type');							
 						// Only show this part if the payment cover the total and in sale or return mode
-						if($pos_mode == '1')
-						{
+
+						if($pos_mode == '1' && $payment_type != $this->lang->line('sales_due') && !isset($customer))
+							{
```
daN4cat: Indentation issues

application/views/sales/register.php
```diff
@@ -474,14 +474,24 @@
...
 						?>
 						<div class='btn btn-sm btn-success pull-right' id='finish_sale_button' tabindex="<?php echo ++$tabindex; ?>"><span class="glyphicon glyphicon-ok">&nbsp</span><?php echo $this->lang->line('sales_complete_sale'); ?></div>
 						<?php
 						}
 						?>
+						<?php							
+						if($pos_mode == '1' && $payment_type = $this->lang->line('sales_due') && isset($customer))
```
jekkos: wonder what the best thing to do is here. Should we disable the button or just not show it at all? I hope the layout doesn't change too much when removing this div?

odiea:
Sale when due is added without a customer.
2019-03-14 15_40_15-Window
After due is added with a customer
2019-03-14 15_40_56-Window

application/views/sales/register.php
WebShells: I see its fine... As you can see below in @odiea screenshots, not a big change ...

WebShells:
Copy Content

@odiea @objecttothis ? Copy info will copy table including html Format to be pasted in Git Issue...

odiea:
I tried your new updated version but I get the same copy paste results as the earlier version. Nothing like what you are displaying in this image.

WebShells:
@odiea the latest comment i showed isn't yet updated on the code, its local screenshot... So if its fine ill re-commit once editions are made...

odiea:
Ok Thanks. And thanks for submitting the payment due issue.

WebShells:
Anytime 🙂

jekkos:
Don't know if feasible but would be nice to [have a copy to clipboard function ](https://hackernoon.com/copying-text-to-clipboard-with-javascript-df4d4988697f)

objecttothis:
@wshells it looks good. I don't think the GitHub issue form will accept HTML and format it properly. I recommend making the copy to clipboard version a simple text version

WebShells:
@jekkos already i added js in order to copy table to clipboard, ill re-update @objecttothis i meant the text alignment not html format :x

WebShells:
@odiea can you try it now please ?

odiea:
I tried it so you can see the issue where I pasted it. Looks good.

objecttothis:
Looks good.  The DB version is still reporting the client version when it should be reporting the DB server version.

objecttothis:
Also it looks like from @odiea's screenshot that the hash is reporting as $id

odiea:
Yes it is because I am using my dev version not the bintray version.

objecttothis:
Ok nevermind then.

odiea:
Posted this new system info copy and paste. Line breaks look good

daN4cat:
Ready for a merge?

daN4cat:
I guess we would need to change the bug report template to adjust to the "go there and copy and paste the stuff here"

WebShells:
@daN4cat yes we should modify the issue template to direct people to copy info, but i'd like first to find a solution for the check list we have before the system info. Ex: Read opened/closed issues, checked FAQ, etc... Maybe i can find a way to add them to the page in a static way...
We can merge what we currently have, until i check other editions...

daN4cat:
Merged

Issue #1139, opened 2017-02-12
Add Docker LEMP stack support


jekkos:
### New Feature / Enhancement

* Add a docker compose alternative for LEMP (nginx + mariadb) support on latest master
* Check portability of the stack to run on embedded (ARM aka Raspberry Pi) devices

jekkos:
See #546 #1130

PR #2410, opened 2019-03-21
docker-compose with nginx reverse proxy and letsencrypt certbot

daN4cat:
Here the docker-compose enhancement to add a nginx reverse proxy with SSL termination based on a Letsencrypt certificate.
The certbot docker image manages the SSL certificate refresh every 3 months.

To run locally you need to use the init-selfcert.sh script.
If you want to try the Letsencrypt SSL, you need to use the init-letsencrypt.sh script.

Please note that myPhpAdmin is enabled on port 8000. If not required the nginx forward rule needs to be commented out.

I didn't touch the docker-compose.dev and .test as I'm not sure how those are used.

@jekkos the issue I had was fixed setting FORCE_HTTPS=true in the environment: docker-compose.yml file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-FROM php:7.2.18-apache
+FROM php:7.2-apache
 MAINTAINER jekkos
+
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     libicu-dev \
     libgd-dev \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2-apache
+FROM php:7.3-apache
 MAINTAINER jekkos
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,8 @@ volumes:
         driver: local
 
 networks:
-    ospos_net:
+    app_net:
+    db_net:
 
 services:
     sqlscript:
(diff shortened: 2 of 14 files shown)


init-letsencrypt.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
```
jekkos: maybe move these scripts to a subdirectory as the project root already contains quite some files?

jekkos:
Thanks for this useful addition. One remark is that maybe we should move some of the init scripts to a separate folder (or reorganize the files in the project root a bit).

Does it renew the certificate everytime the container starts or is the cert persisted after it's initialized?

jekkos:
Also related to #1139

jekkos:
We might also want to add some documentation on this setup in the wiki after it's merged.

data/nginx/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		location / {
+			return 301 https://$host$request_uri;
+		}
+	}
+
+	server {
+		listen 443 ssl;
+		server_name ospos.ospos;
```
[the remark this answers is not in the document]
daN4cat: Yes, something to parametrise

docker-compose.yml
```diff
@@ -51,17 +53,50 @@ services:
...
+          - ospos
+          - phpmyadmin
+          - certbot
+        volumes:
+          - ./data/nginx/nginx.conf:/etc/nginx/nginx.conf:ro
+          - ./data/nginx/error_log.log:/etc/nginx/error_log.log
+          - ./data/certbot/conf:/etc/letsencrypt
+          - ./data/certbot/www:/var/www/certbot
+        ports:
+          - "80:80"
```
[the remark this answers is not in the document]
daN4cat: Required by Letsencrypt to read the ACME file. Only that as the rest is redirected to https

init-letsencrypt.sh
daN4cat: Yes, I believe it's better to move everything to a folder and reorg the docker setup.

init-letsencrypt.sh
```diff
@@ -0,0 +1,75 @@
...
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot-nginx/certbot_nginx/options-ssl-nginx.conf > "$data_path/conf/options-ssl-nginx.conf"
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot/ssl-dhparams.pem > "$data_path/conf/ssl-dhparams.pem"
+  echo
+fi
+
+echo "### Creating dummy certificate for $domains ..."
+path="/etc/letsencrypt/live/$domains"
+mkdir -p "$data_path/conf/live/$domains"
+docker-compose run --rm --entrypoint "\
+  openssl req -x509 -nodes -newkey rsa:1024 -days 1\
```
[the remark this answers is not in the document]
daN4cat: This is just the temp cert, right?

daN4cat:
Yes, it needs some polishing. This is just functioning and a draft for discussion.

The LEMP stack means we would need to change ospos from Apache to nginx. Something I can do as second step.
This is more a reverse proxy to terminate SSL with free cert and auto renewal.

I took some of the docker code from GitHub, there might be other ways to solve the same problem.

daN4cat:
The cert is initially created in a manual way and signed, it stays in a local folder. But I need to countercheck it.

And yes about the wiki documentation.

daN4cat:
@jekkos see if the new setup looks better.

All Docker business is now in a docker folder.

From docker folder:
- for local setup run the script install-local.sh
- for server setup run the script install-server.sh **
- to tidy up run the script uninstall.sh

** to configure all the parameters change the file .env, especially if you run this on a server in the public Internet!

I didn't touch the .dev and .test files because I don't know how those are used. It's probably better if you update them.

jekkos:
@daN4cat I meant that moving only the init scripts to a subfolder would be enough. It might be rather hard to move the Dockerfiles themselves as well (as then you break all references from inside the mounted paths in the container itself). Most of the projects do have those in their root folder anyway.

jekkos:
.dev and .test files are not much different from an actual container setup. They assume that you mount your working directory into the container which is handy for development. It also adds an extra xdebug package that shouldn't be installed in a production setup. I might want to move this into the regular file by adding a flag read from an env variable instead for example.

daN4cat:
Indeed, I need to move Docker files back to the root.
The mounting of folder is a complication.

daN4cat:
I don't need to touch that much dev and test, you don't need the nginx reverse proxy for that.
I might look into touching a bit the networking, ports and the images but not too much else.

objecttothis:
> The LEMP stack means we would need to change ospos from Apache to nginx. Something I can do as second step.
I do think we need to retain general Apache compatibility for OSPOS.

data/nginx/nginx.conf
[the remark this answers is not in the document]
daN4cat: This is done, the .env variable is used to fill in the server name

init-letsencrypt.sh
daN4cat: script moved to docker subfolder.

In order to do install this, from local do:

> sudo docker/install-local.sh

for a server case (remember to change the docker/.env values!!)

> sudo docker/install-server.sh

daN4cat:
@jekkos do you want to give it a try on your server before I merge this?

Dockerfile
```diff
@@ -1,5 +1,6 @@
-FROM php:7.2.18-apache
+FROM php:7.2-apache
```
jekkos: we might want to stick with 7.2.18 here (latest)

daN4cat: I was under the impression that the docker image with this generic version format will pick up the latest.
Also I may bumped it to be 7.3 so we test the latest.

daN4cat: Upgraded to 7.3

daN4cat:
To use this new Docker setup.

From ospos root folder:
- > chmod ugo+x docker/*.sh
- for local setup run the script: > docker/install-local.sh (say y everytime you are asked to renew the cert)
- for server setup run the script: > docker/install-server.sh **
- to tidy up run the script: > docker/uninstall.sh

** to configure all the parameters change the file .env, especially if you run this on a server in the public Internet!
Also make sure you close phpMyAdmin if you don't want to expose that to the public internet.

daN4cat:
@jekkos are you fine for me to merge this?

PR #2412, opened 2019-03-23
Adapt Cashups to new Summary Payments Report

SteveIreland:
This change adapts the Cashups program to the revised Summary Payments report.

It also includes a correction to the Summary Payments report to eliminate the need for the MAX() in the SQL.

### DIFF
--- a/application/controllers/Cashups.php
+++ b/application/controllers/Cashups.php
@@ -106,22 +106,25 @@ class Cashups extends Secure_Controller
 
 			foreach($reports_data as $row)
 			{
-				if($row['payment_type'] == $this->lang->line('sales_cash'))
+				if($row['trans_group'] == $this->lang->line('reports_trans_payments'))
 				{
-					$cash_ups_info->closed_amount_cash += $this->xss_clean($row['payment_amount']);
-				}
-				elseif($row['payment_type'] == $this->lang->line('sales_due'))
-				{
-					$cash_ups_info->closed_amount_due += $this->xss_clean($row['payment_amount']);
-				}
-				elseif($row['payment_type'] == $this->lang->line('sales_debit') || 
-						$row['payment_type'] == $this->lang->line('sales_credit'))
-				{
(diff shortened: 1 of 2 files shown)


odiea:
@SteveIreland I know this is an oldie but goodie but how do I get a return to show in Cashups? I have tried using elseif($row['trans_type'] == $this->lang->line('sales_return')) but nothing shows up where it does in the summary payments report.

odiea:
This only seems to show on a cash return. I tried a return on a credit card and it worked ok. I guess I am looking for the cause of it not showing as a -cash in the cashup cash.

SteveIreland:
@odiea  What happens (unless there is a bug) is that there is only one cash transaction.  That cash transaction has both the amount tendered and the cash refund amount separate on the database record.. In Cashup, Summary Payments, and other transaction reports the net amount for the cash transaction is the amount tendered minus the cash return amount.   So the cash return amount in a cash only scenario isn't reflected as a negative payment.

A negative cash amount is only necessary (to stay in balance) when the original payment is some other payment type.

I thought that would be the "least impact" approach that still made sense.   The payment amounts now reflect the sales amount and there is still only one cash payment (in normal scenarios).

After the work was done it seemed like @jekkos started feeling that a separate transaction should have been logged.  There are two ways to handle this.  First the transaction could still still be a cash transaction type with a negative amount.  The second is to have a distinct transaction type for Refunds and maintain a positive transaction amount.  I think @jekkos leans toward the former.

If you and @jekkos want it rewritten to have separate transactions I'm happy to redo it while the code is fairly fresh on my mind.   Just let me know.

daN4cat:
I have some cases of negative cash transaction because by mistake they made a card transaction for the full amount forgetting a discount then returned the discount as a cash rebate resulting in a negative cash figure.
When I edit that sale the figure is set to 0 instead of negative and I need to fix that bug.
I still need to test mashup behaviour but I hope the cash amount would be correct.

odiea:
I guess I must have been the bug because it is working fine this morning. Cash is being taken out. Thanks

daN4cat:
Good

PR #2415, opened 2019-03-26
Dt performance items1

dtomlins:
**THIS MOD REQUIRES DATABASE CHANGE:
create index trxbydate ospos_inventory (trans_date);**

This is the commented version, with a lot of "commented" code in case re-instatement is necessary,
as well as give a better idea of the changes.

Query times on mine (62,000 items, 1.07 million inventory transactions) drops from
62.5 + 31.6  seconds  (94.1) on opening to a typical  0.62 seconds on opening + .05 if transactions are queried.

Searching by date through ospos_inventory is NOT required or done on initial load, but only if selected in the options box.  Option added for transactions in date range AND no transactions in date range.

Ideally the checkmarks would be mutually exclusive, but has_transactions has precedence.

** Uses SQL_CALC_FOUND_ROWS to populate next query FOUND_ROWS() to eliminate time required to count the records again - saving typically .32 of a second.  (was 32 seconds)
** has to turn off dbprefix temporarily in controller to convince CodeIgniter that it isn't a table needing pre-fixing.

I look forward to your comments!
Dan


dtomlins:
Good!  No idea how I/github did that!  I just cloned and forked on the weekend.

Withdrawing this one.  New version should solve #2416

jekkos:
Ok funny as at first we used SQL_CALC_ROWS as well, but for some reason we factored it out. If I remember well it was because it didn't add any significant performance improvement.

If you want to have a more or less constant query time you can use LIMIT clause which speeds up fetching a resultset. But that leaves you with the issue that you don't know how many rows there were in the resultset.

jekkos:
Also for the PR mind to leave out the code you removed, no need to keep it commented as the version control will handle this for you instead. Also don't add your initials behind every changed line of code (same idea applies there). We can add your name to the copyright list once merged.

PR #2416, opened 2019-03-26
Dt performance patch 1

dtomlins:
Upload version 2.  I have no idea how from a bad repo - I just forked 2 days ago!

THIS MOD REQUIRES DATABASE CHANGE:
create index trxbydate ospos_inventory (trans_date);

This is the commented version, with a lot of "commented" code in case re-instatement is necessary,
as well as give a better idea of the changes.

Query times on mine (62,000 items, 1.07 million inventory transactions) drops from
62.5 + 31.6 seconds (94.1) on opening to a typical 0.62 seconds on opening + .05 if transactions are queried.

Searching by date through ospos_inventory is NOT required or done on initial load, but only if selected in the options box. Option added for transactions in date range AND no transactions in date range.

Ideally the checkmarks would be mutually exclusive, but has_transactions has precedence.

** Uses SQL_CALC_FOUND_ROWS to populate next query FOUND_ROWS() to eliminate time required to count the records again - saving typically .32 of a second. (was 32 seconds)
** has to turn off dbprefix temporarily in controller to convince CodeIgniter that it isn't a table needing pre-fixing.

I look forward to your comments!
Dan

### DIFF
--- a/application/models/Customer.php
+++ b/application/models/Customer.php
@@ -100,7 +100,7 @@ class Customer extends Person
 	{
 		// create a temporary table to contain all the sum and average of items
 		$this->db->query('CREATE TEMPORARY TABLE IF NOT EXISTS ' . $this->db->dbprefix('sales_items_temp') .
-			' (INDEX(sale_id))
+			' (INDEX(sale_id)) ENGINE=MEMORY
 			(
 				SELECT
 					sales.sale_id AS sale_id,
--- a/application/models/Item.php
+++ b/application/models/Item.php
@@ -100,7 +100,10 @@ class Item extends CI_Model
 	*/
 	public function get_found_rows($search, $filters)
 	{
-		return $this->search($search, $filters, 0, 0, 'items.name', 'asc', TRUE);
+		$this->db->select('FOUND_ROWS() as count');	//dmt
+		return $this->db->get()->row()->count;		//dmt
+
+//dmt		return $this->search($search, $filters, 0, 0, 'items.name', 'asc', TRUE);
 	}
 
 	/*
@@ -115,6 +118,7 @@ class Item extends CI_Model
 		}
 		else
 		{
+/*  REPLACED
 			$this->db->select('items.item_id AS item_id');
 			$this->db->select('MAX(items.name) AS name');
 			$this->db->select('MAX(items.category) AS category');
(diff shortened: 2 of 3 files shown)


application/models/Customer.php
```diff
@@ -100,7 +100,7 @@ public function get_stats($customer_id)
 	{
 		// create a temporary table to contain all the sum and average of items
 		$this->db->query('CREATE TEMPORARY TABLE IF NOT EXISTS ' . $this->db->dbprefix('sales_items_temp') .
-			' (INDEX(sale_id))
+			' (INDEX(sale_id)) ENGINE=MEMORY
```
jekkos: Didn't know about this one but looks like a neat little trick. According to docs performance is dependent on how you have configured your database server.

application/models/Item.php
```diff
@@ -137,41 +141,91 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'i
...
+
+			$this->db->select('suppliers.person_id AS person_id');
+			$this->db->select('suppliers.company_name AS company_name');
+			$this->db->select('suppliers.agency_name AS agency_name');
+			$this->db->select('suppliers.account_number AS account_number');
+			$this->db->select('suppliers.deleted AS deleted');
+*/
+
+// Return named fields variant 2
+			$this->db->select('SQL_CALC_FOUND_ROWS items.* ',FALSE);
```
jekkos: a lot shorter, but does it work with 5.7?

application/models/Item.php
```diff
@@ -137,41 +141,91 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'i
...
 		}
 
-		if(empty($this->config->item('date_or_time_format')))
+//		if(empty($this->config->item('date_or_time_format')))
+		if($filters['has_transactions'] != FALSE)
 		{
-			$this->db->where('DATE_FORMAT(trans_date, "%Y-%m-%d") BETWEEN ' . $this->db->escape($filters['start_date']) . ' AND ' . $this->db->escape($filters['end_date']));
+//			$this->db->where('DATE_FORMAT(trans_date, "%Y-%m-%d") BETWEEN ' . $this->db->escape($filters['start_date']) . ' AND ' . $this->db->escape($filters['end_date']));
+			$this->db->where('items.item_id in (select distinct(trans_items) from ospos_inventory use index (trxbydate) where trans_date BETWEEN ' . $this->db->escape($filters["start_date"]) . ' AND ' . $this->db->escape($filters["end_date"] . ' 23:59:59') . '  AND trans_location = ' . $filters["stock_location_id"] . ')' );  //dmt
+		} elseif ($filters['no_transactions'] != FALSE)
```
jekkos: mind indentation here

application/models/Item.php
```diff
@@ -232,7 +286,9 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 'i
 		}
 
 		// avoid duplicated entries with same name because of inventory reporting multiple changes on the same item in the same date range
-		$this->db->group_by('items.item_id');
```
jekkos: Agree that it would be a lot cleaner to get rid of the group by, but it's indeed still needed by attributes.
Also removing the inventory query will probably break the date filtering functionality.

If you can find a way to reconstruct the full resultset by for example appending results of different queries then that would be nice.

application/models/Customer.php
dtomlins: It's one of the "little" things.  Most of them are.  ENGINE=MEMORY drops just about 3/4 of a second ...  not enough to worry about, until it is repeated 25, 50 or 100 times.
I want to test against a different version of storage engine too.  MEMORY works well for THAT query, and would help the next except it isn't allowed.  The GROUP_CONCAT creates a "text" field that cannot be in that engine.
My DevTest server is "small" settings by design.  So I design for good performance once the code escapes into the real world.  When the result exceeds the size of memory allocated it fails badly.  Yes, the error is logged, but the operator screen shows no data.  Not even an error message.

application/models/Item.php
dtomlins: I'll look at the attributes, if I can figure what they are best used for, but a quick look at them didn't show that they would create multiple returned data.  Duplicates would be thrown out by the group by anyway!
I'm still trying to figure out the "operator setup" part of the taxes!  Looks like it could work, sort-of for here (Canada), but figuring out where to put numbers has me stumped.  Putting the tax-rate into each of the thousands of products is a non-starter!  What happens when the rate changes? Edit all the products? I don't think that's yours though :)

daN4cat: I'm sure @steveireland would love to hear about your Taxation requirements. He brilliantly reworked all of this.

daN4cat: The group was required for the inventory issue to remove duplication. Please test that report.

We should really build a test harness for the models to help with any refactoring.

application/models/Item.php
daN4cat: Yes it's supported: https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows

I need to check if the performance improvement is delivered in all cases or just for small deployments. Out of curiosity.

application/models/Item.php
jekkos: thinking about it further, I feel that it might be better to use a separate query if the date filtering is used only. It will simplify the main query a lot and probably yield result quite a bit faster.

daN4cat:
@dtomlins any progress with your improvements?

ebrahimidelawar:
hello dear friends
please Loading speed very slow with large DB in item option please help me dear friend

ebrahimidelawar:
hello dear friend
please Loading speed very slow with large DB in item page  please i don't want show page of item?

ebrahimidelawar:
i install new wamp and new version of opensourcepos .3.4.0 but not solve my problem please help me?

PR #2420, opened 2019-04-07
Fix Sales Tax Migration

SteveIreland:
This fix seems to be working.   I created some 3.0.2 test data and then applied the migration and there weren't any failures.

### DIFF
--- a/application/migrations/20170502221506_sales_tax_data.php
+++ b/application/migrations/20170502221506_sales_tax_data.php
@@ -1,79 +1,79 @@
 <?php if (!defined('BASEPATH')) exit('No direct script access allowed');
-
 class Migration_Sales_Tax_Data extends CI_Migration
 {
+	const ROUND_UP = 5;
+	const ROUND_DOWN = 6;
+	const HALF_FIVE = 7;
+	const YES = '1';
+	const VAT_TAX = '0';
+	const SALES_TAX = '1';
+
 	public function __construct()
 	{
 		parent::__construct();
-
-		$this->load->library('sale_lib');
+		$this->load->library('tax_lib');
 	}
 
 	public function up()
 	{
 		$number_of_unmigrated = $this->get_count_of_unmigrated();
-
 		error_log('Migrating sales tax history.  The number of sales that will be migrated is '.$number_of_unmigrated);
-
 		if($number_of_unmigrated > 0)
 		{
 			$unmigrated_invoices = $this->get_unmigrated($number_of_unmigrated)->result_array();
-
 			foreach($unmigrated_invoices as $key=>$unmigrated_invoice)
 			{
 				$this->upgrade_tax_history_for_sale($unmigrated_invoice['sale_id']);
 			}
 		}
-
 		error_log('Migrating sales tax history.  The number of sales that will be migrated is finished.');
 	}
-
 	public function down()
 	{
-
 	}
-
 	private function upgrade_tax_history_for_sale($sale_id)
 	{
 		$CI =& get_instance();
 		$tax_decimals = $CI->Appconfig->get('tax_decimals', 2);
-		$tax_included = $CI->Appconfig->get('tax_included', 0);
-		$customer_sales_tax_support = $CI->Appconfig->get('customer_sales_tax_support', 0);
-		$tax_type = $tax_included ? Tax_lib::TAX_TYPE_INCLUDED : Tax_lib::TAX_TYPE_EXCLUDED;
-
+		$tax_included = $CI->Appconfig->get('tax_included', Migration_Sales_Tax_Data::YES) == Migration_Sales_Tax_Data::YES;
+		$customer_sales_tax_support = FALSE;
+		if($tax_included)
+		{
+			$tax_type = Migration_Sales_Tax_Data::VAT_TAX;
+		}
+		else
+		{
+			$tax_type = Migration_Sales_Tax_Data::SALES_TAX;
+		}
 		$sales_taxes = array();
 		$tax_group_sequence = 0;
-
 		$items = $this->get_sale_items_for_migration($sale_id)->result_array();
 		foreach($items as $item)
 		{
 			// This computes tax for each line item and adds it to the tax type total
 			$tax_group = (float)$item['percent'] . '% ' . $item['name'];
-			$tax_basis = $this->sale_lib->get_item_total($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], PERCENT, TRUE);
+			$tax_basis = $this->get_item_total($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], TRUE);
 			$item_tax_amount = 0;
 			if($tax_included)
 			{
-				$item_tax_amount = $this->sale_lib->get_item_tax($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], PERCENT, $item['percent']);
+				$item_tax_amount = $this->get_item_tax($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], $item['percent']);
 			}
 			else
 			{
-				$item_tax_amount = $this->tax_lib->get_sales_tax_for_amount($tax_basis, $item['percent'], PHP_ROUND_HALF_UP, $tax_decimals);
+				$item_tax_amount = $this->get_sales_tax_for_amount($tax_basis, $item['percent'], PHP_ROUND_HALF_UP, $tax_decimals);
 			}
 			$this->update_sales_items_taxes_amount($sale_id, $item['line'], $item['name'], $item['percent'], $tax_type, $item_tax_amount);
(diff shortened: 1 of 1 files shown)


application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -1,79 +1,72 @@
...
-
 	public function down()
 	{
-
 	}
-
 	private function upgrade_tax_history_for_sale($sale_id)
 	{
 		$CI =& get_instance();
-		$tax_decimals = $CI->Appconfig->get('tax_decimals', 2);
```
jekkos: I think you should keep the Appconfig->get() here. Issue is that if you are migrating from older versions that do not have the decimals configured yet they also won't be returned from config->config() as this uses cached values and those are only parsed at the beginning of the request, which might span multiple migrations

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -1,79 +1,72 @@
...
-		$tax_included = $CI->Appconfig->get('tax_included', 0);
-		$customer_sales_tax_support = $CI->Appconfig->get('customer_sales_tax_support', 0);
-		$tax_type = $tax_included ? Tax_lib::TAX_TYPE_INCLUDED : Tax_lib::TAX_TYPE_EXCLUDED;
-
+		$tax_decimals = $CI->config->config['tax_decimals'];
+		$tax_included = $CI->config->config['tax_included'];
+		$customer_sales_tax_support = FALSE;
+		if($tax_included)
+		{
+			$tax_type = '0';
```
jekkos: any reason for not using the defined constants here?

application/migrations/20170502221506_sales_tax_data.php
jekkos: Think as a best practice we might just want to use AppConfig in migrations at all times.

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -1,79 +1,72 @@
...
 		$sales_taxes = array();
 		$tax_group_sequence = 0;
-
 		$items = $this->get_sale_items_for_migration($sale_id)->result_array();
 		foreach($items as $item)
 		{
 			// This computes tax for each line item and adds it to the tax type total
 			$tax_group = (float)$item['percent'] . '% ' . $item['name'];
-			$tax_basis = $this->sale_lib->get_item_total($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], PERCENT, TRUE);
+			$tax_basis = $this->get_item_total($item['quantity_purchased'], $item['item_unit_price'], $item['discount_percent'], TRUE);
```
jekkos: Ok I guess it's probably better to keep that specific version of this method within the migration indeed. Then we can just update code elsewhere without worrying that any historical migration can break as a result of a code rewrite.

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +112,183 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+	public function get_sales_tax_for_amount($tax_basis, $tax_percentage, $rounding_mode, $decimals)
+	{
+		$tax_fraction = bcdiv($tax_percentage, 100);
+		$tax_amount = bcmul($tax_basis, $tax_fraction);
+		return $this->round_number($rounding_mode, $tax_amount, $decimals);
+	}
+
+	public function round_number($rounding_mode, $amount, $decimals)
+	{
+		if($rounding_mode == 5)
```
jekkos: same remark here for the constants

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +112,183 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+			array_multisort($sort['print_sequence'], SORT_ASC, $sales_taxes);
+		}
+		$decimals = totals_decimals();
+		foreach($sales_taxes as $row_number => $sales_tax)
+		{
+			$sale_tax_amount = $sales_tax['sale_tax_amount'];
+			$rounding_code = $sales_tax['rounding_code'];
+			$rounded_sale_tax_amount = $sale_tax_amount;
+
+			if ($rounding_code == PHP_ROUND_HALF_UP)
```
jekkos: why not just call `round($sale_tax_amount, $decimals, $rounding_code)` here? Will reduce a 15 lines of code to just 1;

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +112,183 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+			{
+				$rounded_sale_tax_amount = round($sale_tax_amount, $decimals, PHP_ROUND_HALF_EVEN);
+			}
+			elseif($rounding_code == PHP_ROUND_HALF_ODD)
+			{
+				$rounded_sale_tax_amount = round($sale_tax_amount, $decimals, PHP_ROUND_HALF_UP);
+			}
+			elseif($rounding_code == 5)
+			{
+				$fig = (int) str_pad('1', $decimals, '0');
```
jekkos: code duplicaiton: 5 and 6 do exactly the same, so maybe bundle that case together

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +112,183 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+		$total = bcmul($quantity, $price);
+		$discount_fraction = bcdiv($discount, 100);
+		$discount=bcmul($total,$discount_fraction);
+
+		return round($discount, totals_decimals(), PHP_ROUND_HALF_UP);
+	}
+	public function get_item_tax($quantity, $price, $discount_percentage, $tax_percentage)
+	{
+		$price = $this->get_item_total($quantity, $price, $discount_percentage, TRUE);
+		if($this->CI->config->item('tax_included'))
```
jekkos: best to use Appconfig->get here as well

application/migrations/20170502221506_sales_tax_data.php
SteveIreland: not a good one

application/migrations/20170502221506_sales_tax_data.php
SteveIreland: ok

application/migrations/20170502221506_sales_tax_data.php
SteveIreland: I think you're mistaken on this point.

SteveIreland:
Pull request has been updated as requested (mostly).

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +119,179 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+	public function get_sales_tax_for_amount($tax_basis, $tax_percentage, $rounding_mode, $decimals)
+	{
+		$tax_fraction = bcdiv($tax_percentage, 100);
+		$tax_amount = bcmul($tax_basis, $tax_fraction);
+		return $this->round_number($rounding_mode, $tax_amount, $decimals);
+	}
+
+	public function round_number($rounding_mode, $amount, $decimals)
+	{
+		if($rounding_mode == Migration_Sales_Tax_Data::ROUND_UP)
```
jekkos: on a side note: wasn't there any rounding code in a helper somewhere doing a similar thing? No chance of reusing that?

application/migrations/20170502221506_sales_tax_data.php
```diff
@@ -125,13 +119,179 @@ private function update_sales_items_taxes_amount($sale_id, $line, $name, $percen
...
+			array_multisort($sort['print_sequence'], SORT_ASC, $sales_taxes);
+		}
+		$decimals = totals_decimals();
+		foreach($sales_taxes as $row_number => $sales_tax)
+		{
+			$sale_tax_amount = $sales_tax['sale_tax_amount'];
+			$rounding_code = $sales_tax['rounding_code'];
+			$rounded_sale_tax_amount = $sale_tax_amount;
+
+			if ($rounding_code == PHP_ROUND_HALF_UP
```
jekkos: same for this part, think I remember seeing some similar rounding logic elsewhere. Would be worth reusing that as I don't see it changing due to new features.

SteveIreland: There was code that I extracted from the specific release and inserted into the migration script ... as requested.

I believe the principle that I was asked to follow was to NOT use existing code so that we do not need to worry about the impact on migration scripts.    That makes perfect sense to me and I didn't want to cherry pick what I might be able to get away with (because if I had done so then I would be dinged for doing that), although I could also argue your point.

daN4cat: Indeed. The logic needs to be valid at the time of the upgrade hence isolated.
Therefore this is with me. It's not write once, done and dusted.

jekkos: Ok sorry for the confusion, I was a bit in doubt on whether we can still reuse rounding logic onlyt (as you could argue that it's something that won't change over time, there's only a couple of ways in which you round a number and that is guaranteed to stay the same). But also fine if you want to leave it in here.

Issue #2301, opened 2019-01-06
Emailing Quotes/ Invoices


phantomgannon:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0 - 4f5ad57
- OSPOS git commit hash is: 
- PHP version is:  7.2.12
- MySQL or MariaDB version is: 10.1.37-MariaDB - mariadb.org binary distribution
- OS and version is: Windows 7
- WebServer is: 
- Selected language is: English GB
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: Xampp

### Issue / Bug  / Question / New Feature

[...]

PR #2426, opened 2019-04-21
Fix Send Invoice PDF

SteveIreland:
This addresses the issue reported in #2301

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -518,7 +518,7 @@ class Sales extends Secure_Controller
 		$employee_info = $this->Employee->get_info($employee_id);
 		$data['employee'] = $employee_info->first_name . ' ' . mb_substr($employee_info->last_name, 0, 1);
 
-		$data['company_info'] = implode("\n", array(
+		$data['company_info'] = implode("<br/>", array(
 			$this->config->item('address'),
 			$this->config->item('phone')
 		));
@@ -896,7 +896,7 @@ class Sales extends Secure_Controller
 				$data['customer_total'] = empty($cust_stats) ? 0 : $cust_stats->total;
 			}
 
-			$data['customer_info'] = implode("\n", array(
+			$data['customer_info'] = implode("<br/>", array(
 				$data['customer'],
 				$data['customer_address'],
 				$data['customer_location']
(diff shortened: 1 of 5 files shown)


odiea:
I updated the current master with the changes that you made but the Screen version of the Invoice looks as follows. The lower table looks ok.
2019-04-21 11_53_37-Window

SteveIreland:
That's not what I'm seeing in my tests so I will pull the branch down, and test it in other browsers.  I didn't make any change to the css used by the invoice.php view, just the email version.  So I don't know why you would be picking up css changes for the standard invoice view.

It also just now occurred to me that I don't have an email version for the tax invoice (although no one has fussed about that yet).

odiea:
In invoice.php I changed the text area rows to 1 instead of 5 and it got the display I was looking four. Happy Easter

SteveIreland:
Okay.  In the original pull request I replaced the text area in the invoice.php view with a DIV.  The text area control really doesn't play well with embedded `<BR/>` tags.

I downloaded the branch from bintray and tested it in both Chrome and Firefox and it behaved well in both browsers.  So I believe the original pull request is still good.

Edge doesn't work with my virtual hosts and since I hate Edge for any purpose other than displaying PDF files I'm not going to waste any more time trying to figure out how to configure it to work with it.  I saw a microsoft tech support thread on the topic that give me a very good reason not to waste any time with it.

The request was about the send invoice, but I now see that the send quote pdf and the send work order pdf have similar issues so I'm going to adapt those to behave the same and add them to this pull request.

If someone asks for a send tax invoice PDF I'll do that in a later change.

odiea:
Ok The bintray download worked just fine. I sure won't submit issues from my master anymore.

SteveIreland:
No problem @odiea   I've done the same.

The fixes for quotes and work orders won't be ready until next weekend.  Quotes and Work Orders are step-children to invoices and when I started adapting them I realized that they are lagging behind a bit on updates so I'd like to review and make some corrections there.  Plus when I touch them some of the existing code formatting may not make it past @daN4cat

daN4cat:
I trust we are fine with this, confirmed @odiea ?

PR #2425, opened 2019-04-18
fix typo with lang string w/ variable

josh-tf:
Hi guys,

In the `send_pdf()` function in the Sales controller there is a filename string:

`$filename = sys_get_temp_dir() . '/' . $this->lang->line("sales_$type") . '-' . str_replace('/', '-', $number) . '.pdf';`

However `$this->lang->line("sales_$type")` is not valid and should be `$this->lang->line("sales_" . $type)`

Quick PR to change it

Thanks
Josh

PR #2429, opened 2019-04-22
fix typo with lang string

josh-tf:
**Related**: https://github.com/opensourcepos/opensourcepos/pull/2425

Hi Folks, related to above PR, same issue - in the initial PR I didn't notice the subject also had the same issue with lang string.

Summary of fix:
Occurrence of `$this->lang->line("sales_$type")` is not valid and has been changed to `$this->lang->line("sales_" . $type)`

Did a quick search and couldn't find any others

Cheers,
Josh

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -464,7 +464,7 @@ class Sales extends Secure_Controller
 		$quantity = parse_decimals($this->input->post('quantity'));
 		$discount = parse_decimals($this->input->post('discount'));
 		$discount_type = $this->input->post('discount_type');
-		
+
 		$item_location = $this->input->post('location');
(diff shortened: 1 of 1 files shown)


application/controllers/Sales.php
```diff
@@ -793,7 +793,7 @@ public function send_pdf($sale_id, $type = 'invoice')
 		{
 			$to = $sale_data['customer_email'];
 			$number = $sale_data[$type."_number"];
-			$subject = $this->lang->line("sales_$type") . ' ' . $number;
+			$subject = $this->lang->line("sales" . $type) . ' ' . $number;
```
daN4cat: Shouldn't it have an _ ?

josh-tf: ah you're 100% right, oversight on my part.. sorry

`$lang["sales_invoice"] = "Invoice";`
`$lang["sales_receipt"] = "Sales Receipt";`

PR #2431, opened 2019-04-23
Attributes csv import integration

objecttothis:
In addition to modifying the csv when attributes are modified this also includes functionality that modifies the csv when stock locations are modified.


objecttothis:
Let's not merge quite yet.  I forgot a few regression tests.  Let me complete those, but feel free to request changes of what you see so far.

objecttothis:
I found a small bug where the excel import does not work properly if there is an attribute with the same name as a base item.  For example our bookstore has an attribute called category which is the name we have been using since the opening of the bookstore (long before OSPOS) to describe the category of a book (non-fiction, Young Adults, Travel Guides, etc).  I will likely need to have a prefix for attributes similar to what I am doing with stock locations (attribute_ or location_) which would then allow attributes with the same name as a base item attribute without OSPOS getting confused during the import.

objecttothis:
OK, I fixed the bug not allowing attributes of the same name as a base item field.  In the csv attributes are now prefixed as "attribute_" and as before the stock locations are prefixed as "location_".  I still have some regression testing to do with the actual import part, but the file manipulation parts for both stock locations and attributes pass all my tests.

application/controllers/Items.php
```diff
@@ -970,23 +887,126 @@ public function do_excel_import()
...
-	 * Guess whether file extension is not in the table field,
-	 * if it isn't, then it's an old-format (formerly pic_id) field,
-	 * so we guess the right filename and update the table
+	 * @param line
+	 * @param failCodes
+	 * @param attribute_data
+	 */
+	private function save_attribute_data($line)
+	{
+		$definition_names = $this->Attribute->get_definition_names();
```
jekkos: Ok great job moving this to separate methods. Maybe just move them to the helper as they are specific for csv reading?

application/controllers/Items.php
```diff
@@ -970,23 +887,126 @@ public function do_excel_import()
...
+		
+		foreach($definition_names as $definition_name)
+		{
+			if(!empty('attribute_' . $line[$definition_name]))
```
jekkos: Isn't this going to be always true?

application/controllers/Items.php
```diff
@@ -970,23 +887,126 @@ public function do_excel_import()
...
+	 * @param	array	line
+	 * @param			item_data
+	 */
+	private function save_inventory_quantities($line, $item_data)
+	{
+		//Quantities & Inventory Section
+		$employee_id		= $this->Employee->get_logged_in_employee_info()->person_id;
+		$emp_info			= $this->Employee->get_info($employee_id);
+		$comment			= 'Quantity Imported from CSV';
+		//TODO: Right now location_id is coming through as NULL and we need to figure out why.
```
jekkos: Might be perfect to try and add a testcase for this import functionality (if you feel up to that?). It will probably save time now and certainly later on. You could then create some sample CSV and run it through the import, then check if everything ends up where it should be.

application/migrations/20181015100000_attributes.php
```diff
@@ -6,15 +6,47 @@ public function __construct()
...
+			'custom4_name',
+			'custom5_name',
+			'custom6_name',
+			'custom7_name',
+			'custom8_name',
+			'custom9_name',
+			'custom10_name'
+		));
+		
+		$custom_fields = array();
```
jekkos: Nice addition, so we keep the former import functionality and migrate it to this generic template system. I guess that order does not matter then?

application/models/Attribute.php
```diff
@@ -370,6 +378,12 @@ public function save_definition(&$definition_data, $definition_id = -1)
 				}
 			}
 			
+			if($from_definition_name !== $definition_data['definition_name'])
```
jekkos: In one way or another I think it would be more 'flexible' to just generate the template on demand, so we don't need to reflect every operation on a database attribute in the import template.

jekkos: But that might be considered as a new user 'action' on it's own (eg a button in the import dialog that generates the template by looping over attributes and stock locations, then downloads it for the user).

application/controllers/Items.php
objecttothis: So, this and other methods are taking the contents of the CSV row that have already been read and calling the function in the model to save it to the database.  The function specific to CSV reading is called in line 826 of the Items controller and that function is in the helper.

application/controllers/Items.php
objecttothis: oops, good find.  Yeah, that is supposed to read !empty($line['attribute_' . $definition_name]).   I'll fix it.

application/controllers/Items.php
objecttothis: Yes, I have been doing that with sample CSV files.  I have some more testing to do to confirm everything is in order.

application/migrations/20181015100000_attributes.php
objecttothis: Since this is in the attribute migration php file we will need to release this pull request with 3.3.0 otherwise people will not receive this change.  Yes, the idea with this php in particular is that at the same time that custom fields are getting converted to attributes in the database, they are also added to the import_csv file header.  Yes, the way that I rewrote the do_excel_import is such that order should not matter.  There are two nuances though.  One is that when attributes are added, they get appended onto the end of the headers.  When stock locations get added they get inserted after the base item column headers but before attributes.  The reader doesn't care though because it just creates an associative array.

application/controllers/Items.php
jekkos: Ok did you write any code for it or just manual testing?

application/models/Attribute.php
objecttothis: Do you mean generate the template when they click the "download excel import template" link?  Yes, we could do it that way, but can you help me see the advantage of doing it that way is?  I'm not quite sure I understand what you mean by more flexibility.  

One advantage that I can see to making changes as changes are made to the attributes and stock locations is that it reduces the number of times the functions to read/write the file are called.  If we go with what you are suggesting, those functions get called each time someone imports.  Depending on the business, that's maybe several times per month or more compared to one time at attribute/stock location change which is maybe once per year (more for businesses just beginning to use the software.

I think I see now a few major advantages to what you are proposing.
- no more need for delete_csv_column, get_csv_file, rename_csv_column and add_csv_column gets converted into generate_csv_item_import.
- It removes the need for an attribute conversion legwork since we would generate on link click.
- It removes the need for the stock location migration since that too is generated on the fly.

Its a bummer that I did all the work for it the way it is already.  It's not that this is a waste, but it is working the way it is.

jekkos: I didn't think about that upfront either but when seeing the code it feels more 'natural' to bundle these operations in one 'generate' method and get rid of the template file itself in our git repository. 

I was just thinking that this generate would then just be an assembly of what you already did (I guess you only need the add methods in that case then). But maybe if you don't feel like going the extra mile and the current situation works then fine for me as well.

objecttothis:
@jekkos @daN4cat I am now back in the office (we had a baby girl) and will be finishing this up in the next few days.  As per earlier conversations with @jekkos I am going to rework the code to act as a CSV file generator that gets kicked off when someone clicks the "Download Import Excel Template (CSV)" file.  I also need to finish testing to make sure that imports are working properly.  Once merged, I believe everything I was working on will be ready for the 3.3.0 release.  I believe I should be ready to merge this pull request by next Wednesday.

SteveIreland:
Congratulations!

objecttothis:
OK, the conversion to generate the csv file on the fly is done and tested.  Now I just need to finish testing on the import side.  There are several use cases that I need to debug.

application/models/Attribute.php
objecttothis: @jekkos, the code now generates the CSV on the fly.

application/controllers/Items.php
objecttothis: @daN4cat @jekkos I have not written any code for it.   Just manual testing.  The current status is 

- Use cases with correct data all import successfully.
- Use cases with attribute types other than TEXT but where incorrect data is fed still import the data as is.  For example, Attribute-->save_value() does not check to make sure that DECIMAL, DATETIME and DROPDOWN type data includes decimal values, datetime acceptable values or values already in the dropdown.  I think we need to have the code check and return FALSE if the data doesn't match up but wanted to wait to see what you guys thought first.  The big question is where to do error checking and how to report it (not sure if FALSE would be the right return value)

As far as I can tell, the second item there is all that's getting in the way of merging this pull request and that will be the last item I am responsible for to release 3.3.0

application/models/Attribute.php
jekkos: O very cool, thanks for that!

application/models/Attribute.php
```diff
@@ -566,15 +585,33 @@ public function delete_value($attribute_value, $definition_id)
...
+		$this->db->where('definition_id', $definition_id);
 		return $this->db->update('attribute_definitions', array('deleted' => 1));
 	}
 	
 	public function delete_definition_list($definition_ids)
 	{
+		foreach($definition_ids as $definition_id)
+		{
+			//Get the name of the column to remove
+			$this->db->select('definition_name');
```
jekkos: is the result of this select used somewhere? At first sight it looks like it's just running the update as a first thing here.

application/models/Attribute.php
```diff
@@ -566,15 +585,33 @@ public function delete_value($attribute_value, $definition_id)
...
+	 * Deletes an Attribute definition from the database and associated column in the items_import.csv
+	 *
+	 * @param	unknown	$definition_id	Attribute definition ID to remove.
+	 * @return 	boolean					TRUE if successful and FALSE if there is a failure
+	 */
 	public function delete_definition($definition_id)
 	{
-		$this->db->where('definition_id', $definition_id);
+		//Get the name of the column to remove
+		$this->db->select('definition_name');
```
jekkos: same thing here, do we need to get the name to delete by id?

application/helpers/importfile_helper.php
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Generates the header content for the import_items.csv file
+ *
+ * @return	string						Comma separated headers for the CSV file
+ */
+function generate_import_items_csv($stock_locations,$attributes)
+{
+	$csv_headers = pack("CCC",0xef,0xbb,0xbf);	//Encode the Byte-Order Mark (BOM) so that UTF-8 File headers display properly in Microsoft Excel
+	$csv_headers .= 'Barcode,"Item Name",Category,"Supplier ID","Cost Price","Unit Price","Tax 1 Name","Tax 1 Percent","Tax 2 Name","Tax 2 Percent","Reorder Level",Description,"Allow Alt Description","Item has Serial Number",item_image,HSN';
```
jekkos: don't know if we should translate the header names? At least it wasn't like this before, but now we can ofc

application/controllers/Items.php
```diff
@@ -970,23 +886,125 @@ public function do_excel_import()
...
+	 *
+	 * @param	array	line
+	 * @param			item_data
+	 */
+	private function save_inventory_quantities($line, $item_data)
+	{
+		//Quantities & Inventory Section
+		$employee_id		= $this->Employee->get_logged_in_employee_info()->person_id;
+		$emp_info			= $this->Employee->get_info($employee_id);
+		$comment			= 'Quantity Imported from CSV';
```
jekkos: No translation here?

application/controllers/Items.php
```diff
@@ -792,173 +792,89 @@ public function delete()
...
-						//tax 1
-						if(is_numeric($data[7]) && $data[6] != '')
-						{
-							$items_taxes_data[] = array('name' => $data[6], 'percent' => $data[7] );
-						}
-
-						//tax 2
-						if(is_numeric($data[9]) && $data[8] != '')
-						{
-							$items_taxes_data[] = array('name' => $data[8], 'percent' => $data[9] );
```
jekkos: +1 for splitting this up over multiple methods

application/controllers/Items.php
```diff
@@ -792,173 +792,89 @@ public function delete()
...
-
+	
+	/**
+	 * Imports items from CSV formatted file.
+	 */
 	public function do_excel_import()
 	{
-		$non_repeating_element_count = 18;
-
+		//		$non_repeating_element_count = 18; Is this still needed?
```
jekkos: I guess you can get rid of this

application/controllers/Items.php
```diff
@@ -792,173 +792,89 @@ public function delete()
...
-	Items import from excel spreadsheet
-	*/
+	 Items import from excel spreadsheet
+	 */
 	public function excel()
 	{
 		$name = 'import_items.csv';
-		$data = file_get_contents('../' . $name);
-		force_download($name, $data);
+		$data = generate_import_items_csv($this->Stock_location->get_allowed_locations(),$this->Attribute->get_definition_names(FALSE));
```
jekkos: maybe create two variables for the headers (this line a bit long now)

application/controllers/Items.php
```diff
@@ -970,23 +886,125 @@ public function do_excel_import()
...
+				);
+				
+				$this->Inventory->insert($excel_data);
+			}
+			else
+			{
+				$item_quantity_data = array(
+					'item_id' => $item_data['item_id'],
+					'location_id' => $location_id,
+					'quantity' => 0
```
jekkos: Looks like a bit of code duplication: if you extract a variable for `$quantity` and  `$location_id` you can reduce the code in the  if / else to assigning these variables and use them in the `save` and `insert` statements once afterwards?

application/helpers/importfile_helper.php
objecttothis: You're right.  These were changes introduced before the CSV file was being generated on the fly.  I've committed the fixes to the branch.

application/models/Attribute.php
objecttothis: Resolved as per below.

application/controllers/Items.php
objecttothis: That was just code that was in there from the previous version of the function and correct, it did not have translation either.  If you want, we can make this a translated line.

application/controllers/Items.php
objecttothis: Fixed.

application/controllers/Items.php
objecttothis: Fixed.

application/controllers/Items.php
objecttothis: Done.

application/models/Attribute.php
objecttothis: Removed

application/controllers/Items.php
objecttothis: @jekkos, @daN4cat can you respond to my comment above about error checking for imported attributes in CSV files and how/where we should approach that?

application/controllers/Items.php
objecttothis: Never mind, I just made it translatable. :)

application/controllers/Items.php
objecttothis: It seems like the most logical place to check the DECIMAL and DATETIME is in the Attributes->save_value() method.  Checking DROPDOWN is a little more involved and would include pulling acceptable DROPDOWN values and comparing.

The second "problem" is when I do error checking, its almost as if I need to check everything before even calling these methods.  For example, the do_excel_import() function is calling
`						$this->save_tax_data($line, $item_data);`
`						$this->save_inventory_quantities($line, $item_data);`
`						$this->save_attribute_data($line, $item_data);`
But in the case where there is a problem with one of the attributes, the code has already saved tax and inventory data.  I'm not sure how to do all three as a transaction so that it rolls back inventory and tax changes if there is an attribute problem.

daN4cat: Yep, it doesn't sound good to store data, bump into an error and have no roll back plan.
We should indeed sanity check data before any real import and do atomic transactions in bulk operations to avoid broken commits.

I agree it's tricky but I don't see how else we could do.

Opinions?

@SteveIreland @jekkos?

objecttothis: @daN4cat ok, then I'll write a method to error check the entire line before submitting to the methods that are doing the import.  I don't see how we could turn it into an atomic transaction that reverts everything if an error is encountered since these are currently being done in three separate methods.

jekkos: MIght be better to make the whole code transactional then and rollback in case of issues, so a user can correct the error and retry. This might be the best way otherwise he might have duplicate rows or we need to keep track of where we left (like in the session or sth).
So indeed doing some minimal checks on the data and datatype first, and then fail the whole import by throwing an exception and stopping the transaction feels like the best thing to do.

jekkos: More info in [CI docs](https://www.codeigniter.com/user_guide/database/transactions.html)

objecttothis: @jekkos those docs are for making the individual queries transactional, but there doesn't seem to be a built-in way to make the entire file's worth of transactions roll back in case of one failure.

objecttothis: @jekkos agreed that care must be taken to prevent duplicates and one way to prevent that is to integrity-check the whole file before processing then fail the whole file and not just the lines without problems.

Even if we do that we need to report clearly which lines errors appeared on, otherwise users are potentially sifting through a large haystack to find a small needle.

Alternately we could successfully import correctly formatted lines and report which lines fail then leave it to the user to only resubmit corrected lines.

Making the whole file transactional is the safer route of the two.  @daN4cat what do you say?

daN4cat: I believe the current approach is to report the failed rows.
Therefore go row by row in a transactional way and a failure is logged / reported.

jekkos: I agree that the current approach is different but if we can enforce this behavior by adding transaction boundaries, then that might be better.

@objecttothis not sure if that is what is is, to me a database transaction is an all or nothing, meaning that if in between transaction boundaries something fails, it reverts everything that was done before. I wouldn't really call row by row transactional as the transactional is the actual data import of all rows itself.

In java eg you'd get this behavior out of the box. So if you start a tx, and you throw an exception, everything is rolled back, you correct the problem and retry.

daN4cat: That is complicating the whole matter especially if you have multiple point of failures in an import.
People will be forced to try & error many times.
If you import what you can and report the failing lines you have reduced the number of runs.
Personally I would prefer that approach unless there is a reason to keep the list ordered?

objecttothis: OK, so far, there are two things I'm hearing everyone agree on

- data entry into the database must be transactional at minimum on the row level so data doesn't partially get entered into the database then fail.
- Regardless of how we approach it (row-level transactional vs. file-level transactional) reporting of failing rows must be done.

@daN4cat I see no reason for anyone caring about the row order of the import file since each line represents an independent item entry.

Since we are trying to get 3.3.0 out, and its much less work at this point to make the code row-level transactional, that's what I will do and if we then want to adjust the code to make it file-level transactional down the road we can do that later.

daN4cat: Agreed, go for that. It's on par with the current functionality and a pragmatic approach.

jekkos: Hmm I don't see why you wouldn't make it file level? I think that should be just a matter of adding the transaction start and end statements before/after the import starts. You can still do the reporting on row level in that case.

jekkos: @daN4cat I think things will get more complicated for a user once eg 100 different rows in a 1000 row file fail to import. He will need to corect, filter out those rows and retry the import. If you have file level then you just correct and retry?

objecttothis: @jekkos I don't feel strongly enough about it either way.  What I do feel strongly about is that error reporting happens at the line level so that it is very clear where the errors are happening.

As for making it file level, I assumed that adding transaction start/end statements wouldn't work because I'm calling several private functions and wasn't sure that if the transaction statements would roll back multiple querybuilder queries (potentially thousands) if an error was found in one of them.

If the transaction statements are, in fact, that robust, then I still need to check the file in its entirety.  Alternately instead of transaction start/end statements I could just use my error checking function to report errors and fail before any of the queries run.  I think I'd prefer the transaction statements if they work that robustly though.

jekkos: I think that's the idea behind having database transactions, you start it at some point and the database tracks new state from that point on. If something goes wrong it is rolled back. It should work like this (according to the docs) but I must admit I never tested it thoroughly.

jekkos: You might just try it out quickly by adding a throw exception in the middle of the import, and check if there's data in the bb or not. And indeed error reporting will be on row level anyway so that doesn't change at all.

daN4cat: Unless we produce a copy of the file with just the error rows?
That would help more probably...

objecttothis: Ultimately with the error reporting, I am eventually planning to write detailed error reporting into the logs.  For example: "Line 34 not imported: 'red' is not an available dropdown in attribute 'color'"

@daN4cat I think you are saying that doing the import of the good rows, but then kicking back a CSV with all the error rows would allow the user to fix the problems, re-import and no risk of duplicates.  

I really like this and I think might alleviate @jekkos concerns of duplicates.  However, we are really behind on 3.3.0 so i think we should get something out that works well now, but doesn't take a lot of development time.  I don't think it will be difficult down the road for someone to modify the behavior to include that.

daN4cat: Fine. It was an idea... for future generations.

.gitignore
```diff
@@ -9,11 +9,13 @@ application/uploads/*
 public/license/.licenses
 public/license/bower.LICENSES
 public/dist
+generate_langauges.php
```
daN4cat: typo: should languages

daN4cat: BTW, hmmmm why this file?

.gitignore
```diff
@@ -9,11 +9,13 @@ application/uploads/*
...
 dist/
 docs/
 public/bower_components
 *.patch
 patches/
+translations/
```
daN4cat: Not sure I understand this....

application/controllers/Items.php
objecttothis: OK, then I will look into making the import transactional at the file level for now and we can make adjustments in future releases.

objecttothis: @jekkos @daN4cat this is completed.  I've done my regression testing and it should be OK, but it would be good for a few people to test import files to confirm they are getting what they expect.  @jekkos you mentioned writing regression testing code for this.  I admit that I am not well versed in this.  I have just been throwing CSV files at it with various errors and completed rows to confirm it is working.  If you'd like to help me with writing the code for regression testing (tutorials or chat) then I would be grateful.  Otherwise, I think throwing CSV files at it to make sure it's importing things properly when everything is formatted correctly and erroring out properly with no rows being imported when there are errors is as good as I can get.

.gitignore
objecttothis: yeah, these were not my changes.

.gitignore
objecttothis: I don't either.  I didn't add this into .gitignore.

.gitignore
objecttothis: I will rebase and force a commit.  Rebasing in my setup (eclipse) has always been confusing to me, but I guess I need to figure it out sometime.

application/controllers/Items.php
objecttothis: Rebase and Force Push is done.  I'm not sure what I'm doing wrong, but I don't know how to get rid of commits that I didn't make.  e.g. Arabic transations, etc.

daN4cat: Let me check this evening and see if I can sort it out for you. Please make sure you don't have anything more to commit before I look into this tonight.

objecttothis: No commits are outstanding unless someone finds a bug that I missed.

daN4cat:
I squashed all your change and did a forced commit to make it a neat submission.

objecttothis:
> I squashed all your changes and did a forced commit to make it a neat submission.

@daN4cat and @jekkos after deleting the two files I walked again through all of my regression testing and I believe that the import and csv on-the-fly generation are working properly.  Here are a few quick questions.  They can get punted to 3.4.0 or fixed (if needed) here.

- Can someone define for me what Receiving Quantity represents?  I thought it was the combined total quantity of all stock locations, but both my code and the previous method of importing via CSV fills it in as 1.0.
- What does the Allow Alt Description flag do?  There is the checkbox for it, but that doesn't enable an alternate description textbox as far as I can tell.
- Does the Item has Serial Number flag have a function apart from altering the user that this item has a serial number?  I don't think there is functionality in the code to attach a serial number to a sale... or is there?

OK, so other than those questions and barring someone coming across a bug I didn't catch, this Pull Request is ready to merge into the Master.

daN4cat:
Ok, I'll squash again the commit and merge it tonight.
I want to tidy up PRs and get to an end of 3.3.0 by end of the month.

objecttothis:
> Ok, I'll squash again the commit and merge it tonight.

Do all the PR's need to be squashed into a single commit?  I'm assuming if so that this is to keep the commit history clean.  I need to learn how to do this with my current setup (Eclipse/Github) or through the CLI if it's not doable in Eclipse/Github to keep other people from having to do it.

> I want to tidy up PRs and get to an end of 3.3.0 by end of the month.

Hopefully we could even release before that, but then again we've had a few Pull Requests pop up earmarked for 3.3.0.  I think that barring major need, we should freeze any further PR's from going into 3.3.0.

jekkos:
Make sure you keep the original author intact on the commits, it might get lost after the rebase.

jekkos:
Receiving quantity, should be on the wiki somewhere. Funny we have all these features that even our own contributors don't know about.

objecttothis:
> Receiving quantity, should be on the wiki somewhere. Funny we have all these features that even our own contributors don't know about.

Yeah it's mentioned in https://github.com/opensourcepos/opensourcepos/wiki/Items but all it says is that it cannot be set to 0 (zero).  There is no description as to what it represents.  Regarding this Pull request it's only relevant in that the code sets the receiving quantity to 1 on import from CSV, but I don't know if that is the correct amount since I don't know what it is.

daN4cat:
I emerge without further squash, it's not really required.

PR #2434, opened 2019-04-27
Support updates to payments

SteveIreland:
Okay, this is a rather trivial change that allows payments to be adjusted "post sale".

You simply go to Daily Sales to locate the invoice containing the payments to be adjusted and click on the "Update" icon.  You will be presented with the the Update form which now has a place to enter a payment adjustment.

Of course, there are a couple of problems I have with this entire process.  Payments are not transaction oriented by which I mean that if I add a payment transaction there isn't a record of the transaction written.  Instead the transaction is added to the collection of sales payments.  Since the key to the sales_payments table is the payment type there can only be one entry per payment type - so additional payments for a sale that all have the same payment type are accumulated into one record that reflects that payment type.

That's the negative side of this change.  The positive side is that you can now update the balance due for each sale if payments are received after the sale.

There are numerous improvements that can be made, but I wanted to offer this as both something that will allow some sites to temporarily "get by" and as a prototype for discussions about what future improvements might be necessary so that the job of payment tracking is done right.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -4,6 +4,7 @@ require_once("Secure_Controller.php");
 
 define('PRICE_MODE_STANDARD', 0);
 define('PRICE_MODE_KIT', 1);
+define('PAYMENT_TYPE_UNASSIGNED', '--');
 
 class Sales extends Secure_Controller
 {
@@ -1166,6 +1167,11 @@ class Sales extends Secure_Controller
 		$data['selected_customer_name'] = $sale_info['customer_name'];
 		$data['selected_customer_id'] = $sale_info['customer_id'];
 		$data['sale_info'] = $sale_info;
+		$balance_due = $sale_info['amount_due'] - $sale_info['amount_tendered'];
+		if($balance_due < 0)
+		{
+			$balance_due = 0;
+		}
 
 		$data['payments'] = array();
 		foreach($this->Sale->get_sale_payments($sale_id)->result() as $payment)
@@ -1177,8 +1183,19 @@ class Sales extends Secure_Controller
 			$data['payments'][] = $payment;
 		}
 
+		$data['payment_type_new'] = PAYMENT_TYPE_UNASSIGNED;
+		$data['payment_amount_new'] = $balance_due;
+
+		$data['balance_due'] = $balance_due != 0;
+
 		// don't allow gift card to be a payment option in a sale transaction edit because it's a complex change
 		$data['payment_options'] = $this->xss_clean($this->Sale->get_payment_options(FALSE));
+
+		// Set up a slightly modified list of payment types for new payment entry
+		$new_payment_options = $this->Sale->get_payment_options(FALSE);
+		$new_payment_options["--"] = $this->lang->line('common_none_selected_text');
+		$data['new_payment_options'] = $this->xss_clean($new_payment_options);
+
 		$this->load->view('sales/form', $data);
 	}
 
@@ -1240,6 +1257,7 @@ class Sales extends Secure_Controller
 	public function save($sale_id = -1)
 	{
 		$newdate = $this->input->post('date');
+		$employee_id = $this->Employee->get_logged_in_employee_info()->person_id;
 
 		$date_formatter = date_create_from_format($this->config->item('dateformat') . ' ' . $this->config->item('timeformat'), $newdate);
 
@@ -1251,39 +1269,28 @@ class Sales extends Secure_Controller
 			'invoice_number' => $this->input->post('invoice_number') != '' ? $this->input->post('invoice_number') : NULL
 		);
 
-		// go through all the payment type input from the form, make sure the form matches the name and iterator number
+		// In order to maintain tradition the only element that can change on prior payments is the payment type
 		$payments = array();
 		$number_of_payments = $this->input->post('number_of_payments');
 		for($i = 0; $i < $number_of_payments; ++$i)
 		{
+			$payment_id = $this->input->post('payment_id_' . $i);
 			$payment_amount = $this->input->post('payment_amount_' . $i);
 			$payment_type = $this->input->post('payment_type_' . $i);
-			// remove any 0 payment if by mistake any was introduced at sale time
-			if($payment_amount != 0)
-			{
-				// search for any payment of the same type that was already added, if that's the case add up the new payment amount
-				$key = FALSE;
-				if(!empty($payments))
-				{
-					// search in the multi array the key of the entry containing the current payment_type
-					// NOTE: in PHP5.5 the array_map could be replaced by an array_column
-					$key = array_search($payment_type, array_map(function ($v)
-					{
-						return $v['payment_type'];
-					}, $payments));
-				}
 
-				// if no previous payment is found add a new one
-				if($key === FALSE)
-				{
-					$payments[] = array('payment_type' => $payment_type, 'payment_amount' => $payment_amount);
-				}
-				else
-				{
-					// add up the new payment amount to an existing payment type
-					$payments[$key]['payment_amount'] += $payment_amount;
-				}
-			}
+			// To maintain tradition we will also delete any payments with 0 amount assuming these are mistakes
+			// introduced at sale time.  This is now done in Sale.php
+
+			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'payment_user' => $employee_id);
+		}
+
+		$payment_id = -1;
+		$payment_amount = $this->input->post('payment_amount_new');
+		$payment_type = $this->input->post('payment_type_new');
+
+		if($payment_type != PAYMENT_TYPE_UNASSIGNED && $payment_amount <> 0)
+		{
+			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'payment_user' => $employee_id);
 		}
 
 		if($this->Sale->update($sale_id, $sale_data, $payments))
@@ -1480,7 +1487,7 @@ class Sales extends Secure_Controller
 	{
 		foreach($array as $key => $val)
 		{
-			if ($val['item_id'] === $id)
+			if($val['item_id'] === $id)
 			{
 				return $key;
 			}
--- a/application/migrations/sqlscripts/3.3.0_paymenttracking.sql
+++ b/application/migrations/sqlscripts/3.3.0_paymenttracking.sql
@@ -1,0 +1,26 @@
+-- Improve payment tracking
+
+RENAME TABLE ospos_sales_payments TO ospos_sales_payments_backup;
+
+CREATE TABLE `ospos_sales_payments` (
+  `payment_id` int(11) NOT NULL AUTO_INCREMENT,
+  `sale_id` int(10) NOT NULL,
+  `payment_type` varchar(40) NOT NULL,
+  `payment_amount` decimal(15,2) NOT NULL,
+  `payment_user` int(11) NOT NULL DEFAULT 0,
+  `payment_date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `reference_code` varchar(40) NOT NULL DEFAULT '',
+  PRIMARY KEY (`payment_id`),
+  KEY `payment_sale` (`sale_id`, `payment_type`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO ospos_sales_payments (sale_id, payment_type, payment_amount, payment_user)
+SELECT payments.sale_id, payments.payment_type, payments.payment_amount, sales.employee_id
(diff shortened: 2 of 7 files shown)


SteveIreland:
The improvement I'd like to see made in the next release is to have the primary key to sales_payments be transaction_id (similar to the inventory table) and then have sale_id as a foreign key as well as additional columns for an "when entered" timestamp and a reference number for those who want to track check number, credit transaction number, etc.

WebShells:
@SteveIreland i submitted a pull request long time ago related to payments updates from the form, it was including an empty input box and update button once clicked it deduct the debt amount and add to cash amount, if you want to have a look... Definitely timestamp is needed for tracking, but i cant tell how to link it...

odiea:
Works good. The only suggestion I would have would be to show the current balance due on the form so the payment amount would be adjusted by that.

SteveIreland:
@odiea. Agreed. I started to add that but pulled back when I realized that the change would no longer be trivial. B cause we don’t track invoice totals I will need to basically run a complex SQL similar to what we do for the sales report to get the invoice total before I can start deducting payments to compute the balance due.

I believe if we move forward with enhancements to this effort I will add a ‘sales_stats’ to the system which will capture some of the summary data for a sale so that changes like this are much more efficient and easier

odiea:
@SteveIreland . I just added the following to the form.
```
			<?php echo form_label($this->lang->line('sales_total'), 'sales_total', array('class'=>'control-label col-xs-3')); ?>
			<div class='col-xs-4'>
				<div class="input-group">
					<span class="input-group-addon input-sm"><b><?php echo $this->config->item('currency_symbol'); ?></span>
					<?php echo form_input(array(
							'name'=>'sales_total',
							'id'=>'sales_total',
							'class'=>'form-control input-sm', 'readonly'=>'true',
							'value'=>$sale_info['amount_due'])
							);?>
				</div>
			</div>
		</div>	
```

SteveIreland:
I've expanded this pull request. 

@odiea I went ahead and implemented the sales_stats and modified sales_payments to show where I'm headed with it.

I believe I might need to adapt one or more of the sales reports to the new structure of the sales_payment tables but so far everything seems to be functional.

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
```diff
@@ -0,0 +1,38 @@
...
+CREATE TABLE `ospos_sales_payments` (
+  `payment_id` int(11) NOT NULL AUTO_INCREMENT,
+  `sale_id` int(10) NOT NULL,
+  `payment_type` varchar(40) NOT NULL,
+  `payment_amount` decimal(15,2) NOT NULL,
+  `payment_user` int(11) NOT NULL DEFAULT 0,
+  `payment_date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `reference_code` varchar(40) NOT NULL DEFAULT '',
+  PRIMARY KEY (`payment_id`),
+  KEY `paymnt_sale` (`sale_id`, `payment_type`)
```
jekkos: minor typo here

application/models/Sale.php
```diff
@@ -112,6 +112,35 @@ public function get_info($sale_id)
...
+		if($query->num_rows() == 1)
+		{
+			return $query->row();
+		}
+		else
+		{
+			//Get empty base object,
+			$stats_obj = new stdClass();
+
+			//Get all the fields from supplier table
```
jekkos: comment is not completely correct, maybe just remove it

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
```diff
@@ -0,0 +1,38 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO `ospos_sales_payments` (`sale_id`, `payment_type`, `payment_amount`)
+SELECT `sale_id`, `payment_type`, `payment_amount` FROM `ospos_sales_payments_backup`;
+
+DROP TABLE IF EXISTS `ospos_sales_payments_backup`;
+
+CREATE TABLE `ospos_sales_stats` (
+  `sale_id` int(10) NOT NULL,
```
jekkos: so this table will only be filled by new completed sales? Wouldn't there be a way to make it an sql view of the data instead? In that case it will always be uptodate and we need no code to manage it.

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
```diff
@@ -0,0 +1,38 @@
+-- This is to cleanup any orphaned tax migration tables
+
+RENAME TABLE `ospos_sales_payments` TO `ospos_sales_payments_backup`;
```
jekkos: any specific reason for recreating the table here? If possible maybe just add the new columns on top of the old one?
Also I think you forgot to add the foreign key constraints so integrity checks are disabled here.

application/controllers/Sales.php
```diff
@@ -1251,39 +1293,28 @@ public function save($sale_id = -1)
...
+			// introduced at sale time.  This is now done in Sale.php
+
+			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'payment_user' => $employee_id);
+		}
+
+		$payment_id = -1;
+		$payment_amount = $this->input->post('payment_amount_new');
+		$payment_type = $this->input->post('payment_type_new');
+
+		if($payment_type != '--' && $payment_amount <> 0)
```
jekkos: don't know if we have a list of payment constant types but this '--' would then be the NONE constant.

application/controllers/Sales.php
```diff
@@ -1167,6 +1189,19 @@ public function edit($sale_id)
 		$data['selected_customer_id'] = $sale_info['customer_id'];
 		$data['sale_info'] = $sale_info;
 
+		$stats_info = $this->Sale->get_stats_info($sale_id);
+		$data['payment_type_new'] = '--';
+		$data['payment_amount_new'] = $stats_info->net_due;
+
+		if($data['payment_amount_new'] != 0)
+		{
+			$data['balance_due'] = TRUE;
```
jekkos: you could make this more concise (a oneliner), like `$data['balance_due'] = $data['payment_amount_new'] != 0`

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
jekkos: I guess it was done to fill the columns with default values. I'd just run an update query instead here. Also the default user id doesn't necessarily exist, so it could be a data integrity violation, best to make it nullable for the past.

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: done

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: I thought I might get some push back on this table, however I'd like argue for the existence of this table although I'm certainly okay with dropping my use of the table for tracking payments.  In fact I'll go ahead and make that change.   

When I checked the function for `Sale->get_info()` the SQL overhead to retrieve information that will never change is enormous.  Once a sale is complete each of the amounts on the stats table won't change.  So calling get_info on sales (building two temporary tables and then joining multiple tables with group) seems like an unnecessary drain on resources when it can be replaced by a quick retrieve of two records (one from `sales` and one from `sales_stats`.

The sales_statistics table can also be used for user defined reporting that others, who are intimidated by the use of SQL to recalculate the sales totals,  might want to implement.

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: Since I was changing the primary key my traditional approach is to populate the new version of the table using an insert so that the key can be auto assigned.

Good point on the user id, but I can actually populate it as part of the migration from the `sales` table.  That was something I hadn't yet thought about but will include that in the migration.  I feel it is important for audit purposes to be able to associate a user with each payment.

application/controllers/Sales.php
SteveIreland: done

application/controllers/Sales.php
SteveIreland: done

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: I just pushed my latest changes as described previously.  

I do believe the panel renders faster than before, but I'm biased so I even though it makes sense to me that the panel should render faster I'll let others be the judge.  It has to read the sales record, the customer record, the employee record and the set of payments before it can render so there were a couple of more reads than I first thought.   But still, it's no longer building temporary tables and reading the sales_items table, etc.

More testing is needed but I'm done for the weekend so I'll try to test some during the week.  I still have tested in a included tax scenario or with the destination based tax scenario.

jekkos: Personally I'd like to stick to one plan of attack here. 

* Either we go for a stats table everywhere (and get rid of the temporary tables), and so we try to reuse the temporary tables for this case specifically. 

Reusing the temporary ones makes sense if we move it's creation to a piece of common code that can be called to find any aggregate value related to sales or payments.

Another argument for sticking with one approach is testability and correctness. Using one approach requires validation of only one resultset for further use, while mixing different approaches (temp and stats table) will lead to more complexity and bugs.

jekkos: Also I think we can tackle performance issue with these temp tables by tweaking the time they can be kept in memory eg. (see some of the hints in the other perfomance PR here)

SteveIreland: okay, I'll redo it.

SteveIreland: done

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: Done

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
SteveIreland: Backed out the stats table.

SteveIreland:
I'm not sure if it's the result of changes with this pull request (I can't think of a reason why it would) but there seems to be an issue with the Payment Summary report and I will fix that tomorrow as part of this pull request regardless of what introduced the problem.

SteveIreland:
I found the issue with the payments report and fixed it as part of this pull request.  It wasn't caused by the changes in this pull request.  It turns out that the problem manifested itself in a "tax included" scenario.

Oh, and the issue was that the sales amount wasn't being included in the payments report. 

The pull request has been updated.  And if no one disapproves, since I've addressed all of the issues that @jekkos requested I will go ahead and merge it tomorrow night (around 24 hours from now) so that it has a wider audience for testing.

application/migrations/sqlscripts/3.3.0_paymenttracking.sql
```diff
@@ -0,0 +1,24 @@
...
+CREATE TABLE `ospos_sales_payments` (
+  `payment_id` int(11) NOT NULL AUTO_INCREMENT,
+  `sale_id` int(10) NOT NULL,
+  `payment_type` varchar(40) NOT NULL,
+  `payment_amount` decimal(15,2) NOT NULL,
+  `payment_user` int(11) NOT NULL DEFAULT 0,
+  `payment_date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `reference_code` varchar(40) NOT NULL DEFAULT '',
+  PRIMARY KEY (`payment_id`),
+  KEY `payment_sale` (`sale_id`, `payment_type`)
```
jekkos: ok thanks for taking care of this, but we're still missing the foreign key constraints here

SteveIreland: okay, I'll plug that in and test it before merging.

odiea:
@SteveIreland I don't know if anyone else gets this error. When I go to edit a new Invoice payment I receive the following error. It appears to be caused by an empty  $sale_info['amount_tendered']. It does not show if an additional payment has been made.
Severity: Warning --> A non-numeric value encountered C:\wamp64\www\my-master\application\controllers\Sales.php 1197 probably not the same line in Master.

SteveIreland:
Thanks. I will try to replicate the problem tonight and fix it

odiea:
Ok. Tested this am and no Issues. Thanks @SteveIreland

SteveIreland:
@odiea So the big question is did the changes fix the problems you found?

odiea:
So far. When I leave it in dev mode now I do not see any errors nor do I see any log errors..

PR #2448, opened 2019-05-19
Track initial cash refunds

SteveIreland:
This pull request will capture the initial cash refund amount when the sale is completed.  I believe this should help the reports make more sense where cash  refunds are involved.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -582,6 +582,21 @@ class Sales extends Secure_Controller
 		}
 		$data['amount_change'] = $data['amount_due'] * -1;
 
+		if($data['amount_change'] > 0)
+		{
+			// Save cash refund to the cash payment transaction if found, if not then add as new Cash transaction
+
+			if(array_key_exists($this->lang->line('sales_cash'), $data['payments']))
+			{
+				$data['payments'][$this->lang->line('sales_cash')]['cash_refund'] = $data['amount_change'];
+			}
+			else
+			{
+				$payment = array($this->lang->line('sales_cash') => array('payment_type' => $this->lang->line('sales_cash'), 'payment_amount' => 0, 'cash_refund' => $data['amount_change']));
+				$data['payments'] += $payment;
+			}
+		}
+
 		$data['print_price_info'] = TRUE;
 
 		$override_invoice_number = NULL;
@@ -1281,7 +1296,7 @@ class Sales extends Secure_Controller
 			// To maintain tradition we will also delete any payments with 0 amount assuming these are mistakes
 			// introduced at sale time.  This is now done in Sale.php
 
-			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'payment_user' => $employee_id);
+			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'employee_id' => $employee_id);
 		}
 
 		$payment_id = -1;
@@ -1290,7 +1305,7 @@ class Sales extends Secure_Controller
 
 		if($payment_type != PAYMENT_TYPE_UNASSIGNED && $payment_amount <> 0)
 		{
-			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'payment_user' => $employee_id);
+			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'employee_id' => $employee_id);
 		}
 
 		if($this->Sale->update($sale_id, $sale_data, $payments))
--- a/application/migrations/20190502100000_refundtracking.php
+++ b/application/migrations/20190502100000_refundtracking.php
@@ -1,0 +1,110 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_RefundTracking extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		execute_script(APPPATH . 'migrations/sqlscripts/3.3.0_refundtracking.sql');
+
+		// Add missing cash_refund amounts to payments table
+
+		$decimals = totals_decimals();
+
+		$trans_amount = 'ROUND(SUM(CASE WHEN sales_items.discount_type = ' . PERCENT
+			. ' THEN sales_items.item_unit_price * sales_items.quantity_purchased * (1 - sales_items.discount / 100) '
+			. 'ELSE sales_items.item_unit_price * sales_items.quantity_purchased - sales_items.discount END), ' . $decimals . ') AS trans_amount';
+
+		$cash_payment = $this->lang->line('sales_cash');
+
+		$this->db->query('CREATE TEMPORARY TABLE IF NOT EXISTS ' . $this->db->dbprefix('migrate_taxes') .
+			' (INDEX(sale_id))
+			(
+				SELECT sales.sale_id, SUM(sales_taxes.sale_tax_amount) AS total_taxes
+				FROM ' . $this->db->dbprefix('sales') . ' AS sales
+				LEFT OUTER JOIN ' . $this->db->dbprefix('sales_taxes') . ' AS sales_taxes
+					ON sales.sale_id = sales_taxes.sale_id
+				WHERE sales.sale_status = \'' . COMPLETED . '\' AND sales_taxes.tax_type = \'1\'
+				GROUP BY sale_id
+			)'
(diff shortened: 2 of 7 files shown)


application/models/Customer.php
```diff
@@ -118,10 +118,10 @@ public function get_stats($customer_id)
 		$quantity_decimals = quantity_decimals();
 
 		$this->db->select('
-						SUM(sales_payments.payment_amount) AS total,
-						MIN(sales_payments.payment_amount) AS min,
-						MAX(sales_payments.payment_amount) AS max,
-						AVG(sales_payments.payment_amount) AS average,
+						SUM(sales_payments.payment_amount - sales_payments.cash_refund) AS total,
```
jekkos: why aren't we just showing sale total here instead of payments?

application/migrations/sqlscripts/3.3.0_refundtracking.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE `ospos_sales_payments`
+	ADD COLUMN `cash_refund` decimal(15,2) NOT NULL DEFAULT 0 AFTER `payment_amount`,
+	CHANGE `payment_user`  `employee_id` int(11) NOT NULL DEFAULT 0,
```
jekkos: Shouldn't this column be non nullable? having default set to 0 might not represent any valid employee?

application/migrations/20190502100000_refundtracking.php
```diff
@@ -0,0 +1,110 @@
...
+			(
+				select a.sale_id, total_payments - trans_amount as refund_amount
+				from ospos_migrate_sales a
+				join ospos_migrate_payments b on a.sale_id = b.sale_id
+				where total_payments > trans_amount and number_payments = 1
+			)'
+		);
+
+		// Update existing cash transactions with refund amount
+		$this->db->query('update ospos_sales_payments a
```
jekkos: so if I understand correctly we're doing two things here
* Add a cash refund column to existing payments to store the excess payment per line (existing cash payments only)
* Add a new line for the cash refund to the payments table (existing non cash payments)

So we're kind of modelling the change differently depending on the type of payment the refund is linked to. Is there any specific reason for doing it that way?

application/controllers/Sales.php
```diff
@@ -582,6 +582,25 @@ public function complete()
...
+			if($key == $this->lang->line('sales_cash'))
+			{
+				$data['payments'][$key]['cash_refund'] = $data['amount_change'];
+				$count = 0;
+				break;
+			}
+		}
+		if($count > 0)
+		{
+			$payment = array($this->lang->line('sales_cash') => array('payment_type' => $this->lang->line('sales_cash'), 'payment_amount' => 0, 'cash_refund' => $data['amount_change']));
```
jekkos: Is the payment refund amount negative here? It would make sense to give it the opposite sign at least. Also I think we only should add the negative cash transaction if there is an amount due. 

Maybe do a simple array_search for cash payment type key instead of using the count variable, might be more concise.

application/models/Customer.php
SteveIreland: I tried to initiate a discussion regarding this panel and no one seemed interested in discussing the value of payments vs sales.  I find the information doesn't have much value if invoicing is a possible type of sales transaction.  But since we don't have a sales stats table (cough - cough) there would be significant performance hit to includes sales totals on the panel just to edit a customer.

So I just fixed it to factor in cash refunds.

SteveIreland:
I've made the specific changes you've requested.  I believe it is ready to merge.  Can I proceed?

daN4cat:
I'm terribly late with all of this. Have you searched all the code for the keyword refund or due?
Given what you implemented there was another place in the payment report where any due payment was assumed to be cash.
I know the two might not the same but please make sure there is no weird inconsistency that might cost time to debug.

It's all about end cash calculations.
It also affects the daily cashup feature.

But maybe you fixed that already somewhere and I have missed?

jekkos:
@SteveIreland I didn't see I had another comment pending, posted it now. What was the reason behind adding the refund both as a column and as a row to payments table?

SteveIreland:
@jekkos I thought it would have the least impact on the user if I added a cash refund column.  That way if the original payment is made in cash then I still only have one payment that now reflects the amount refunded.   So cash payment minus cash refund equals net cash payment and all of that is on one record.

In the weird case where the original payment is made via a non-cash payment then that doesn't work.  So I had to add a cash payment type, but the cash payment amount is zero and the cash refund amount reflects WHAT IS ASSUMED to have been a refund made via cash.  Someone mentioned that this actually does make sense if there was some agreement that allows for providing some cash to the customer

The formula of subtracting the cash refund from the 0 cash payment amount results in a negative cash payment which makes sense when shown on the reports.

SteveIreland:
@daN4cat I check the payment summary report a couple of times and it was looking good, but it wouldn't hurt for me to spend a bit more time in this area checking things out so I'll do that this weekend.

daN4cat:
Please check even the cash up functionality. I'm speaking in general terms because I don't remember all the details.

SteveIreland:
Even though the summary payments report was producing correct results, I make a change to it to take advantage of the new cash_refund field in payments.

I also reviewed the cashups process and it appears to still be producing correct results.

daN4cat:
Sounds all good then, thanks for confirming.

SteveIreland:
Don't merge yet ... I just ran a migration test against a much larger database and not all of the cash refunds that I expected to be generated were generated.

daN4cat:
Ok

SteveIreland:
Just to note, there wasn't a problem with the migration script.  The bug is actually in MySQL so I modified the migration to work around it.   Specifically the bug manifests itself only when temporary tables are used.  So my fix was to change one of the migration tables from a temporary table to a permanent table.

SteveIreland:
I am concerned that moving the migration table for refunds to "memory" will flip it back into the "temporary table" table treatment by MySQL and re-introduce the issue I spent a few hours trying to identify and resolve.

SteveIreland:
Excellent ... changing it to a non-temporary memory table still avoided the temporary table issue.

The test updated the sales_payments table as expected.

Thanks for the cleanup.

jekkos:
> @jekkos I thought it would have the least impact on the user if I added a cash refund column. That way if the original payment is made in cash then I still only have one payment that now reflects the amount refunded. So cash payment minus cash refund equals net cash payment and all of that is on one record.
> 
> In the weird case where the original payment is made via a non-cash payment then that doesn't work. So I had to add a cash payment type, but the cash payment amount is zero and the cash refund amount reflects WHAT IS ASSUMED to have been a refund made via cash. Someone mentioned that this actually does make sense if there was some agreement that allows for providing some cash to the customer
> 
> The formula of subtracting the cash refund from the 0 cash payment amount results in a negative cash payment which makes sense when shown on the reports.

I indeed see the different scenarios but was wondering whether we couldn't model both cases as a negative entry in the payments table for cash payment type.. In that case we'd just sum the amounts grouped by sale id and the total is the balance due.

daN4cat:
@SteveIreland I believe there is a problem with cash_refund, please check if editing a sale and then submit it's fine. I get an error in my test installation:

```
A PHP Error was encountered
Severity: Notice

Message: Undefined index: cash_refund

Filename: models/Sale.php

Line Number: 549

Backtrace:

File: /home/colibri2/public_html/pos3/application/models/Sale.php
Line: 549
Function: _error_handler

File: /home/colibri2/public_html/pos3/application/controllers/Sales.php
Line: 1311
Function: update

File: /home/colibri2/public_html/pos3/public/index.php
Line: 315
Function: require_once
```

daN4cat:
@SteveIreland I debugged and the views/sale/form.php doesn't deal with cash_refund per se (as you would expect as it's generic / any payments). controller/Sales.php save function doesn't add any cash_refund to the $payments[]. Not sure what you want to do here, set a 'cash_refund' to 0 so the model/Sale.php update doesn't break?
Or the logic should be something else? But the payment form has no concept of showing cash refunds unless it's a cash payment with negative number.

You need to look into this and fix according to the logic you wanted to implement.

jekkos:
If we always model the refund as a new payment (just a negative one), we might be fine here..

daN4cat:
I'll temporary fix setting cash_refund 0 in the $payments[] inside controller/Sales.php in order to avoid exceptions.

Also, could you please confirm that function get_sales_manage_payments_summary($payments, $sales) in tabular_helper.php is correct given the WARNING I put at the time?

SteveIreland:
Thanks. Just saw this and I agree with the change you needed to make. Sorry for the oversight

daN4cat:
But what if somebody sets a negative cash payment which is basically a cash_refund?
I tested a case and it's not managed. When you open that sale form it's shown as zero.
I need to test a bit more and see if it's a legacy case support problem or current behaviour.

PR #2461, opened 2019-06-13
Database Fix

SteveIreland:
This adds a couple of missing constraints.  I haven't been able to test with the most recent master yet because of the bintray issues so please don't merge it yet.

### DIFF
--- a/application/migrations/20190612100000_dbfix.php
+++ b/application/migrations/20190612100000_dbfix.php
@@ -1,0 +1,20 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_DBFix extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		execute_script(APPPATH . 'migrations/sqlscripts/3.3.0_dbfix.sql');
+	}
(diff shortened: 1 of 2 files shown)


SteveIreland:
Don't merge just yet.  I'm planning on adding a couple of constraints to the tax rate table.   I just wanted to someone to take a look at these changes to make sure that I'm meeting expectations.

SteveIreland:
The last changes that I identified have been added to this pull request.  If there are any other missing database constraints just let me know and I'll be happy to add them.

daN4cat:
I've added a comment even if not related specifically to this change, but given we are adding indexes I wonder if we should check we have everything in place. For instance trans_time is used for Items data range selection and there is no index and probably non foreign key neither.

SteveIreland:
I've applied an index to three of the timestamps I've found.  That should certainly improve the performance of any filters or queries that involve the timestamp.

daN4cat:
Thanks for the work @SteveIreland. I've added a date index to expenses table, tested and merged.

jekkos:
Not sure if adding indexes for singe fields will always help us out here. We probably need to check which queries are ran and then explain them to see which (combined) indexes we need to create.

PR #2462, opened 2019-06-13
Update system_info.php

objecttothis:
Remove check for write permissions on import_items.csv file since the file no longer exists and is generated on-the-fly.


objecttothis:
Pretty straight forward.  I just removed the lines of code associated with checking and displaying the write permission status.

PR #2484, opened 2019-07-03
Change newlines into html breaks (invoice template)

jekkos:
Some minor changes to the invoice template

* Found that some templates still use textarea for their variables. This doesn't look very nice in a generated PDF and so I removed the last remniscants in the sale invoice email template.
* Some of them use the `pre` tag which I think means the rendered contents does not need any html markup for line breaks. This also holds for most email templates (work order, quote, etc)
* Some of them use a regular div. In this case I think the contents should be rendered using `nl2br`

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -520,7 +520,7 @@ class Sales extends Secure_Controller
 		$employee_info = $this->Employee->get_info($employee_id);
 		$data['employee'] = $employee_info->first_name . ' ' . mb_substr($employee_info->last_name, 0, 1);
 
-		$data['company_info'] = implode("<br/>", array(
+		$data['company_info'] = implode("\n", array(
 			$this->config->item('address'),
 			$this->config->item('phone')
 		));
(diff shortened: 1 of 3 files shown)


WebShells:
@jekkos that will make the sale comment non-editable field i think?

SteveIreland:
I like the changes, except I was wondering if you needed the State to be on a separate line.  In the U.Ss it is normally just separated by a comma.  I don't think it's a big deal either way.

I'm not a big fan of glopping data together like that since we lose control of the rendering of the individual elements but I assume that's the way it has always been and will always be.

@WebShells the comment is maintainable only from the register and not from the rendered invoice view.  The use of textarea in invoice.php was a misuse of the HTML tag.

WebShells:
@SteveIreland to me actually it was helpful most of the time ;p ! One of the good misuses ;p

SteveIreland:
So I suppose you didn't have a need to save changes to the comments to the database?

I think we still have a a "post sale" way of changing the comment and reprinting the document.

jekkos:
Can we actually save changes in the invoice itself? I have only removed that textarea in the email version.

WebShells:
@SteveIreland no actually using it for instant comments like delivery notes or availability time etc... No need to save these info to db, see useful 😋.

SteveIreland:
Clever ... I hadn't thought of that use case - I just assumed it was a markup blunder.  Is the last entry lost when you send it via email?  The changes that @jekkos made won't have an impact.  I'm guessing that those are lost and if they are is that a problem for you?

Sorry for dragging this out here.  I'm just learning.

jekkos:
No indeed, the idea behind having textarea in the invoice template was indeed allowing us to change something once and the print it. I don't think these changes were applied when you would send it as an email, as this has it's own template.

WebShells:
True, it's not related to emailing invoices as it's related to direct printing, that's the only part where you can make the right use of the textarea... Changing it to divs wont allow text editing anymore since the tags were removed... It's fine to me if no one else is making use of it we can remove.

PR #2488, opened 2019-07-09
3.3.0 patch datetime to date

objecttothis:
These are fixes to bugs introduced with datetime modifications.  Please review and merge into the master if there are no problems.

### DIFF
--- a/application/helpers/locale_helper.php
+++ b/application/helpers/locale_helper.php
@@ -561,6 +561,11 @@ function dateformat_bootstrap($php_format)
 	return strtr($php_format, $SYMBOLS_MATCHING);
 }
 
+function valid_date($date)
+{
+	return preg_match('/^([0-9]{2,4})-([0-1][0-9])-([0-3][0-9])(?:( [0-2][0-9]):([0-5][0-9]):([0-5][0-9]))?$/', $date);
+}
+
 function valid_decimal($decimal)
 {
 	return preg_match('/^(\d*\.)?\d+$/', $decimal);
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -305,7 +305,7 @@ class Attribute extends CI_Model
 		{
 			if($to_type === DATE || $to_type === DECIMAL)
 			{
-				$field = ($to_type === DATETIME ? 'attribute_date' : 'attribute_decimal');
+				$field = ($to_type === DATE ? 'attribute_date' : 'attribute_decimal');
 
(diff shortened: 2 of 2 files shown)


odiea:
And where is the datetime to date stored in the database.

objecttothis:
No, this doesn't fix the reports issue. I just ran across this bug while trying to find the source of the reports issue.

objecttothis:
I'm not sure I understand what you mean by datetime to date. Originally @jekkos added the datetime attribute type but it was later decided that datetime was overkill and that just a date was needed.

odiea:
Ok. I was under the impression that it was in the db but I saw it referred to the attributes datetime function. Thanks

objecttothis:
Attribute_values has a column

application/models/Attribute.php
```diff
@@ -305,7 +305,7 @@ private function convert_definition_type($definition_id, $from_type, $to_type)
 		{
 			if($to_type === DATE || $to_type === DECIMAL)
 			{
-				$field = ($to_type === DATETIME ? 'attribute_date' : 'attribute_decimal');
+				$field = ($to_type === DATE ? 'attribute_date' : 'attribute_decimal');
```
daN4cat: Good catch

application/helpers/locale_helper.php
```diff
@@ -561,6 +561,11 @@ function dateformat_bootstrap($php_format)
 	return strtr($php_format, $SYMBOLS_MATCHING);
 }
 
+function valid_date($date)
```
daN4cat: Where is this used?

application/models/Attribute.php
```diff
@@ -305,7 +305,7 @@ private function convert_definition_type($definition_id, $from_type, $to_type)
 		{
 			if($to_type === DATE || $to_type === DECIMAL)
 			{
-				$field = ($to_type === DATETIME ? 'attribute_date' : 'attribute_decimal');
```
jekkos: Ok I probably missed that one. Thanks for fixing.

application/helpers/locale_helper.php
jekkos: Did I break anything by removing this?

objecttothis: Yes, the text to date attribute conversion in the Attribute model calls it to validate the date.  It's in the locale_helper because it can/should be used in any other date validations that need to get done, just like the decimal validator.

Issue #2471, opened 2019-06-24
Empty Reports in 3.3.0


objecttothis:
### Background information
- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: dbf311
- PHP version is: 7.3.3
- MySQL or MariaDB version is: 5.7.25
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English

### Bug
I've noticed this problem during the entirety of the 3.3.0 development, but I am just now getting to it because I'm not sure if it's just my install or if other people are having the same issue.  I have a 3.2.0 database migrated to 3.3.0 and I'm not able to get any data to show up in any of the reports for any date range.  All reports generate with an empty dataset.  Is anyone else experiencing this issue?  I've even tried creating an item after the migration and selling it but it too does not appear in the report.

objecttothis:
Looking at the Query logs and comparing them between each other I can see that the 3.2.0 logs are including two queries per sale ID in the date range given:

[...]

PR #2490, opened 2019-07-12
fix empty reports

objecttothis:
This corrects an error in a previous migration (2.3_to_2.3.1.sql) that assigned a null value to receivings_stock and sales_stock for the location_id when it should have been the location_id of that particular stock_location.

### DIFF
--- a/application/views/reports/date_input.php
+++ b/application/views/reports/date_input.php
@@ -70,7 +70,7 @@ if(isset($error))
 	?>
 
 	<?php	
-	if (!empty($stock_locations) && count($stock_locations) > 1)
+	if (!empty($stock_locations) && count($stock_locations) > 2)
 	{
 	?>
 		<div class="form-group form-group-sm">
--- a/application/migrations/20190712150200_fix_empty_reports.php
+++ b/application/migrations/20190712150200_fix_empty_reports.php
@@ -1,0 +1,30 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_fix_empty_reports extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		$this->db->select('location_name');
+		$this->db->from('stock_locations');
+		$this->db->where('location_id', 1);
+		$this->db->limit(1);
+		$location_name = $this->db->get()->result_array()[0]['location_name'];
+
+		$location_name = str_replace(' ', '_', $location_name);
+		$this->db->set('location_id',1);
+		$this->db->where('permission_id','receivings_' . $location_name);
+		$this->db->or_where('permission_id', 'sales_' . $location_name);
+		$this->db->update('permissions');
+	}
+
+	public function down()
+	{
+
+	}
+}
+?>


objecttothis:
We need to decide if we need to fix the bug in the old migration script that created this problem in the first place. If it's a supported use case to do partial upgrades, then we really need to also fix the old migration script in addition to this migration script.  If it is an unsupported case to do a partial upgrade, then no need to fix the bug in the old migration script.

objecttothis:
See #2471 for more details.

application/views/reports/date_input.php
```diff
@@ -70,7 +70,7 @@
 	?>
 
 	<?php	
-	if (!empty($stock_locations) && count($stock_locations) > 1)
+	if (!empty($stock_locations) && count($stock_locations) > 2)
```
daN4cat: Good one!

daN4cat:
We move forward as there are for sure people like you in "halfway" status. Better this approach so every "halfway" gets the fix once forever.

Well done to spot this issue.

objecttothis:
> 
> 
> We move forward as there are for sure people like you in "halfway" status. Better this approach so every "halfway" gets the fix once forever.

I think everyone agrees that the migration script that I just created needs to be there.  I think we are saying two different things by "halfway" though.  Everyone also agrees that there is a bug in the migration script causing receiving_stock and sales_stock to remain NULL, the question is let's say there is someone who wants to upgrade to, say 3.2.3 but doesn't want to upgrade to 3.3.0 just yet because they don't want attributes just yet.  Since that person won't hit this patch, they will still have the bug.  We have two options

1- Anything prior to the latest version of OSPOS is unsupported, and while people are allowed to upgrade just to a specific version, we won't retroactively fix bugs, but instead fix them in the latest version which means anyone asking for support for this bug or any other bug will be asked to upgrade to the latest version before we can help them.

2- Allow versions of OSPOS prior to the latest release.  This means that we would need to both release this migration patch and go back and fix the bug in the migration script that caused the problem in the first place (ostensibly by removing the IGNORE keyword.

@daN4cat I think you're saying that the 1st one is what we should do.  Is that accurate?

jekkos:
Could we fix it in the old migration script and then do an idempotent fix in the migration to 3.3? I'm indeed referring to scenario 2) here. Would be ideal if we could solve it for both cases.

daN4cat:
Maybe it's me, but I don't get how we could retroactively fixed a release that is out, unless you add that specific fix to produce a 3.2.4 instead of 3.3.0.

However I would push for 3.3.0 and upgrade instead of branching sub releases, simply because we can't cope with that type of scenario.

I'll merge this as at least will do for what we have.

objecttothis:
@daN4cat you can't retroactively fix a release that someone has already migrated to. You can only retroactively fix a release that someone hasn't already migrated to. For example, user A is on release 2.0 and in the 2.3 release a bug is introduced. There is a migration in 3.3 that fixes the bug, but if user A only wants to upgrade to 3.2.3, they get the bug when they hit 2.3 but never get rid of the bug.

You fix this by releasing the migration in 3.3 for everyone who already has the bug AND fixing the 2.3 migration so that anyone who doesn't have the bug never gets it.  This covers both use cases.

daN4cat:
But who gets 3.3.0 automatically upgrades to it. So I still don't see the value of 3.2.3 fix unless you expect people to apply patches to old versions.
At that point issue 3.2.4.

objecttothis:
There is no 3.2.3 fix. Only a 3.3.0 fix (already merged) and a 2.3 change of the migration sql so that the bug never gets introduced for anyone pre-2.3 who does not migrate all the way to 3.3.0.  The bug is that the original introduction of receivings_stock and sales_stock into permissions was set as NULL and the 2.3.1 migration script that has the sql to set it as location_id of 1 has the IGNORE constraint.  That is what I'm suggesting changing in addition to the 3.3.0 migration that we just merged

objecttothis:
@daN4cat hopefully this will clarify everything.  There are three types of users that apply to the question at hand:

1. Users on any version previous to 2.3.2 who want to migrate only to a version before 3.3.0.
2. Users on any version previous to 2.3.2 who want to migrate all the way to the latest version.
3. Users on any version after 2.3.2.

The code creating the bug is on lines 38 and 29 of /database/2.3_to_2.3.1.sql

`INSERT INTO `ospos_permissions` (`permission_id`, `module_id`) VALUES`
`...`
`('sales_stock', 'sales'),`
`('receivings_stock', 'receivings');`

Note that location_id is not set so it defaults to NULL.  For user's 2 and 3 they have to migrate to 3.3.0 to fix the bug and nothing else is needed. If we change that code to properly assign a location_id of 1 to those rows in the 2.3_to_2.3.1.sql then user 1 will be bug free when the run 2.3_to_2.3.1.sql.  We only need to do this if we want user 1 to be an acceptable use case.  If we don't want to allow users to migrate past 2.3 but only up to 3.2.3 then we don't fix the bug in the original script and just tell everyone that wants to do this (which I don't think will be many users) that they need to go to 3.3.0 if they want the bug fixed.

The only caveat to fixing the bug in 2.3_to_2.3.1.sql as well as the 3.3.0 migration is that we actually need to fix it in the previous releases of OSPOS available on the site (2.4 - 3.2.3). The reason for this is that if we just change those lines of code in the 3.3.0 release, no one will be able to use that code upgrade from pre-2.3.1 to say 3.2.3.  This is because the code changes in the 3.3.0 won't work without the database also being 3.3.0.  So if someone wants to only upgrade to 3.2.3 for example, they need to download the 3.2.3. release package which would need to have the code change in the 2.3_to_2.3.1.sql file.

daN4cat:
That's the point, the 3.2.3 release package is already out. You can't change it.

objecttothis:
OK, that's fine.  Let's just make partial updates be not a supported use case of OSPOS and then we don't need to worry about retroactive fixes since they seem to introduce too much complication.  I'm guessing it is a rare situation anyway.

PR #2496, opened 2019-07-20
Improves the behavior of dinner table selection.

SteveIreland:
This is an attempt to resolve the reported issue with dinner table selection.  I found one or two other issues with it that I also resolved.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -176,11 +176,23 @@ class Sales extends Secure_Controller
 			$this->sale_lib->set_sale_type(SALE_TYPE_RETURN);
 		}
 
+		if($this->config->item('dinner_table_enable') == TRUE)
+		{
+			$occupied_dinner_table = $this->input->post('dinner_table');
+			$released_dinner_table = $this->sale_lib->get_dinner_table();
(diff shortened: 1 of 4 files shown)


odiea:
It does not appear to be clearing the status flag in the dinner tables The status is unchanged after cancelling a Sale.
2019-07-20 13_47_27-Window

odiea:
Both  2 and 3 were cancelled at the register after being Suspended. It appears the drop down is now showing the current un suspended Table.

odiea:
Both tables 2 and 3 were cancelled. From the image you can see the Tables are not available for use.

2019-07-20 14_03_53-Window

SteveIreland:
Okay. I will test it again.  Thanks for checking it out

SteveIreland:
I'm happy you sent it back.  In addition to fixing the issue you found I realized that there were a couple of other items that needed to be addressed.

The Delivery and Takeaway should never be flagged as occupied, so I've corrected that.

And I also only update the dinner_tables table if the dinner table configuration is enabled.

odiea:
Works great now.

daN4cat:
Thanks @SteveIreland for the good work done. Shall we merge this?

odiea:
I would say go ahead. It is a good improvement.

PR #2531, opened 2019-09-08
Remove the redundancy of cart resequencing for POS receipts.

SteveIreland:
When I made my first changes to OSPOS I was hesitant to touch anything other than what I absolutely knew I had to change.  So when I reworked the cart resorting and filtering logic that takes place prior to the actual printing of the invoice in order to avoid re-retrieving from the database at the last second I could not bring myself to delete what was taking place for POS receipts.

This pull request will finally remove that line of code and thus improve the performance a bit (probably not noticeable but it's the right thing to do in order to keep the code clean).

It would be nice to have this included in the 3.3.0 release.  It should be low risk since this is already being done for invoices, quotes, and work orders.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -802,10 +802,6 @@ class Sales extends Secure_Controller
 			else
 			{
 				$data['barcode'] = $this->barcode_lib->generate_receipt_barcode($data['sale_id']);
-
-				// Reload (sorted) and filter the cart line items for printing purposes
-				$data['cart'] = $this->get_filtered($this->sale_lib->get_cart_reordered($data['sale_id_num']));
-
 				$this->load->view('sales/receipt', $data);
 				$this->sale_lib->clear_all();
 			}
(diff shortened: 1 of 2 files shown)


SteveIreland:
@daN4cat even though this is a simple change I didn't want to merge it myself without you getting your eyeballs on it first.

daN4cat:
Have you tested the scenario where you reprint a receipt from the past?
I know it's stating the obvious but there are two printing scenarios: a) current transaction b) past transaction.
The latter requires a reload of the data, etc etc.

odiea:
I tested that scenario and all looks happy to me. The first 1 was dated 2015 and todays 2019.

daN4cat:
It sounds like we are happy with this. I'm sorry if I need to rely on you guys but I struggle to find any time for OSPOS since June/July.

daN4cat:
Merged

SteveIreland:
Thank you both!

PR #2532, opened 2019-09-08
fix #2522 by reverting cosmetic change in f235ced

mckaygerhard:
by the moment we need to this report to work, the reference select box in that moment have the conditional in the view, element "all" in the select box are not a part of the locations.. ("all" after counting if there's any location presented)

so dont care if the select box have only one element with "all" and only one location... **just need to this work!**

### DIFF
--- a/application/views/reports/date_input.php
+++ b/application/views/reports/date_input.php
@@ -70,7 +70,7 @@ if(isset($error))
 	?>
 
 	<?php	
-	if (!empty($stock_locations) && count($stock_locations) > 2)
+	if (!empty($stock_locations) && count($stock_locations) > 1)
 	{
 	?>
 		<div class="form-group form-group-sm">


jekkos:
You might be right that this needs to be reverted as the final fix for this issue was somewhere in the migration if I remember well.

@objecttothis could you confirm if this is fine?

odiea:
I can confirm that the few reports that were not showing are now showing. @andriux1990. Thanks for your persistence. @mckaygerhard Thanks for the pull request.

daN4cat:
The problem is what was that change fixing what?

WebShells:
That's what I first thought of...

odiea:
Update date_input.php

Corrected this from 1 to 2.  'All' is counted in the array which means that if there is only one stock location, this dropdown will still appear.  Setting this to 2 means that the dropdown only shows when there is more than 1 stock location.

mckaygerhard:
>  Setting this to 2 means that the dropdown only shows when there is more than 1 stock location.

that's right that's why i revert to 1.. please pull merge this to continue.. i need some other chages to integrate a mini accounitng simple system to avoid that shit of Quickbooks ... seems are in all the bussines ...

mckaygerhard:
> Setting this to 2 means that the dropdown only shows when there is more than 1 stock location.

thta's right! that's why with one location doe snot show nothing.. that's why i revert to 1.. please pull merge this to continue..  (reedited due to langauaje mistakes)

daN4cat:
@odiea therefore does it make sense to show it anyway even with only one location?
To me it sounds more a lack of default if nothing is shown.
The location variable is not defaulted and therefore when not shown has nothing and the filter doesn't work.
If I find the time this eve I'll check that scenario.

odiea:
In my test version I have 1 stock location and "all" and "stock" show just fine in the dropdown after changing it back to 1 and the Reports are working properly now. He had good intentions but  it did cause some Major issue.

daN4cat:
See my comment in #2522, I believe I fixed properly and committed to master.

mckaygerhard:
see my comment does not contains the commit that fix.. so are in  64eb7ab puff

testing later..

mckaygerhard:
seeems the commit does not fix complety https://github.com/opensourcepos/opensourcepos/commit/64eb7abf2f77247a21bdb8224a46d94837341f70

seems @jekkos already explain  in the issue comments a reference.. so i restore the path ..

PR #2534, opened 2019-09-10
EG to LB

WebShells:
Although it fits for all arabic language countries, yet it's the most accurate locale to be used since
most of AR Translation were made for LB language.


daN4cat:
Why not duplicate? If I remember right there was a guy who created the ar-EG and would be better retain it.

WebShells:
@daN4cat i handled the translation part for the past period, it was around 28% translated as for ar-EG. Am not sure if it's better to keep both; in this case, we should have some one from EG willing to skim double check the strings... Whatever you find more suitable.

jekkos:
He might be notified through weblate once strings are updated, some let's hope he comes back and improves the translation :)

WebShells:
Added back ar-EG to the list 👍 we can merge...

Issue #2452, opened 2019-05-25
Very serious error in the items


andriux1990:
Creating an item with the name: OXIMETASOLINA 0.05% ADULTO
Category: GOTAS NASALES
Wholesale Price: 44
Retail Price: 55
![Screenshot_2019-05-24 Save Mart Powered by OSPOS 3 3 0](https://user-images.githubusercontent.com/29256343/58363708-e036df80-7e65-11e9-8436-f0979154d52e.png)
And it does not allow me to save, so
To the item name delete the symbol%, then that way if I keep it.
But when looking for the item it does not appear and when you put the last sheet, No Items to display.
![Screenshot_2019-05-24 Save Mart Powered by OSPOS 3 3 0(1)](https://user-images.githubusercontent.com/29256343/58363780-e5e0f500-7e66-11e9-89ca-c98be32d933c.png)
As I noticed it was very strange I did the same thing in the ospos demo and it happened exactly the same.
No Items to display.

Check the database with phpmyadmin and the item was duplicated many times
![Screenshot_2019-05-24 localhost localhost final ospos_items phpMyAdmin 4 8 5](https://user-images.githubusercontent.com/29256343/58363844-ee85fb00-7e67-11e9-84ae-1bd532843433.png)
How can I recover the items that are not shown and fix the database
Please help

daN4cat:
Version?

andriux1990:
I did tests in several versions of ospos 3.3.0 up to the ones that are in https://demo.opensourcepos.org
That currently is 3.3.0 - 39639c

andriux1990:
On the site https://demo.opensourcepos.org/items says No Items to display. and it's not because there is no element, because I also did the same thing there, They are hidden so to speak, you can create a new item and always say no Items to display.

odiea:
It does appear to be a bug but if you wrap the item name or % in a single or double quote it does work.

andriux1990:
The duplicates that were created were due to the number of times you try to save the item

SteveIreland:
It's probably the problem created in the CodeIgniter CSR prevention function xss_clean.  I recall needing to work around it on an earlier issue.  It does not like the % sign followed by a specific characters.

mckaygerhard:
this are not a Codeigniter bug, the allowe chars to send in a request are configured in applications/config due the ajax injection.. so then i suggest to stidy if there a better ways to do that event configured to allow the "%" char in the request put/requst call

[...]

PR #2545, opened 2019-09-21
WIP: part1 fix bug #2452 in right way due there's no need of

mckaygerhard:
The issue is that in the lines around 610 the response coded in json occurs if the information was saved or failed, this is an output a response to the user, the only data that comes from "outside" is the input of the description (and that prevously was cleaned from xss) .. is the only place where the xss code would come, so the first thing I see is that it is not necessary to treat the entire character string of the message but only that of the name of the item, the second is that instead of applying the xss simply removed the html characters .i'll explain my arguments why are secure and allowed:

**Think about it .. @jekkos , @objecttothis and @daN4cat :  the xss filtering was already applied before saving in the database, the data that comes in the name would be cracked only by the technique of "man in the middle" or similar .. so therefore I just have to protect of possible html codes since sql will not occur (if I click again on edit item this will bring the info of the db .. not the interface)** of course.. so much "bla bla bla" for something obvius: unnecessary code, this pull will property fixed the #2452 

1. we already know that this can allow some form of xss? *not at all*, i'll use only in the "name" of the resulting problem.. not in all the message that comes from the lang file.. the lang file are internally.. to property crack the attacker must be inside of the opensourcepos installation so this are nonsense, if are inside there's no salvation..
2. so then i applied only a `strip_tags($item_data['name'])` in the messsage and remove the xss cleaning.. (due the reason previously explained there's no sense) that's all, solution recommended in the issue are nonsense,, due response from opensourcepos team are "set to false the xss clean" let ALL the opensourcepos system without protection.. 
3. due the line are just a message of response.. the only xss posibilities are just html codes.. so this hack can be improved with a config flag (permit not allowed chars in the items descriptions), i can later after disscuss that commit more changes to peermit that...

### DIFF
--- a/application/controllers/Items.php
+++ b/application/controllers/Items.php
@@ -612,20 +612,20 @@ class Items extends Secure_Controller
 
 			if($success && $upload_success)
 			{
-				$message = $this->xss_clean($this->lang->line('items_successful_' . ($new_item ? 'adding' : 'updating')) . ' ' . $item_data['name']);
+				$message = $this->lang->line('items_successful_' . ($new_item ? 'adding' : 'updating') . ' ' . strip_tags($item_data['name']) );
 
 				echo json_encode(array('success' => TRUE, 'message' => $message, 'id' => $item_id));
 			}
 			else
 			{
-				$message = $this->xss_clean($upload_success ? $this->lang->line('items_error_adding_updating') . ' ' . $item_data['name'] : strip_tags($this->upload->display_errors()));
+				$message = $upload_success ? $this->lang->line('items_error_adding_updating') . ' ' . strip_tags($item_data['name']) : strip_tags($this->upload->display_errors());
 
 				echo json_encode(array('success' => FALSE, 'message' => $message, 'id' => $item_id));
 			}
 		}
 		else // failure
 		{
-			$message = $this->xss_clean($this->lang->line('items_error_adding_updating') . ' ' . $item_data['name']);
+			$message = $this->lang->line('items_error_adding_updating') . ' ' . strip_tags($item_data['name']);
 
 			echo json_encode(array('success' => FALSE, 'message' => $message, 'id' => -1));
 		}


mckaygerhard:
ah i forgoted.. before inserting in the DB we need to encode the input.. that are another path but i dont want to apply if this will not apply.. in my case solves complety, using base64 encoding before save to db and base64 decoding after retreive from db

application/controllers/Items.php
```diff
@@ -612,20 +612,20 @@ public function save($item_id = -1)
...
 			{
-				$message = $this->xss_clean($this->lang->line('items_successful_' . ($new_item ? 'adding' : 'updating')) . ' ' . $item_data['name']);
+				$message = $this->lang->line('items_successful_' . ($new_item ? 'adding' : 'updating') . ' ' . strip_tags($item_data['name']) );
 
 				echo json_encode(array('success' => TRUE, 'message' => $message, 'id' => $item_id));
 			}
 			else
 			{
-				$message = $this->xss_clean($upload_success ? $this->lang->line('items_error_adding_updating') . ' ' . $item_data['name'] : strip_tags($this->upload->display_errors()));
+				$message = $upload_success ? $this->lang->line('items_error_adding_updating') . ' ' . strip_tags($item_data['name']) : strip_tags($this->upload->display_errors());
```
jekkos: the clean should be kept here

application/controllers/Items.php
```diff
@@ -612,20 +612,20 @@ public function save($item_id = -1)
...
 
 				echo json_encode(array('success' => FALSE, 'message' => $message, 'id' => $item_id));
 			}
 		}
 		else // failure
 		{
-			$message = $this->xss_clean($this->lang->line('items_error_adding_updating') . ' ' . $item_data['name']);
+			$message = $this->lang->line('items_error_adding_updating') . ' ' . strip_tags($item_data['name']);
```
jekkos: same here

jekkos:
@daN4cat indeed the change would introduce a potential xss vuln. Closed it for now.

application/controllers/Items.php
mckaygerhard: part 2 (now due reject only solved to me) will coded the string! but well .. it's woring for all of the rest!

Issue #2515, opened 2019-08-13
import_items.csv barcode not import


andriux1990:
The barcode for alphanumeric numbers is not imported to the database when importing a csv file.
Use all types of codes 39, 128, EAN 8 and EAN 13, from only 4 digit numbers etc, without any result.

All the data in the table are imported to Ospos without any problem, but the barcodes are not, and the maximum number of products that can be imported without generating problems is 150 or 200 maximum.

WebShells:
@andriux1990 can you please copy the info from System info tab in the configuration section if available so we can help you in a better way ? or fill in the issue template please....

andriux1990:
![Screenshot_2019-08-13 Tienda Gales Potenciado por NMP 3 3 0](https://user-images.githubusercontent.com/29256343/62985601-5871c900-bdf5-11e9-84b1-c5216721f1d1.png)
I am using this compilation opensourcepos.20190806222050.master.8bca8f.zip

andriux1990:
also download the latest version of bintray opensourcepos.20190811152407.master.7ebae8.zip

I did tests on another operating system and I always had the same problem, no matter the bar codes

WebShells:
Maybe there is a duplicate? Once it reaches this specific row it's interrupting the import process? Specifically you mentioned that you are having a problem inserting more than 150 or 100 item.

andriux1990:
As I had that problem of duplicates what I did was create a new db and import 50 items for each import.
I currently have no duplicates, I have tried to import a single product with a barcode and the barcode does not matter.
I get all the data in the table except the barcode

andriux1990:
I'm not a good programmer, in fact, I'm just learning

since it is very difficult for me to solve the problem of importing barcodes.
I created an attribute with the name codes, type of text. my intention to put the codes, since it allows me to import attributes.
The problem in this is that in the sales module, the search engine only looks for barcode and the names of the items, but does not look for a product by the name of the attribute.
I think that in order to search for the attribute from the register, something must be changed in models / Item.php but I still don't know exactly where.

odiea:
Apparently the Barcode is not being read during the import. I inserted a column before Barcode and it finally saved the barcode. I had to name the column to get it to work.

daN4cat:
The import was changed recently, the logic is different. There must be a keyword or mapping bug.

izeblood:
Can we add the qr codes since the library of it is available?

daN4cat:
[...]

PR #2546, opened 2019-09-22
Fix barcode import (#2515)

jekkos:
Fix for the barcode import issue. For some reason first three characters are skipped, and that's why we fail to match the column header (it sees 'code' instead of 'Barcode').

@daN4cat any idea why the byte order marker snippet was added? Is it ok to remove or should it be kept in a specific case?

### DIFF
--- a/application/helpers/importfile_helper.php
+++ b/application/helpers/importfile_helper.php
@@ -62,9 +62,6 @@ function get_csv_file($file_name)
 	
 	if(($csv_file = fopen($file_name,'r')) !== FALSE)
 	{
-		//Skip Byte-Order Mark
-		fseek($csv_file, 3);
-		
 		while (($data = fgetcsv($csv_file)) !== FALSE)
 		{
 			$line_array[] = $data;


odiea:
I downloaded this master from bintray and all works great now.I had no issues with the excel import and barcodes showed up.

daN4cat:
No idea why that was added.
Only @objecttothis can say....

daN4cat:
Unless it's a Windows, Linux, Mac issue?

But I would merge since @odiea tested it.

objecttothis:
> 
> 
> Fix for the barcode import issue. For some reason first three characters are skipped, and that's why we fail to match the column header (it sees 'code' instead of 'Barcode').
> 
> @daN4cat any idea why the byte order marker snippet was added? Is it ok to remove or should it be kept in a specific case?

This is long overdue but I only just came back.  I added BOM because excel doesn't read unicode CSV properly without it.  Since we are calling it an import from Excel, I figured that it should be fully compliant with what Excel is expecting for UTF-8  content.  That shouldn't cause the first three characters to skip though... was that ever figured out?

objecttothis:
> It's probably needed if you edit the template in Excel and save it as utf8 csv? I must admit that i didn't try that. ( Just read up on this special excel encoding )

I wish Microsoft would get their aces in gear and start implementing a normal CSV standard.  The BOM business just adds extra complexity, but it's needed for excel to not render UTF8 characters as junk symbols.

daN4cat:
@objecttothis wishful thinking, we know that some companies love to twist the open standards to muddy the waters.

jekkos:
I won't use Excel for editing csv as I'm on Linux, but i suppose most of the windows users will.

So if that is the way Excel saves those files then i suppose we need to add this back in some way, but only skip the first three bytes if this is indeed an Excel byte order Mark. Alas i have no access to a decent computer now so feel free to submit something in the mean time.

daN4cat:
What if we call it import from .csv so we are more open source? It makes more sense to me.

objecttothis:
> 
> 
> What if we call it import from .csv so we are more open source? It makes more sense to me.

I prefer that option for the naming.  Yet, I also think we should make it excel compatible out of the box though and just check for the BOM and skip it on import if it exists.  The reason for this is that I believe most people will be using excel to populate the generated csv template.

objecttothis:
> 
> 
> I won't use Excel for editing csv as I'm on Linux, but i suppose most of the windows users will.
> 
> So if that is the way Excel saves those files then i suppose we need to add this back in some way, but only skip the first three bytes if this is indeed an Excel byte order Mark. Alas i have no access to a decent computer now so feel free to submit something in the mean time.

I can work on it.  Is everyone ok with renaming "Import from Excel" to "CSV Import" but me adding in the BOM on the generated CSV and skipping it on the import, but only if it exists to avoid the risk of skipping "bar" in barcode when the BOM doesn't exist?

daN4cat:
Fine with renaming and make it flexible to work with Excel type of .csv and non Excel types.
I would say we add this to 3.3.1.

Issue #2521, opened 2019-08-27
Duplicate barcoded items created in the database when duplicates are not allowed.


klisle:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.2.0

### Issue / Bug  / Question / New Feature

[...]

PR #2550, opened 2019-09-24
Trigger validation when creating items in bulk (#2521)

jekkos:
@klisle here a fix for the duplicate barcode issue.

Indeed I can confirm that validation was disabled after the first submit. The jQuery validation logic needs to be initialized again after. I verified this locally but would be good if someone can confirm if it is fine.

### DIFF
--- a/application/views/items/form.php
+++ b/application/views/items/form.php
@@ -457,8 +457,6 @@ $(document).ready(function()
 		focus: fill_value
 	});
 
-
-
 	var fill_value = function(event, ui) {
 		event.preventDefault();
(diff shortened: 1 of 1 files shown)


klisle:
good to know. I am however still running 3.2.0. I hope the issue I am having ith 3.3.0 can be resolved soon. Thanks or the energy.

klisle:
Issue still persists in the newest version :(
Getting duplicates when Enter key is rapidly repeatedly pressed when adding new item even though duplicates are disabled.
image

daN4cat:
@jekkos are you still around?

PR #2560, opened 2019-10-01
Update invoice_email.css

tuthanika:
Error show utf-8 font

### DIFF
--- a/public/css/invoice_email.css
+++ b/public/css/invoice_email.css
@@ -5,11 +5,11 @@
 */
 #menubar, #footer { display: none; }
 * { margin: 0; padding: 0; }
-body { font-family: Helvetica; font-size: 13px; }
+body { font-family: DejaVu Sans; font-size: 13px; }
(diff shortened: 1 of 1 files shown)


daN4cat:
Thank you for your contribution. What is the problem you are fixing with this change?

tuthanika:
> Thank you for your contribution. What is the problem you are fixing with this change?

When sending the invoice by email, the generated pdf file has the font error utf-8

SteveIreland:
Is anyone else experiencing this issue?  If this is an error then it seems like we would need a more complete solution (something that works for all PDF renderings).

jekkos:
@SteveIreland indeed we should define the fonts in order of preference, if one is not available it should fallback to another.

daN4cat:
@SteveIreland is this something to merge with the other invoice email change?

WebShells:
I think so

SteveIreland:
The changes in this pull request have been included in the #2660 pul request

PR #2562, opened 2019-10-01
Update manage_tables.js

tuthanika:
Change the color to match the dark background


tuthanika:
@daN4cat When in the darkly interface, the color of the text is white, so the original color overwriting is quite difficult to see, the animation time is too long to be uncomfortable for the user.

jekkos:
If this change only applies to darkly, then we can't just change it for all themes, as it will look bad in others. Not sure if there is a way to override this behavior in CSS by using a specific selector that is only applied if darkly is active. But at least it feels like you'd have to solve it that way.

PR #2579, opened 2019-10-11
Add a code to indicate a taxable item on a sales receipt

SteveIreland:
This is the "more complete" solution to print a code on a sales receipt that will indicate whether an item was taxable.  The language files can be used to establish what the code is in case you don't like the "T" or blank" combo ("T" = taxable, blank = non-taxable).


daN4cat:
Looks ok, I merge.

Issue #2574, opened 2019-10-05
Changing currency code in Localization Configuration


wd-code:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 0b9a76
- PHP version is: 7.3
- MySQL or MariaDB version is: MarisDB 5.5
- OS and version is: Debian 9
- WebServer is: Nginx 1.14.1
- Selected language is: French / English

### Issue / Bug
![image](https://user-images.githubusercontent.com/6681481/66261455-dcc94280-e7cd-11e9-9849-cf71378d5921.png)

In the Localizationconfiguration, when I want to change the currency code from USD to EUR then hit the submit button, the currency code comes back to USD.

SteveIreland:
I'll take a look a this issue.

[...]

PR #2581, opened 2019-10-17
Fix numeric locale configuration.

SteveIreland:
This should address  a couple of maintenance issues with numeric locale configuration ... hopefully without introducing new ones.  This addresses issue #2574


daN4cat:
Thanks @Ste, I merged the code. Let's wait for any test feedback.

PR #2583, opened 2019-10-19
Fix numeric locale configuration.

SteveIreland:
This is another attempt to fix the locale configuration issue #2574.

### DIFF
--- a/application/controllers/Config.php
+++ b/application/controllers/Config.php
@@ -209,6 +209,7 @@ class Config extends Secure_Controller
 		$data['tax_category_options'] = $this->tax_lib->get_tax_category_options();
 		$data['tax_jurisdiction_options'] = $this->tax_lib->get_tax_jurisdiction_options();
 		$data['show_office_group'] = $this->Module->get_show_office_group();
+		$data['currency_code'] = $this->config->item('currency_code');
 		
 		$data = $this->xss_clean($data);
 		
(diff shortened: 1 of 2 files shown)


application/controllers/Config.php
```diff
@@ -308,10 +309,21 @@ public function save_general()
 
 	public function ajax_check_number_locale()
 	{
-		$number_locale = $this->input->post('number_locale');
-		$fmt = new \NumberFormatter($number_locale, \NumberFormatter::CURRENCY);
-		$currency_symbol = empty($this->input->post('currency_symbol')) ? $fmt->getSymbol(\NumberFormatter::CURRENCY_SYMBOL) : $this->input->post('currency_symbol');
-		$currency_code = empty($this->input->post('currency_code')) ? $fmt->getTextAttribute(\NumberFormatter::CURRENCY_CODE) : $this->input->post('currency_code');
+        $number_locale = $this->input->post('number_locale');
```
daN4cat: There is an indentation issue with the newly added code.

odiea:
I can actually see changes when selecting a different language code  for localization now. It also appears to be saved with the Submit button and the currency symbol can be changed after selecting the new localization. Thanks @SteveIreland

application/controllers/Config.php
SteveIreland: Fixed

SteveIreland:
Thanks for testing it @odiea!

daN4cat:
Good job, @SteveIreland

Issue #2537, opened 2019-09-12
item type disabled


andriux1990:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.

![Captura de pantalla de 2019-09-12 10-40-56](https://user-images.githubusercontent.com/29256343/64803851-ebb03280-d54a-11e9-9f78-7bb97095bc98.png)

[...]

PR #2603, opened 2019-11-11
Fix item type maintenance issue regardless.

SteveIreland:
This pull request attempts to address issue #2537.  I can't explain how it ever worked, but it did work at one time.  It could be that the addition of the additional item types broke it.  But regardless I took advantage of this opportunity to rework it a bit and retest all of the various item type scenarios and it seems to work now.


odiea:
Works great. Thanks @SteveIreland

PR #2606, opened 2019-11-12
type="username" is not valid html and breaks firefox autofill; changed to "text"

bricas:


public/css/login.css
```diff
@@ -36,7 +36,7 @@ form
 	padding:6% 4%;
 }
 
-input[type="username"], input[type="password"]
+input[name="text"], input[type="password"]
```
daN4cat: Not sure it's name="text". Isn't type="text" and name="username"?

daN4cat:
Good catch but css needs fixing.

public/css/login.css
bricas: sigh. the mind i strange thing sometimes! :) In any case, this should now be fixed.

daN4cat:
Merged and thanks for the contribution.

public/css/login.css
daN4cat: That's why we review :-)

PR #2611, opened 2019-11-18
3.3.1 patch csv import bom check

objecttothis:


objecttothis:
The refactor may require us to do some work on the weblate side (Pull maybe)

daN4cat:
That's for the contribution, I was wondering whether we could change Excel for CSV in all the translations to easy the weblate sync.

objecttothis:
> 
> 
> Thanks for the contribution, I was wondering whether we could change Excel for CSV in all the translations to easy the weblate sync.

I would think that it's as simple as doing a pull from weblate, but I don't know.  Perhaps @jekkos is more familiar with it?  Either way, you wouldn't be able to do it until we merged the pull request.  I think I have permissions on weblate to do a pull, but admittedly, I am less familiar than @jekkos, so it might be better if I didn't do that part.

daN4cat:
My thinking was more on doing a replacement in the translations directly as part of this PR.
But I would need to search in the translation code to get an idea whether Excel to CSV is a straight forward thing to do.

objecttothis:
> 
> 
> My thinking was more on doing a replacement in the translations directly as part of this PR.
> But I would need to search in the translation code to get an idea whether Excel to CSV is a straight forward thing to do.

OK, then I'll assign this to you for now to research whether weblate can just pull the changes in from the main language and propagate them to the other languages or if we have to change the variable names in each of the other language files for it to work properly.  I, unfortunately won't have time to do this for another week since I'm under a deadline on getting the Third Party Integrations scaffolding in place and the CLCdesq Integration written.

objecttothis:
@daN4cat so do I need to refactor variable names in the non-source languages in order for Weblate to figure it out or will Weblate just overwrite the variable names after the Pull operation?  If we need to rebase the branch onto the master after Weblate pushes any changes to translations to the master.  Then once we merge this branch into the master we won't lose any translation changes.  Immediately after merging into the master we will need to do a pull in Weblate.  If weblate will automatically adjust the variable names in other languages then it's a little simpler.

daN4cat:
I would change the variable names in the code.

objecttothis:
OK, I just refactored the variable names in all of the languages and replaced Excel with CSV in the translation files.  We need to test then merge quickly or freeze all translations in weblate so that we don't end up losing changes that people make to translations between now and when we pull into weblate.

objecttothis:
OK, I just locked the local repository from changes on weblate.  I don't know if that will prevent translations or not, but I don't want people doing work that gets lost.  Let me download the build and test it to make sure I didn't miss anything.

objecttothis:
@daN4cat I've downloaded this build and tested it.  Now it needs to be signed off on and merged.  Once merged I think we should be able to perform a pull from weblate and then confirm that it worked.  Once confirmed we should be able to unlock the local repository on weblate.  The ball is in your court.

daN4cat:
Thanks @objecttothis for all of this, I know it was a bit of a pain by you have done a great job. Excel was a legacy thing from the old OSPOS and don't know why it was like that. We probably never prioritised during the "big tidy up".
I'll merge and check weblate side.

After this I can release this new version as we should be good now.

Issue #2608, opened 2019-11-13
CI Events for Third-Party integrations


objecttothis:
### Background information

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.1
- OSPOS git commit hash is: 
- PHP version is: 7.3.6
- MySQL or MariaDB version is: 5.7.24
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### New Feature

[...]

Issue #2381, opened 2019-02-26
Update Items updates the database even when no changes are made


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: efd4e9
- PHP version is: 5.6.38
- MySQL or MariaDB version is: MySQL 5.7.24
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue

Current behavior when submit is pressed on the Item form view is an update is made to the database regardless of whether anything was changed in the form or not.  It also updates all fields and not just changed fields.

[...]

Issue #2315, opened 2019-01-18
Preparing the way for Third Party Integration


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 
- PHP version is: 5.6.3
- MySQL or MariaDB version is: 5.7
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### New Feature
** I renamed the issue because it's a bigger question than just the UI.

Is anyone against me renaming the Mailchimp tab in Config to "Third Party Integration" in order to make way for additional integration configurations such as the one I am working on for CCIDesq CMS? (#1680 )  

[...]

Issue #1680, opened 2017-10-15
CCIDesq One-way integration


objecttothis:
### New Feature / Enhancement
Our bookstore is implementing a website redevelopment using CCIDesq eCommerce CMS.  I want to write the integration to OSPOS to send CCIDesq changes to inventory (add, remove, update) when they happen on the OSPOS side.  This would be a one-way integration for now.  Changes are made via RESTful API calls to CCIDesq.  Here are the parameters that I have so far:
- This would be an optional feature turned off by default in OSPOS (similar to the way mailchip is optional).
- When turned on, and configured API calls would be triggered on sale of an item, receiving, return, excel import of items and any other inventory change that I haven't thought of yet.
- Currently there are no plans for two-way data exchange as that would require REST being implemented in OSPOS and my vision of the implementation is that OSPOS, not the CMS would be the master of the data.

Please share your thoughts and objections.  Also, if anyone has ideas about the best way to fit this into the code base (hook, helper, item module, etc.) that would be good too.  I want to do this as unobtrusively as possible.

SteveIreland:
I gave a bit of thought to this awhile back and decided that it should be a two step process.

First, if OSPOS needs to have an interface to a third party package (and it could be numerous popular and obscure packages) then we just set an configuration to "Enable Third Party Interface".

This option would then write the transactions to a common file keyed by a timestamp and identified by a transaction type code.  Then based on the transaction type the body of the transaction would include JSON encoded data needed by third party packages.  Then whoever needs to write an interface could periodically pick up the data in the transaction file and write it to the external package (in the format required by the external package).

Additional complexities could be configured (such as capturing only certain types of transactions) and purging the transaction file of transactions older than what would be of interest.

But that was my thought.

jekkos:
Hmm if you want a reactive flow and have another application react to events from OSPOS then indeed it makes sense to hook another application's API call inside of OSPOS. But that will of course only work for one integration and so it would still be nice to abstract away the way events are generated.

Of course as @SteveIreland suggests it could be done through a dump but some bookkeeping will need to be done at the other side to see up to which point data was imported. However it could be rather straightforward there. This choice doesn't have the benefit that you can't react to events immediately and thus you'll have to poll a REST interface on OSPOS to feed in the data

[...]

PR #2619, opened 2019-12-09
Event generator (obsolete with CI4 Events)

objecttothis:
This is the CRUD event generator that @jekkos suggested be the driver for third-party integrations.  Per #2608 this has been implemented using the CI Events library and is pretty lean in its functionality so as not to bog the code down. The goal is to provide a layer of separation between the base code and third party integrations.

The concept is that "Triggers" are placed throughout the code (in the controllers) which trigger Create, Read, Update and Delete Methods.  Simultaneously "Listeners" receive the appropriate data and pass it off to the third-party integrations.

An example of this is that there is a trigger in the Items controller so that when a user successfully creates an item, the trigger puts the event out there with the item_data.  The event_create listener defined in the events/Integrations library then picks up the event and sends it to integrations_create defined in the same library which in turn will call any third party functions needing to use the data.  The idea is that that integrations_create function would do no heavy lifting but serve to call the third-party integrations functions only.  So, for example integrations_create($data) will look at $data[type] and when it's of type CUSTOMERS it might call, say mailchimp_add_email($data[data]) which would be defined separately.

There are a couple of question marks:
- Currently I have not added any "Read" triggers, because it was unclear to me how or if third-party integrations would use a read operation.  We can either remove the code or keep it as a place-holder for future use.  Alternately if someone can think of a good use case, I'll be glad to insert the triggers in the appropriate controllers.
- ~~Currently I have third-party integrations specific functions being created in the events/Integrations at the bottom of the class, but I'm thinking that will quickly get messy.  I'm thinking that we should extrapolate each integration into it's own file.  I was thinking that this could be done with a helper or library for each integration.  What is the best way to do this?  A helper would make the scope of each function global to the whole code, and I don't think we want that.  A library would require it be nested into its own class.  I'm not sure if that creates it's own scope problems.  Thoughts?~~ I went ahead and made changes so that each third party integration needs to be defined in its own library to create better code separation.
- This seems to be consistent with the rest of the code, but the errors on third-party integration fails are written to the log with a hard-coded English error message.  Should these be switched to toast notifications and not written in the log?  Another question is whether the error messages should be translatable regardless of whether they are toast notifications or written to the error log.  It seems that error log notifications are always in English and toast notifications are always translatable.

### DIFF
--- a/application/controllers/Customers.php
+++ b/application/controllers/Customers.php
@@ -11,7 +11,9 @@ class Customers extends Persons
 		parent::__construct('customers');
 
 		$this->load->library('mailchimp_lib');
-
+		$this->load->library('events');
+		$this->load->event('integrations');
+		
 		$CI =& get_instance();
 
 		$this->_list_id = $CI->encryption->decrypt($CI->Appconfig->get('mailchimp_list_id'));
@@ -269,24 +271,33 @@ class Customers extends Persons
 
 		if($this->Customer->save_customer($person_data, $customer_data, $customer_id))
 		{
-			// save customer to Mailchimp selected list
+		//Save customer to Mailchimp selected list
 			$this->mailchimp_lib->addOrUpdateMember($this->_list_id, $email, $first_name, $last_name, $this->input->post('mailchimp_status'), array('vip' => $this->input->post('mailchimp_vip') != NULL));
 
-			// New customer
+		// New customer
 			if($customer_id == -1)
 			{
 				echo json_encode(array('success' => TRUE,
 								'message' => $this->lang->line('customers_successful_adding') . ' ' . $first_name . ' ' . $last_name,
 								'id' => $this->xss_clean($customer_data['person_id'])));
+				$event_failures = Events::Trigger('event_create', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
 			}
-			else // Existing customer
+		//Existing customer
+			else
 			{
 				echo json_encode(array('success' => TRUE,
 								'message' => $this->lang->line('customers_successful_updating') . ' ' . $first_name . ' ' . $last_name,
 								'id' => $customer_id));
+				$event_failures = Events::Trigger('event_update', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
+			}
+			
+			if($event_failures)
+			{
+			    log_message("ERROR","Third-Party Integration failed during Customer create or update: $event_failures");
 			}
 		}
-		else // Failure
+	//Failure
+		else
 		{
 			echo json_encode(array('success' => FALSE,
 							'message' => $this->lang->line('customers_error_adding_updating') . ' ' . $first_name . ' ' . $last_name,
@@ -328,13 +339,21 @@ class Customers extends Persons
 		{
 			if($this->Customer->delete($info->person_id))
 			{
-				// remove customer from Mailchimp selected list
+			//Remove customer from Mailchimp selected list
 				$this->mailchimp_lib->removeMember($this->_list_id, $info->email);
 
+			//Event triggers for Third-Party Integrations
+				$event_failures = Events::Trigger('event_delete', array("type"=> "CUSTOMERS", "data" => $customers_to_delete), 'string');
+				
 				$count++;
 			}
 		}
-
+		
+		if($event_failures)
+		{
+		    log_message("ERROR","Third-Party Integration failed during Customer delete: $event_failures");
+		}
+		
 		if($count == count($customers_to_delete))
 		{
 			echo json_encode(array('success' => TRUE,
--- a/application/controllers/Customers.php
+++ b/application/controllers/Customers.php
@@ -280,7 +280,7 @@ class Customers extends Persons
 				echo json_encode(array('success' => TRUE,
 								'message' => $this->lang->line('customers_successful_adding') . ' ' . $first_name . ' ' . $last_name,
 								'id' => $this->xss_clean($customer_data['person_id'])));
-				$event_failures = Events::Trigger('event_create', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
+				Events::Trigger('event_create', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
 			}
 		//Existing customer
 			else
@@ -288,12 +288,7 @@ class Customers extends Persons
 				echo json_encode(array('success' => TRUE,
 								'message' => $this->lang->line('customers_successful_updating') . ' ' . $first_name . ' ' . $last_name,
 								'id' => $customer_id));
-				$event_failures = Events::Trigger('event_update', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
-			}
-			
-			if($event_failures)
-			{
-			    log_message("ERROR","Third-Party Integration failed during Customer create or update: $event_failures");
+				Events::Trigger('event_update', array("type"=> "CUSTOMERS", "data" => $customer_data), 'string');
 			}
 		}
 	//Failure
@@ -343,17 +338,12 @@ class Customers extends Persons
 				$this->mailchimp_lib->removeMember($this->_list_id, $info->email);
 
 			//Event triggers for Third-Party Integrations
-				$event_failures = Events::Trigger('event_delete', array("type"=> "CUSTOMERS", "data" => $customers_to_delete), 'string');
+				Events::Trigger('event_delete', array("type"=> "CUSTOMERS", "data" => $customers_to_delete), 'string');
 				
 				$count++;
 			}
 		}
-		
-		if($event_failures)
-		{
-		    log_message("ERROR","Third-Party Integration failed during Customer delete: $event_failures");
-		}
-		
+
 		if($count == count($customers_to_delete))
 		{
 			echo json_encode(array('success' => TRUE,
--- a/application/controllers/Items.php
+++ b/application/controllers/Items.php
@@ -7,19 +7,21 @@ class Items extends Secure_Controller
 	public function __construct()
 	{
 		parent::__construct('items');
-
+		
 		$this->load->library('item_lib');
+		$this->load->library('events');
+		$this->load->event('integrations');
 	}
-
+	
 	public function index()
 	{
 		$this->session->set_userdata('allow_temp_items', 0);
-
+		
 		$data['table_headers'] = $this->xss_clean(get_items_manage_table_headers());
-
+		
 		$data['stock_location'] = $this->xss_clean($this->item_lib->get_item_location());
 		$data['stock_locations'] = $this->xss_clean($this->Stock_location->get_allowed_locations());
-
+		
 		// filters that will be loaded in the multiselect dropdown
 		$data['filters'] = array('empty_upc' => $this->lang->line('items_empty_upc_items'),
 			'low_inventory' => $this->lang->line('items_low_inventory_items'),
@@ -28,10 +30,10 @@ class Items extends Secure_Controller
 			'search_custom' => $this->lang->line('items_search_attributes'),
 			'is_deleted' => $this->lang->line('items_is_deleted'),
 			'temporary' => $this->lang->line('items_temp'));
-
+		
 		$this->load->view('items/manage', $data);
 	}
-
+	
 	/*
 	 Returns Items table data rows. This will be called with AJAX.
 	 */
@@ -42,11 +44,11 @@ class Items extends Secure_Controller
 		$offset = $this->input->get('offset');
 		$sort = $this->input->get('sort');
 		$order = $this->input->get('order');
-
+		
 		$this->item_lib->set_item_location($this->input->get('stock_location'));
-
+		
 		$definition_names = $this->Attribute->get_definitions_by_flags(Attribute::SHOW_IN_ITEMS);
-
+		
 		$filters = array('start_date' => $this->input->get('start_date'),
 			'end_date' => $this->input->get('end_date'),
 			'stock_location_id' => $this->item_lib->get_item_location(),
@@ -58,15 +60,15 @@ class Items extends Secure_Controller
 			'is_deleted' => FALSE,
 			'temporary' => FALSE,
 			'definition_ids' => array_keys($definition_names));
-
+		
 		// check if any filter is set in the multiselect dropdown
 		$filledup = array_fill_keys($this->input->get('filters'), TRUE);
 		$filters = array_merge($filters, $filledup);
-
+		
 		$items = $this->Item->search($search, $filters, $limit, $offset, $sort, $order);
-
+		
 		$total_rows = $this->Item->get_found_rows($search, $filters);
-
+		
 		$data_rows = array();
 		foreach($items->result() as $item)
 		{
@@ -76,19 +78,19 @@ class Items extends Secure_Controller
 				$this->_update_pic_filename($item);
 			}
 		}
-
+		
 		echo json_encode(array('total' => $total_rows, 'rows' => $data_rows));
 	}
-
+	
 	public function pic_thumb($pic_filename)
 	{
 		$this->load->helper('file');
 		$this->load->library('image_lib');
-
+		
 		// in this context, $pic_filename always has .ext
 		$ext = pathinfo($pic_filename, PATHINFO_EXTENSION);
 		$images = glob('./uploads/item_pics/' . $pic_filename);
-
+		
 		// make sure we pick only the file name, without extension
 		$base_path = './uploads/item_pics/' . pathinfo($pic_filename, PATHINFO_FILENAME);
 		if(sizeof($images) > 0)
@@ -111,7 +113,7 @@ class Items extends Secure_Controller
 			$this->output->set_output(file_get_contents($thumb_path));
 		}
 	}
-
+	
 	/*
 	 Gives search suggestions based on what is being searched for
 	 */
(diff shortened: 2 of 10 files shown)


objecttothis:
#2315, #2381 and #1680 are also relevant to the conversation.

application/controllers/Customers.php
```diff
@@ -328,13 +339,21 @@ public function delete()
...
+				$event_failures = Events::Trigger('event_delete', array("type"=> "CUSTOMERS", "data" => $customers_to_delete), 'string');
+				
 				$count++;
 			}
 		}
-
+		
+		if($event_failures)
+		{
+		    log_message("ERROR","Third-Party Integration failed during Customer delete: $event_failures");
```
jekkos: I guess logging error messages in plain English is fine. Not sure but maybe there is a different method that directly logs the message as error (log_error or sth) so you can use one arg only?

jekkos: Wouldn't it make sense to handle the failure in the trigger call itself? I mean log the message there if a call to a specific listener fails?

It would mean less code duplication and a more detailed context ( the listener that failed the update ). Maybe you can extend the events library to do this logging or sth ( not sure how or it works exactly ).

application/controllers/Items.php
```diff
@@ -432,62 +434,62 @@ public function attributes($item_id)
...
 				$link_value = $this->Attribute->get_link_value($item_id, $definition_id);
 				$values['selected_value'] = (empty($link_value)) ? '' : $link_value->attribute_id;
 			}
-
+			
 			if (!empty($definition_ids[$definition_id]))
 			{
 				$values['selected_value'] = $definition_ids[$definition_id];
 			}
-
```
jekkos: Not sure if indentation is as it should be here? I think we use tabs but maybe double check it wasn't changed to spaces or sth.

application/controllers/Items.php
```diff
@@ -613,34 +615,49 @@ public function save($item_id = -1)
...
+				}
+				else
+				{
+				    $event_failures = Events::Trigger('event_update', array("type"=> "ITEMS", "data" => $item_data), 'string');
+				}
+				
+				if($event_failures)
+				{
+				    log_message("ERROR","Third-Party Integration failed during item save: $event_failures");
+				}
```
jekkos: Same remark as before, might not be the best place to log errors.

application/events/Integrations.php
```diff
@@ -0,0 +1,168 @@
...
+	 */
+	public function __construct()
+	{
+		//Third-Party Integrations by adding all code related to the integration into a library and referencing the library here with $this->ci->load->library();
+		$this->CI =& get_instance();
+		//$this->CI->load->library('LIB_NAME_HERE');
+		
+		Events::register('event_create', array($this,'integrations_create'));
+		Events::register('event_read', array($this,'integrations_read'));
+		Events::register('event_update', array($this,'integrations_update'));
```
jekkos: Can't we make the events more specific? Like use items_update so that we don't need to do the huge switch a bit further on?

Or does that imply changing a lot of code?
Edit: the amount of events could explode so might not be the best idea.

application/events/Integrations.php
```diff
@@ -0,0 +1,168 @@
...
+	 */
+	public function integrations_update(array $data)
+	{
+		$results = NULL;
+		
+		//calls to update functions for the different Integrations
+		switch($data['type'])
+		{
+			case 'ITEMS':
+				break;
```
jekkos: Maybe we can use a convention in naming the integration controller with the entity it will update and then lookup the file and call that method from here. Then we can cleanly keep related logic together in one file.

application/controllers/Customers.php
objecttothis: > I guess logging error messages in plain English is fine. Not sure but maybe there is a different method that directly logs the message as error (log_error or sth) so you can use one arg only?

According to https://codeigniter.com/user_guide/general/errors.html there is just log_message.  There is show_error but that displays it to the screen.

objecttothis: > 
> 
> Wouldn't it make sense to handle the failure in the trigger call itself? I mean log the message there if a call to a specific listener fails?
> 
> It would mean less code duplication and a more detailed context ( the listener that failed the update ). Maybe you can extend the events library to do this logging or sth ( not sure how or it works exactly ).

The error message needs to be specific to where it failed, so if I refactor the error message inside the event (e.g. delete), then the error tells me information about the failure, but it doesn't tell me what kind of operation.  This is because multiple Controllers use the same events.  For example both Suppliers and Customers controllers have event_delete triggers, so if an error gets kicked back, it will only tell you that person so-and-so couldn't be deleted, but it won't tell you which controller called it.  Having the error where it is will kick a log message that tells you that it was the Customer Delete operation that person so-and-so couldn't be deleted.  But perhaps I didn't fully understand what you were suggesting, so let me know if you meant something else.

application/events/Integrations.php
objecttothis: > 
> 
> Can't we make the events more specific? Like use items_update so that we don't need to do the huge switch a bit further on?
> 
> Or does that imply changing a lot of code?
> Edit: the amount of events could explode so might not be the best idea.

Currently there are four different event types (one for each CRUD operation).  event_create, for example, is used by five different controllers, so refactoring the events out to more specific naming does make the number of events go up significantly and code re-use go down.  You wouldn't need the switch statements then, but unless I'm not looking at it right, you end up with more code instead of less.

> Maybe the switch can be made generic by dispatching to a method in the integrations controller the we define using a convention? Like if you want to hook items, you place the controller in a folder integration/Items.php and redefine the method you want to hook. That way you know to which method you can dispatch to from here.

Sorry, I don't completely understand.  Right now the switch has calls to each integration function based on the type.  For example, for the CLCdesq API I'm writing, in the 'ITEMS' switch I have $results = $this->CI->clcdesq_integration_lib->new_product_push($data['data']); which is defined in libraries/Clcdesq_integration_lib.php.  Can you clarify what you think should change?

application/events/Integrations.php
objecttothis: > 
> 
> Maybe we can use a convention in naming the integration controller with the entity it will update and then lookup the file and call that method from here. Then we can cleanly keep related logic together in one file.

Right now I have each integration in a library rather than a controller. The naming convention I'm using currently is [integration name]_integration_lib.php and as you've suggested, all member functions related to that integration need to remain in that class.  Do you think we should use a different naming convention?

daN4cat: Not sure library is the correct format and location. This is an extension and libraries are more something you use in your code.

objecttothis: > 
> 
> Not sure library is the correct format and location. This is an extension and libraries are more something you use in your code.

So, should it be a controller, as @jekkos referred to?

application/controllers/Items.php
objecttothis: > 
> 
> Not sure if indentation is as it should be here? I think we use tabs but maybe double check it wasn't changed to spaces or sth.

Not sure how those got inserted.  Fixed.

application/events/Integrations.php
jekkos: My idea was to organize the integration files bt grouping together the logic in files that have the same name as the file of the method it is triggered by.

Putting all logic in one library file will probably make the file too big.

You know which files you need to check for when you enter the switch statement.

application/controllers/Customers.php
jekkos: Don't you already have all that context in the events lib? You have the switch case there which allows you to interpret the arguments using the correct context (whether it is a person, supplier)? You can reinterpret the argument accordingly.

It just feels cleaner as you will have a more generic way of handling the exception and the logic is where the call is dispatched.

Otherwise we need to add those if statements in every file separately.

application/events/Integrations.php
jekkos: It could be a controller or maybe just a class that extends a basic crud interface. I think we already have such an interface but not sure if it suits our need here.

daN4cat: Extend by calling an interface. In the old days that was called "callback".

objecttothis: > 
> 
> Extend by calling an interface. In the old days that was called "callback".

This is what I love about my job.  I get to learn stuff all the time.  I haven't touched interfaces since abstract classes with C++ in college, so I'll need to do some research and dive into it.  I admit that I have much to learn about software design, so I appreciate you and @jekkos helping with the design elements of this piece.  I'm glad to do the heavy lifting, but I might have some questions.  I don't see that we are using interfaces in other parts of the code here (at least my grep search for the interface keyword didn't return anything), so I'll need to look for example code elsewhere.  If you happen to have articles/examples you think translates well to what we are doing here, please feel free to send them along.

objecttothis: BTW if anyone has a favorite software design text, article or otherwise, please feel free to send me the title/url.

daN4cat: I'll lookup something over the weekend. This is the last week of work for the year so busy with social events and closing tasks.

jekkos: Check the bottom of the `Secure_controller.php`. It contains the abstract definitions of save (write), view (read), search ( read multiple). I think we might just want to take that out again and move out to an interface in its own file. If the integrations 'controller' then implements this interface, we can find the right method to call from the events lib switch case. We might need to do a manual include and then cast to the interface before the call itself. Not sure if CI offers a better mechanism there, but this is already sth 'external' to their framework and we need to find the file anyway so I think this will be just fine.

jekkos: Glad to help a bit on this one btw, if there's still stuh unclear I'll try to add some example code of what I have in mind for this `dispatch` thingy.

objecttothis: > 
> 
> Glad to help a bit on this one btw, if there's still stuh unclear I'll try to add some example code of what I have in mind for this `dispatch` thingy.

@jekkos, thank you.  The thing that is most fuzzy for me right now is what the file/data structures of the integrations need to be.  Currently each integration is confined to its own library file in libraries/[INTEGRATION]_lib.php, but it sounds like perhaps each integration should have it's own Controller/Model pair?  If this is the case, should those be stored in the Controllers/Models directories respectively or should we come up with something more appropriate

The secondary part of this is the callbacks to determine what kind of CRUD operation it is (item, customer, etc) and tertiary would be error handling in a more elegant fashion to reduce code duplication.  In short, any of those areas that you want to help provide some more detail or code snippets, feel free.  I just learned about Github gists.  It might be easier to provide code snippets in a gist rather than in the comments here.

jekkos: Well we could move this integration logic per enitity to some file that I'd call a 'controller hook'. I would like to write out a skeleton but as i don't have a decent keyboard at hand I fear this will be a bit hard.

We won't need new models as we will get all the data needed passed in through the call to the hook. What you will need to do is write some file lookup logic in the eventlib switch case that based on the case it is in

* Finds all integration controller hook files for that entity and calls `include($path)` 
* The $path can be built by looking for `integrations/*/$enitiy.php`. the wildcard is a globbing pattern and there is probably a way in php to find all files matching that pattern.
* Finds the method to call ( write, update, ...) in that loaded file (class) by casting it to the crud interface
* Calls the retrieved method within that
Casted object with the data that was passed from the original call in the controller ( could be an array with updated values )
* Catches any exception and log the message returned

application/controllers/Customers.php
jekkos: @objecttothis still following? Or should I wrap up some code to show how this could work?

objecttothis: > @objecttothis still following? Or should I wrap up some code to show how this could work?

Sorry, I'm out of the office until January 6th. Feel free to post a gist of what an example would be and I'll do the grunt work from there.  Otherwise I'll try to figure it out when I get in on the 6th.

jekkos:
I thought about this a bit further and think that the way we should try to make this code as modular as possible. When you add a 3rd party integration, you will need to sync with all data you already have in the ospos db. That's why we should use a publisher / subscriber pattern here. Currently I see following uses

* A publisher that reads existing records from the database and dispatches this data as an 'event' through the defined CRUD hooks. This one will allow you to sync existing data to an integration 'connector'. You'll actually replay all past events through this publisher by 
* Next we have the same events being fired from our controllers, meaning that once something is changed by a user in ospos, we can immediately dispatch those changes to the 3rd party integration.
* Also how should we hande an event dispatch failure? I guess that we will have to stop the sync at that point as our data could become inconsistent. All messages should be dispatched in order but the process should keep track of what was processed and ideally be able to resume from where it left of.

jekkos:
Same idea for the mailchimp integration: it's failry limited at this time and will only add customers to mailchimp that are created in ospos, if you have the integration setup at that time. It seems there is no way of syncing data that is already there, which is a pity. So once we have the groundwork done here this should be reworked to use the plugin structure.

daN4cat:
@jekkos yep, it's push only and not bidirectional syncing. The syncing requires comparisons and policies about what to do in conflicting scenarios.

objecttothis:
Agreed. The event generator does not include an integration listener. So that you all know, this is high on my priority list, but I came down with a fever on Monday and am waiting for COVID-19 test results. I'm almost back to 100% and since I work from home, I will likely be back at it tomorrow. The bad news is that the crap has hit the fan and I am trying to get my staff geared to work from home.

odiea:
Sorry to hear that. Get well soon.

application/controllers/Customers.php
objecttothis: Changes made.  The error handling is now done in the trigger handler as suggested.

objecttothis:
> 
> 
> I thought about this a bit further and think that we should try to make this code as modular as possible. When you add a 3rd party integration, you will need to sync with all data you already have in the ospos db. That's why we should use a publisher / subscriber pattern here. 
I apologize, I agree that the 3rd party integrations should be removed from the core code, which is why I wrote the code using event generators to trigger events.  The disadvantage of this method rather than a completely modular design is that this design requires the completion of 3rd-party integrations before the main application completes.  I can see long-running integrations, slowing the application down significantly.  A design that triggers an integration to run a playback of events would get around that, but it is out of my league to design and I would need some help with the code design.

> Currently I see following uses
> 
>     * A publisher that reads existing records from the database and dispatches this data as an 'event' through the defined CRUD hooks. This one will allow you to sync existing data to an integration 'connector'. You'll actually replay all past events through this publisher one-by-one.
I like the idea, but will need help with the design.

>     * Next we have the same events being fired from our controllers, meaning that once something is changed by a user in ospos, we can immediately dispatch those changes to the 3rd party integration.
> 
>     * Also how should we hande an event dispatch failure? I guess that we will have to stop the sync at that point as our data could become inconsistent. All messages should be dispatched in order but the process should keep track of what was processed and ideally be able to resume from where it left of.
The current method I am taking to handle integration failures is to report but not prevent the core application from processing the data.  The downside to this is that it potentially means stale data in the integration side.  We could potentially run the trigger as a transaction and fail the operation if the integration fails.  This would involve reworking the trigger location itself since I believe most or all are placed in the various controllers only after the application completes its work successfully, but not during its processing of the data.

Short form is that the biggest help I need is that we would reach a decision on exactly how the event generator should work, exactly how the design should be implemented (e.g., what goes in libraries, controllers, etc).

RFC: @jekkos @daN4cat

application/events/Integrations.php
objecttothis: > 
> 
> My idea was to organize the integration files by grouping together the logic in files that have the same controller name as the controller of the method it is triggered by.
> 
> So in this case we'd have sth like `integrations/clcdesq/Items.php` that contains all item crud related logic for that integration. You can then just include all Items.php in every subfolder of the `integrations/` folder and have the controller implement an interface that you call from the event lib.

@jekkos I think I see what you are getting at, but depending on the integration, it may be running the same or similar operations against multiple controllers.  For example, my CLCdesq integration on a Create operation (CRUD) builds a RESTful API request that utilizes operations against Items, Attributes, Suppliers and Config Controllers to gather all the data.   Right now the Class for my CLCdesq integration is sitting at 822 lines of code which is half of the mammoth Reports Controller.

> Putting all logic in one library file will probably make the file too big.
> 
> You know which files you need to check for when you enter the switch statement (Items.php, Suppliers.php, etc)

I don't know how much time you have on your hands, but I could use some help stubbing out what your dream is for the organization of this.  I want it to be well designed, and I get the sense you have more experience with it, but I also want to make sure I understand the implications of what you are proposing.

jekkos:
I was checking whether it would not be better to  integrate an existing library to take care of this
example: https://eventsauce.io/

jekkos:
Intereseting read: https://martinfowler.com/eaaDev/EventSourcing.html

objecttothis:
> 
> 
> I was checking whether it would not be better to integrate an existing library to take care of this
> example: https://eventsauce.io/

The code in this PR is based off the existing library https://github.com/nathanmac/CodeIgniter-Events.  I'll take a look at Event Sauce.

jekkos:
I think the eventsauce looks like it has some more functionality like support for testing, exception handling, event persistence, mesage generation from definition.. 
it feels like if we use something like this we don't need to reinvent the wheel as the current lib seems rather limited in that regard (just dispatching events but no clear way on how to do these things). Of course the specific integration with CI will remain (in this case wiring the 3rd party controllers as consumers) and handling event registration.

objecttothis:
> 
> 
> I think the eventsauce looks like it has some more functionality like support for testing, exception handling, event persistence, mesage generation from definition..
> it feels like if we use something like this we don't need to reinvent the wheel as the current lib seems rather limited in that regard (just dispatching events but no clear way on how to do these things). Of course the specific integration with CI will remain (in this case wiring the 3rd party controllers as consumers) and handling event registration.

Agreed that EventSauce seems to be more robust.  A few questions come to my mind:
- What mechanism do 3rd party plugins need to use to run?  The current implementation triggers the event with the data passed to the appropriate integrations.  I skimmed the article that you sent and it appears that each 3rd party integration will need a parser to read the event log, but admittedly, I'm fuzzy on this, because it's new to me.
- Will the event logs give enough information for 3rd party integrations to do their work?  If the event log simply tells the integrations that an item was sold, without item_ids, sale amounts, etc, that could prove problematic.
- Are you available to help get this running?  I'm fine with getting my hands dirty, but I suspect this will go a lot slower if I need to implement all of it because of the learning curve.  
- The cost overhead in terms of memory and time for the current implementation is fairly low, because the data is being passed to the 3rd party integrations.  Does EventSauce add significant "go get it yourself" overhead against the database for the 3rd party integrations?

jekkos:
Ok I think the next step would be to map out what we have in the PR here against what this library offers, as you suggest.

Basically this means redefining the events in this eventsauce DSL or yaml file and then running a code generator to see how it comes out. Once that is done we need to have some callbacks that push these events to the 3rd party API, and map them out to the correct fields on that side.

But I think that having event persistence could be an important one here, if it comes out of the box then that is quite a leap forward. As once an exception occurs, you know exactly up to which point you are synchronized, and can choose to resolve the error before you start a replay from that snapshot.
I'll see if I can free up some time maybe later next week to take a closer look. I can't make any hard promises there though, but I'll try my best.

objecttothis:
> 
> 
> Ok I think the next step would be to map out what we have in the PR here against what this library offers, as you suggest.
> 
> Basically this means redefining the events in this eventsauce DSL or yaml file and then running a code generator to see how it comes out. Once that is done we need to have some callbacks that push these events to the 3rd party API, and map them out to the correct fields on that side.

sounds reasonable.  

> 
> But I think that having event persistence could be an important one here, if it comes out of the box then that is quite a leap forward. As once an exception occurs, you know exactly up to which point you are synchronized, and can choose to resolve the error before you start a replay from that snapshot.

Agreed, but it's unclear to me if each integration will have it's own log to draw on or if not, how one integration is supposed to keep track of what it has processed and what it hasn't.  e.g., integration A processes 10 of 20 events in the logfile does integration B still have those first 10 to process.  Also, does each integration need to have a listener?  Currently the integration is kicked off by the event generator, but event sourcing sounds more like eventsauce just writes the persistent log and the integrations are on their own.

> I'll see if I can free up some time maybe later next week to take a closer look. I can't make any hard promises there though, but I'll try my best.

Thanks.  My CLCdesq API 3rd party integration PR is running off this event manager, so merging it is dependent on this being in place.  I believe I understand the rough idea of Event sourcing, but my hesitation here is 25% not wanting to spend a bunch of time on another proof of concept if we don't know we are going to use it and 75% the learning curve.  Hence asking for your assistance with getting the big pieces in place and I can probably take it from there.  

If our bookstore didn't have the quarantine absolutely destroying our sales causing me to rush getting the API code written (CLCdesq API PR), I would have a go at it with just the understanding that this might just be learning a new tool, but I've got a crazy backlog of "urgent".

objecttothis:
To clarify, clearly changing this component means I would adapt the CLCdesq API 3rd party integration.

objecttothis:
> 
> 
> I see what you mean, you might want to have something minimal working first in that case so you can start using it already.

This PR as it stands now works.  I'm using it in #2754. 

> I understand the urgency and it's probably more important then to have a nicely designed framework for doing integrations.

I 100% agree that a nicely designed framework for integrations is absolutely important.  It sounds like EventSauce fits that description much better than the PR I have here.  We can take one of two routes.  

1. We can merge this PR as it is, and then rip it out and replace it when we get EventSauce in place.  The 3rd party integration code itself ideally wouldn't be affected anyway.
2. We can not merge this PR and I can do a rebase dance to the master, this PR and #2754 then use #2754 in production until we can get EventSauce running.

Thoughts?

> The integration you talk about is not something in this PR here I suppose (unless I missed that)? In that case I'd think it'd be just the mapping of these 'events' to the API that you are integrating?

#2754 is the 3rd party integration that I referred to.  I have just rebased that branch onto this one to make it work.

PR #2624, opened 2019-12-16
onclick selection item

andriux1990:
It frustrates me a lot to have to be double clicking to be able to select all the content when changing the price or quantity.

a simple but useful change

### DIFF
--- a/application/views/customers/form.php
+++ b/application/views/customers/form.php
@@ -73,6 +73,7 @@
 									'name'=>'discount',
 									'id'=>'discount',
 									'class'=>'form-control input-sm',
+									'onClick'=>'this.select();',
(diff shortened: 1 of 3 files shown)


WebShells:
@andriux1990 can you please add the line on all forms, clients, suppliers, etc...
So we can merge all at once ? 
Thanks, all prefers a single click rather than 2, somehow it's frustrating.

andriux1990:
ok I'll do it now

andriux1990:
I think onclick is only necessary in these places

jekkos:
Indeed this code is already in place in the sales register. I'll merge this one.

daN4cat:
Thanks @andriux1990 for the contribution!

PR #2656, opened 2020-01-03
use github raw urls

jawadhoot:
curl fetches html content for non-raw urls


jekkos:
Indeed that looks more like it. Thanks for the fix!

daN4cat:
Thanks @jawadhoot for the contribution!

PR #2660, opened 2020-01-04
Add payments to emailed invoice and fix issue with unsupported fonts

SteveIreland:
This is the fix for adding payments to emailed invoices (provided by @odiea and I hope I didn't fat finger anything).  I'll try to do some testing on it Saturday.  This also includes applies a consistent font family assignment of "DejaVu Sans, Helvetica, Sans-Serif" to the invoice email style sheet.

### DIFF
--- a/application/views/sales/invoice_email.php
+++ b/application/views/sales/invoice_email.php
@@ -6,12 +6,13 @@
 </head>
 
 <body>
+
 <?php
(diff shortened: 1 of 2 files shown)


odiea:
I enjoy learning from all of you. I like to see the proper way of doing things even if they are very subtle. Thanks for the PR.

Issue #2653, opened 2020-01-02
"Allow sales price change" for employees in sales


andriux1990:
So that an employee in ospos does not see the items in the item module but at the same time can create a product from the sales module, it is done in this way
![Captura el 2020-01-02 a las 07 06 28](https://user-images.githubusercontent.com/29256343/71668557-f9263e00-2d2e-11ea-9a17-6f3bf7471605.png)

But I have a problem here in sales. the employee can change the price of the item even if allow delete this unchecked

![Screenshot_2020-01-02 Hasba Systemsdfsdfds JEJEJE Powered by OSPOS 3 4 0](https://user-images.githubusercontent.com/29256343/71668743-b4e76d80-2d2f-11ea-9c07-cb7ac9fca343.png)
I think this "allow delete" should only be when the employee can see the item module.

I was trying if I could change this condition but I was not successful at the moment, if anyone has an idea I would appreciate it.

Thanks a lot

daN4cat:
@andriux1990 "Allow delete" means allowing people to delete a transaction. You can see the delete button when you go to Sales/Transactions and you edit a Sale. The button is visible but you can't delete the Transaction is the check box is unchecked.

andriux1990:
Well it's true, you're right. So what can be done so that this employee cannot change the original price of an item. because the discount is supposed to be for this

daN4cat:
@andriux1990 what you can do? Train your employee. To tailor generic tools to specific business processes is a lot of work.
You'll end up having tons of switches and options and nobody understanding anymore what to configure. I mean, it's already complex now. Sometimes I can't even remember what a configuration switch is supposed to do.

andriux1990:
I'm sorry, this is not what I was trying to say.

I meant that this field where the price is should not be editable. because employees can steal, charge less than the item costs and not be reflected in the discount details.

I will open a issue by expressing myself well

odiea:
@andriux1990 . You can do that if you change the following from 'items' to'config' in application/controllers/sales. I believe it is line 1114. You would not want to give an employee that permission.
		$data['items_module_allowed'] = $this->Employee->has_grant('items', $this->Employee->get_logged_in_employee_info()->person_id);

daN4cat:
@andriux1990 if the change is simple and just needs an extra permission, feel free to suggest a PR for that.

andriux1990:
[...]

PR #2662, opened 2020-01-04
disable edit price to sales employees

andriux1990:
This disabled the option so that employees cannot change the price of a product. and make discounts in the traditional way in sales

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -1084,7 +1084,7 @@ class Sales extends Secure_Controller
 			$data['payment_options'] = $this->Sale->get_payment_options();
 		}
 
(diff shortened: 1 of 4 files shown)


andriux1990:
https://github.com/opensourcepos/opensourcepos/issues/2653

daN4cat:
The permission change is not the correct one, there should be probably a newly created permission so it can be selected or not.

However the rest of the changes looked good, so I revert the permission and retained the rest and going to merge.

daN4cat:
@andriux1990 see my comment in #2653 to get an understanding of what is roughly required.
I leave to you because I can't spend too much time on this today.
You can make it for 3.3.2 if you work on it and test before submitting a PR.

Last and not least, thanks for the help with testing and suggesting changes.

Issue #2666, opened 2020-01-06
DATETIME attribute type problems


objecttothis:
### Bug

- When creating a new item, if you add an attribute that is of DATETIME type and change the value, the date reverts to 1.1.2010 as soon as that attribute loses focus.

- In the date selector 1.1.2010 is the earliest selectable date.  Dates prior to that are not selectable.  When dealing with publication dates, this is too restrictive since it's pretty common to come across book publication dates prior to 2010.

OSPOS Installation Info: | 3.3.1 - ae8450
-- | --
en-US
» GD: Enabled ✓» BC Math: Enabled ✓» INTL: Enabled ✓» OpenSSL: Enabled ✓» MBString: Enabled ✓» Curl: Enabled ✓
.Browser:  			Firefox  				.Server Software: Server 				.PHP Version: 7.3.6 				.DB Version: 5.7.26-log 				.Server Port: 80 				.OS: FreeBSD 11.2-STABLE FreeBSD  11.2-STABLE #0 r325575+c9231c7d6bd(HEAD): Mon Nov 18 22:46:47 UTC 2019       root@nemesis:/freenas-releng/freenas/_BE/objs/freenas-releng/freenas/_BE/os/sys/FreeNAS.amd64  amd64
» [application/logs:] 				 -  0775 \|    Writable ✓  				» [public/uploads:] 				 -  0750 \|   Not Writable ✗  				» [public/uploads/item_pics:] 	 				 -  0750 \|   Not Writable ✗  				» [import_customers.csv:]  				 -  0640 \|   Not Writable ✗

jekkos:
I guess it would be better to have the default date set to today instead? Check my PR for that change.

objecttothis:
Also, to clarify point 1 in the bug report.  It only reverts when losing focus to adding another attribute.  If I click another permanent field, it doesn't revert to 1.1.2010.

jekkos:
Can you try my latest commit? I removed the startDate initializer in the widget.

daN4cat:
That change will affect all the views and forms.

Can you please verify the effect on the other parts of the code using that default value in the search box?

daN4cat:
I vaguely remember that there was a reason behind that change.

objecttothis:
@jekkos, I just tried it on the demo.  Here are my results:
- The date selector is defaulting to todays date.
- With two or more attributes setup, If I create an item, add a date attribute, change the date, then add another attribute, the date reverts back to todays date.

[...]

PR #2668, opened 2020-01-07
Use today as default date in DATETIME attribute (#2666)

jekkos:


odiea:
I am using php 7.2.4. When I added the define('NOW', mktime()); it errors and states you should use the time function. I changed that to time and it appears to work ok

application/views/partial/datepicker_locale.php
```diff
@@ -3,7 +3,6 @@
 var pickerconfig = function(config) {
     return $.extend({
         format: "<?php echo dateformat_bootstrap($this->config->item('dateformat')) . ' ' . dateformat_bootstrap($this->config->item('timeformat'));?>",
-        startDate: "<?php echo to_date(); ?>",
```
daN4cat: When you open Items view, what is the default start date?
That was the reason why this was set, otherwise it selects today and you don't see the Items.

I can't test this on demo website, it doesn't look like having this version right?

jekkos: You mean the inventory datetimepicker ? It does seem to be set to the correct date range.

It's on demo now. The attribute datetime selector now opens with current date as a default.

jekkos: I added a datetime attribute so you can double check the behavior there

odiea:
I enjoyed it with the current date and filtered  out todays sales which also show up on a daily basis.

daN4cat:
I merge, if any issue we will fix.

PR #2678, opened 2020-01-12
show employee

andriux1990:
to know which employee is attending the table


andriux1990:
What happened to ospos 3.4.0?

application/views/sales/suspended.php
```diff
@@ -47,6 +48,21 @@
 					}
 					?>
 				</td>
+				<td>
+					<?php
+					if (isset($suspended_sale['employee_id']))
```
jekkos: It's not a good idea to do this performancewise as you'll have a n+1 situation here (you run one extra query for every row in the original resultset). Better would be to fetch this data in the initial query and read from memory here.

andriux1990: What would it be like then?

jekkos: Just query for the first and last name in the model or controller instead. Then you can extract the fields from the resultset in one go.

If we leave it like this then the db will be flooded with queries when the list is a bit longer.

odiea:
I don't recall if I had to do any Model changes so here is what I used to show that info.
```
                               <td>
				<?php
				$employee_info = $this->Employee->get_info($suspended_sale['employee_id']);
		               echo $data['employee'] = $employee_info->first_name . ' ' . 
                               mb_substr($employee_info->last_name, 0, 1);
				?>
				</td>
```

andriux1990:
> <td>
> 				<?php
> 				$employee_info = $this->Employee->get_info($suspended_sale['employee_id']);
> 		               echo $data['employee'] = $employee_info->first_name . ' ' . 
>                                mb_substr($employee_info->last_name, 0, 1);
> 				?>
> 				</td>

this works great too

andriux1990:
it's more simple i like it

andriux1990:
I think it is important to know who is attending a table. do we push this or cancel the pr?

application/views/sales/suspended.php
```diff
@@ -47,6 +48,21 @@
...
+					{
+						$employee = $this->Employee->get_info($suspended_sale['employee_id']);
+						echo $employee->first_name . ' ' . $employee->last_name;
```
daN4cat: It's probably simpler to just fetch this string in the controller and echo here instead of having the employer I'd and then fetch the employee name. Just one variable will do.

daN4cat:
The addition conceptually is fine, just that it's in my opinion a bit more complicated than it should and we should simplify it.

andriux1990:
I think it's very good because we don't do this

application/views/sales/suspended.php
andriux1990: It's very good

daN4cat:
I commited this change to master.

andriux1990:
> I commited this change to master.

thanks @daN4cat

PR #2701, opened 2020-01-28
cash receivings for owner

Hamzaqureshi401:
it is the module same like cashups but the difference is in this opening and closing is auto fill but someone has to create edit functionality and lang work is in complete


jekkos:
As said before, try to bundle things together and submit all at once.

hoststatic:
Nice thanks for sharing, what needs to be auto fill is the total, dues. @Hamzaqureshi401, any update have you developed it further

PR #2707, opened 2020-01-29
create 3.3.0 with cash receiving

Hamzaqureshi401:
cash receiving db


jekkos:
You should only add migrations for SQL changes.. we can't merge anything in this form.

Issue #2702, opened 2020-01-28
serious error in Inventory Summary


andriux1990:
serious error in
Inventory Summary
![DeepinScreenshot_20200128162332](https://user-images.githubusercontent.com/29256343/73310827-a71bef80-41ea-11ea-8241-871816c1aaec.png)
https://demo.opensourcepos.org/reports/inventory_summary/all/more_than_zero

jekkos:
Sorry but, what is the issue here?  I don't understand spanish well enough. Also try to adhere to the template.

odiea:
He is correct. There are no amounts shown in the Summary Totals.
![2020-01-28 16_48_39-Window](https://user-images.githubusercontent.com/10962177/73315766-337be180-41ee-11ea-84c0-a196bf366006.png)

odiea:
@jekkos If you get a chance to fix the above issue could you also fix Low Inventory so that it does not show any deleted locations. Thanks @odiea .

odiea:
I figured it out. In Models Low_Inventory I added the following below AND item_quantities.quantity <= items.reorder_level.
AND stock_locations.deleted = 0  Now it just shows the current stock location.

jekkos:
Thanks for that @odiea. I applied your fix + submitted a PR for this other issue, not sure why this was changed recently and will wait for the review before merge.

odiea:
After applying your fix. Looks Good to me. Also no logged errors now.
![2020-01-29 15_43_06-Window](https://user-images.githubusercontent.com/10962177/73404216-5cfc4200-42ae-11ea-92e6-fb99dbe69f4f.png)

andriux1990:
> I figured it out. In Models Low_Inventory I added the following below AND item_quantities.quantity <= items.reorder_level.
> AND stock_locations.deleted = 0 Now it just shows the current stock location.

this did not work for me

andriux1990:
the problem is here. in controllers / Reports on line 1630
'summary_data' => $this->xss_clean($model->getSummaryData($inputs))
you just have to replace $inputs with "report_data

there are 3 lines 'summary_data' => $this->xss_clean($model->getSummaryData($inputs)) of these in reports

daN4cat:
I need to check this over the weekend because it's not following the rest of the reports pattern somehow.

daN4cat:
Merged the fix, I close this issue.

PR #2708, opened 2020-01-29
Fix inventory summary regression (#2702)

jekkos:
@daN4cat I just found that this change fixes the report. These variables were swapped recently, any idea why that was needed? Just to make sure I'm not overlooking anything here.


jekkos:
It also seems the PR contains my work for adding unit tests, will try to rewrite the commit today.

jekkos:
@daN4cat were you able to verify this one? As it does seem to fix the list of warnings the page gives now..

daN4cat:
This fix is correct, however I don't understand why the inventory reports have been designed that way. At least I don't remember this story.
Anyway I merge as the fix is correct.

PR #2718, opened 2020-02-10
Keyboard events in sales

andriux1990:
F2 ... Search Item
F7 ... Amount tendered
F8 ... Add payment and complete sale
ESC..Cancel sale

### DIFF
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -957,4 +957,48 @@ function check_payment_type()
 }
 </script>
 
+<script>
+// Keyboard events
+$(document).keydown(function(event)
+{	
+	var mycode = event.keyCode;
+
+	//F2
+	if (mycode == 113)
+	{
+		$("#item").focus();
+		$("#item").select();
+		return;
+	}
+
+	//F7
+	if (mycode == 118)
+	{
+		event.preventDefault();
+		$("#amount_tendered").focus();
+		$("#amount_tendered").select();
+		return;
+	}
+
+	//F8
+	if (mycode == 119)
+	{
+		event.preventDefault();
+		$("#add_payment_button").click();
+		$("#finish_sale_button").click();
+		return;
+	}
+
+	//ESC
+	if (mycode == 27)
+	{
+		event.preventDefault();
+		$("#cancel_sale_button").click();
+		return;
+	}
+
+});
+
+</script>
+
 <?php $this->load->view("partial/footer"); ?>
--- a/application/views/sales/register.php
+++ b/application/views/sales/register.php
@@ -959,46 +959,25 @@ function check_payment_type()
 
 <script>
 // Keyboard events
-$(document).keydown(function(event)
-{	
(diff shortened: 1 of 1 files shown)


application/views/sales/register.php
```diff
@@ -957,4 +957,48 @@ function check_payment_type()
 }
 </script>
 
+<script>
+// Keyboard events
+$(document).keydown(function(event)
+{	
+	var mycode = event.keyCode;
```
daN4cat: I would simply do switch(event.keyCode) ... instead of all these if and returns.

It's just a numeric value.

mckaygerhard: well .. i guess @andriux1990 are not developer.. so it looks good to me! its a collaboration! so! merge please!

andriux1990: It works very well, it is true that the code is something long for what is done. but it works and is clean
it's very good for me

andriux1990: > I would simply do switch(event.keyCode) ... instead of all these if and returns.
> 
> It's just a numeric value.

**This works**

_document.body.onkeyup = function(e){
    switch(e.keyCode) {
        case 113:
            $("#item").focus();
			$("#item").select();
            break;
        case 118:
            $("#amount_tendered").focus();
			$("#amount_tendered").select();
            break;
        case 119:
            $("#add_payment_button").click();
			$("#finish_sale_button").click();
            break;
        case 27:
            $("#cancel_sale_button").click();
            break;
    }
}_

daN4cat: Yep, please commit that code to the PR.

andriux1990: already did

application/views/sales/register.php
```diff
@@ -957,4 +957,27 @@ function check_payment_type()
 }
 </script>
 
+<script>
+// Keyboard events
+document.body.onkeyup = function(e){
+    switch(event.keyCode) {
+        case 113:
+            $("#item").focus();
```
jekkos: indentation is not completely right. Also I think `<script>` has some required attributes

mckaygerhard: If for every contribution we are going to complain about it, will ever advance, I see that opensourcepos has 1.3K forks but the contributions of those are almost limited to fews to null respect that number... I understand that you have to have a quality control but ..

oh! i remenber why i dont write anymore to the wiki..  i just write it in that comment! and the wiki has a lot of broken links now!

andriux1990: > indentation is not completely right. Also I think `<script>` has some required attributes

Have you tried it yet? How would it be then?
You know a better way. please contribute...

jekkos: Let's try to keep things consistent. We usually add an attribute `type="text/javascript"` to the script tag.

daN4cat: @jekkos might be an old approach. If my memory is right I changed all of them to be <script> only but please check the code.

PR #2719, opened 2020-02-11
F9 return to sales

andriux1990:
F9 to return to sales after completing.


WebShells:
This is already enabled through go to sales after print option... But maybe dunno..

andriux1990:
> This is already enabled through go to sales after print option... But maybe dunno..

Well, it never worked for me, in any of the browsers. Does it work for you?
Also the F9 key is fine for this. because it is easier than moving the mouse

WebShells:
Well it was working fine before, maybe browser updates?

PR #2720, opened 2020-02-12
uploaded the png when I should have uploaded the ico

objecttothis:


objecttothis:
Pretty straight forward.  deleted the favicon.png as that is not needed and overwrote the favicon.ico with the correct file.

PR #2727, opened 2020-02-16
Add tax group constraint

SteveIreland:
This simple change will add a constraint on the tax_group field to force it to be unique so that duplicate tax groups cannot be added.  The update will not take place unless the tax_groups are unique.

### DIFF
--- a/application/controllers/Taxes.php
+++ b/application/controllers/Taxes.php
@@ -443,6 +443,8 @@ class Taxes extends Secure_Controller
 
 		$array_save = array();
 
+		$unique_tax_groups = [];
+
 		foreach($jurisdiction_id as $key => $val)
 		{
 			$array_save[] = array(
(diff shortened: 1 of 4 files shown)


daN4cat:
No constraint from UI perspective?
This fix will be a reactive failure not proactive at input time.

SteveIreland:
I'll work on a more user friendly version of this, but can you explain why the CI build failed?  I can't find it in the details.  Obviously I'm missing something.

SteveIreland:
This pull request has been updated with the more "user friendly" version.

daN4cat:
@jekkos I cannot obviously find how to change Travis setup. I was under the impression it was configure with a html file but I don't seem to find any.
Can you please look into the build failure? It seems related to composer/composer:latest docker image.

PR #2739, opened 2020-02-25
Implement CHECKBOX attribute type

objecttothis:
Attribute Configuration view
Attributes section of item management view
Attributes section of CSV item import


objecttothis:
@opensourcepos/org I need some help with this.  The attribute is in place.  I can create/delete the attribute type.  I can add the attribute to an item and check it.  When I click submit, it gets added to the database properly.  The bug I've been trying to fix is that if I uncheck the attribute, rather than change the attribute value to 0 for that attribute_link, it removes the attribute from the item completely.  the attribute item view logic is sound because I've I manually change the link in the database to an attribute_value that is 0 then edit the item it pops it up unchecked.  Yet, if I submit with a CHECKBOX attribute unchecked, then edit again, the attribute is no longer there.  Ideas?

jekkos:
@objecttothis I think a checkbox does not send it's value back to the server if you unselect it . The view might work in such a way that if it does not receive the value, it removes the attribute link itself.

jekkos:
A solution might be to send the checkbox value not matter what, by adding a new input field at submit time using jquery, which then has the val(`:checked` ) for that control as a value. That way the backend will process the value properly without unlinking it.

objecttothis:
> 
> 
> @objecttothis I think a checkbox does not send it's value back to the server if you unselect it . The view might work in such a way that if it does not receive the value, it removes the attribute link itself.

@jekkos thanks for the help.  I'll go that route.

application/views/attributes/item.php
```diff
@@ -16,37 +16,47 @@
     <?php echo form_label($definition_value['definition_name'], $definition_value['definition_name'], array('class' => 'control-label col-xs-3')); ?>
     <div class='col-xs-8'>
         <div class="input-group">
-
             <?php
+	            echo form_hidden("attribute_ids[$definition_id]", $definition_value['attribute_id']);
```
objecttothis: @jekkos (or @daN4cat) can you tell me what the purpose of this line is?  If I add a hidden input field above the checkbox with value 0, then the unchecked checkbox works (without any jquery), but this is inserting a hidden field with value of "" for previously unchecked checkboxes, and I think may interfere with my fix.  What I'm a little unclear on is why each attribute needs a hidden field above it with the name of attribute_ids[$definition_id] and value of whatever the value was on load since this is already prepopulated in the forms themselves.

daN4cat: Hidden usually is not shown in the UI but gets HTTP POSTed.
It's a means for instance to store an array counter.

objecttothis: Thanks @daN4cat.  I know what the hidden input form field does (doesn't show in the UI but gets submitted), but what was unclear is what ospos is doing with the field, since the input with the same name is getting submitted anyway.  It's not a big deal, I left it in and it isn't interfering with my fix to get unchecked CHECKBOX attributes to submit a value to the model.

daN4cat: Sorry to have sounded obvious, it was a shallow reading mistake.

objecttothis:
@jekkos and @daN4cat this should be ready for review and testing.  I've run through tests as follows:
- Creation of new CHECKBOX attribute
- Deletion of existing CHECKBOX attribute
- Conversion of checkbox from TEXT or DROPDOWN behavior (anything that is not empty() gets converted to 1 indicating checked).
- Unchecking retains the value in the database.
- Checking retains the value in the database.

objecttothis:
@jekkos I assume that the travis-ci/push is failing purposely so that it doesn't push the build to the demo?  travis-ci/pr builds fine.

objecttothis:
#2744 just reminded me that I need to test the CSV import to make sure that checkbox attributes are being imported properly.

jekkos:
@objecttothis I think that behavior was changed. Initially pull requests builds were deployed to bintray, as long as they were done from the project's repository.

objecttothis:
This pull request can be reviewed and merged.  The CSV import is now handling checkbox attribute types.  I also added in a change to have empty lines in CSV files skipped.

objecttothis:
@jekkos @daN4cat are either of you available to review my code and merge?

application/controllers/Attributes.php
```diff
@@ -8,124 +8,126 @@ public function __construct()
...
-
+		
 		if($this->Attribute->save_definition($definition_data, $definition_id))
 		{
 			//New definition
 			if($definition_id == -1)
 			{
 				$definition_values = json_decode($this->input->post('definition_values'));
-
+
```
jekkos: hm not sure but seems like your editor is reindenting these lines here

application/views/attributes/item.php
```diff
@@ -16,37 +16,56 @@
...
-            }
-            else if ($definition_value['definition_type'] == DECIMAL)
-            {
-				$value = (empty($attribute_value) || empty($attribute_value->attribute_decimal)) ? $definition_value['selected_value'] : $attribute_value->attribute_decimal;
-				echo form_input("attribute_links[$definition_id]", $value, "class='form-control valid_chars' data-definition-id='$definition_id'");
-            }
+				//Sends 0 if the box is unchecked instead of not sending anything.
+	            	echo form_input(array(
+		            	'type' => 'hidden',
+						'name' => "attribute_links[$definition_id]",
```
jekkos: indentation here

application/models/Attribute.php
```diff
@@ -300,109 +301,165 @@ private function check_data_validity($definition, $from, $to)
...
 			{
 				$success = TRUE;
 			}
+			else if($to_type === CHECKBOX)
+			{
+				$checkbox_attribute_values = $this->checkbox_attribute_values($definition_id);
+				
+				$this->db->trans_start();
+				
+				$query = 'UPDATE ospos_attribute_values a ';
```
jekkos: cosmetic, maybe use a more descriptive table alias here (like attribute_links and attribute_values')

application/models/Attribute.php
```diff
@@ -300,109 +301,165 @@ private function check_data_validity($definition, $from, $to)
...
+				$this->db->trans_complete();
+				
+				//To CHECKBOX
+				if($to_type === CHECKBOX)
+				{
+					$checkbox_attribute_values = $this->checkbox_attribute_values($definition_id);
+					
+					$this->db->trans_start();
+					
+					$query = 'UPDATE ospos_attribute_values a ';
```
jekkos: same here

objecttothis:
> 
> 
> Thanks for the work! Looks fine to me. Maybe just try to get rid of the indentation and then good to go.

After some searching I finally figured out how/why Eclipse was inserting the tabs.  I had already turned off the feature adding tabs to blank lines, but apparently there is another setting telling it to adjust indentation when I pasted code.  I disabled it, and no more adding tabs into empty lines.  I made a commit to fix the Attributes controller.  I'll do the same for the other files.

application/models/Attribute.php
objecttothis: done

application/models/Attribute.php
objecttothis: done

objecttothis:
Changes made.  I think all of them are cosmetic.  Let me know if there is anything else.

application/controllers/Attributes.php
```diff
@@ -68,7 +67,7 @@ public function save_definition($definition_id = -1)
 			$definition_flags |= $flag;
 		}
 
-		//Save definition data
+	//Save definition data
```
daN4cat: Comments throughout the PR have indentation issues.

objecttothis: @daN4cat can you clarify what the issues are? The changes to the comment indentation I submitted were to correct the auto-formatting my IDE did when I pasted the code.

objecttothis: > 
> 
> @daN4cat can you clarify what the issues are? The changes to the comment indentation I submitted were to correct the auto-formatting my IDE did when I pasted the code.

@daN4cat did you see my question?

PR #2741, opened 2020-02-25
Make multistage docker build

jekkos:
Refactor and restructure the docker setup a bit further

* Currently we can install / run tests using `docker-compose -f docker-compose.test.yml up`
* One unit test added as an example, still has an issue in index.php that needs to be solved before it can be ran in CI build
* Dockerfile contains build instructinos for standard, dev and test container

@daN4cat if I just start using `docker-compose up` it complains about missing env variables and it does not seem to start properly. Did it work before or do I need to go through another script?

Maybe we can also provide an easy lightweight way for a more technical person to spinup and test the app (just http and the mysql?). Current setup is more something you'd do if you want to use it like that in a production setup (however not sure where your actual db data will be stored then)


docker-compose.test.yml
```diff
@@ -1,31 +1,22 @@
...
+            target: ospos_test
+        depends_on: 
           - mysql
-        ports:
-          - "80:80"
+        container_name: ospos_test
         environment:
+          - CI_ENV=testing
+          - ENCRYPTION_KEY=
+          - MYSQL_HOST_NAME=127.0.0.1
```
daN4cat: This should stay mysql because you create that image.

daN4cat:
Normal docker requires the script in docker folder to load .env.

docker-compose.test.yml
jekkos: For some reason the setup does not work when changing this back. Also I need to add something to the setup that makes sure all migrations are executed before starting the tests (to have db unit tests run against an uptodate schema)

daN4cat: You probably need to create a startup sequence adding the dependency.

daN4cat: Right, just checked and there is the depends_on.

Weird....

daN4cat: Thought about this, maybe you need to add the network.

jekkos: yeah the dependency is one thing, it will wait for the mysql to come up and load the initial script, but we also have to do a login to start the migration itself (which I will probably do through a curl request here).

You mean the hostname might work if I add the network? Might be worth a try indeed.

daN4cat: I would try the network anyway.
However please take into account that there might be some delay before you get everything up.

jekkos: well it seems that compose sets up a default network anyway, so not sure if it's really needed. I tried to add it but didn't succeed making the containers connect to eachother..

daN4cat: Yep, there is something weird. Have you tried an earlier version of docker?

application/config/config.php
```diff
@@ -65,9 +65,7 @@
 | a PHP script and you can easily do that on your own.
 |
 */
-$config['base_url'] = ((isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') || (isset($_ENV['FORCE_HTTPS']) && $_ENV['FORCE_HTTPS'] == 'true')) ? 'https' : 'http';
-$config['base_url'] .= '://' . $_SERVER['HTTP_HOST'];
-$config['base_url'] .= str_replace(basename($_SERVER['SCRIPT_NAME']), '', $_SERVER['SCRIPT_NAME']);
+$config['base_url'] = '';
```
jekkos: @daN4cat this section was causing issues when loading the tests using docker, it crashed on  the fact that `HTTP_HOST`  did not exist. In default CI this is empty. Do you know why we changed this in the first place?

I also think it's not the right place to enforce https. This should be done in a reverse proxy instead. Typically you configure apache or nginx to redirect all http to https using 301.

application/tests/libraries/Barcode_lib_test.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+/**
+ * @backupGlobals disabled
+ */
```
jekkos: Temporary workaround for a serialization issue, I will however try to set this globally in `phpunit.xml` instead

bin/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
jekkos: standard wait strategy so the test container does not proceed until the mysql socket is available

application/tests/phpunit.xml
```diff
@@ -0,0 +1,26 @@
+<phpunit
+	bootstrap="./Bootstrap.php"
+	backupGlobals="false"
```
jekkos: This file seems to be overwritten inside of the container (once the tests files are copied inside). Will try to see if this can be remedied.

docker-compose.test.yml
jekkos: seems all good now, I think the problem was that I was mixing different mysql and mariadb in the compose files, while giving them the same name. That caused the db to fail when switching setup. I have them all aligned now (mariadb 10.3)

daN4cat: Oh good, I didn't review the complete file.

application/config/config.php
daN4cat: This is something historical, I always saw it even in PHPPOS vs 1.1 if my memory is right.
In the normal docker I needed to set FORCE_HTTPS flag to true to get it to workaround this.
Forcing a reverse proxy is not always what people need (e.g. local install).
Therefore we need to make sure this is not causing a problem.

jekkos: Indeed I don't want to force anything regarding setup, we should be as lenient as possible. I think CI contains all the logic needed to resolve the base url, unless we have a good reason to override, I think we should not redo it ourselves.

jekkos:
> Probably this is required for security purposes and to avoid URL rewriting? It's like a pinning done from the application itself.... just a thought.

That might be an issue in itself. If you need rewrite rules then you define that in apache itself. If I setup an application in a particular way then I would not expect the service to rewrite it's own url. It might cause extra issues and confusion?

daN4cat:
There was also the CLI patch that had a similar issue. Probably you need to set the Environment to be Test and then avoid that part of the code?

application/config/config.php
jekkos: fixed this one by changing the behavior if we are in testing mode (as you suggested)

jekkos:
Ok I think it's mostly done now, last issue is that the test itself does not pass. If it does then I will add the docker-compose up to travis so that also new unit tests are ran in the CI build.

daN4cat:
And where are the tests?

jekkos:
@daN4cat Build is green here! From now on we can start adding unit tests to the codebase. I'll try to add some guidelines to the dev section for this one. I think it's fine to merge it to 3.3.2 instead, as it does not change any functionality. It's purely a refactoring and build pipeline improvement, so no functional impact.

jekkos:
> Normal docker requires the script in docker folder to load .env.
It would be better to make the compose file self contained. It could read the .env file by mounting it in the container itself. That way we make it work using a simple `docker-compose up` and can get rid of those extra shell scripts in the repo.

daN4cat:
Re your last comment, how .env will be loaded for myphpadmin in docker container? The problem is to inject those environment variables to all images.

jekkos:
@daN4cat in that case you can load the file using env section in docker-compose. It will apply variable values to all containers.

 ```
ospos:
  env_file:
    - docker.env
```

daN4cat:
Yep sound like the way. I don't remember having seen that in the past.

PR #2742, opened 2020-02-26
btn...btn-sm

andriux1990:
Sometimes it is very uncomfortable to select these buttons because they are very small. This makes them look prettier and helps faster operation. 
Screenshot_2020-02-26 Farmacia Nios Powered by OSPOS 3 3 2

skip database.php. by mistake I did a push this


jekkos:
Thanks and I like that fact that you made the clickable area bigger, but personally I'm not a fan of the coloring (as the labels indicate you use colors that we normally used to indicate succes or failure of an action). It might be distracting and inconsistent. I'm not a UI designer but using a softer scheme would feel better.

andriux1990:
> Thanks and I like that fact that you made the clickable area bigger but personally I'm not a fan of the coloring (as the labels indicate you use colors that we normally used to indicate succes or failure of an action). It might be distracting and inconsistent. I'm not a UI designer but using a 60-30-10 scheme would feel better.

Ok, I'll make those changes to see how it looks

daN4cat:
Personally I prefer a minimalistic design. The existing buttons are not that difficult to click.

andriux1990:
I get it. No problem

daN4cat:
Sorry @andriux1990, I know you are putting a lot of effort in this and proposing changes but please keep going. I see good things coming and we all appreciate your contribution to the project.
Unfortunately I can dedicate little time to OSPOS, just bug fixes or something required by my wife's shop employees.

PR #2753, opened 2020-03-09
Resolve bug not updating attributes correctly

objecttothis:
Prior to this fix, if the attribute was of DECIMAL or DATE type and the action was an UPDATE, then it would try to update the attribute_value, not the corresponding attribute_decimal or attribute_date.

### DIFF
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -530,7 +530,7 @@ class Attribute extends CI_Model
 
 		if(empty($attribute_id) || empty($item_id))
 		{
-			if($definition_type == TEXT || $definition_type == DROPDOWN)
+			if(in_array($definition_type, [TEXT, DROPDOWN], TRUE))
 			{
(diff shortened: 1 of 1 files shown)


jekkos:
Ok so that meant that you weren't able to update any datetime or decimal attribute value once created?

objecttothis:
> 
> 
> Ok so that meant that you weren't able to update any datetime or decimal attribute value once created?

unless I'm reading the code wrong, updating a decimal or datetime attribute type would have updated the attribute_value, not the attribute_decimal or attribute_date.  I only found the bug because I happened to update a date attribute and used a value that already existed, so when I checked the logs while debugging my CLCdesq integration code it showed up and I thought that it was odd.  The code is creating the value correctly, but just not updating it right.

jekkos:
You're right, I didn't spot that. Thanks for the fix. I'll merge and close.

PR #2754, opened 2020-03-09
CLCdesq API

objecttothis:
This will need to be completely reworked and implemented after CI4 Event triggers have been implemented.


application/libraries/Events.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * Checks if the event has listeners
+	 *
+	 * @access	public
+	 * @param	string	The name of the event
+	 * @return	bool	Whether the event has listeners
+	 */
+	public static function has_listeners($event)
+	{
+		if (isset(self::$_listeners[$event]) AND count(self::$_listeners[$event]) > 0)
```
jekkos: you can just return this statement, no if needed

objecttothis: I've corrected this in this branch and in the event_generator branch.

odiea:
What is CLCdesq api anyway?

objecttothis:
> 
> 
> What is CLCdesq api anyway?

It's a third-party integration.  It has the event_generator for third-party integrations as a dependency.  In short, it sends JSON RESTful API calls to a proprietary web platform so that we can sync inventory with the website.  Several people have been asking for integrations similar to it for publicly available CMSs.  Those, like this will need to integration through whatever the final product of our event_generator ends up being.

odiea:
Thanks. I tried googling it and did not see much about it.

jekkos:
@objecttothis It seems this branch has some changes that are already in the import branch? Do you want me to have a look and sync it with latest master? Maybe it's better to wait until we merge your import changes.

Issue #2665, opened 2020-01-06
Colon ':' symbol not allowed in attributes or item value on Item View


objecttothis:
### Background information

OSPOS Installation Info: | 3.3.1 - ae8450
-- | --
en-US
» GD: Enabled ✓» BC Math: Enabled ✓» INTL: Enabled ✓» OpenSSL: Enabled ✓» MBString: Enabled ✓» Curl: Enabled ✓
.Browser:  			Firefox  				.Server Software: Server 				.PHP Version: 7.3.6 				.DB Version: 5.7.26-log 				.Server Port: 80 				.OS: FreeBSD chiraq.chickenkiller.com 11.2-STABLE FreeBSD  11.2-STABLE #0 r325575+c9231c7d6bd(HEAD): Mon Nov 18 22:46:47 UTC 2019       root@nemesis:/freenas-releng/freenas/_BE/objs/freenas-releng/freenas/_BE/os/sys/FreeNAS.amd64  amd64
» [application/logs:] 				 -  0775 \|    Writable ✓  				» [public/uploads:] 				 -  0750 \|   Not Writable ✗  				» [public/uploads/item_pics:] 	 				 -  0750 \|   Not Writable ✗  				» [import_customers.csv:]  				 -  0640 \|   Not Writable ✗

### Bug

In attributes where values include a colon (':'), the error checking in the Items->New Item view is throwing an error.  Yet, this symbol should be allowed.  For example, anyone needing to record the Aspect Ratio of a movie (16:9, 4:3) or Book titles with colons ("Programmers Guide the Galaxy: How to Become a Better Programmer") can't.

Is there any reason why I shouldn't change the code to allow a colon?

WebShells:
@objecttothis did you copy paste the issue template or the browser isn't rendering the code as expected? I see it's messed up

objecttothis:
@WebShells I had to copy and paste it from the browser because the copy info link isn't working.
![Annotation 2020-01-06 171830](https://user-images.githubusercontent.com/17935339/71820480-a0e19a00-30a8-11ea-81f5-95169db521fd.jpg)

WebShells:
@objecttothis clear man... That comment/question was before you opened the sys info related issue...

jekkos:
I think there is a validation message that says you can't use the semicolon? It's used as a separator for aggregating the attribute value rows in the resultset. I could use another character ofcourse but I suppose the better way would be to read out the result from a data structure instead.

[...]

PR #2760, opened 2020-03-13
Removing the restriction for colons in the name

objecttothis:
Please see #2665


objecttothis:
Is there any reason we can't do this?

jekkos:
Did you check the GROUP_CONCAT section in the queries? at a certain point I was using this semicolon as a separator there. So it might still work unless you want to split out multiple values (or maybe I reworked that already)

PR #2791, opened 2020-04-12
loginButton Update login.php

blutme:
the "loginButton" valule had no translation when changing languages.


daN4cat:
Thanks for the contribution and fix.

Issue #2788, opened 2020-04-10
Problems with Item->get_all() returning null for category


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.2.2
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Bug
My 3rd Party integration is using Item->get_all() in the item model to pull all items in the database and send them to the API (via JSON).  The problem is that unlike Item->get_info(), Item->get_all() is returning null for the category.

Here is the get_all() function:
````
	/*
	Returns all the items
	*/
[...]

PR #2792, opened 2020-04-13
Removing LEFT JOIN of supplier table

objecttothis:
per #2788 the Supplier table results from Item->get_all() and Item->get_info() are not used anywhere and causing a bug where the duplicate fields between the items table and the suppliers table (deleted and category) are returning just one value to the results.  This creates the undesired effect of expecting items.deleted and items.category but getting values from suppliers.deleted and suppliers.category.

### DIFF
--- a/application/models/Item.php
+++ b/application/models/Item.php
@@ -259,7 +259,6 @@ class Item extends CI_Model
 	public function get_all($stock_location_id = -1, $rows = 0, $limit_from = 0)
 	{
 		$this->db->from('items');
-		$this->db->join('suppliers', 'suppliers.person_id = items.supplier_id', 'left');
 
 		if($stock_location_id > -1)
 		{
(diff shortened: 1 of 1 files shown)


daN4cat:
I would be surprised if it's not really used. Are we sure it's really like that?

objecttothis:
> 
> 
> I would be surprised if it's not really used. Are we sure it's really like that?

I'm absolutely sure that the LEFT JOIN is causing the behavior. You can see it in the database. Any time you join two tables that each have an identical column name but different data, the result object or array will only contain one result value with that name.

As for whether I'm sure that the results of the query are not used anywhere else, I did a grep search for the function in the code and carefully checked each instance. None of them use 95% of what gets returned by that query, so there should be 0 repercussion from removing the left join.

@jekkos is working on a unit test for get_all() but I don't see why this can't be merged unless someone spots something I missed.

jekkos:
I did a quick search and get_all seems to be used only in the Item's controller save() method?

objecttothis:
> 
> 
> I did a quick search and get_all seems to be used only in the Item's controller save() method?

I couldn't find a reference to Item->get_all in the Item controller at all.  There are three references to a get_all method in the Controller.  Two are to the Supplier->get_all() and one is to the Tax_category->get_all().

Item->get_all() is used in the Stock_location model on line 122 and only the item_id is used there from the results.
````
			$items = $this->Item->get_all();
			foreach($items->result_array() as $item)
			{
				$quantity_data = array('item_id' => $item['item_id'], 'location_id' => $location_id, 'quantity' => 0);
				$this->db->insert('item_quantities', $quantity_data);
			}
````
It's also being used more robustly in Clcdesq_integration_lib in my code, but that's where I found the issue.  We haven't seen the issue, because nothing is actually trying to use it.

jekkos:
Ok I merge, we'll fix if something breaks.

PR #2797, opened 2020-04-17
Barcode format rework

jekkos:
This rework allows us to use barcode formats with the token system originally developed by @SteveIreland . Also an extra $price variable was added that can be used with a barcode scanner to fill in item parameters in sales and receivings directly.

Also added a couple of units tests to show how the functionality is working.

### DIFF
--- a/application/libraries/Token_lib.php
+++ b/application/libraries/Token_lib.php
@@ -88,10 +88,10 @@ class Token_lib
 			foreach($barcode_formats as $barcode_format)
 			{
 				$parsed_results = $this->parse($item_id_or_number_or_item_kit_or_receipt, $barcode_format, $barcode_tokens);
-				$quantity = (isset($parsed_results['W'])) ? (int) $parsed_results['W'] / 1000 : 1;
+				$quantity = (isset($parsed_results['W'])) ? (int) $parsed_results['W'] / pow(10, $this->CI->config->item('quantity_decimals')) : 1;
 				$item_id_or_number_or_item_kit_or_receipt = (isset($parsed_results['I'])) ?
 					$parsed_results['I'] : $item_id_or_number_or_item_kit_or_receipt;
-				$price = (isset($parsed_results['P'])) ? (double) $parsed_results['P'] : NULL;
+				$price = (isset($parsed_results['P'])) ? (double) $parsed_results['P']  / pow(10, $this->CI->config->item('currency_decimals') - 1) : NULL;
 			}
 		}
 	}
--- a/application/libraries/Token_lib.php
+++ b/application/libraries/Token_lib.php
@@ -78,6 +78,24 @@ class Token_lib
 		return $token_tree;
 	}
 
+	public function parse_barcode(&$quantity, &$price,  &$item_id_or_number_or_item_kit_or_receipt)
+	{
+		$barcode_formats = json_decode($this->CI->config->item('barcode_formats'));
+		$barcode_tokens = Token::get_barcode_tokens();
+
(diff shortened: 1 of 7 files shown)


SteveIreland:
I started to test it, but when I entered the format @WebShells gave as an example ("02{P:2}{W:6}{I:3}") into the Input Format ended up being rendered as follows.  Is this correct?
image

WebShells:
@SteveIreland I guess jekkos changed the format after the rework to "02{P:2}{W:6}{I:3}"
The old barcode format we used to set was 02(\d{5})(\w{6}) d standing for item barcode and w stands for weight/qty...

application/libraries/Token_lib.php
```diff
@@ -78,6 +78,24 @@ public function scan($text)
 		return $token_tree;
 	}
```
objecttothis, line 80: This is more a philosophical question than a statement.  What are our best practices for function comments?  I've been trying to generally use this (https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods) but I noticed that it isn't consistent throughout the code.  Should we be asking everyone to comment their non-obvious functions in PR's?

application/libraries/Token_lib.php
```diff
@@ -78,6 +78,24 @@ public function scan($text)
...
 
+	public function parse_barcode(&$quantity, &$price,  &$item_id_or_number_or_item_kit_or_receipt)
```
objecttothis, line 81: Is the reason for passing these by reference instead of a return value because there are multiple values being modified and you would have to put them into an array or object if you used return?

application/tests/libraries/Token_lib_test.php
```diff
@@ -0,0 +1,73 @@
+<?php
```
objecttothis, line 1: Sorry, another philosophical question.  Do we want to begin asking the PR submitter to have Unit Tests written when new methods are being submitted?

SteveIreland: I would hope it would only be for areas that are delicate and complicated by configuration options so that we can have confidence that when someone makes a change in those areas that the appropriate unit tests can be run to cover all possible configuration permutations.  I don't see a need to make simple changes so burdensome that we need to have unit tests for stuff that only have a single test scenario.

application/libraries/Token_lib.php
jekkos: Yes indeed that's the reason.

application/tests/libraries/Token_lib_test.php
jekkos: I think that adding unit tests really adds a lot of value, we can only do it step by step from now on but it really helps to detect regressions early on, especially in more complex and critical sections of the code (eg calculations).

I think we should really try to add more unit tests. It's just a more effective way of writing code (test driven development) and validating features in the code. Probably not everything is testable in this project, like logic in the views is probably out of scope. But queries in the datamodel, functions in libraries and helpers all seem to be good and easy testable components.

jekkos:
> I started to test it, but when I entered the format @WebShells gave as an example ("02{P:2}{W:6}{I:3}") into the Input Format ended up being rendered as follows. Is this correct?
> image

Not sure if it works as expected, I'll have to review .

application/libraries/Token_lib.php
jekkos: Well I would say better to have no comments at all (except if you really have no other choice). Code should be self explanatory where possible. Most of the time it's considered a code smell.
But for most php formatting I think we try to follow the CI guidelines.

PR #2798, opened 2020-04-19
Updating System Info & Clipboard function 

WebShells:
Updating System Info to check file permissions:
1st Case:
If all files permissions are set correctly no warnings will be displayed.
2nd Case:
If permissions weren't set correctly / But no security risks, a warning will be displayed
3rd Case:
If permissions were set correctly but causing security issues > 750, a warning will be displayed.

Clipboard function now fixed to work on Mozilla and other browsers. Text format now displayed in original format.


application/views/configs/system_info.php
```diff
@@ -76,15 +78,51 @@ function get_browser_name($user_agent)
...
+						echo '<br><font color="red"><strong>Security Vulnerability Warning </strong> <br>Permissions higher than 750 leaves this software at risk.<br>';
+					} 
+					else { 
+					    echo '<br><font color="green">No security/vulnerability risks.';
+					}
+					if(substr(decoct(fileperms($logs)), -4) > 750) {
+						echo '<br><font color="red"> => [application/logs:] is writable, but the permissions are higher than 750.</font>';						
+					}
+					if(substr(decoct(fileperms($uploads)), -4) > 750) {
+						echo '<br><font color="red"> => [public/uploads:] is writable, but the permissions are higher than 750.</font>';
```
jekkos, line 115: no translations here?

WebShells: Yes, am not done yet... Keep it open

WebShells:
image

@opensourcepos/org shall we ? merge them all under a tab called System or Config or something ? 
@jekkos @daN4cat @odiea @objecttothis @SteveIreland

application/language/en-US/config_lang.php
```diff
@@ -280,3 +279,11 @@
 $lang["config_website"] = "Website";
 $lang["config_work_order_enable"] = "Work Order Support";
 $lang["config_work_order_format"] = "Work Order Format";
+$lang["config_system_conf"] = "Setup & Conf";
```
objecttothis, line 282: @jekkos @daN4cat can confirm this or not and I might be mistaken, but for future reference, I think you only need to add the lines in the en-US translation and it will be picked up by Weblate, then added into the rest of the translations.

WebShells: @objecttothis did you have the time to check it on a local install ? Tried all the scenarios ? I guess something is missing here... If the file have 702 permissions can still be writable by world, it's a security vulnerability as well... no ? 
Maybe we should consider additional case... What do you think we should do here ?

objecttothis:
I can check it when I'm back in the office tomorrow.



application/language/en-US/config_lang.php
jekkos: yes indeed en-US is the 'source' language. So once a string is added in there, it should become available in other languages.

objecttothis: > 
> 
> @objecttothis did you have the time to check it on a local install ? Tried all the scenarios ? I guess something is missing here... If the file have 702 permissions can still be writable by world, it's a security vulnerability as well... no ?
> Maybe we should consider additional case... What do you think we should do here ?

@WebShells because your pull request is from a branch outside of this repository (your master), bintray isn't building your pull request.  Unfortunately I don't have the time to download the modified files and patch them onto a different build right now, but I will commend of the permissions question.  For those files, giving world anything besides 0 is a bad idea, so that makes it an easier check.

On some servers, PHP is running in the same group, but not the owner of the file which means that it will give a warning for 770 permission, but the server admin may choose to ignore it.  In that case, it's only a security risk if other members of that group cannot be trusted.

WebShells: I'll do the changes and create PR/push them to this repo.. Maybe that's why things arent appearing in the demo upon PR submission :/... I've been wondering since a year LOOOL

PR #2806, opened 2020-04-22
Values is a mysql keyword

objecttothis:
Corrected the name of the alias from values to vals to prevent DB errors on conversion from DROPDOWN to CHECKBOX.


objecttothis:
This should be merged before 3.3.2 since currently the conversion from DROPDOWN to CHECKBOX is broken without the fix.  I think this somehow didn't get committed when I made the changes because it worked on my development box when I tested it.

PR #2814, opened 2020-04-30
Added ORDER BY to sort various dropdowns

objecttothis:
This sorts alphabetically:
- Attributes on the Attribute->Item view
- Dropdowns for adding attributes in the Attribute->Item view
- Dropdowns for the attribute values themselves.

### DIFF
--- a/application/models/Attribute.php
+++ b/application/models/Attribute.php
@@ -138,7 +138,8 @@ class Attribute extends CI_Model
 		$this->db->where('receiving_id');
 		$this->db->where('sale_id');
 		$this->db->where('deleted', 0);
-
+		$this->db->order_by('definition_name','ASC');
+		
 		$results = $this->db->get()->result_array();
 
 		return $this->_to_array($results, 'definition_id');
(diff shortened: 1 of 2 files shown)


objecttothis:
sortorder
It's the dropdown values below that I'm trying to figure out how to sort alphabetically and can't figure out which function in the attribute model is controlling it.

objecttothis:
OK, I figured out the problem:
````
	var definition_values = <?php echo json_encode($definition_values) ?>;
	$.each(definition_values, function(index, element) {
		add_attribute_value(element);
	});
````
in /views/attribute/form.php starting on line 178, the definition_values variable is an array with the indexes set to the attribute_id and the values set to the attribute_value.  The $definition_values associative array is indeed sorted by the value, but the .each() function is processing the variable by the indexes.  I'm guessing that in assigning the definition_values variable its converting it to an indexed array and in doing so it resorts it.

objecttothis:
fixed.  I just needed to reindex the array, since the attribute_id indexes are not needed, reindexing using array_values() works great.

Issue #2368, opened 2019-02-19
Allow Category to be a Dropdown


objecttothis:
### Background information

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.0
- OSPOS git commit hash is: 
- PHP version is: 5.6.38
- MySQL or MariaDB version is: MySQL 5.7
- OS and version is: FreeBSD 11.2
- WebServer is: Apache 2.4
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### New Feature
Description and category in the items table feel like attributes to me.  Is there a reason why we shouldn't migrate them to attributes?
#851 as least seems to be requesting this for category so they can make it a dropdown.  For me at least this is important because then I can make category a dropdown so my staff can stop messing up the data.

jekkos:
[...]

PR #2816, opened 2020-05-04
Category as dropdown

objecttothis:


objecttothis:
The PR is almost ready for review.  I forgot that I need to change the logic in the Attribute form view to prevent the dropdown type from being converted to another type or renamed.

objecttothis:
@opensourcepos/org OK, all done and ready for review.  

Earlier in the conversation @daN4cat had expressed that category remain a part of the core of an item and therefore I couldn't completely rewrite the element to fully make category an attribute which could be freely changed between TEXT and DROPDOWN types.  However since the infrastructure was already in place to make a DROPDOWN and manage acceptable values through attributes, I chose to keep the backend of saving/retrieving from the database identical while borrowing from the attribute dropdown structure.  I think this is the best hybrid because it involved the least amount of code rewrite and yet kept interchangability between the traditional text input and dropdown while not losing category as a core part of the item.

After not receiving feedback on #2368 I implemented this by tapping into the Attributes->DROPDOWN type.  From the user standpoint, if the checkbox in Settings->General for 'Show Category as a dropdown is checked then Category behaves just like a DROPDOWN attribute which the exception that you cannot convert it to any other attribute type.  In the backend, it is still a part of the ospos_items table and is still stored/retrieved just like it is when it's default behavior (input form) is selected.

The drawback to its current implementation is that you can delete the category_dropdown "attribute" from the table of attributes.  Doing does not change the behavior in the item view until or unless you disable the setting.  If you then re-enable the setting it "undeletes" the category_dropdown in attribute_definitions table.  The only adverse effect of deleting the attribute is that you can't add or remove dropdown values until you disable and re-enable the feature.  I would prefer to disable the delete checkbox when the definition_id == -1 is true.

objecttothis:
This should be live on the demo, but for some reason the demo is currently spitting out a database error.  It was live, but not sure what changed.  I don't know that it's related to my changes.

jekkos:
Ok it semed to be an issue in the docker networking, the mysql container did not allow access from an ip in the local subnet. I guess that happened because the containers did not have a network defined. I added ths and it works again now. I'll have a look somewhere later today at your changes .

jekkos:
> @opensourcepos/org OK, all done and ready for review.
> 
> Earlier in the conversation @daN4cat had expressed that category remain a part of the core of an item and therefore I couldn't completely rewrite the element to fully make category an attribute which could be freely changed between TEXT and DROPDOWN types. However since the infrastructure was already in place to make a DROPDOWN and manage acceptable values through attributes, I chose to keep the backend of saving/retrieving from the database identical while borrowing from the attribute dropdown structure. I think this is the best hybrid because it involved the least amount of code rewrite that also kept interchangability between the traditional text input and dropdown while not losing category as a core part of the item.
> 
> After not receiving feedback on #2368 I implemented this by tapping into the Attributes->DROPDOWN type. From the user standpoint, if the checkbox in Settings->General for 'Show Category as a dropdown is checked then Category behaves just like a DROPDOWN attribute which the exception that you cannot convert it to any other attribute type. In the backend, it is still a part of the ospos_items table and is still stored/retrieved just like it is when it's default behavior (input form) is selected.
> 
> The drawback to its current implementation is that you can delete the category_dropdown "attribute" from the table of attributes. Doing does not change the behavior in the item view until or unless you disable the setting. If you then re-enable the setting it "undeletes" the category_dropdown in attribute_definitions table. The only adverse effect of deleting the attribute is that you can't add or remove dropdown values until you disable and re-enable the feature. I would prefer to disable the delete checkbox when the definition_id == -1 is true.

This sounds fine to me, you keep the category column in the items table but try to leverage the attributes mechanics for defining values and for populating the UI logic. It will be backward compatible in the other views (as they can keep reading from the category column). I'll try to have a more thorough look later today or tomorrow.

objecttothis:
> 
> 
> Ok it semed to be an issue in the docker networking, the mysql container did not allow access from an ip in the local subnet. I guess that happened because the containers did not have a network defined. I added ths and it works again now. I'll have a look somewhere later today at your changes .

OK, I just tested this PR in the demo and it is working as expected.  Thank you for taking a closer look at the code changes when you have time.

objecttothis:
> This sounds fine to me, you keep the category column in the items table but try to leverage the attributes mechanics for defining values and for populating the UI logic. It will be backward compatible in the other views (as they can keep reading from the category column). I'll try to have a more thorough look later today or tomorrow.

Yes, it should be completely backward compatible and reversible since none of the database information regarding the category is stored differently.

application/controllers/Items.php
```diff
@@ -227,6 +227,17 @@ public function view($item_id = -1)
 
 		$use_destination_based_tax = (boolean)$this->config->item('use_destination_based_tax');
 		$data['include_hsn'] = $this->config->item('include_hsn') == '1';
+		$data['category_dropdown'] = $this->config->item('category_dropdown');
+
+		if($data['category_dropdown'] == 1)
+		{
+			$categories 		= array('' => $this->lang->line('items_none'));
+			$category_options 	= $this->Attribute->get_definition_values(-1);
+			$category_options	= array_combine($category_options,$category_options);	//Overwrite indexes with values for saving in items table instead of attributes
```
objecttothis, line 236: If this can be done more eloquently, please let me know.  The attributes mechanism give you a key/value array with the key as the attribute_id.  Without this, the attribute_id gets stored in ospos_items.category instead of the text of the category.

jekkos:
On another note, your merge target is 3.4, so I guess it's time to finally get 3.3.2 out.

objecttothis:
> 
> 
> On another note, your merge target is 3.4, so I guess it's time to finally get 3.3.2 out.

Yes, I've been doing that on purpose.  I don't want to hold up 3.3.2 which is supposed to be a bugfix release I think.  If it's ready before 3.3.2 gets released then let's include it, but not required.

application/controllers/Items.php
jekkos: Maybe you considered it, but in items we'll see the attribute 'value' in case we opt in on the new config setting, but this means that where ever we query the category field we need to consider the state of this config option to know whether we need to fetch the attribute value instead? 

So this means we need to go over the attributes tables anyway. Maybe we can just save it then by adding a 'category' definition and then query that one the regular way if the option is set? Or would this be too much rework? Or if it's not all that simple, let me know.

Would be nice to have @daN4cat input also.

jekkos:
I didn't review thoroughly yet, but have some first comments / questions. I'll try to take a further look this week.

application/controllers/Config.php
```diff
@@ -44,13 +44,13 @@ private function _licenses()
 				if($fileinfo->getExtension() == 'version')
 				{
 					++$i;
-
```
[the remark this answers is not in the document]
objecttothis: > 
> 
> some indentation issues here

Can you clarify what you mean by indentation issues?  I thought we are not supposed to have tabs on blank lines?  I removed ones that I found.  Github shows them as removing a line that has tabs and adding a line with no tabs.  Perhaps I'm missing something here.

application/controllers/Config.php
```diff
@@ -846,71 +864,72 @@ private function _handle_logo_upload()
...
 	}
-	
+
 	public function backup_db()
 	{
 		$employee_id = $this->Employee->get_logged_in_employee_info()->person_id;
 		if($this->Employee->has_module_grant('config', $employee_id))
 		{
 			$this->load->dbutil();
-
```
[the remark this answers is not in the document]
objecttothis: I'm not sure how to make those indentation fixes not show up in the diff.  I do see how it makes it more difficult to see changes to the code itself.  I think we said we don't want empty lines with tabs in them.  Is there a different way that I should be conducting that code cleanup?

application/controllers/Items.php
objecttothis: > 
> 
> Maybe you considered it, but in items we'll see the attribute 'value' in case we opt in on the new config setting, but this means that where ever we query the category field we need to consider the state of this config option to know whether we need to fetch the attribute value instead?
Do you mean that it would be visible in the item table view as an attribute if enabled?  I set the flags as 0 and hide the flags section in the attribute/form.php view so that users cannot make it appear twice in the table.

> 
> So this means we need to go over the attributes tables anyway. Maybe we can just save it then by adding a 'category' definition and then query that one the regular way if the option is set? Or would this be too much rework? Or if it's not all that simple, let me know.
This PR does create a definition in attribute_definitions. It also creates attribute links in attribute_links for the acceptable dropdown values.  The only thing it isn't doing is creating and storing the attribute links for the value of category.  For backward compatibility that value has to be stored in items.category, so storing it also in attribute_values would be redundant and if the setting is disabled, would become stale so that you have inconsistent states if you ever re-enabled the feature.  Maybe I'm not understanding what you are suggesting so please feel free to clarify.
> 
> Would be nice to have @daN4cat input also.
Agreed

application/controllers/Attributes.php
```diff
@@ -85,7 +85,7 @@ public function save_definition($definition_id = -1)
 		if($this->Attribute->save_definition($definition_data, $definition_id))
 		{
 		//New definition
-			if($definition_id == -1)
```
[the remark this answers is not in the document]
objecttothis: > 
> 
> any reason to change this to -1? I think that's the default used throughout the application.

EXCELLENT question.  Yes, in fact there is.  So it turns out that you can do an insert in a database with autoincrement and specify the index you want.  I originally wanted 0, since, yes -1 was used in the code to specify "no definition".  The problem is that when you specify 0 as the index you want to give it, MySQL interprets that as specifying no index and it gives the definition the next index in the series.  This is why I changed the code so that definition -1 is the definition_id of category_dropdown and 0 indicates "no definition".  This only affects attributes controller and model functions.  Items is still the way it was, because that's outside the scope of this change.

application/controllers/Items.php
jekkos: Oh ok I might have missed that, I thought I read that the attribute_id was stored in the category field. Ok so you save the list of possible values in the atttribute values and then copy the selected value to the category field, using the attributes as your set of values to choose from. Makes sense.

objecttothis: > 
> 
> Oh ok I might have missed that, I thought I read that the attribute_id was stored in the category field. Ok so you save the list of possible values in the atttribute values and then copy the selected value to the category field, using the attributes as your set of values to choose from. Makes sense.

What you read was that if I don't do the array_combine, it leaves the key as the attribute_id and stores it in the items.category field.  For backward compatibility we need to store the actual value, so I essentially use the array_combine to overwrite the keys with the values.

application/controllers/Attributes.php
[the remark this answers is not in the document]
jekkos: Ok thanks for the explanation here, now let's make this clear in the code as well. Maybe you should replace the 0 with a constant defined from which we can understand this behavior?

Like define a constant for 0 in the attribute model named `CATEGORY_DEFINITION_ID`?

application/controllers/Config.php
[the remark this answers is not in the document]
jekkos: Ok fine so the file itself was wrongly formatted then?

WebShells:
@objecttothis can you please apply any change so i can test it in the demo?

objecttothis:
> 
> 
> @objecttothis can you please apply any change so i can test it in the demo?

OK, just pushed a commit.  Travis-CI should build it and deploy to the demo in the next 5-10 minutes.  It should be live until/unless someone else pushes a commit.

WebShells:
@objecttothis I ran through the whole process all seems working as it should be, even after disabling the show category as drop down, category changes back to text field and the category list stays untouched.

application/controllers/Attributes.php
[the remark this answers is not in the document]
objecttothis: Super Idea.  I'll push a commit probably tomorrow sometime with constants.

objecttothis:
> 
> 
> Not sure I understand wy in certains cases definition is -1 or verified to be -1.

Here is my comment to @jekkos about the same question above.  Short form is that you can't set an index of 0 in an autoincrement column in the database or I would have used 0.  So I used 0 to indicate no definition and -1 to reference category_dropdown in the db when enabled.  Per @jekkos' request, I will be making it a constant in the code to make it more clear that this is what it's doing.

>    any reason to change this to -1? I think that's the default used throughout the application.

>EXCELLENT question. Yes, in fact there is. So it turns out that you can do an insert in a database with autoincrement and specify the index you want. I originally wanted 0, since, yes -1 was used in the code to specify "no definition". The problem is that when you specify 0 as the index you want to give it, MySQL interprets that as specifying no index and it gives the definition the next index in the series. This is why I changed the code so that definition -1 is the definition_id of category_dropdown and 0 indicates "no definition". This only affects attributes controller and model functions. Items is still the way it was, because that's outside the scope of this change.

andriux1990:
When will this be ready?
I like how it looks

objecttothis:
> 
> 
> When will this be ready?
> I like how it looks

I think I just need to submit one pull request to convert a value to a constant.  It shouldn't take me long, but It'll need to wait until the morning.

application/controllers/Config.php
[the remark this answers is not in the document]
objecttothis: OK, I've added the Global constants to the Model and referenced them in the Model and the Controller where appropriate.

objecttothis:
@jekkos I made the change you requested.  Are we good to go?

application/controllers/Config.php
```diff
@@ -305,11 +308,27 @@ public function save_general()
...
 		);
 
 		$this->Module->set_show_office_group($this->input->post('show_office_group') != NULL);
 
+		if($batch_save_data['category_dropdown'] == 1)
+		{
+			$definition_data['definition_name'] = 'ospos_category';
+			$definition_data['definition_flags'] = 0;
+			$definition_data['definition_type'] = 'DROPDOWN';
+			$definition_data['definition_id'] = -1;
```
jekkos, line 322: maybe use the constant here as well?

application/controllers/Config.php
```diff
@@ -305,11 +308,27 @@ public function save_general()
...
+			$definition_data['deleted'] = 0;
+
+			$this->Attribute->save_definition($definition_data, -1);
+		}
+		else if($batch_save_data['category_dropdown'] == 0)
+		{
+			$this->Attribute->delete_definition(-1);
```
jekkos, line 329: same thing here

application/controllers/Config.php
```diff
@@ -613,7 +632,8 @@ public function save_tables()
 				{
 					$dinner_table_id = preg_replace("/.*?_(\d+)$/", "$1", $key);
 					$not_to_delete[] = $dinner_table_id;
-					// save or update
+
+				// save or update
```
jekkos, line 636: these comments have like one tab less , not sure why

application/controllers/Config.php
```diff
@@ -305,11 +308,27 @@ public function save_general()
...
+			$definition_data['definition_name'] = 'ospos_category';
+			$definition_data['definition_flags'] = 0;
+			$definition_data['definition_type'] = 'DROPDOWN';
+			$definition_data['definition_id'] = -1;
+			$definition_data['deleted'] = 0;
+
+			$this->Attribute->save_definition($definition_data, -1);
+		}
+		else if($batch_save_data['category_dropdown'] == 0)
+		{
```
jekkos, line 328: maybe use constant here as well

objecttothis: Done

application/controllers/Config.php
objecttothis: Done

application/controllers/Config.php
objecttothis: Done

application/controllers/Config.php
objecttothis: This was a code formatting change.  I can change it back if you don't want it.  The idea behind it is that comments on the line above are regarding the block of code below them.  the indentation offset from that block helps visually encapsulate the block in the same way that { } having an indentation offset help do the same thing  For example:

````
//Here's a comment about the block below
    $variable = "foo";
    $variable_more = "bar";
    $result = $variable . $variable_more;
````
vs.
````
//Here's a comment about the block below
$variable = "foo";
$variable_more = "bar";
$result = $variable . $variable_more;
````

let me know what you want to do.

daN4cat: I spotted that too with other parts of the code.
If possible I would stick to what we have already and it was inherited from older implementations.
Therefore, yes please make all the comments to be in line with the code, even for older submissions.

Sorry for the pain of changing it but I would focus more on making the comments automatic documentation generation rather than shifting indentation here and there.

objecttothis: OK, I'll change it then merge since this was the last thing... unless anyone else has code changes they want for this PR.

PR #2828, opened 2020-05-09
Configure a serial device readout over websockets

jekkos:
@objecttothis I have briefly updated this patch as the original code also relied on another functionality that I won't probably merge in upstream (which was adding quantity 'units' to the system). I think in this version the allocation logic for deciding which input should go where is probably broken.

The idea was to have a set of predefined field identifiers in the config section that are then linked to a serial device input. You can then 'flag' those fields throughout the application by adding a html class attribute to them. A value readout was then triggered by hitting a predefined key on the keyboard.
At that point a scale readout was done and sent to this field. The readout format parsing is done through a regex and can be easily made configurable.

I'm not sure whether this can be useful for a bookstore but I think it certainly is for other types of stores. Currently I have on real usecase for it myself, so maybe I feel up to it I'll polish and test it so we can merge. Otherwise anyone else can pick it up and drive it home.

### DIFF
--- a/application/views/configs/serial_config.php
+++ b/application/views/configs/serial_config.php
@@ -1,0 +1,150 @@
+<div id="page_title"><?php echo $this->lang->line('config_serial_configuration'); ?></div>
+<?php
+echo form_open('config/save_units/',array('id'=>'serial_config_form'));
+?>
+    <div id="config_wrapper">
+        <fieldset id="config_info">
+            <div id="required_fields_message"><?php echo $this->lang->line('common_fields_required_message'); ?></div>
+            <ul id="serial_error_message_box" class="error_message_box"></ul>
+            <legend><?php echo $this->lang->line("config_unit_info"); ?></legend>
+
+			<div class="field_row clearfix">
+				<?php echo form_label($this->lang->line('config_serialport_server_url').':', 'config_serialport_server_url',array('class'=>'wide')); ?>
+				<div class='form_field'>
+					<?php echo form_input(array(
+						'name'=>'serialport_server_url',
+						'id'=>'serialport_server_url',
+						'value'=>'ws://localhost:8989/ws',
+						'type'=>'text')); ?>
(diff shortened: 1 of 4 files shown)


application/views/configs/serial_config.php
```diff
@@ -0,0 +1,150 @@
...
+
+			</div>
+
+			<div class="field_row hidden">
+				<?php echo form_label($this->lang->line('config_serialport_picker').':', 'config_serialport_picker',array('class'=>'wide')); ?>
+				<div class='form_field'>
+					<?php echo form_dropdown('input_device_picker', array(), ' ','id="input_device_picker"');?>
+					<?php echo form_dropdown('baud_rate_picker', array(), ' ','id="baud_rate_picker"');?>
+					<?php
+						foreach($item_units as $unit_id => $unit_data )
```
jekkos, line 30: I suppose this code now won't work anymore. It might need to be changed with just a `field identifier` per device. So javascript can bind the readout logic to it once it is present in a view.

objecttothis:
> 
> 
> @objecttothis I have briefly updated this patch as the original code also relied on another functionality that I won't probably merge in upstream (which was adding quantity 'units' to the system). I think in this version the allocation logic for deciding which input should go where is probably broken.
> 
> The idea was to have a set of predefined field identifiers in the config section that are then linked to a serial device input. You can then 'flag' those fields throughout the application by adding a html class attribute to them. A value readout was then triggered by hitting a predefined key on the keyboard.
> At that point a scale readout was done and sent to this field. The readout format parsing is done through a regex and can be easily made configurable.
> 
> I'm not sure whether this can be useful for a bookstore but I think it certainly is for other types of stores. Currently I have on real usecase for it myself, so maybe I feel up to it I'll polish and test it so we can merge. Otherwise anyone else can pick it up and drive it home.

I added a unit of measure to attributes, so that a DECIMAL attribute type can have the unit of definition which is specified in the attribute.  Could that be used instead?

jekkos:
@objecttothis that could be an option indeed. It really depends on where you want to process the inputs from this device. I think most of the time this will be the register or order screen.

PR #2830, opened 2020-05-12
Fix default quantity #2829

jekkos:
This should fix the issues with default quantity = 0 @andriux1990 ?


WebShells:
@jekkos this is working now in the demo version... I think it's good now.

jekkos:
Thanks for the test @WebShells

Issue #2734, opened 2020-02-20
Newly added items not displaying


bowenjnr:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.1
- OSPOS git commit hash is: c786d4
- PHP version is: 7.2.24-0ubuntu0.18.04.3
- MySQL or MariaDB version is: 5.7.29-0ubuntu0.18.04.1
- OS and version is: Ubuntu 18
- WebServer is: Apache2
- Selected language is: en-US
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.
[...]

PR #2840, opened 2020-05-27
Minor Changes & Footer Update

### LINKED ISSUE #2831: Updating Footer Info
@opensourcepos/org i'd like to change the current footer to something more simple and better formatted, also change the website link from github's project to our official website https://opensourcepos.org.

Will submit a PR with the changes, thoughts ?

Current footer: You are using Open Source Point Of Sale version 3.3.2 - 00020b. Please visit the website to learn the latest information about the project.

Suggestions: 

      · All rights reserved © 2010 - 2020 · 
 [OSPOS - Open Source Point of Sale](https://opensourcepos.org) v3.3.2 - 00020b

WebShells:
Added timezone checkup to system info

1st Case:
Timezones are the same ( no warning )
2nd Case:
Difference in timezones: Warning will be displayed.

Simplifying OSPOS footer / replacing git project link by OSPOS website.
Closes #2831 #2734


WebShells:
@opensourcepos/org this PR should get the timezone thing done and is functional now, yet still need 1 final fix not to display the timezone at the end of the file.

objecttothis:
> 
> 
> @opensourcepos/org this PR should get the timezone thing done and is functional now, yet still need 1 final fix not to display the timezone at the end of the file.

Yeah, change that and from a visual perspective it looks like it should.  I'll look at the code now.  One quick thing to note is that you are using green checks everywhere, except in the "user setup" section.  I know it wasn't the focus of this PR, but I would change those checks from grey to green to be consistent with the rest of the layout.

application/views/configs/system_info.php
```diff
@@ -46,108 +50,110 @@ function get_browser_name($user_agent)
...
-					?>
-					<br>
-					&#187; [public/uploads:]
-					<?php
-						if (is_writable($uploads)) {
-							echo ' -  ' . substr(sprintf("%o",fileperms($uploads)),-4) . ' |  ' . '<font color="green">	 Writable &#x2713 </font>';
-						} else {
-							echo ' -  ' . substr(sprintf("%o",fileperms($uploads)),-4) . ' |  ' . '<font color="red"> Not Writable &#x2717 </font>';
+						&#187; [application/logs:]
+						<?php $logs = '../application/logs/';
```
objecttothis, line 54: Let's place the "<?php" declaration on it's own line to be consistent with the rest of the code.

application/views/configs/system_info.php
```diff
@@ -2,30 +2,34 @@
...
-				<?php 
-					echo "&#187; GD: ", extension_loaded('gd') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
-					echo "&#187; BC Math: ", extension_loaded('bcmath') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
-					echo "&#187; INTL: ", extension_loaded('intl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
-					echo "&#187; OpenSSL: ", extension_loaded('openssl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
-					echo "&#187; MBString: ", extension_loaded('mbstring') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br>';
-					echo "&#187; Curl: ", extension_loaded('curl') ? 'Enabled &#x2713' : 'MISSING &#x2717', '<br> <br>';		
-				?>
+				<div id="TimeError"></div>
+				Extensions & Modules:<br> <?php
```
objecttothis, line 25: Let's place the "<?php" declaration on it's own line to be consistent with the rest of the code.

application/views/configs/system_info.php
```diff
@@ -46,108 +50,110 @@ function get_browser_name($user_agent)
...
+								echo ' -  ' . substr(sprintf("%o",fileperms($importcustomers)),-4) . ' |  ' . '<font color="red"> Not Readable &#x2717 </font>';
+							}
+							clearstatcache(); 
+							if (is_writable($importcustomers) && substr(decoct(fileperms($importcustomers)), -4) >= 751  ) {
+								echo ' | <font color="red">Vulnerable &#x2717</font>';
+							} else {
+								echo ' | <font color="green">Security Check Passed &#x2713 </font>';						
+							}	
+							clearstatcache();
+						?><br>
```
objecttothis, line 120: Let's place the "?>" declaration on it's own line to be consistent with the rest of the code.

application/views/partial/footer.php
```diff
@@ -3,11 +3,9 @@
 
 	<div id="footer">
 		<div class="jumbotron push-spaces">
-			<strong><?php echo $this->lang->line('common_you_are_using_ospos'); ?>
+			<strong><?php echo $this->lang->line('common_copryrights') . $year = date("Y"); ?> ·
```
objecttothis, line 6: Let's pass the Year to the language line as a variable rather than concatenating it.  Here's an example from items_lang.php on how to do this in the language file
````
$lang["items_csv_import_partially_failed"] = "There were %1 item import failure(s) on line(s): %2. No rows were imported.";
````
 and in the items controller that one gets called like this:
````
$message = $this->lang->line('items_csv_import_partially_failed', count($failCodes), implode(', ', $failCodes));
````

WebShells:
> > @opensourcepos/org this PR should get the timezone thing done and is functional now, yet still need 1 final fix not to display the timezone at the end of the file.
> 
> Yeah, change that and from a visual perspective it looks like it should. I'll look at the code now. One quick thing to note is that you are using green checks everywhere, except in the "user setup" section. I know it wasn't the focus of this PR, but I would change those checks from grey to green to be consistent with the rest of the layout.

@objecttothis do you think we should make the whole Enabled green ? since the check's color only won't be visible

application/views/partial/footer.php
WebShells: @objecttothis why are we doing it this way ? Whats the problem in adding it as is? Just to know

objecttothis: > 
> 
> @objecttothis why are we doing it this way ? Whats the problem in adding it as is? Just to know

Good question.  The reason to put the variables into the lang->line() function call is that different languages order words in sentences differently, which means that using a concatenation like you have it limits other languages from properly ordering words for their language.  Azerbaijani and languages like it, put the verb at the end of the sentence which cannot happen if there is a forced concatenation with something other than a verb.  Russian and languages like it are somewhat agnostic and word order matters less than endings of words.

In this particular case, we could likely get away with the concatenation, but in a Right-to-Left language such as Arabic and Japanese, are the dates flip flopped?  I don't know the answer to that, but it's a way that sending the ending year as a variable even in this case is the way it should be.

Even if there are no languages that flip-flop the dates, the reason we still want to do it here is for consistency with the rest of the code where it definitely matters.  Does that explanation help?

objecttothis: This stack exchange article (https://linguistics.stackexchange.com/questions/6287/are-hebrew-numeric-ranges-written-right-to-left-or-left-to-right) seems to indicate that Hebrew is an example of where the copyright dates should read 2020-2010 instead of 2010-2020 which you can't do without sending the end year as a variable and letting the translator rearrange the translation.

objecttothis: @WebShells did my explanation make sense?  Let me know if you get stuck.  It's a pretty straight forward change, so I'm guessing it will take you no time at all.

daN4cat:
@WebShells can you please make the last changes?

application/views/partial/footer.php
objecttothis: @WebShells do you want me to do this last change?  We are pretty far behind on releasing 3.3.2, so we need to get this last change done and merged into the master so that we can release 3.3.2.

objecttothis: I went ahead and made the change so that we can move forward with merging and getting 3.3.2 out the door.

WebShells:
@objecttothis thanks for the fixes, just saw the PR was merged :+1:

PR #2849, opened 2020-06-03
Change the locations of global constants to make them global across files

objecttothis:
- Items Constants
- Attributes Constants
- Sale Constants


objecttothis:
This is a pretty straight forward change.  CI best practices say that we should be defining Global Constants in /config/constants.php but we have them defined in three models.  The problem is that when one of those constants gets used in a different controller or model, the application doesn't know what to do with it.  The solution is to move all Global Constants to config/constants.php.

daN4cat:
Well done, good idea.

PR #2871, opened 2020-06-28
Update Item.php

objecttothis:
Set group_concat_max_len session variable to prevent count totals at the bottom of the tables becoming `1 to null of null items` when database includes significant number of item rows.


objecttothis:
@jekkos do you have time to approve this PR?  It's just adding the one-line temporary fix that we talked about.

jekkos:
Looks fine. Thanks for the f!x

Issue #2820, opened 2020-05-06
Invoice number is not visible in sales register


jekkos:
### Background information

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

test on latest master

### Issue / Bug  / Question / New Feature

When creating invoices, we used to have the invoice_number field in the view. For some reason it has disappeared and if a generated sequence clashes with one already in the db, then there is no way to correct it and complete the sale

andriux1990:
@jekkos Also the latest versions of bintray, when you add products to the shopping cart .. it adds 0. no matter how much you add since registering, the quantity always remains at 0

objecttothis:
> 
> 
> @jekkos Also the latest versions of bintray, when you add products to the shopping cart .. it adds 0. no matter how much you add since registering, the quantity always remains at 0

I'm using 3.3.2 - 9d2fd1 and not experiencing this behavior.  What steps are you taking to reproduce the problem?

I'm seeing
[...]

PR #2880, opened 2020-07-04
Fix custom invoice number (#2820)

jekkos:
I have tested the custom invoice number override in sale_invoice mode and found the following

* Custom invoice number was not set when overriding it, probably because some js code was not triggerd when the view was loaded with invoice number enabled
* For some reason we ignore custom invoice numbers with less then 3 chars. Not sure what the rationale behind this is (if someone knows feel free to comment)
* Still need to fix one issue in that after you override an invoice number once, it seems that the next time the auto generated token is not resolved anymore (it just takes the token as invoice number as-is)


SteveIreland:
I'm a little confused.  In receipt mode I don't believe it is any longer possible to assign an invoice number.   Am I wrong on that point?  I did quite a bit of testing and haven't yet found it to be possible.   I'm thinking that this section of code could actually be removed because the invoice number when in invoice mode is generated a few lines further down.

SteveIreland:
Just to clarify (I hope), the reason that segment of code isn't necessary in receipt mode is because there is no longer any user expectation of invoice numbers when in receipt mode.  I believe that was added when there still was some support for it, but when we added receipt, invoice, work order, and quote modes.

jekkos:
@SteveIreland indeed, I also realised this and have done some more cleanup here. Do we also support the override in work orders, quotes etc? As I didn't test in these modes. At first sight the code seems to ignore the invoice number override.

But I do think those type of sales have their own sequence generation no? So in a way it might make sense to enable the override for them as well?

SteveIreland:
I've been testing your changes and your changes still look very good.  However, I included some testing in the area of quotes and work orders and noticed that after the work order is cleared the automatically generated work order number needs to be cleared.  The last work order number I created was being used to update the sales record work order number field for subsequent invoices I created.  Quotes are working fine.  Just checked the code and I see there is a missing "clear_work_order_number" ins Sale_lib.php.   Must be an oversight that's been there for a looooong time.  I can take care of that 
 this week if you don't want to make this change any more complicated.

As for being able to manually override the work order number or quote number I deliberately left that out because I couldn't think of a valid use case for it.  In the States here it is frowned on to allow generating a document out of sequence.  But apparently it's a requirement by some companies elsewhere for purposes of invoicing.  No one has yet complained, so I'm good with leaving it as is.  :-)

jekkos:
Ok great thanks for the testing @SteveIreland . I have added that last clear for the work order session variable.

Your point is that usually indeed you don't need to override anything as you just want to stick to the sequence generation. Now in my case for some reason, this sequence went out of sync and then for all invoices this year, it is trying to generate a number that is already there . The override allows me to fix that every time. So in this case I still stick to this sequence as you say, but have a problem elsewhere.I think that the issue might be because I have a deleted sale with invoice number still filled in . I have checked the code and indeed, this condition was not taken into account in the query. I have updated this and will check whether this solves my issue.

In that case I won't need the invoice field anymore either..

SteveIreland:
On the configuration page we have the ability to set the "last number used" for invoice, quote, and work orders.  So if that number somehow gets out of sync then you can adjust that number to reflect the actual last number used.   I needed that yesterday when I accidentally used the QSEQ token in the invoice number field instead of the ISEQ token.

jekkos:
Ok cool thanks for the tip there, I didn't know that one. So that sort of makes the override field obsolete I guess. 

I have also reverted the check on the deleted flag as in fact it does not exist on the sale table. So the issue might have been something else. I'll upload this version to my server to give this a try now.

jekkos:
Ok I have deployed and tested the thing, it seems fine, I can override this number manually, but for some reason the sequence fix does not work for me.

PR #2885, opened 2020-07-07
Update system_info.php

### LINKED ISSUE #2860: impossible to fill field in chrome and derivatives
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

OSPOS Installation Info: 3.3.2 - d6df08

Language Code: en-US

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓

User Configuration:
.Browser: Chrome
.Server Software: Apache/2.4.38 (Debian)
.PHP Version: 7.3.14-1~deb10u1
.DB Version: 5.5.5-10.3.22-MariaDB-0+deb10u1
.Server Port: 80
.OS: Linux 4.19.0-9-amd64

File Permissions:
» [application/logs:] - 0777 | Writable ✓ | Vulnerable ✗
» [public/uploads:] - 0777 | Writable ✓ | Vulnerable ✗
» [public/uploads/item_pics:] - 0777 | Writable ✓ | Vulnerable ✗
» [import_customers.csv:] - 0777 | Readable ✓ | Vulnerable ✗

There are problems with file permissions please fix and reload this page.

Security Vulnerability Warning
Permissions higher than 750 leaves this software at risk.

» [application/logs:] is writable, but the permissions are higher than 750.
» [public/uploads:] is writable, but the permissions are higher than 750.
» [public/uploads/item_pics:] is writable, but the permissions are higher than 750.
» [import_customers.csv:] is writable, but the permissions are higher than 750.

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.

Chrome and Opera browser error. and perhaps in browsers derived from chrome.
Office -> Configuration -> General, Localization, Info ...
Cannot fill in the fields. every time the pulsor is placed it is ejected.
chrome

in demo ospos the same thing happens

Firefox works fine

jekkos:
@WebShells for you to review, but I think that removing that code will fix #2860

### DIFF
--- a/application/views/configs/system_info.php
+++ b/application/views/configs/system_info.php
@@ -173,13 +173,6 @@
 			clipboard.on('error', function(e) {
 				console.log(e);
 			});
-
-			function clearSelection() {
-				window.getSelection().removeAllRanges();
-			}
-			document.onmouseup = function() {
-				window.setTimeout(clearSelection, 100);
-			};
 			
 			document.getElementById("timezone").innerText = Intl.DateTimeFormat().resolvedOptions().timeZone;	
 					


objecttothis:
@jekkos please see my comment in #2860 before merging.  This may change how you want to handle the problem.

WebShells:
@jekkos we can reverse it, but this will keep the copied text selected after clicking "Copy Info". If it's fine this way we can use it as a fast solution until our next release. Thoughts ?

jekkos:
Just had some extra thought on the read permission 'risk'. I'm not sure whether it makes sense to say there is a risk if this file is world readable. Maybe indeed we should stick to the idea that world writable and executable files can be harmful?

Then we should just check files on which we require write to be below or equal to 750, for read this can be 750 as well.
And then only warrn that we are at risk if any permission is higher than 750.

The currentt implementation deviates from this as we warn and indicate a risk if we have more than 640 for read, which might not make sense.

daN4cat:
Definition time:

Issue -> we know it's a problem and therefore we mandate something.

Risk -> we don't know for sure it's a problem but we perceive it could and therefore we recommend something.

objecttothis:
> 
> 
> Just had some extra thought on the read permission 'risk'. I'm not sure whether it makes sense to say there is a risk if this file is world readable. Maybe indeed we should stick to the idea that world writable and executable files can be harmful?
> 
> Then we should just check files on which we require write to be below or equal to 750, for read this can be 750 as well.
> And then only warrn that we are at risk if any permission is higher than 750.
> 
> The currentt implementation deviates from this as we warn and indicate a risk if we have more than 640 for read, which might not make sense.

@jekkos, the reason I recommended the default permission of the log file is that while reading the log file itself is not an attack vector, information gained from the log file could be used to find and exploit a weakness in the code.  This would be especially true if log_level is set higher than 1.  A potential attacker could try various things and then look at the world readable log to see the results of their efforts and make tweaks.  Hence, world readable log files, in my opinion, should be considered a risk.

With that said, different people are going to have different server setups and therefore may legitimately need to have 660 permissions on their log files or have 644 permissions on upload_image files.  In my case, my upload image files need to be world readable because a 3rd party API is fed URL's to the images and it then goes and picks them up.  Unfortunately the $config['log_file_permissions'] found in config.php controls not only the permissions for log files, but for ALL files that the app places.

objecttothis:
Here is an article describing a vulnerability with nginx where it was setting log permissions to world readable by default (https://www.csoonline.com/article/2134121/nginx-sets-world-readable-logs-by-default.html). Given, webserver software will have even more importance than a webapp on not making logs world readable, but the concept is the same.

jekkos:
@objecttothis Agree that log files should not be world readable. Actually I was talking about the import template for customers. Idon't think it really matters if that one is world readable as it does not contain any sensitive information?

jekkos:
So just wanted to close on this one before pushing the release button.

objecttothis:
> 
> 
> @objecttothis Agree that log files should not be world readable. Actually I was talking about the import template for customers. Idon't think it really matters if that one is world readable as it does not contain any sensitive information?

OK, that makes sense.  Agreed that import_customers.csv template file does not contain sensitive data.  I think eventually we will likely implement people attributes which would mean generating the customer import file in the same way that we do the item import file.  That will get rid of that file entirely.

jekkos:
Agree, my point is that I don' really consider it a risk if this file is world readable.. so I'll change that in the code accordingly.

jekkos:
I don't think the current implementation is correct as it stands.

we consider a risk if a writable folder is > 750. The same for application logs. As we discussed it's also at risk if a readable folder is world-readable (in this case that can be 644 for example, which is lower then 750) 

I guess this approach of checking readable files to be greater then 750 is not sound. We need to check the last three (world) bits instead for both readable and writable folders.

jekkos:
@WebShells just want to bring this to your attention.

WebShells:
Shall we release as is and fix in the next release ? Until we agree on a different way to check the file permissions or what we need exactly... I totally agree that 644 lower than 750 and still world readable which is also an issue. Either i should add another "&" or we should change the whole thing.

objecttothis:
My take: Anything above 0 in the world octet for a log file is risk because it can give an attacker more information about the effectiveness of their attacks or alert an attacker of a vulnerability.  The group octet on the other hand is different.  In the case of shared hosting and many others, you don't want group octets higher than 4 or 5 for most files and definitely don't want 7.  However in the case that the web server is not the owner of the file, but the webserver is the one writing a log, you can't make the group octet 0 for example.  In most shared hosting, I believe, the web server instance is run as the user, so you just make the log 600 and that prevents any other users in the group (maybe another shared hosting client) or world from viewing your log files. in the case where you need to give the web server permissions to write to the log then it needs to have 660 permissions.

customer_import.csv should have 644 permissions regardless of setup.  Anything higher is a vulnerability (not just a risk) since you don't want a scenario where world can write to your customer_import.csv file.

In the end, I think if we change the code so that it's looking at the individual octets, that would be a more robust way of looking at it.  Then it's just a matter of coming to an agreement on what file permissions need to be for various files and folders and what constitutes risk vs vulnerability.

jekkos:
Ok I'll be off for the weekend, maybe we need to create a ticket for this so it can be reviewed and reworked. I think it's not blocking but what it reported felt rather confusing to me.

objecttothis:
> 
> 
> Ok I'll be off for the weekend, maybe we need to create a ticket for this so it can be reviewed and reworked. I think it's not blocking but what is reported felt rather confusing to me.

OK Sounds good.  @WebShells can you create an issue for someone to rework the world octet checking.  @jekkos does this mean we can release 3.3.2?  I assume that it would need to wait until Monday since you are out for the weekend?

WebShells:
`							echo '<br><font color="red"><strong>' . $this->lang->line('**_config_security_issue_**') . '</strong> <br>' . $this->lang->line('config_perm_risk') . '</font><br>';
						} 
						else { 
							echo '<br><font color="green">' . $this->lang->line('**_config_security_issue_**') . '</strong> <br>' . $this->lang->line('config_no_risk') . '</font>';
						}`

I guess i found it, we are using the same phrase even if there is no issue; i think permission wise there is no issue unless we agree on different permissions for the files.

PR #2821, opened 2020-05-06
Item CSV Import Update and Optimizations

objecttothis:

PR #2886, opened 2020-07-08
Quote improvements

jekkos:
Some initial improvements to use yearly variables in invoice numbers for quotes. Also I have reviewed the quote template and brought it in line with how the invoice looks.

I think one last issue here is the quote PDF generation. It does not seem to work properly with this template. i'll have look into that a bit further.

### DIFF
--- a/application/models/Sale.php
+++ b/application/models/Sale.php
@@ -454,19 +454,29 @@ class Sale extends CI_Model
 		return $this->db->get();
 	}
 
+	public function get_invoice_number_for_year($year = '', $start_from = 0)
+	{
+		return $this->_get_number_for_year('invoice_number', $year, $start_from);
+	}
+
+	public function get_quote_number_for_year($year = '', $start_from = 0)
+	{
+		return $this->_get_number_for_year('quote_number', $year, $start_from);
+	}
+
 	/**
 	 * Gets invoice number by year
 	 */
-	public function get_invoice_number_for_year($year = '', $start_from = 0)
+	public function _get_number_for_year($field, $year = '', $start_from = 0)
 	{
 		$year = $year == '' ? date('Y') : $year;
-		$this->db->select('COUNT( 1 ) AS invoice_number_year');
+		$this->db->select('COUNT( 1 ) AS number_year');
 		$this->db->from('sales');
 		$this->db->where('DATE_FORMAT(sale_time, "%Y" ) = ', $year);
-		$this->db->where('invoice_number IS NOT NULL');
+		$this->db->where("$field IS NOT NULL");
 		$result = $this->db->get()->row_array();
 
-		return ($start_from + $result['invoice_number_year']);
+		return ($start_from + $result['number_year']);
 	}
 
 	/**
--- a/application/models/Sale.php
+++ b/application/models/Sale.php
@@ -456,18 +456,18 @@ class Sale extends CI_Model
 
 	public function get_invoice_number_for_year($year = '', $start_from = 0)
 	{
-		return $this->_get_number_for_year('invoice_number', $year, $start_from);
+		return $this->get_number_for_year('invoice_number', $year, $start_from);
 	}
 
 	public function get_quote_number_for_year($year = '', $start_from = 0)
 	{
-		return $this->_get_number_for_year('quote_number', $year, $start_from);
+		return $this->get_number_for_year('quote_number', $year, $start_from);
 	}
 
 	/**
 	 * Gets invoice number by year
 	 */
-	public function _get_number_for_year($field, $year = '', $start_from = 0)
+	private function get_number_for_year($field, $year = '', $start_from = 0)
 	{
 		$year = $year == '' ? date('Y') : $year;
 		$this->db->select('COUNT( 1 ) AS number_year');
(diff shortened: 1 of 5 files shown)


application/models/Sale.php
```diff
@@ -471,19 +471,29 @@ public function get_sale_by_invoice_number($invoice_number)
...
+	public function get_quote_number_for_year($year = '', $start_from = 0)
+	{
+		return $this->_get_number_for_year('quote_number', $year, $start_from);
+	}
+
 	/**
 	 * Gets invoice number by year
 	 */
-	public function get_invoice_number_for_year($year = '', $start_from = 0)
+	public function _get_number_for_year($field, $year = '', $start_from = 0)
```
objecttothis, line 487: Maybe I'm missing something here with the change, but is this Hungarian notation (leading the function name with a single underscore)?  If so, isn't that used to denote a protected method?  Here is listed as a public method.  Let me know if I'm missing something.

jekkos: You're right, it should be private.. I'll change that.

objecttothis: @jekkos my understanding is that the leading underscore was used mostly in PHP 4 and lower because at the time there was no Private or Protected designation for functions.  Do we need to use the leading underscore in the naming convention?

objecttothis: https://stackoverflow.com/questions/663350/whats-the-deal-with-a-leading-underscore-in-php-class-methods

jekkos: Ok good point, but if we change this rule then I might need to skim over the whole codebase to have these occurences replaced? I'm in favor of using only the access modifiers in this case. We could log a technical debt ticket for this one.

objecttothis: PR #2821 is already packed with these kinds of fixes.  I can do a grep search and find functions that start with an underscore then refactor them in that PR and submit it after 3.3.2 gets released.  The PR is ready to go, but I'm waiting so that 3.3.2 doesn't get weighed down with it.

daN4cat: Leading underscore makes it consistent with Python style.
I can't remember at the moment if I added some of them to make it clear and consistent with the rest. But I remember I added the explicit public and private to improve encapsulation.

objecttothis: @daN4cat Sorry, I didn't know we had any python code in this code base. Would you be OK if I submitted a PR to bring the function naming in line with PHP PSR standards?

objecttothis: Just for proof of concept, I was able to eliminate all of the leading underscores in function and variable names except where the underscore is part of a function or variable that we are overriding in the Code Igniter framework or where the variable is a server variable (e.g., $_POST, $_SERVER, etc).  It all works correctly, so the change isn't catastrophic.  I think the reason for me pushing for the change is that it doesn't make sense to me to adhere to Python naming styles in a PHP web app that isn't using Python.  Thoughts?

SteveIreland: I agree ... for what it's worth.

daN4cat: @objecttothis sorry we don't use Python here.

daN4cat: Renaming the variables is a risk, if somehow we added a _ to copy the value of a parameter then removing the _ makes it the param.
We might not have the case but we need to review all the changes.

objecttothis: > Leading underscore makes it consistent with Python style.
> I can't remember at the moment if I added some of them to make it clear and consistent with the rest. But I remember I > added the explicit public and private to improve encapsulation.

> 
> 
> @objecttothis sorry we don't use Python here.

That's my point though.  :) Above you said that the leading underscores were added to keep things consistent with Python, but we agree that we don't have Python code in this code base, hence suggesting we refactor to remove the leading underscores.

> Renaming the variables is a risk, if somehow we added a _ to copy the value of a parameter then removing the _ makes it the param.

In the changes that I did in the development version that I'm working on I did a grep file search for every instance of "->_", " _" and "$_".  Aside from system variables, there were only two variables in the mailchimp api that removing the underscore would have caused the variable name to be the same as another variable.  In that case I didn't remove the underscore.  We could refactor the variable name entirely and it would be fine. 

Other than that the only functions I left untouched were overridden functions from CI classes (there were maybe two files).

> We might not have the case but we need to review all the changes.

Yes, most definitely.

jekkos: OK I have renamed the method. Maybe we should create a ticket for further followup on this discussion.

objecttothis: > 
> 
> OK I have renamed the method. Maybe we should create a ticket for further followup on this discussion.

I already have the proposed renaming of almost all the leading underscores built into #2821.  I've been using it for a little while now, so once 3.3.2 releases we can do more code reviews on #2821 and if we don't like the effect revert the changes.

jekkos:
@SteveIreland I still need to test it a bit further but decided to put it out here already. Specifically I need to check how the PDF looks (as I bumped dompdf library here)

jekkos:
Just got a PDF quote from the demo, looks fine in fact (using dompdf 0.8.5 now)

jekkos:
[Quote-Q20000004.pdf](https://github.com/opensourcepos/opensourcepos/files/5126488/Quote-Q20000004.pdf)

odiea:
@jekkos Did this need to be changed in the Composer.lock file also. The current demo shows it is still at 7.

jekkos:
Ah ok you might be right there. So my test was probably worthless.

odiea:
I don't think it was worthless.

jekkos:
Well it does show that the quote changes I did were fine indeed.

PR #2903, opened 2020-07-22
Update system_info.php (#2885)

jekkos:
@WebShells I had a closer look and I think this is exactly as we want it to be. Have the text deselectd after the copy button is clicked. It was a matter of adding that code to the event handler linked to that button.
I checked locally on chrome and it seems fine.

### DIFF
--- a/application/language/en-GB/config_lang.php
+++ b/application/language/en-GB/config_lang.php
@@ -294,3 +294,7 @@ $lang["config_security_issue"] = "Security Vulnerability Warning";
 $lang["config_perm_risk"] = "Permissions higher than 750 leaves this software at risk.";
 $lang["config_all_set"] = "All file permissions are set correctly!";
 $lang["config_no_risk"] = "No security/vulnerability risks.";
+$lang["config_timezone_error"] = "OSPOS Timezone is Different from your Local Timezone.";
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN echo -e “$(hostname -i)\t$(hostname) $(hostname).localhost” >> /etc/host
 WORKDIR /app
 COPY . /app
 RUN ln -s /app/*[^public] /var/www && rm -rf /var/www/html && ln -nsf /app/public /var/www/html
-RUN chmod -R 750 /app/public/uploads /app/application/logs && chown -R www-data:www-data /app/public /app/application
+RUN chmod -R 750 /app/public/uploads /app/application/logs && chown -R www-data:www-data /app
```
jekkos, line 17: Hopefully this one will render all checks ok in the docker setup

jekkos:
Following screenshot taken on demo

@WebShells it seems all security setup is ok now in docker, but still we have a warning that some permissions are incorrect. MIght be a bug there?

Screenshot from 2020-07-23 11-08-50

jekkos:
Ok I caught that last one, let's wait and see when installed on demo.

PR #2909, opened 2020-07-28
Update INSTALL.md

objecttothis:
Added support for PHP 7.4

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,7 +1,7 @@
 Server Requirements
 -------------------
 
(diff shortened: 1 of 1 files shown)


jekkos:
@objecttothis I never did any tests with PHP7.4 but guess it works fine?

jekkos:
Also I still need to change the CHANGELOG for 3.3.2

objecttothis:
I've been running it in production and development for a couple of months now.  I haven't run into any problems.

PR #2910, opened 2020-07-28
Modify input validator

objecttothis:
- Matched the allowed inputs to what is found in the validator for the
Attribute form (/views/attributes/form.php) for consistency.
- Removed the restriction on colon and added the restriction on
underscore.
- Modified the language line to indicate correct error.

### DIFF
--- a/application/language/en-GB/attributes_lang.php
+++ b/application/language/en-GB/attributes_lang.php
@@ -1,6 +1,6 @@
-<?php 
+<?php
 
-$lang["attributes_attribute_value_invalid_chars"] = "Attribute value cannot contain ':' or '|'";
+$lang["attributes_attribute_value_invalid_chars"] = "Attribute value cannot contain '_' or '|'";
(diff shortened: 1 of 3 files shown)


objecttothis:
This change is ready for review.  It's pretty straight forward.  We got rid of the limitation against colons in attribute values awhile ago in the /views/attributes/form.php but did not make the same change in /views/items/form.php.  This change is just bringing everything in line.  Without the change TEXT attributes were unable to include the colon.

jekkos:
Ok sounds good. I wanted to ask whether we already removed the separator in the query but I guess so.

objecttothis:
> 
> 
> Ok sounds good. I wanted to ask whether we already removed the separator in the query but I guess so.

I'll double-check it before merging, but when I looked I didn't see it.  I believe you have the separator in the CONCAT, but that's a pipe, not a colon.

objecttothis:
I checked and the only time that a colon is used in an explode statement is related to an item_id.  It has no effect here.

PR #2926, opened 2020-08-18
add farsi language

Ali-Rnik:
I tried to add Farsi language files with the help of google translate with the following script:
https://gitlab.com/ali_rnik/auto_i18n_ospos


application/helpers/locale_helper.php
```diff
@@ -67,6 +67,7 @@ function get_languages()
 		'es:spanish' => 'Spanish',
 		'es-MX:spanish' => 'Spanish (Mexico)',
 		'fr:french' => 'French',
+		'fa-IR:persian' => 'Farsi (Iran'),
```
daN4cat, line 70: There is a mistake here. The single quote ' should go after the ).

Strongsaxophone: What a bad mistake, now it has been fixed and I hope it pass the integration.

Ali-Rnik:
in the the Demo, After I changed locale to "Farsi ( Iran )"  and then moved to Home menu, only a blank page is loaded.

objecttothis:
> 
> 
> in the the Demo, After I changed locale to "Farsi ( Iran )" and then moved to Home menu, only a blank page is loaded.
> what is the possible issue of this commit?

You'll need to look at the error log.  This can't be done on the demo, so you'll need to download this build in bintray (https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=desc#files) and install it on your server.  Enable the logs in the config.php.  When you get the blank screen you should be able to go to /application/logs/.  That log will give you a clue as to what is going on.  Is the language setup in weblate also?

Issue #2918, opened 2020-08-11
Cannot Delete Receiving if item has Attribute


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

OSPOS Installation Info: 3.3.2 - cb67b5
Language Code: en-US

OSPOS Timezone is Different from your Local Timezone.

Local Timezone:
Asia/Baku

OSPOS Timezone:
America/New_York

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓

User Configuration:
.Browser: Firefox
.Server Software: Server
.PHP Version: 7.4.8
.DB Version: 8.0.21
.Server Port: 443
[...]

PR #2927, opened 2020-08-19
Allow ON DELETE CASCADE for receiving_id foreign key in links table

objecttothis:
per #2918


jekkos (approved):

application/migrations/20200819000000_modify_attr_links_constraint.php
```diff
@@ -0,0 +1,24 @@
...
+class Migration_modify_attr_links_constraint extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		error_log('Migrating modify_attr_links_constraint');
```
jekkos, line 12: Might be nitpicking but would be nicer if there is a way to add this logging in the code that calls this method for every migration.. I guess it could make things more concise

PR #2939, opened 2020-09-01
fix Let's Encrypt fail on host install using docker

jerry2714:
nginx container keep crashing while installing using `docker/install-server.sh`.


jekkos:
Thanks for the fix @jerry2714 !

PR #2946, opened 2020-09-07
Add rounding unit test

SteveIreland:
After much anguish I finally managed to get this unit test running.  It passes all 10 assertions.  I tried using the instructions you provided but there was a lot of setup I was missing and while I'm not sure I still have it the way you expect, I know that this runs with the configuration for ci-phpunit-test that I finally managed to execute without crashing.  I'm hopeful I can recreate it, but will try that next week.  I promised I would have one this weekend and here it is!

### DIFF
--- a/application/tests/models/Rounding_mode_test.php
+++ b/application/tests/models/Rounding_mode_test.php
@@ -2,19 +2,28 @@
 /**
  * @backupGlobals disabled
  */
+
+
 class Rounding_mode_test extends UnitTestCase
 {
+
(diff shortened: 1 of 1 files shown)


SteveIreland:
Well, that's interesting.  When I run the unit test on my system all 10 assertions work fine, but this travis thing seems to have a failure on the first one.  I'm way past my allocated time for this so I'll dive back in to it next weekend.

jekkos:
Ok thanks

Issue #2889, opened 2020-07-12
Tax calculation issue after changing Tax percent 


w002qaa:
Hello everyone,

Now I am using OSPOS v3.3.1

I created two sales for demo. The first one was with 5% of VAT with Amount Due ($0.00). I changed the VAT value to 7%, and I did another sale with new VAT percent.

If go back to check the receipt sale # 1, it show me Amount Due ($17.80).

Here is photos to show the problem

https://ibb.co/yYGRS1x

https://ibb.co/WDQfGWQ

daN4cat:
There was a problem with VAT calculation in 3.3.1. That's changed with 3.3.2, please test with latest master build.

odiea:
The latest Master is 3.3.1. The last Bintray unstable build is defective.

jekkos:
@odiea latest master seems fine no? I mean it does not have the issue on the demo anymore. @w002qaa any feedback here?

w002qaa:
Hello

Where I can download 3.3.2? 🤔

jekkos:
just fetch latest on [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/3.3.2#files).

w002qaa:
@jekkos  
I did test v3.3.2 and has the same issue in calculation of tax

w002qaa:
Check this link, please

https://drive.google.com/file/d/1_Qia-n1jnVYVmK301uOmSCaB5PqdtpIY/view?usp=sharing

It is video clear for you the problem for you ^_^

objecttothis:
@w002qaa The build you are using in the video is 2F00A6 which is a branch that has not been merged with the master and not rebased with the latest master, which means any changes to fix this issue that were made in the master may not reflect in that build.  Please download the latest master in BinTray (currently b65f68) and check to see if the issue still exists.

As a sidenote, this is why we provide a issue template, which you erased when you created this issue.  Knowing what build you are using to replicate the issue is very important.

objecttothis:
> 
> 
> just fetch latest on [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/3.3.2#files).

@jekkos I think he took your advice, but didn't know that he should only fetch the latest master.

w002qaa:
Hello 
I did fetch(b65f68) and has same issue.

objecttothis:
[...]

PR #2950, opened 2020-09-13
Fix for #2889 Tax calculation issue after changing Tax percent

SteveIreland:
This resolves issue #2889.  Document reprint was retrieving current tax rates instead of retrieving the tax rates that were applied to the original sale.

### DIFF
--- a/application/models/Sale.php
+++ b/application/models/Sale.php
@@ -807,6 +807,20 @@ class Sale extends CI_Model
 	}
 
 	/**
+	 * Return the taxes applied to a sale for a particular item
+	 */
+	public function get_sales_item_taxes($sale_id, $item_id)
+	{
+		$this->db->select('item_id, name, percent');
+		$this->db->from('sales_items_taxes');
+		$this->db->where('sale_id',$sale_id);
+		$this->db->where('item_id',$item_id);
+
+		//return an array of taxes for an item
+		return $this->db->get()->result_array();
+	}
+
+	/**
 	 * Deletes list of sales
 	 */
 	public function delete_list($sale_ids, $employee_id, $update_inventory = TRUE)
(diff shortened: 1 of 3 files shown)


application/models/Sale.php
```diff
@@ -806,6 +806,20 @@ public function get_sales_taxes($sale_id)
 		return $query->result_array();
 	}
 
+	/**
+	 * Return the taxes applied to a sale for a particular item
+	 */
+	public function get_sales_item_taxes($sale_id, $item_id)
+	{
+		$this->db->select('item_id, name, percent');
```
jekkos, line 814: Are we selecting all the fields here? In that case we might just omit this line.

jekkos (approved):

application/models/Sale.php
jekkos: Not sure if it's an issue, but the get_info selects all fields while this one only three.. If I recall correctly the receipt will be loaded using some tax info (if the feature is enabled). So it might be safer to just select all fields in the table to be consistent with the `get_info`

jekkos (changes requested):

application/models/Sale.php
SteveIreland: The get_info selects item_id, name, and percent from the Item_taxes table ... which is exactly why I limited this to the same 3 fields, so that there wouldn't be any accidental introduction of issues caused by the returned results from the two calls being different.

jekkos: Ok then I might have missed that, I had the impression that it was selecting all fields. In that case we should be good. Let's also do a smoke test on the receipt with tax details enabled to see if everything renders fine.

jekkos (approved):

application/models/Sale.php
SteveIreland: Okay - will do.

On another note, I was focused on the "basic tax" system which I'm assuming is still over 95% of your user base.   It never even occurred to me to test the destination based tax document reprint (until this morning while I was in the shower).  So I'll also follow up on that.

Issue #2951, opened 2020-09-14
about Permission of files after upgraded from 3.3.1 to 3.3.2


ekardian:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is:  3.3.2
- OSPOS git commit hash is: 
- PHP version is: 7.2.33
- MySQL or MariaDB version is: 10.0.38-MariaDB-cll-lve
- OS and version is: Linux 3.10.0-962.3.2.lve1.5.28.el7.x86_64
- WebServer is: Apache
- Selected language is: Spanish
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

i just ask, i changed the requirements of files perm. and all working, but i stilll receiving this msg, see the perm of my files.

[...]

PR #2954, opened 2020-09-18
Updates to System Info

WebShells:
Updates to System Info Fixing misplaced error messages (language variables). 
#2951 #2885


WebShells:
Do you want us to put it specifically the read permissions = to 640 or 660 ? That makes it strict for the user to set it to 660 for example.
Am worried about permission 0007 for example which is write permission for world the system info will display:
» [import_customers.csv:] - 0007 | Not Readable ✗ | Security Check Passed ✓ <--- here there is a security risk, so i prefer to make it strict to 640 or 660. 
thoughts ? @opensourcepos/org

objecttothis:
> 
> 
> Do you want us to put it specifically the read permissions = to 640 or 660 ? That makes it strict for the user to set it to 660 for example.
> Am worried about permission 0007 for example which is write permission for world the system info will display:
> » [import_customers.csv:] - 0007 | Not Readable ✗ | Security Check Passed ✓ <--- here there is a security risk, so i prefer to make it strict to 640 or 660.
> thoughts ? @opensourcepos/org

The permission that we really need to test against more than others is the world bit (000x).  If we are testing a log file permission, then anything above 0 is a risk.  If we are testing the import_customers.csv then anything above 4 is a risk.

The second bit we may want to look at is the Group bit (00x0).  This one is tricky.  If it's an environment where untrusted users are in the same group then the rules should be the same as the world bit. Some setups however will require the group bit to be the same as the user bit.  For example, when Apache is not run as the user, but run in the same group, your group bit may legitimately need to be 6.

WebShells:
Then i guess we should be strict with the permissions set because we have thousands of scenarios here, we make it check if the permissions for those specific files is different from what we set for each will display an error.

application/views/configs/system_info.php
```diff
@@ -113,7 +113,7 @@ function get_browser_name($user_agent)
 								echo ' -  ' . substr(sprintf("%o",fileperms($importcustomers)),-4) . ' |  ' . '<font color="red"> Not Readable &#x2717 </font>';
 							}
 							clearstatcache(); 
-							if (is_writable($importcustomers) && substr(decoct(fileperms($importcustomers)), -4) >= 751  ) {
+							if (is_writable($importcustomers) && substr(decoct(fileperms($importcustomers)), -4) > 660  ) {
```
jekkos, line 116: @WebShells what if we have permission 647? in that case the file is world readable and writable, but it will show as ok?

jekkos:
> Then i guess we should be strict with the permissions set because we have thousands of scenarios here, we make it check if the permissions for those specific files is different from what we set for each will display an error.
> This will be set on both group and world so we consider 2 specific files permissions if it's set to any other it will display an error.

I think as @objecttothis states it might be hard to make any assumptions about how the group bit should be set, this depends on the server setup (shared hosting). We might just change the translation for the group bit so the users can decide for themselves?

application/views/configs/system_info.php
WebShells: @jekkos that's why am saying we should be strict with what permissions should be set

WebShells: If we made it specifically set to 640 or 660 then anything else will display a security error

jekkos: Better to make it strict as you say. If group is the same as user we can still issue a warning telling that this might be fine depending on how their server is setup.z

WebShells:
Still got a small fix and will resubmit with changes, are we going for 3.3.3 or 3.4.0 ?

daN4cat:
I vote for a quick 3.3.3

WebShells:
> Do you mind to fix even en-UK language file? Thanks

@daN4cat done.

jekkos (approved):

jekkos:
Thanks @WebShells . Looks good to me .

Issue #2965, opened 2020-10-13
New stock location contains items from previously deleted location


andriux1990:
if we delete a stock location and lego creates a new stock location ... the new stock location contains the items from the old stock location.

the new stock location should have quantities of items 0

jekkos:
You mean you delete and create it without saving in between? Does the location have the same name?

andriux1990:
I mean that I delete the location, and days later I create a new one with another name .. but the amounts I had remain and are shown in the new location

jekkos:
Ok I'm quite surprised. I would expect a new location to have a different id then the created one. Two things I can check here
* Whether stock_location has AUTO_INCREMENT set on the ID field
* Whether we are overriding the ID field when inserting a new stock location

jekkos:
Seems the autoincrement is set correctly. It could be a glitfch in the stock location model (had a quick look and it seems that might need a further look to verify if it works the way it should)

andriux1990:
> Seems the autoincrement is set correctly. It could be a glitfch in the stock location model (had a quick look and it seems that might need a further look to verify if it works the way it should)

ok .. thanks for watching .. the truth is uncomfortable creating a new location and noticing that it has quantities of the last location that was previously .. also developing something to show search suggestions note that locations that are already deleted are shown . I checked the configuration and had only one. but in development look at all the locations that have been created

jekkos:
I had a closer look and indeed, the id is generated in the frontend. So this should not happen there as it does not know the exact database id it should use. In this case I thought removing the id from the field might help, but then you won't be able to create multiple fields in one go, as the input field name in the form is basically the same.

An option would be to use an array as form type and append only new stock locations. In this case updates would need to be done separately. Let me think about and I'll try to come back here with a solution.

jekkos:
@andriux1990 any update? I can't reproduce what you described as a last thing, seems all fine to me?

jekkos:
Would be good if anyone else could do a quick test here so we can hopefully close..

jekkos:
No further response, so let's assume all is fine. Will reopen if needed.

PR #2969, opened 2020-10-16
Fix stock location crud (#2965)

jekkos:
A fix done for the issue reported by @andriux1990 . Tested locally and this time you should end up with a new location id once you add an unexisting one.


andriux1990:
Thanks @jekkos for fixing this problem

PR #2980, opened 2020-10-22
Update bower.json

jekkos:
Bumping bower dependencies (#2980 )

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,7 @@
     "bootswatch": "3.4.1+1",
     "smalot-bootstrap-datetimepicker": "~2.4.4",
     "bootstrap-select": "~1.13.18",
-    "bootstrap-table": "~1.18.0",
+    "bootstrap-table": "~1.18.1",
     "bootstrap-daterangepicker": "~2.1.27",
     "tableExport.jquery.plugin": "1.10.20",
     "jspdf": "~1.3.4",
@@ -70,11 +70,11 @@
     },
     "bootstrap-table": {
       "main": [
-        "dist/bootstrap-table.js",
+        "dist/bootstrap-table.min.js",
         "dist/bootstrap-table.css",
-        "dist/extensions/export/bootstrap-table-export.js",
-        "dist/extensions/mobile/bootstrap-table-mobile.js",
-        "dist/extensions/sticky-header/bootstrap-table-sticky-header.js",
+        "dist/extensions/export/bootstrap-table-export.min.js",
+        "dist/extensions/mobile/bootstrap-table-mobile.min.js",
+        "dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js",
         "dist/extensions/sticky-header/bootstrap-table-sticky-header.css"
       ]
     }
--- a/bower.json
+++ b/bower.json
@@ -63,6 +63,11 @@
         "tableExport.js"
       ]
     },
+    "chartist-axis-plugin": {
+        "main": [
+            "dist/chartist-plugin-axistitle.js"
+        ]
+    },
     "bootstrap-table": {
       "main": [
         "dist/bootstrap-table.js",
--- a/bower.json
+++ b/bower.json
@@ -33,12 +33,12 @@
     "bootswatch": "3.4.1+1",
     "smalot-bootstrap-datetimepicker": "~2.4.4",
     "bootstrap-select": "~1.13.18",
-    "bootstrap-table": "~1.12.2",
+    "bootstrap-table": "~1.18.0",
     "bootstrap-daterangepicker": "~2.1.27",
-    "tableExport.jquery.plugin": "1.5.1",
-    "jspdf": "~1.0.272",
-    "chartist": "0.9.8",
-    "chartist-plugin-axistitle": "0.0.5",
+    "tableExport.jquery.plugin": "1.10.20",
+    "jspdf": "~1.3.4",
+    "chartist": "0.11.4",
+    "chartist-plugin-axistitle": "0.0.7",
     "chartist-plugin-pointlabels": "0.0.4",
     "chartist-plugin-tooltip": "https://github.com/tmmdata/chartist-plugin-tooltip.git#0.0.18",
     "chartist-plugin-barlabels": "https://github.com/mtgibbs/chartist-plugin-barlabels.git#0.0.5",
@@ -48,7 +48,8 @@
     "bootstrap-toggle": "^2.2.2"
   },
   "resolutions": {
-    "jquery": "~1.12.4"
+    "chartist": "0.9.8",
+    "jspdf": ">=1.3.4"
   },
   "overrides": {
     "jquery-ui": {
@@ -64,16 +65,13 @@
     },
     "bootstrap-table": {
       "main": [
-        "src/bootstrap-table.js",
-        "src/bootstrap-table.css",
+        "dist/bootstrap-table.js",
+        "dist/bootstrap-table.css",
         "dist/extensions/export/bootstrap-table-export.js",
         "dist/extensions/mobile/bootstrap-table-mobile.js",
         "dist/extensions/sticky-header/bootstrap-table-sticky-header.js",
         "dist/extensions/sticky-header/bootstrap-table-sticky-header.css"
       ]
     }
-  },
-  "resolutions": {
-    "chartist": "0.9.8"
   }
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,7 +11,8 @@ module.exports = function(grunt) {
 		bower_concat: {
 			all: {
 				mainFiles: {
-					'bootstrap-table': [ "src/bootstrap-table.js", "src/bootstrap-table.css", "dist/extensions/export/bootstrap-table-export.js", "dist/extensions/mobile/bootstrap-table-mobile.js", "dist/extensions/sticky-header/bootstrap-table-sticky-header.js", "dist/extensions/sticky-header/bootstrap-table-sticky-header.css"]
+					'bootstrap-table': [ "dist/bootstrap-table.min.js", "dist/bootstrap-table.css", "dist/extensions/export/bootstrap-table-export.min.js", "dist/extensions/mobile/bootstrap-table-mobile.min.js", "dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js", "dist/extensions/sticky-header/bootstrap-table-sticky-header.css"],
+					'chartist-plugin-axistitle': [ "./dist/chartist-plugin-axistitle.min.js"]
 				},
 				dest: {
 					'js': 'tmp/opensourcepos_bower.js',
--- a/application/views/partial/header.php
+++ b/application/views/partial/header.php
@@ -42,10 +42,10 @@
 		<script src="bower_components/jasny-bootstrap/dist/js/jasny-bootstrap.js"></script>
 		<script src="bower_components/smalot-bootstrap-datetimepicker/js/bootstrap-datetimepicker.min.js"></script>
 		<script src="bower_components/bootstrap-select/dist/js/bootstrap-select.js"></script>
-		<script src="bower_components/bootstrap-table/dist/bootstrap-table.js"></script>
-		<script src="bower_components/bootstrap-table/dist/extensions/export/bootstrap-table-export.js"></script>
-		<script src="bower_components/bootstrap-table/dist/extensions/mobile/bootstrap-table-mobile.js"></script>
-		<script src="bower_components/bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.js"></script>
+		<script src="bower_components/bootstrap-table/dist/bootstrap-table.min.js"></script>
+		<script src="bower_components/bootstrap-table/dist/extensions/export/bootstrap-table-export.min.js"></script>
+		<script src="bower_components/bootstrap-table/dist/extensions/mobile/bootstrap-table-mobile.min.js"></script>
+		<script src="bower_components/bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js"></script>
 		<script src="bower_components/moment/moment.js"></script>
 		<script src="bower_components/bootstrap-daterangepicker/daterangepicker.js"></script>
 		<script src="bower_components/es6-promise/es6-promise.js"></script>
@@ -84,7 +84,7 @@
 	<?php } ?>
 
 		<!-- start minjs template tags -->
-		<script type="text/javascript" src="dist/opensourcepos.min.js?rel=f4ddaa235e"></script>
+		<script type="text/javascript" src="dist/opensourcepos.min.js?rel=2982efc631"></script>
 		<!-- end minjs template tags -->
 	<?php endif; ?>
 
--- a/application/views/partial/header.php
+++ b/application/views/partial/header.php
@@ -14,7 +14,7 @@
 		<link rel="stylesheet" href="bower_components/jasny-bootstrap/dist/css/jasny-bootstrap.css" />
 		<link rel="stylesheet" href="bower_components/smalot-bootstrap-datetimepicker/css/bootstrap-datetimepicker.min.css" />
 		<link rel="stylesheet" href="bower_components/bootstrap-select/dist/css/bootstrap-select.css" />
-		<link rel="stylesheet" href="bower_components/bootstrap-table/src/bootstrap-table.css" />
+		<link rel="stylesheet" href="bower_components/bootstrap-table/dist/bootstrap-table.css" />
 		<link rel="stylesheet" href="bower_components/bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.css" />
 		<link rel="stylesheet" href="bower_components/bootstrap-daterangepicker/daterangepicker.css" />
 		<link rel="stylesheet" href="bower_components/chartist/dist/chartist.min.css" />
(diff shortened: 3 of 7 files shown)


bower.json
```diff
@@ -29,15 +29,15 @@
...
     "jquery-ui": "~1.11.4",
     "bootstrap3-dialog": "https://github.com/nakupanda/bootstrap3-dialog.git#master",
-    "jasny-bootstrap": "~3.1.3",
-    "bootswatch": "^3.4.1",
+    "jasny-bootstrap": "~3.2.0",
+    "bootswatch": "^3.4.1+1",
     "smalot-bootstrap-datetimepicker": "~2.4.4",
     "bootstrap-select": "~1.12.4",
-    "bootstrap-table": "~1.12.2",
+    "bootstrap-table": "~1.18.0",
```
daN4cat, line 36: This change causes problems, grunt minify step crashes.

bower.json
```diff
@@ -29,15 +29,15 @@
     "jquery": "~2.1.4",
     "jquery-ui": "~1.11.4",
     "bootstrap3-dialog": "https://github.com/nakupanda/bootstrap3-dialog.git#master",
-    "jasny-bootstrap": "~3.1.3",
-    "bootswatch": "^3.4.1",
+    "jasny-bootstrap": "~3.2.0",
```
daN4cat, line 32: Items form image loading gets changed with this.

bower.json
```diff
@@ -29,15 +29,15 @@
...
+    "bootswatch": "^3.4.1+1",
     "smalot-bootstrap-datetimepicker": "~2.4.4",
     "bootstrap-select": "~1.12.4",
-    "bootstrap-table": "~1.12.2",
+    "bootstrap-table": "~1.18.0",
     "bootstrap-daterangepicker": "~2.1.27",
-    "tableExport.jquery.plugin": "1.5.1",
-    "chartist": "0.9.8",
-    "chartist-plugin-axistitle": "0.0.5",
+    "tableExport.jquery.plugin": "1.10.20",
```
daN4cat, line 38: causes a jpdf conflict

bower.json
```diff
@@ -29,15 +29,15 @@
     "jquery": "~2.1.4",
     "jquery-ui": "~1.11.4",
     "bootstrap3-dialog": "https://github.com/nakupanda/bootstrap3-dialog.git#master",
-    "jasny-bootstrap": "~3.1.3",
-    "bootswatch": "^3.4.1",
+    "jasny-bootstrap": "~3.2.0",
+    "bootswatch": "^3.4.1+1",
```
daN4cat, line 33: This format doesn't work, ^ must be removed

daN4cat (changes requested):

daN4cat:
It's better if you rebase this with the changes I committed. Sorry, didn't notice/remember you did something similar months ago.

bower.json
jekkos: indeed bower complains that there are minified files in the main[] attribute

bower.json
jekkos: fixed

bower.json
jekkos: For some reason this dependency doesn't load properly.. looking into it

bower.json
jekkos: Bumped jspdf to a newer version and seems fine now

odiea:
I get the following errors now when using Composer. Any Ideas? After I remove the lock file it works.
Microsoft Windows [Version 10.0.19041.685]
(c) 2020 Microsoft Corporation. All rights reserved.

C:\WINDOWS\system32>cd c:\wamp64\www\masterpos

c:\wamp64\www\masterpos>composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for phpdocumentor/reflection-common 2.2.0 -> satisfiable by phpdocumentor/reflection-common[2.2.0].
    - phpdocumentor/reflection-common 2.2.0 requires php ^7.2 || ^8.0 -> your PHP version (7.1.16) does not satisfy that requirement.
  Problem 2
    - Installation request for phpdocumentor/reflection-docblock 5.2.2 -> satisfiable by phpdocumentor/reflection-docblock[5.2.2].
    - phpdocumentor/reflection-docblock 5.2.2 requires php ^7.2 || ^8.0 -> your PHP version (7.1.16) does not satisfy that requirement.
  Problem 3
    - Installation request for phpdocumentor/type-resolver 1.4.0 -> satisfiable by phpdocumentor/type-resolver[1.4.0].
    - phpdocumentor/type-resolver 1.4.0 requires php ^7.2 || ^8.0 -> your PHP version (7.1.16) does not satisfy that requirement.
  Problem 4
    - Installation request for phpspec/prophecy 1.12.1 -> satisfiable by phpspec/prophecy[1.12.1].
    - phpspec/prophecy 1.12.1 requires php ^7.2 || ~8.0, <8.1 -> your PHP version (7.1.16) does not satisfy that requirement.
  Problem 5
    - Installation request for theseer/tokenizer 1.2.0 -> satisfiable by theseer/tokenizer[1.2.0].
    - theseer/tokenizer 1.2.0 requires php ^7.2 || ^8.0 -> your PHP version (7.1.16) does not satisfy that requirement.
  Problem 6
    - phpunit/php-code-coverage 6.1.4 requires theseer/tokenizer ^1.1 -> satisfiable by theseer/tokenizer[1.2.0].
    - phpunit/php-code-coverage 6.1.4 requires theseer/tokenizer ^1.1 -> satisfiable by theseer/tokenizer[1.2.0].
    - theseer/tokenizer 1.2.0 requires php ^7.2 || ^8.0 -> your PHP version (7.1.16) does not satisfy that requirement.
    - Installation request for phpunit/php-code-coverage 6.1.4 -> satisfiable by phpunit/php-code-coverage[6.1.4].


c:\wamp64\www\masterpos>

bower.json
daN4cat: Test item edit form and image upload. It's completely changed.

bower.json
daN4cat: I've committed this change to master, just rebase.

daN4cat:
It probably expects PHP7.2 min. If I'm not mistaken 7.1 is now deprecated.

odiea:
PhpDocumentor requires PHP 7.2 or higher to run. However, code of earlier PHP versions can be analyzed.

bower.json
jekkos: yes this version does not have the files we need. It seems that it does not contain the dist or compiled versions of the files

bower.json
jekkos: ok thanks, I fixed it by rebasing

bower.json
jekkos: I have reverted this one as the archive with dist files is nowhere to be found

bower.json
jekkos: the css file has moved  location, I changed this in the grunt script now, let's see

jekkos:
So now the bootstrap table works in debug mode, but not with minified javascript due to this jspdf that has some es6 import statements. The grunt uglify does not play well with that, browser complains that they should be declared on top of the file

jekkos:
Sadly this grunt uglify solution dates back from 2017 and was never well integrated after..

jekkos:
Tables are working again now after my last push here.

jekkos:
@odiea @daN4cat @objecttothis 
This branch is ready for review and further testing.

Verified the following
* Bootstrap tables loads both using minified and debug js
* Table export version bumped and seems to work find in minified version
* Chartist bumped and dependency bower overrides added, seems to work in both minified and debug js

jekkos:
I did forget to add one more change, committed that now. If anyone has some time to play with this branch.. would be good. Otherwise let's merge and fix any issue after.

daN4cat:
Tested this PR:
- tables are not showing the white and grey alternate lines anymore
- header sticks at the top but stretches to match the width of the window and don't keep the fixed size
- the look doesn't match the rest of the style anymore, not at least with the default theme.

odiea:
I reverted back to the older opensource.min css and js and it appears to be working the way it was.

daN4cat:
Maybe I didn't copy the new css....

jekkos:
From bootstrap table changelog: Remove(js): Removed striped option and use classes instead. So indeed the stripe option was deprecated and removed.

What do mean with the look does not match? Is it more then the striping? I suppose they have also updated their stylesheets so it might be expected.

odiea:
Here is the previous Items table with sticky header.
2020-12-21 07_54_31-Window

Here is the current Items Table with sticky header. It now appears to go full screen.
2020-12-21 07_55_20-Window

jekkos:
ok clear now thanks for the screencaps @odiea

daN4cat:
Sorry, I'm bit on and off doing some DIY at home. I didn't have the time to expand on this but thanks @odiea for the help.

odiea:
@jekkos What changes are needed with this update to include the newer Bootswatch and Bootstrap? I would be willing to try it out.

daN4cat:
@odiea you mean version 4? Probably quite some changes.

daN4cat:
BTW, I tested and the alternate colours are back as the border. It's the sticky headers that are still not fine. I'll see if I find some time later to search a bit about this issue unless it's solved earlier.

odiea:
The only way I could get stripes to show was using the following code in Manage and Tabular.
```
<div id="table_holder">
    <table id="table" class= "table table-striped"></table>
</div>
```

jekkos:
@odiea you need to run grunt on the latest commit before it would work. Otherwise did you try in debug mode? The css classes were added in javascript in my case.

jekkos:
@odiea @daN4cat I fixed the stick header behavior in my last commit. Found [this issue](https://github.com/wenzhixin/bootstrap-table/issues/4814) on bootstrap tables that offered a working solution.

jekkos:
I propose to merge open PR's now and do a quick xmas release. Then maybe a 3.3.4 with the cash rounding fixes of  @SteveIreland .

odiea:
@jekkos Thanks for the tip. So I should run grunt -f after I update this branch? I learn something new everyday.

daN4cat:
Testing.... if fine I'll merge.

daN4cat:
Not working as expected I'm afraid, the header is shifted completely right:

image

image

daN4cat:
Also please update .../application/views/partial/header.php:

image

jekkos:
Screenshot from 2020-12-22 14-16-13

@daN4cat weird as it seems to look good here. I have updated the header.php. Can you try in debug mode?

jekkos:
> @jekkos Thanks for the tip. So I should run grunt -f after I update this branch? I learn something new everyday.

@odiea indeed you need to run the minifcation yourself in this case. `grunt package` or simply `grunt` might be fine.

odiea:
Now when I export to pdf all the Items are showing instead of just the visible ones. Nice improvement

jekkos:
Ah ok that's cool in fact, do you have multiple pages in the pagination then?

odiea:
Also I just noticed after doing an export the sticky header looks great but after a refresh it returns to showing pulled to the right as @daN4cat shows.

jekkos:
@odiea do you mean row refresh or table refresh?

odiea:
Table refresh but this seemed to only appear once. I tried other exports and after that the sticky header went back to it's right alignment.

daN4cat:
pdf export, as it is, is a problem. There is no way to export only the selected, which means you end up exporting everything always. If you have 2000 customers it hangs. I don't dare to try with Items :-O

That needs to be fixed.

jekkos:
@daN4cat what do you mean with 'selected'? Do you just mean what is in the current page?

I think we have exportAll flag set to true so it might behave as expected?

odiea:
@daN4cat In Items I filtered it to just the out of stock Items and that is what it shows now in the pdf.

jekkos:
so as expected you see all items that are out of stock, not just the visible ones?

daN4cat:
@jekkos with selected I mean the ticked boxes, so you can just select the urrent page ticking all the boxes.

daN4cat:
@jekkos in my current setup I don't have all the files uploaded to enable ?debug=true.

daN4cat:
@jekkos in my current setup I don't have all the files uploaded to enable ?debug=true.

daN4cat:
Try to scroll down a bit and then refresh, see if you can reproduce the issue with that. I can see that when you refresh the shifting gets worse.

objecttothis:
> 
> 
> pdf export, as it is, is a problem. There is no way to export only the selected, which means you end up exporting everything always. If you have 2000 customers it hangs. I don't dare to try with Items :-O
> 
> That needs to be fixed.

If we don't already have an issue for that, we should probably create one to remind us of the technical debt.

jekkos:
Alignment issue should be fixed in my last commit. I had to load a bigger database dump and then found the issue.

daN4cat:
Good, I'll test tomorrow.
We are left with the pdf export issue.

odiea:
I tried left left but for some reason the header did not float in the Items table. I then tried right right and that worked for me.
2020-12-22 16_12_41-Window

jekkos:
> 
> 
> @jekkos with selected I mean the ticked boxes, so you can just select the urrent page ticking all the boxes.

Ok I didn't know that was a feature. So in the previous version you would only export the selected rows in that case?

jekkos:
> 
> 
> I tried left left but for some reason the header did not float in the Items table. I then tried right right and that worked for me.
> 2020-12-22 16_12_41-Window

@odiea did you need to modify the code then to make it show? In the bootstrap table github issue they suggest using left offset for both left and right indeed.

odiea:
Yes I did. with left left the Items Header was no longer sticky and I will include an image of Customer header with that using left left. The top image is set to right right and bottom set to left left
2020-12-23 04_18_10-Window
2020-12-23 04_16_36-Window

daN4cat:
@jekkos yes, I get the same as @odiea.

daN4cat:
@jekkos in the previous version it was exporting only what visible, there was no export selected. However, given the current export all situation, it's indeed better to add that part from the example so people can export only a portion of it based on the selected by ticked checkbox.

daN4cat:
> > I tried left left but for some reason the header did not float in the Items table. I then tried right right and that worked for me.
> > 2020-12-22 16_12_41-Window
> 
> @odiea did you need to modify the code then to make it show? In the bootstrap table github issue they suggest using left offset for both left and right indeed. Can you try to see if [this example ](https://live.bootstrap-table.com/code/wenzhixin/1658)works for you?

Yes, this example works fine for me.

daN4cat:
> > @jekkos with selected I mean the ticked boxes, so you can just select the urrent page ticking all the boxes.
> 
> @daN4cat Ok I didn't know that was a feature. So in the previous version you would only export the selected rows in that case?
> In the following example an extra dropdown is shown that allows you to select the type of export (selected only or full). I could make it smarter by triggering selected in case at least one row in the table is selected.
> 
> https://github.com/wenzhixin/bootstrap-table-examples/blob/develop/extensions/export.html

@jekkos in the previous version it was exporting only what visible, there was no export selected. However, given the current export all situation, it's indeed better to add that part from the example so people can export only a portion of it based on the selected by ticked checkbox.

jekkos:
@daN4cat not so straightforward, but I hacked my way around it. So a small gift to you today. Should work in the last commit.

odiea:
It does now export selected Items. The issue I see now is on the table refresh. It is quite slow and does not show the check boxes until a  manual refresh is done.

jekkos:
Yes I have also noticed that behavior. I guess in this case the refresh is slow after the export is triggered? I think that happens because it queries the backend without applying the offset and limit query parameters, so basically it still loads the whole result set in browser and then does the filtering afterwards.

Not sure if this can be fixed in a decent way as the table export feature is in the frontend and streaming the data requires a different (end-to-end) approach. As a last fix I could try to bump the table export libraries again (as I saw one of them included some file that would enable webworkers to run this export in a browser thread, so maybe then we can get rid of that refresh ).

If that doesn't work then I'd just revert to the situation that we had as it won't scale well with larger datasets anyway.

jekkos:
Ok i had a look and we are on the latest version of table export, so no improvements possible there. I will change the 'all' to 'basic' as a default case then instead. I guess in case of the export the selection is lost after the table refresh, not sure we can do anything about that.

daN4cat:
Sticky headers are fine, but partial export is terribly slow after export. Web browser times out.

daN4cat:
What I see after the export is a search&search= means no offset and not limit on a page refresh. Maybe that should be fixed, a page refresh with correct offset and limits.

objecttothis:
> 
> 
> Sticky headers are fine, but partial export is terribly slow after export. Web browser times out.

I believe some improvement can be had with my optimizations once that PR gets merged, but I believe it's waiting on 3.3.3 before we merge into the master.

odiea:
You will also notice when it uses the pdf export the check box column disappears until  you refresh the module.

daN4cat:
> > Sticky headers are fine, but partial export is terribly slow after export. Web browser times out.
> 
> I believe some improvement can be had with my optimizations once that PR gets merged, but I believe it's waiting on 3.3.3 before we merge into the master.

I'm not sure it's an optimisation of the database query issue, it's really fetching all entries which is not correct. At the moment it uses pagination but for that case it is not.

objecttothis:
> 
> 
> > > Sticky headers are fine, but partial export is terribly slow after export. Web browser times out.
> > 
> > 
> > I believe some improvement can be had with my optimizations once that PR gets merged, but I believe it's waiting on 3.3.3 before we merge into the master.
> 
> I'm not sure it's an optimisation of the database query issue, it's really fetching all entries which is not correct. At the moment it uses pagination but for that case it is not.

Yeah, if it's fetching all items, there is no db optimization that's going to fix that.  It either needs to use pagination to limit the result or some other filtering.  That's one of those bugs that won't get caught unless you run it against a full-sized database.

jekkos:
> You will also notice when it uses the pdf export the check box column disappears until you refresh the module.

I tried the bootstrap-table demo and it's [the same there](https://examples.bootstrap-table.com/#extensions/export.html)..

jekkos:
> What I see after the export is a search&search= means no offset and not limit on a page refresh. Maybe that should be fixed, a page refresh with correct offset and limits.

@daN4cat did you recompile the minified js? Otherwise you need to set in debug mode.
In my latest version the export does not do any extra request.. (so just dumps the data in the DOM)

daN4cat:
@jekkos the problem is when you randomly select two or three, if you export just the page it's fine.

jekkos:
> > > Sticky headers are fine, but partial export is terribly slow after export. Web browser times out.
> > 
> > 
> > I believe some improvement can be had with my optimizations once that PR gets merged, but I believe it's waiting on 3.3.3 before we merge into the master.
> 
> I'm not sure it's an optimisation of the database query issue, it's really fetching all entries which is not correct. At the moment it uses pagination but for that case it is not.

I found that it only removes the limit and offset query parameters in case the exportType is set to 'all'. In the latest version it's set to basic so this should be fixed.

jekkos:
Ok no you guys were right. It did do this unbounded request but only for the export when rows were selected. I have removed the feature and back to the old setup now.

I think a full table export could be done but then we should use end-to-end streaming. In this case a db query can use a cursor and stream batches of data in chunked http response to the client.

odiea:
I still get the following after a pdf export. You can see that the check boxes no longer show.
2020-12-27 14_19_32-Window

jekkos:
Good catch indeed this only happens after a PDF export..

daN4cat:
Shall we merge? It's like it used to be before excluding the disappearing checkboxes with pdf export.

daN4cat:
@jekkos how can I get Customers sorted by Total Spent?

odiea:
@daN4cat In tabular helper it appears as the following. function get_customer_manage_table_headers()
```
array('total' => $CI->lang->line('common_total_spent'), 'sortable' => FALSE)
``

jekkos:
@daN4cat Indeed let's merge and release. It's been since september and there's quite some fixes up til now.

daN4cat:
@odiea thanks for the help, now I remember why I didn't implement it because the stats query is separate from the customers lookup.

daN4cat (approved):

PR #2982, opened 2020-10-22
Fix notifications

jekkos:
It seems like some of the notifications code wasn't working (at least in the invoice screen) and as the newer code adheres to this setup, I did a search/replace across the codebase to align this.


jekkos:
@here anyone some spare time to take a look here? I found that after this change some of the notifications started working again. At least now everything is consistent.

jekkos:
I went ahead and merged this. Will fix if any issue.

Issue #2996, opened 2020-10-29
Warning in domPdf after 7.4 upgrade


jekkos:
I'm currently seeing this warning after upgrading to php 7.4 when generating PDFs

```
Message: Function get_magic_quotes_runtime() is deprecated

Filename: helpers/dompdf_helper.php

Line Number: 10

Backtrace:

File: /home/runwalk/domains/ospos.runwalk.be/public_html/application/helpers/dompdf_helper.php
Line: 10
Function: _error_handler

File: /home/runwalk/domains/ospos.runwalk.be/public_html/application/controllers/Sales.php
Line: 792
Function: create_pdf 
```

jekkos:
This `magic quote` feature was already deprecated in PHP 5.3, so I'll just remove it.

PR #2997, opened 2020-10-29
Remove magic quotes check in dompdf helper (#2996)

jekkos:
Tested on my own server, PDF generation looks good + no more warning in the server output


daN4cat:
Still many bugs...

jekkos:
This oen popped up after upgrading to php 7.4..

Issue #2949, opened 2020-09-13
Theme Darkly CSS Fix


minhazulOO7:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.2
- OSPOS git commit hash is: bb309c
- PHP version is: 7.2.24
- MySQL or MariaDB version is: 5.5.5-10.3.17-MariaDB
- OS and version is: Linux 4.18.0-193.14.2.el8_2.x86_64
- WebServer is: Apache/2.4.37 (centos)
- Selected language is: en-US
- (If applicable) Docker installation: N/A
- (If applicable) Installation package for the LAMP/LEMP stack is: N/A

### Issue

[...]

PR #3001, opened 2020-11-02
Adjust a theme to handle feature development using non-theme supported classes.

SteveIreland:
This resolves the issue reported with #2949.  It provides a mechanism to override the color style that might have been used to make the flatly theme work well with the custom development.  The stylesheet has the same name as the theme but resides in the public/css folder.  This includes darkly.css as an example.  It uses the provided color override specs provided by the issue reporter Syed.

### DIFF
--- a/application/views/partial/header.php
+++ b/application/views/partial/header.php
@@ -77,6 +77,12 @@
 		<link rel="stylesheet" type="text/css" href="dist/jquery-ui/jquery-ui.min.css"/>
 		<link rel="stylesheet" type="text/css" href="dist/opensourcepos.min.css?rel=397f582d3d"/>
 		<!-- end mincss template tags -->
+
+		<!-- Tweaks to the UI for a particular theme should drop here  -->
+	<?php if ($this->config->item('theme') != 'flatly' && file_exists($_SERVER['DOCUMENT_ROOT'] . '/public/css/' . $this->config->item('theme') . '.css')) { ?>
+		<link rel="stylesheet" type="text/css" href="<?php echo 'css/' . $this->config->item('theme') . '.css' ?>"/>
+	<?php } ?>
+
 		<!-- start minjs template tags -->
 		<script type="text/javascript" src="dist/opensourcepos.min.js?rel=1ca975c3b9"></script>
 		<!-- end minjs template tags -->
--- a/public/css/darkly.css
+++ b/public/css/darkly.css
@@ -1,0 +1,33 @@
+/* Since flatly is the default theme, all css design should use flatly as the starting point. */
+/* Any css color attributes that need to be overridden for a particular theme */
+/* should be included in a css file file named. */
+
+.fixed-table-container tbody .selected td {
+	background-color:#757575
+}
+
+/* Receivings & Sales */
+
+#register td {
+	background-color:#757575;
+}
+
+#add_item_form,#overall_sale {
+	background-color:#757575
+}
+
+#mode_form,#payment_details {
+	background-color:#757575
+}
+
+/* Reports > Graphical Reports */
+
+.ct-label {
+	fill:#eee;
+	color:#eee;
+}
+
+text.ct-axis-title{
+	fill:#eee;
+	color:#eee;
+}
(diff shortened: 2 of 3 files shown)


application/views/partial/header.php
```diff
@@ -77,6 +77,12 @@
 		<link rel="stylesheet" type="text/css" href="dist/jquery-ui/jquery-ui.min.css"/>
 		<link rel="stylesheet" type="text/css" href="dist/opensourcepos.min.css?rel=397f582d3d"/>
 		<!-- end mincss template tags -->
+
+        <!-- Tweaks to the UI for a particular theme should drop here  -->
+    <?php error_log('>>> theme-' . $this->config->item('theme') . ', ' . $_SERVER['DOCUMENT_ROOT'] . '/public/css/' . $this->config->item('theme') . '.css'); if ($this->config->item('theme') != 'flatly' && file_exists($_SERVER['DOCUMENT_ROOT'] . '/public/css/' . $this->config->item('theme') . '.css')) { ?>
```
daN4cat, line 82: php error log?

public/css/darkly.css
```diff
@@ -0,0 +1,33 @@
+/* Since flatly is the default theme, all css design should use flatly as the starting point. */
+/* Any css color attributes that need to be overridden for a particular theme */
+/* should be included in a css file file named. */
+
+.fixed-table-container tbody .selected td {
```
jekkos, line 5: I was reflecting, I think it will fix indeed the issue at hand, but was wondering whether you considered if there could be a CSS only fix in which you override these styles based on a theme-specific selector? It would be nicer in a way as if it is possible, then you can just minify the stylesheet here as well

application/views/partial/header.php
```diff
@@ -77,6 +77,12 @@
...
+        <link rel="stylesheet" type="text/css" href="<?php echo 'css/' . $this->config->item('theme') . '.css' ?>"/>
```
jekkos, line 83: minor remark here, but this line would give a 404 if there is no overridde present for the file. Not that much of a dealbreaker, just an unnessecary request from the browser (as we know beforehand whether an override is there or not).

I was thinking to maybe suggest whether we can contruct this file as a variable before this view is loaded and check if it exists and needs to be loaded .. or in case we can have a CSS specific selector it might not be needed anymore at all.

jekkos: Ok as you suggested before in that case we might need to setup a bootswatch build ourselves to include our modifications, which might be a story on its own..

jekkos (approved):

jekkos:
Thanks for the work here @SteveIreland

application/views/partial/header.php
SteveIreland: Aaaarrgghhhh!!!!

application/views/partial/header.php
SteveIreland: Because I mucked up and left the error log statement in there, the fact that the link to the theme override style sheet is conditional ONLY if it exists was not obvious.   I will be removing the debug statement tonight.

SteveIreland:
@jekkos Including it in the 3.3 release is fine.  I believe this to be a very low risk change.  I'll pull the build file from bin tray tomorrow and confirm it works one more time, but I've tested the changes several times (even after removing the debug statement) and it still works.

application/views/partial/header.php
jekkos: Ok indeed II didn't read that line completely.. looks good to me

jekkos:
I guess it

> @jekkos Including it in the 3.3 release is fine. I believe this to be a very low risk change. I'll pull the build file from bin tray tomorrow and confirm it works one more time, but I've tested the changes several times (even after removing the debug statement) and it still works.

I guess it'll be fine, might be faster if you correct the statement then merge, we'll check the dev server then after.

SteveIreland:
I finally found time to download it and test the theme stylesheet override functionality and it is working so I merged it.  I didn't test everything to make sure that darkly has been fully adapted (I personally am dark-themed challenged).  But if someone identifies something else that needs to be tweaked then it is now a simple task to add it to the theme based stylesheet override.

PR #3003, opened 2020-11-02
Add Retail Price to Items Summary Report

chunter2:


jekkos (approved):
Thanks for the contribution, looks fine and let's check if it works on dev after merge.

PR #3011, opened 2020-11-15
Added Transactions column to Employees Summary Report and Transactions Summary Report

chunter2:
Not sure I've done this right but the file diffs look correct.  This also included a fix for quantity sorting in Categories, Employees and Transactions reports.


jekkos:
Looks good to me in fact, thanks again for this addtiion @chunter2! I'll merge and then do a quick validation on dev.

jekkos (approved):

Issue #3019, opened 2020-11-19
Change in Email Bill format


saurabh261187:
### Installation information

OSPOS version : opensourcepos.org · 3.3.3 - 2393c2.
Server Software: Apache/2.4.29
PHP Version: 7.2.24-0ubuntu0.18.04.7
MYSQL DB Version: 5.7.32-0ubuntu0.18.04.1

### Issue / Bug  / Question / New Feature

Before updating my ospos version from 3.3.2 to the above version from the unstable build opensourcepos.20201115225224.master.2393c2 , the Email bills in PDF format sent to the customers looked like this :

![earlierpdfformat](https://user-images.githubusercontent.com/15091342/99664723-a1034200-2a8e-11eb-8d47-47202a91bfbb.png)

However, after the update, although all the settings are the same, it is looking like this :

![currentpdfformat](https://user-images.githubusercontent.com/15091342/99664783-baa48980-2a8e-11eb-83c7-dc1466fc126d.png)

Please guide me if I have done something wrong, or have gone wrong in some configuration.

odiea:
It appears you are missing the invoice.css file.

jekkos:
@odiea how did you conclude that?

odiea:
It appears that there is no Invoice.css formatting.

jekkos:
@saurabh261187 can you try master build and report back if this fixes the issue?

It might be that in certain setups the css file was not found (as I think that base_url did not include the public part in the URL). In fact if the css lookup happens relative to the current page it should be fine. It's also consistent with the CSS href's in the header.php now.

jekkos:
I have done a quick test locally and PDF generation seems fine now. I close.
[Invoice.pdf](https://github.com/opensourcepos/opensourcepos/files/5694469/Invoice.pdf)

odiea:
@jekkos I just tried sending an Invoice.pdf and it still appears to be an issue. I am using the current bower update branch. 
[Invoice-INV-20000599.pdf](https://github.com/opensourcepos/opensourcepos/files/5736600/Invoice-INV-20000599.pdf)

jekkos:
Ok I'm no certain but think bower-update was branched some time ago.. so it might not contain this fix yet.

jekkos:
Might be better to verify on master instead? At least I addded evidence after testing and closing.

odiea:
@jekkos Did you try sending yourself an Invoice.

jekkos:
[...]

PR #3021, opened 2020-11-19
Space removal

WebShells:
Additional spaces removal/alignment

### DIFF
--- a/application/views/sales/invoice.php
+++ b/application/views/sales/invoice.php
@@ -83,15 +83,15 @@ $(document).ready(function()
 		<table id="meta">
 			<tr>
 				<td class="meta-head"><?php echo $this->lang->line('sales_invoice_number');?> </td>
-				<td><textarea rows="5" cols="6"><?php echo $invoice_number; ?></textarea></td>
+				<td><div><?php echo $invoice_number; ?></div></td>
 			</tr>
 			<tr>
 				<td class="meta-head"><?php echo $this->lang->line('common_date'); ?></td>
-				<td><textarea rows="5" cols="6"><?php echo $transaction_date; ?></textarea></td>
+				<td><div><?php echo $transaction_date; ?></div></td>
 			</tr>
 			<tr>
 				<td class="meta-head"><?php echo $this->lang->line('sales_invoice_total'); ?></td>
-				<td><textarea rows="5" cols="6"><?php echo to_currency($total); ?></textarea></td>
+				<td><div><?php echo to_currency($total); ?></div></td>
 			</tr>
 		</table>
 	</div>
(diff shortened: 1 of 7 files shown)


jekkos:
@WebShells thanks for this fix! While you are at it, I think that having the textarea might bring some issues with the PDF rendering. Would you mind to check and compare how the fields are rendered in regular templates? I think it was changed to a normal div not so long ago. Thanks

jekkos (changes requested):
If you find the time maybe double check the textarea thingy.. thx

WebShells:
> If you find the time maybe double check the textarea thingy.. thx
@jekkos wait let me apply it on all templates before we merge.

WebShells:
@jekkos shall i replace also for number fields ? Can i have a screenshot how does it look so i can know what to fix?

jekkos:
> @jekkos shall i replace also for number fields ? Can i have a screenshot how does it look so i can know what to fix?

Yes I'd replace all textarea items. It does not play well with dom2pdf.

jekkos:
Thanks for the modifications done @WebShells .. 

I might have been wrong asking to change textarea for a div, in case of a column (td) we might not need to next an extra div.. I'll quickly verify if it's required and then wrap this one up.

WebShells:
I can take care of that man no worries, ill check and confirm it

jekkos:
Ok great thanks for going the extra mile here, I want to make sure we are not using any unneeded elements. Also I vaguely remember something in [html standard saying it is not something recommended ](https://stackoverflow.com/questions/1110915/is-a-div-inside-a-td-a-bad-idea).

jekkos:
[Invoice-1.pdf](https://github.com/opensourcepos/opensourcepos/files/5625699/Invoice-1.pdf)

I have deployed this version on the dev server and did a quick test (check attached PDF). It seems the PDF generation is not working as it should (does not have any styles properly applied)

jekkos:
This is a pdf generated from demo (3.2.2)

[2.pdf](https://github.com/opensourcepos/opensourcepos/files/5625748/2.pdf)

jekkos:
Not sure but it could be that the invoice_email.css is not being imported. We need to check if this issue is also in master.. It might also be due to the dompdf upgrade

jekkos:
EDIT: did a quick check and this 0.8.5 domPDF was already in 3.3.2, so it's probably not that

jekkos (changes requested):
First we need to resolve the PDF generation issue

jekkos:
@WebShells this might be same issue as #3019

jekkos:
[Quote1.pdf](https://github.com/opensourcepos/opensourcepos/files/5694647/Quote1.pdf)
[Invoice1.pdf](https://github.com/opensourcepos/opensourcepos/files/5694652/Invoice1.pdf)
Invoice and quote PDFs look fine now. Proof attached here.

Issue #2355, opened 2019-02-08
Round with out Decimals


truchosky:
hi guys, i use ospos 3.2.3, this is not a bug, is more like a feature

I checked Rounding_mode.php but I couldn't accomplish what I am looking for.

my currency has not decimals, there are int numbers only, but i want to round anyways.

for example: 131 should be 130, 134 should be 130 too and 135 should be 140.

this can be accomplish using -1 on round function, like this: 

round( 136, -1) = 140
round (134,-1) = 130

any advise or guidance on how to procced?

greetings

SteveIreland:
Are you asking to add -1 to the currency decimal options?

That is simple enough to add, but the big question is whether or not it is handled properly by the NumberFormatter PHP class.  I cannot find any documentation where it is supported.  Maybe someone else knows for sure.

```
function to_decimals($number, $decimals, $type=\NumberFormatter::DECIMAL)
{
	// ignore empty strings and return
	// NOTE: do not change it to empty otherwise tables will show a 0 with no decimal nor currency symbol
	if(!isset($number))
	{
		return $number;
	}

	$config = get_instance()->config;
	$fmt = new \NumberFormatter($config->item('number_locale'), $type);
	$fmt->setAttribute(\NumberFormatter::MIN_FRACTION_DIGITS, $config->item($decimals));
	$fmt->setAttribute(\NumberFormatter::MAX_FRACTION_DIGITS, $config->item($decimals));
	if(empty($config->item('thousands_separator')))
	{
		$fmt->setAttribute(\NumberFormatter::GROUPING_SEPARATOR_SYMBOL, '');
	}
	$fmt->setSymbol(\NumberFormatter::CURRENCY_SYMBOL, $config->item('currency_symbol'));

	return $fmt->format($number);
}
```

 Just out of curiosity what country requires this or is it a regional thing?

truchosky:
this is mandatory in chile, https://www.legalchile.cl/como-funciona-la-regla-del-redondeo/

where I add that function?, on Rounding_mode.php ?

SteveIreland:
Thanks for the link which clarified that this is a "cash rounding" application.  I'll see what it takes.

SteveIreland:
I explored the option where I added -1 as a cash decimals option and it seems to work just fine.  Any change I make will only go to version 3.3.   Can you locate `local_config.php` and make the following change and let us know how it works out for you?  If it works for you then I'll go ahead and make the corresponding change in version 3.3.

[...]

PR #3026, opened 2020-11-22
Correct issues with cash rounding.

SteveIreland:
This is a trial run for feedback and to see if it will build.  @odiea pointed out that this could impact cashups but that is not an area I'm familiar with.  There is still an open question regarding Summary reporting (do I need to do something with those reports). The detail reports should be properly reflecting cash rounding.

### DIFF
--- a/application/controllers/Sales.php
+++ b/application/controllers/Sales.php
@@ -2,10 +2,6 @@
 
 require_once("Secure_Controller.php");
 
-define('PRICE_MODE_STANDARD', 0);
-define('PRICE_MODE_KIT', 1);
-define('PAYMENT_TYPE_UNASSIGNED', '--');
-
 class Sales extends Secure_Controller
 {
 	public function __construct()
@@ -263,7 +259,7 @@ class Sales extends Secure_Controller
 		$data = array();
 
 		$payment_type = $this->input->post('payment_type');
-		if($payment_type != $this->lang->line('sales_giftcard'))
+		if($payment_type !== $this->lang->line('sales_giftcard'))
 		{
 			$this->form_validation->set_rules('amount_tendered', 'lang:sales_amount_tendered', 'trim|required|callback_numeric');
 		}
@@ -274,7 +270,7 @@ class Sales extends Secure_Controller
 
 		if($this->form_validation->run() == FALSE)
 		{
-			if($payment_type == $this->lang->line('sales_giftcard'))
+			if($payment_type === $this->lang->line('sales_giftcard'))
 			{
 				$data['error'] = $this->lang->line('sales_must_enter_numeric_giftcard');
 			}
@@ -285,7 +281,7 @@ class Sales extends Secure_Controller
 		}
 		else
 		{
-			if($payment_type == $this->lang->line('sales_giftcard'))
+			if($payment_type === $this->lang->line('sales_giftcard'))
 			{
 				// in case of giftcard payment the register input amount_tendered becomes the giftcard number
 				$giftcard_num = $this->input->post('amount_tendered');
@@ -316,7 +312,7 @@ class Sales extends Secure_Controller
 					$this->sale_lib->add_payment($payment_type, $amount_tendered);
 				}
 			}
-			elseif($payment_type == $this->lang->line('sales_rewards'))
+			elseif($payment_type === $this->lang->line('sales_rewards'))
 			{
 				$customer_id = $this->sale_lib->get_customer();
 				$package_id = $this->Customer->get_info($customer_id)->package_id;
@@ -346,6 +342,20 @@ class Sales extends Secure_Controller
 
 						$this->sale_lib->add_payment($payment_type, $amount_tendered);
 					}
+				}
+			}
+			elseif($payment_type === $this->lang->line('sales_cash'))
+			{
+				$amount_due = $this->sale_lib->get_total();
+				$sales_total = $this->sale_lib->get_total(FALSE);
+
+				$amount_tendered = $this->input->post('amount_tendered');
+				$this->sale_lib->add_payment($payment_type, $amount_tendered);
+				$cash_adjustment_amount = $amount_due - $sales_total;
+				if($cash_adjustment_amount <> 0)
+				{
+					$this->session->set_userdata('cash_mode', CASH_MODE_TRUE);
+					$this->sale_lib->add_payment($this->lang->line('sales_cash_adjustment'), $cash_adjustment_amount, CASH_ADJUSTMENT_TRUE);
 				}
 			}
 			else
@@ -571,22 +581,22 @@ class Sales extends Secure_Controller
 		$data['payments_total'] = $totals['payment_total'];
 		$data['payments_cover_total'] = $totals['payments_cover_total'];
 		$data['cash_rounding'] = $this->session->userdata('cash_rounding');
+		$data['cash_mode'] = $this->session->userdata('cash_mode');
 		$data['prediscount_subtotal'] = $totals['prediscount_subtotal'];
 		$data['cash_total'] = $totals['cash_total'];
 		$data['non_cash_total'] = $totals['total'];
 		$data['cash_amount_due'] = $totals['cash_amount_due'];
 		$data['non_cash_amount_due'] = $totals['amount_due'];
 
-		if($data['cash_rounding'])
+		if($data['cash_mode'])
 		{
-			$data['total'] = $totals['cash_total'];
 			$data['amount_due'] = $totals['cash_amount_due'];
 		}
 		else
 		{
-			$data['total'] = $totals['total'];
 			$data['amount_due'] = $totals['amount_due'];
 		}
+
 		$data['amount_change'] = $data['amount_due'] * -1;
 
 		if($data['amount_change'] > 0)
@@ -609,6 +619,7 @@ class Sales extends Secure_Controller
 		if($this->sale_lib->is_invoice_mode())
 		{
 			$invoice_format = $this->config->item('sales_invoice_format');
+
 			// generate final invoice number (if using the invoice in sales by receipt mode then the invoice number can be manually entered or altered in some way
 			if(!empty($invoice_format) && $invoice_number == NULL)
 			{
@@ -787,6 +798,7 @@ class Sales extends Secure_Controller
 
 			// generate email attachment: invoice in pdf format
 			$html = $this->load->view("sales/" . $type . "_email", $sale_data, TRUE);
+
 			// load pdf helper
 			$this->load->helper(array('dompdf', 'file'));
 			$filename = sys_get_temp_dir() . '/' . $this->lang->line("sales_" . $type) . '-' . str_replace('/', '-', $number) . '.pdf';
@@ -903,7 +915,9 @@ class Sales extends Secure_Controller
 	private function _load_sale_data($sale_id)
 	{
 		$this->sale_lib->clear_all();
-		$this->sale_lib->reset_cash_flags();
+		$cash_rounding = $this->sale_lib->reset_cash_rounding();
+		$data['cash_rounding'] = $cash_rounding;
+
 		$sale_info = $this->Sale->get_info($sale_id)->row_array();
 		$this->sale_lib->copy_entire_sale($sale_id);
 		$data = array();
@@ -922,18 +936,18 @@ class Sales extends Secure_Controller
 
 		// Returns 'subtotal', 'total', 'cash_total', 'payment_total', 'amount_due', 'cash_amount_due', 'payments_cover_total'
 		$totals = $this->sale_lib->get_totals($tax_details[0]);
+		$this->session->set_userdata('cash_adjustment_amount', $totals['cash_adjustment_amount']);
 		$data['subtotal'] = $totals['subtotal'];
-		$data['total'] = $totals['total'];
 		$data['payments_total'] = $totals['payment_total'];
 		$data['payments_cover_total'] = $totals['payments_cover_total'];
-		$data['cash_rounding'] = $this->session->userdata('cash_rounding');
+		$data['cash_mode'] = $this->session->userdata('cash_mode');
 		$data['prediscount_subtotal'] = $totals['prediscount_subtotal'];
 		$data['cash_total'] = $totals['cash_total'];
 		$data['non_cash_total'] = $totals['total'];
 		$data['cash_amount_due'] = $totals['cash_amount_due'];
 		$data['non_cash_amount_due'] = $totals['amount_due'];
 
-		if($this->session->userdata('cash_rounding'))
+		if($data['cash_mode'] && ($data['selected_payment_type'] === $this->lang->line('sales_cash') || $data['payments_total'] > 0))
 		{
 			$data['total'] = $totals['cash_total'];
 			$data['amount_due'] = $totals['cash_amount_due'];
@@ -943,6 +957,7 @@ class Sales extends Secure_Controller
 			$data['total'] = $totals['total'];
 			$data['amount_due'] = $totals['amount_due'];
 		}
+
 		$data['amount_change'] = $data['amount_due'] * -1;
 
 		$employee_info = $this->Employee->get_info($this->sale_lib->get_employee());
@@ -1013,6 +1028,11 @@ class Sales extends Secure_Controller
 			$sale_id = -1;
 			$this->session->set_userdata('sale_id', -1);
 		}
+		$cash_rounding = $this->sale_lib->reset_cash_rounding();
+
+		// cash_rounding indicates only that the site is configured for cash rounding
+		$data['cash_rounding'] = $cash_rounding;
+
 		$data['cart'] = $this->sale_lib->get_cart();
 		$customer_info = $this->_load_customer_data($this->sale_lib->get_customer(), $data, TRUE);
 
@@ -1027,22 +1047,29 @@ class Sales extends Secure_Controller
 		$data['taxes'] = $tax_details[0];
 		$data['discount'] = $this->sale_lib->get_discount();
 		$data['payments'] = $this->sale_lib->get_payments();
+
 		// Returns 'subtotal', 'total', 'cash_total', 'payment_total', 'amount_due', 'cash_amount_due', 'payments_cover_total'
 		$totals = $this->sale_lib->get_totals($tax_details[0]);
+
 		$data['item_count'] = $totals['item_count'];
 		$data['total_units'] = $totals['total_units'];
 		$data['subtotal'] = $totals['subtotal'];
 		$data['total'] = $totals['total'];
 		$data['payments_total'] = $totals['payment_total'];
 		$data['payments_cover_total'] = $totals['payments_cover_total'];
-		$data['cash_rounding'] = $this->session->userdata('cash_rounding');
+
+		// cash_mode indicates whether this sale is going to be processed using cash_rounding
+		$cash_mode = $this->session->userdata('cash_mode');
+		$data['cash_mode'] = $cash_mode;
 		$data['prediscount_subtotal'] = $totals['prediscount_subtotal'];
 		$data['cash_total'] = $totals['cash_total'];
 		$data['non_cash_total'] = $totals['total'];
 		$data['cash_amount_due'] = $totals['cash_amount_due'];
 		$data['non_cash_amount_due'] = $totals['amount_due'];
 
-		if($data['cash_rounding'])
+		$data['selected_payment_type'] = $this->sale_lib->get_payment_type();
+
+		if($data['cash_mode'] && ($data['selected_payment_type'] == $this->lang->line('sales_cash') || $data['payments_total'] > 0))
 		{
 			$data['total'] = $totals['cash_total'];
 			$data['amount_due'] = $totals['cash_amount_due'];
@@ -1052,11 +1079,12 @@ class Sales extends Secure_Controller
 			$data['total'] = $totals['total'];
 			$data['amount_due'] = $totals['amount_due'];
 		}
+
 		$data['amount_change'] = $data['amount_due'] * -1;
 
 		$data['comment'] = $this->sale_lib->get_comment();
 		$data['email_receipt'] = $this->sale_lib->is_email_receipt();
-		$data['selected_payment_type'] = $this->sale_lib->get_payment_type();
+
 		if($customer_info && $this->config->item('customer_reward_enable') == TRUE)
 		{
 			$data['payment_options'] = $this->Sale->get_payment_options(TRUE, TRUE);
@@ -1143,8 +1171,8 @@ class Sales extends Secure_Controller
 		$data['selected_employee_id'] = $sale_info['employee_id'];
 		$data['selected_employee_name'] = $this->xss_clean($employee_info->first_name . ' ' . $employee_info->last_name);
 		$data['sale_info'] = $sale_info;
-		$balance_due = $sale_info['amount_due'] - $sale_info['amount_tendered'];
-		if($balance_due < 0)
+		$balance_due = round($sale_info['amount_due'] - $sale_info['amount_tendered'] + $sale_info['cash_refund'], totals_decimals(), PHP_ROUND_HALF_UP);
+		if(!$this->sale_lib->reset_cash_rounding() && $balance_due < 0)
 		{
 			$balance_due = 0;
 		}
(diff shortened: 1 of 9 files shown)


application/controllers/Sales.php
```diff
@@ -341,14 +343,32 @@ public function add_payment()
...
-						$this->sale_lib->add_payment($payment_type, $amount_tendered);
+						$this->sale_lib->add_payment($payment_type, $amount_tendered, CASH_ADJUSTMENT_FALSE);
 					}
 				}
 			}
+			elseif($payment_type == $this->lang->line('sales_cash'))
+			{
+				$cash_adjustment_amount = $this->session->userdata('cash_adjustment_amount');
+				$amount_tendered = $this->input->post('amount_tendered');
+				$this->sale_lib->add_payment($payment_type, $amount_tendered, 0);
```
jekkos, line 354: maybe use the constants here as well?

application/controllers/Sales.php
```diff
@@ -341,14 +343,32 @@ public function add_payment()
...
+				else
+				{
+					$this->session->set_userdata('cash_mode', 0);
+				}
+			}
 			else
 			{
 				$amount_tendered = $this->input->post('amount_tendered');
-				$this->sale_lib->add_payment($payment_type, $amount_tendered);
+				$this->sale_lib->add_payment($payment_type, $amount_tendered, CASH_ADJUSTMENT_FALSE);
```
jekkos, line 371: the defautl is false I suppose? Maybe we can just specify that in the method itself as a default parameter value or would you prefer to set it automatically

application/migrations/20201108100000_cashrounding.php
```diff
@@ -0,0 +1,20 @@
...
+class Migration_cashrounding extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		$this->db->query('ALTER TABLE ' . $this->db->dbprefix('sales_payments') . ' ADD COLUMN `cash_adjustment` tinyint NOT NULL DEFAULT 0 AFTER `cash_refund`');
```
jekkos, line 12: Do we need to backfill this field for existing sales in some way?

application/models/Sale.php
```diff
@@ -1126,15 +1133,17 @@ public function create_temp_table(array $inputs)
 		$sales_tax = 'IFNULL(SUM(sales_items_taxes.sales_tax), 0)';
 		$internal_tax = 'IFNULL(SUM(sales_items_taxes.internal_tax), 0)';
 
+		$cash_adjustment = 'IFNULL(SUM(payments.sale_cash_adjustment), 0)';
+
 		if($this->config->item('tax_included'))
 		{
-			$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ')';
+			$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') - ' . $cash_adjustment;
```
jekkos, line 1140: so the adjustment is already saved as rounded amount? What if the precision changes? It might be better to save as much detail as possible and round here in the end?

application/libraries/Sale_lib.php
```diff
@@ -533,6 +550,15 @@ public function get_totals($taxes)
 		$totals['item_count'] = $item_count;
 		$totals['total_units'] = $total_units;
 
+		if($totals['payments_cover_total'])
+		{
+			$totals['cash_adjustment_amount'] = round($totals['total'] - $totals['payment_total'], totals_decimals(), PHP_ROUND_HALF_UP);
```
jekkos, line 555: Ok so we do the rounding here and then save it in the db

application/controllers/Sales.php
```diff
@@ -1037,17 +1062,26 @@ private function _reload($data = array())
 		{
 			$data['total'] = $totals['cash_total'];
 			$data['amount_due'] = $totals['cash_amount_due'];
+			$this->session->set_userdata('cash_adjustment_amount', $totals['amount_due'] - $totals['cash_amount_due']);
```
jekkos, line 1065: any reason to use a session variable for this one?

application/libraries/Sale_lib.php
```diff
@@ -533,6 +550,15 @@ public function get_totals($taxes)
...
+		}
+		else
+		{
+			$totals['cash_adjustment_amount'] = 0;
```
jekkos, line 559: you could set this one upfront (and get rid of the else statement)

SteveIreland:
The reason this isn't being resolved as quickly as I would like is because the cash adjustment payment type approach is requiring me to address a couple of new areas.  At the end of the day yesterday I had the register, the receipt, and the receipt reprint working the way I want.  The daily sales report is still looking good.  I was startled when I tested the scenario with a cash payment that required change due as well as adjusting for cash rounding.  That was an oversight on my part but has been resolved.

jekkos:
@SteveIreland indeed I think this change is a bit more involved (as it touches different parts in the app)..

SteveIreland:
I've updated this pull request with several fixes I made this morning.  But it's not ready for testing yet.  I have a problem with the "success" response not be delivered on a successful sales update and I'm checking to see if a rebuild clears that problem.  Also, I have not yet added the migration or looked to see if there is a negative impact on the cashup component.

jekkos:
Success response not showing is typically the server side that is sending a different status code (non 200)? So not sure if you were able to spot anything when you enable logging?

SteveIreland:
Thanks @jekkos, I believe I've resolved the issue with the success response.  There was an issue with one of the database updates.

SteveIreland:
Found the issue with #3041 and the next update to this pull request will include the fix for that.

SteveIreland:
This last update to this pull request also includes an attempt to address #3031.  Hopefully it works as needed.  There are so many test scenarios that could be tested but I wasn't able to get through them this weekend.

truchosky:
this will work with cash rounding -1 ? https://github.com/opensourcepos/opensourcepos/issues/2355

application/libraries/Sale_lib.php
```diff
@@ -378,11 +384,20 @@ public function add_payment($payment_id, $payment_amount)
...
+				'cash_refund' => 0, 'cash_adjustment' => $cash_adjustment));
 
 			$payments += $payment;
 		}
 
+		if($this->CI->session->userdata('cash_mode'))
+		{
+			if($this->CI->session->userdata('cash_rounding') && $payment_id != $this->CI->lang->line('sales_cash') && $payment_id != $this->CI->lang->line('sales_cash_adjustment'))
+			{
+				$this->CI->session->set_userdata('cash_mode', 0);
```
daN4cat, line 397: Should cash_mode values defined instead of using magic numbers?

daN4cat (changes requested):

daN4cat:
Was this change tested in any way outside Steve's run? @odiea ?

daN4cat:
@jekkos is this the last to close 3.3.3?

jekkos:
I have moved to 3.3.4 for now.

SteveIreland:
@truchosky I will test this week to make sure that -1 still works, but I'm pretty sure it will.

application/libraries/Sale_lib.php
SteveIreland: That correction has been made.

application/controllers/Sales.php
```diff
@@ -1257,6 +1297,20 @@ public function save($sale_id = -1)
 			$refund_type = $this->input->post('refund_type_' . $i);
 			$cash_refund = $this->input->post('refund_amount_' . $i);
 
+			if($payment_type == $this->lang->line('sales_cash_adjustment'))
+			{
+				$cash_adjustment = 1;
```
daN4cat, line 1302: Is this supposed to be 1 or one of the defines? Same comment for the 0 case below.

application/controllers/Sales.php
```diff
@@ -1283,7 +1334,24 @@ public function save($sale_id = -1)
 
 		if($payment_type != PAYMENT_TYPE_UNASSIGNED && $payment_amount <> 0)
 		{
-			$payments[] = array('payment_id' => $payment_id, 'payment_type' => $payment_type, 'payment_amount' => $payment_amount, 'cash_refund' => 0.00, 'employee_id' => $employee_id);
+			$cash_refund = 0;
+			if($payment_type == $this->lang->line('sales_cash_adjustment'))
+			{
+				$cash_adjustment = 1;
```
daN4cat, line 1340: Same comment 1 or a define? Same applies for the case 0 below.

application/views/sales/register.php
```diff
@@ -904,9 +905,9 @@ function check_payment_type() {
 		$(".non-giftcard-input").attr('disabled', true);
 		$(".giftcard-input:enabled").val('').focus();
 	}
-	else if($("#payment_types").val() == "<?php echo $this->lang->line('sales_cash'); ?>" && cash_rounding)
+	else if(($("#payment_types").val() == "<?php echo $this->lang->line('sales_cash'); ?>" && cash_mode == '1'))
```
daN4cat, line 908: should the '1' be one of the defines?

daN4cat (changes requested):

application/views/sales/register.php
SteveIreland: I didn't touch this one because it falls into the JavaScript territory.  I didn't see any precedence of magic names in JavaScript that I could replicate and I didn't want to start a new tradition without direction.  :-)

daN4cat: You are correct

odiea:
@SteveIreland is there any reason that you remove the column cash_rounding in the migration script. After it is removed there is a db error completing the sale

odiea:
I was incorrect. Apparently the migration script did not run so I had to manually add that row.

jekkos:
@SteveIreland any update on what the state is here? Do you know if it is bug free or do you need support with anything? Thanks!

objecttothis:
@SteveIreland @jekkos How close are we to merging this? Are you stuck and need help?

application/libraries/Sale_lib.php
```diff
@@ -402,11 +417,28 @@ public function edit_payment($payment_id, $payment_amount)
...
+	 * @param $payment_id
+	 */
 	public function delete_payment($payment_id)
 	{
 		$payments = $this->get_payments();
-		unset($payments[urldecode($payment_id)]);
+		$decoded_payment_id = urldecode($payment_id);
+		unset($payments[$decoded_payment_id]);
+		$cash_rounding = $this->reset_cash_rounding();
+		if($cash_rounding)
```
objecttothis, line 431: Can we add a blank line between 430 and 431 for readability?

application/libraries/Sale_lib.php
```diff
@@ -402,11 +417,28 @@ public function edit_payment($payment_id, $payment_amount)
...
+		{
+			if($decoded_payment_id == $this->CI->lang->line('sales_cash'))
+			{
+				unset($payments[$this->CI->lang->line('sales_cash_adjustment')]);
+			}
+			if($decoded_payment_id == $this->CI->lang->line('sales_cash_adjustment'))
```
objecttothis, line 437: The current code logic will always run even if the prior if statement is true.  Is that what we want?  I can't quite tell but it appears that the second if statement will never evaluate to true if the first evaluates to true, so we can replace it with elseif which will cause PHP to skip it and save the processing time.  Thoughts?

application/libraries/Sale_lib.php
```diff
@@ -416,31 +448,41 @@ public function empty_payments()
...
+			$this->CI->session->set_userdata('cash_mode', CASH_MODE_TRUE);
 		}
 
 		return $subtotal;
 	}
 
 	/**
 	 * Returns 'subtotal', 'total', 'cash_total', 'payment_total', 'amount_due', 'cash_amount_due', 'paid_in_full'
-	 * 'subtotal', 'discounted_subtotal', 'tax_exclusive_subtotal', 'item_count', 'total_units'
+	 * 'subtotal', 'discounted_subtotal', 'tax_exclusive_subtotal', 'item_count', 'total_units', 'cash_adjustment_amount'
```
objecttothis, line 482: Two thoughts.
1- If we document this, we should try to use a PHPdoc format.  I'm not sure what we have agreed on for this project (https://docstore.mik.ua/orelly/webprog/pcook/ch21_09.htm)
2- From a clean coding perspective, one of the pitfalls associated with these kinds of comments is that they can accidentally be misleading.  For example if a code contributor changes what the function returns for a good reason, but then doesn't update this list, another code contributor might look at this list to see what the function returns not knowing that the information in it is stale.  It might be better to instead document as 

`@return    array    Values needed for calculating payments in sales`

Or something similar to that.  Then in the code, as much as possible, group the associative array assignments together.

application/libraries/Sale_lib.php
```diff
@@ -1006,9 +1058,34 @@ public function copy_entire_sale($sale_id)
 			$this->add_item($row->item_id, $row->quantity_purchased, $row->item_location, $row->discount, $row->discount_type, PRICE_MODE_STANDARD, NULL, NULL, $row->item_unit_price, $row->description, $row->serialnumber, $sale_id, TRUE, $row->print_option);
 		}
 
+		$this->CI->session->set_userdata('cash_mode', CASH_MODE_FALSE);
+
+		// Establish cash_mode for this sale by inspecting the payments
+		if($this->CI->session->userdata('cash_rounding'))
+		{
+			$cash_types_only = true;
+			foreach($this->CI->Sale->get_sale_payments($sale_id)->result() as $row)
```
objecttothis, line 1067: Blank line needed between 1066 and 1067 for readability.

application/libraries/Sale_lib.php
```diff
@@ -1006,9 +1058,34 @@ public function copy_entire_sale($sale_id)
...
+			{
+				if($row->payment_type != $this->CI->lang->line('sales_cash') && $row->payment_type != $this->CI->lang->line('sales_cash_adjustment'))
+				{
+					$cash_types_only = FALSE;
+				}
+
+			}
+			if($cash_types_only)
```
objecttothis, line 1075: Blank line needed between 1074 and 1075 for readability

application/libraries/Sale_lib.php
```diff
@@ -1051,18 +1128,13 @@ public function clear_cash_flags()
 		$this->CI->session->unset_userdata('payment_type');
 	}
 
-	public function reset_cash_flags()
+	/**
+	 * Determines if cash rounding should be a consideration for this site
+	 * It also set resets the cash mode to disabled which will then be re-evaluated when
```
objecttothis, line 1133: From a clean coding perspective, functions should do one thing as much as possible.  There may be risk in having one function both determine whether cash rounding should be used on the site AND reset the cash mode to disabled but then call the function "reset_cash_rounding".  Food for thought... Perhaps @jekkos or @daN4cat have an opinion.

application/libraries/Sale_lib.php
```diff
@@ -1249,9 +1329,12 @@ public function get_total()
 			}
 		}
 
-		if($cash_rounding)
+		if($include_cash_rounding)
 		{
-			$total = $this->check_for_cash_rounding($total);
+			if($cash_mode)
```
objecttothis, line 1334: rather than nesting an if statement inside another if statement, it might be better to combine the two

`if($include_cash_rounding && $cash_mode)`

application/models/Sale.php
```diff
@@ -149,9 +159,10 @@ public function search($search, $filters, $rows = 0, $limit_from = 0, $sort = 's
 		$tax = 'IFNULL(SUM(sales_items_taxes.tax), 0)';
 		$sales_tax = 'IFNULL(SUM(sales_items_taxes.sales_tax), 0)';
 		$internal_tax = 'IFNULL(SUM(sales_items_taxes.internal_tax), 0)';
+		$cash_adjustment = 'IFNULL(SUM(payments.sale_cash_adjustment), 0)';
 
 		$sale_subtotal = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') - ' . $internal_tax;
-		$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') + ' . $sales_tax;
+		$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') + ' . $sales_tax . ' + ' . $cash_adjustment;
```
objecttothis, line 165: Rather than using the concatenation operator (.), this is a great place to instead use the PHP parser by replacing ' with " and ditching the concatination.  165 becomes

`$sale_total = "ROUND(SUM($sale_price)), $decimals) + $sales_tax + $cash_adjustment";` 

IMO is much more readable.  Thoughts?  You might also change 164 and other lines to do the same thing.

objecttothis: You're doing it in line 211 as an example if that helps.

application/models/Sale.php
```diff
@@ -550,17 +561,19 @@ public function update($sale_id, $sale_data, $payments)
...
 				if($payment_id == -1 && $payment_amount != 0)
 				{
 					// Add a new payment transaction
 					$sales_payments_data = array(
-						'sale_id' => $sale_id,
-						'payment_type' => $payment_type,
-						'payment_amount' => $payment_amount,
-						'cash_refund' => $cash_refund,
-						'employee_id' => $employee_id
+						'sale_id'		  => $sale_id,
```
objecttothis, line 571: I personally love using the tab to line up all the `=>` operators and wish that we could do it everywhere.  What I'm finding is that unfortunately various IDEs don't render the tabs properly and they end up looking like garbage.  It pains me to say but we should probably stick to single spaces before and after the `=>` operators rather than tabs before and single space after.  Thoughts?

application/models/Sale.php
```diff
@@ -1126,15 +1140,17 @@ public function create_temp_table(array $inputs)
 		$sales_tax = 'IFNULL(SUM(sales_items_taxes.sales_tax), 0)';
 		$internal_tax = 'IFNULL(SUM(sales_items_taxes.internal_tax), 0)';
 
+		$cash_adjustment = 'IFNULL(SUM(payments.sale_cash_adjustment), 0)';
+
 		if($this->config->item('tax_included'))
 		{
-			$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ')';
+			$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') + ' . $cash_adjustment;
```
objecttothis, line 1147: Same as my earlier comment about using the PHP parser instead of concatenation operator.

application/models/Sale.php
```diff
@@ -1126,15 +1140,17 @@ public function create_temp_table(array $inputs)
...
 			$sale_subtotal = $sale_total . ' - ' . $internal_tax;
 		}
 		else
 		{
-			$sale_subtotal = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') - ' . $internal_tax;
-			$sale_total = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') + ' . $sales_tax;
+			$sale_subtotal = 'ROUND(SUM(' . $sale_price . '), ' . $decimals . ') - ' . $internal_tax . ' + ' . $cash_adjustment;
```
objecttothis, line 1152: Here too.

objecttothis (changes requested):
Great work @SteveIreland.  See my comments.

objecttothis:
@jekkos @daN4cat I'm starting on this today.  I got a bit of a late start.  I'll begin with my code review, then resolve the other code review comments and begin testing against the issues list to make sure this fixes all of those bugs.

application/controllers/Sales.php
```diff
@@ -5,6 +5,11 @@
 define('PRICE_MODE_STANDARD', 0);
 define('PRICE_MODE_KIT', 1);
 define('PAYMENT_TYPE_UNASSIGNED', '--');
+define('CASH_ADJUSTMENT_TRUE', 1);
```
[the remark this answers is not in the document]
jekkos: Ok yes let's keep it in one place for now. Ideally we separate it out further later on..

application/controllers/Sales.php
```diff
@@ -348,6 +353,20 @@ public function add_payment()
 					}
 				}
 			}
+			elseif($payment_type == $this->lang->line('sales_cash'))
```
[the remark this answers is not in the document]
SteveIreland: I changed it to use the identical and not identical operators for comparisons involving payment type.   And didn't take it any further lest things start breaking.

application/controllers/Sales.php
```diff
@@ -922,18 +943,18 @@ private function _load_sale_data($sale_id)
...
-		$data['cash_rounding'] = $this->session->userdata('cash_rounding');
+		$data['cash_mode'] = $this->session->userdata('cash_mode');
 		$data['prediscount_subtotal'] = $totals['prediscount_subtotal'];
 		$data['cash_total'] = $totals['cash_total'];
 		$data['non_cash_total'] = $totals['total'];
 		$data['cash_amount_due'] = $totals['cash_amount_due'];
 		$data['non_cash_amount_due'] = $totals['amount_due'];
 
-		if($this->session->userdata('cash_rounding'))
+		if($data['cash_mode'] && ($data['selected_payment_type'] == $this->lang->line('sales_cash') || $data['payments_total'] > 0))
```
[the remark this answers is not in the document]
SteveIreland: done

application/controllers/Sales.php
```diff
@@ -1013,6 +1035,10 @@ private function _reload($data = array())
 			$sale_id = -1;
 			$this->session->set_userdata('sale_id', -1);
 		}
+		$cash_rounding = $this->sale_lib->reset_cash_rounding();
+		// cash_rounding indicates only that the site is configured for cash rounding
```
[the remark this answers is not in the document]
SteveIreland: done

application/libraries/Sale_lib.php
SteveIreland: done

application/models/Sale.php
SteveIreland: I would prefer not to spend time time fussing with tabs to line things up but I am good either way.

However, that's a decision that's above my pay grade.

objecttothis: To clarify, my understanding of the project practice is that we aren't worrying about adding tabs left and right of the => operator.  Just tabs left of the line being important.

SteveIreland: I recall being asked once to line them up, but that was years ago.  Things change.

application/controllers/Sales.php
```diff
@@ -1029,20 +1055,25 @@ private function _reload($data = array())
...
 		$totals = $this->sale_lib->get_totals($tax_details[0]);
+
 		$data['item_count'] = $totals['item_count'];
 		$data['total_units'] = $totals['total_units'];
 		$data['subtotal'] = $totals['subtotal'];
 		$data['total'] = $totals['total'];
 		$data['payments_total'] = $totals['payment_total'];
 		$data['payments_cover_total'] = $totals['payments_cover_total'];
-		$data['cash_rounding'] = $this->session->userdata('cash_rounding');
+		// cash_mode indicates whether this sale is going to be processed using cash_rounding
```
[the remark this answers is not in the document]
SteveIreland: There were a handful in Sales.php that weren't preceded by a blank line, so I "fixed" them all.  however, if it was preceded by an open or close brace then I figured that was close enough to a blank line.

application/models/Sale.php
objecttothis: @jekkos @daN4cat What do you want to do here?  Do we align the left side of the => operators with tabs or have a space on either side.  Aligning with tabs looks better IMO, but it doesn't always render lined up on all IDEs.  I've seen both in the code, but consistency is probably important.

daN4cat: Let's not get too bothered about this detail. It's fine like it is, there are plenty of other examples where on this specific case we followed different approaches.
If one day I have time to do some cosmetic work, I'll give a pass to everything.

objecttothis: OK

objecttothis (approved):
Looks like that's everything. @SteveIreland I think you said that the PR already resolves the issues that it set out to.

SteveIreland:
After some testing to make sure that I didn't fat finger the corrections, I've pushed it back up. But there is still an issue with correcting payments if the user does something odd like making a partial cash payment against an invoice and then enters a subsequent credit card payment (which should invalidate the cash adjustment).  However, the payment adjustment functionality isn't currently robust enough to handle that scenario.  I believe I'd still like to get this pull request merged and then I'll follow up with another change to improve the payment adjustment to handle the odd scenario (rather than dragging this change out any longer).

objecttothis:
> After some testing to make sure that I didn't fat finger the corrections, I've pushed it back up. But there is still an issue with correcting payments if the user does something odd like making a partial cash payment against an invoice and then enters a subsequent credit card payment (which should invalidate the cash adjustment). However, the payment adjustment functionality isn't currently robust enough to handle that scenario. I believe I'd still like to get this pull request merged and then I'll follow up with another change to improve the payment adjustment to handle the odd scenario (rather than dragging this change out any longer).


Agreed. It seems like making it more robust falls more in the category of enhancement than bugfix.

SteveIreland:
Okay, I have a travis-ci issue and I'm in deeper water than I'm used to.  I can't find anything that is code related and what I can find suggests that it's a permission issue.  Can someone provide some insight?  Is this related to travis-ci shutting down soon?

objecttothis:
> 
> 
> Okay, I have a travis-ci issue and I'm in deeper water than I'm used to. I can't find anything that is code related and what I can find suggests that it's a permission issue. Can someone provide some insight? Is this related to travis-ci shutting down soon?

This seems to indicate that the build passed (https://travis-ci.org/github/opensourcepos/opensourcepos/builds/757954759).  What exactly is the issue that's happening?

SteveIreland:
Okay, something changed and it was restarted and completed.  I am NOT insane.  I know there was an issue.  I knew I should have taken an image of the screen.  Someone is gas-lighting me.  :-)  

I'm ready to merge unless someone objects to this.

objecttothis:
> I'm ready to merge unless someone objects to this.

@SteveIreland I see what you did there. ;) Merge away!

jekkos:
> > Okay, I have a travis-ci issue and I'm in deeper water than I'm used to. I can't find anything that is code related and what I can find suggests that it's a permission issue. Can someone provide some insight? Is this related to travis-ci shutting down soon?
> 
> This seems to indicate that the build passed (https://travis-ci.org/github/opensourcepos/opensourcepos/builds/757954759). What exactly is the issue that's happening?

I have restarted the build and it went fine after.. there is a rate limitation on docker hub usage and it seems to kick in now and then..

Issue #3016, opened 2020-11-17
Default register mode question


chunter2:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.2
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Is there a way to set the default register mode in the sales module?  We switched it to return, ran a return but forgot to switch it back to sales.  I'd be nice to have it go back to sales by default.  

Thanks

WebShells:
Hello there, it's available there but i guess there is an issue with the functionality needs to be re-checked. 
You can set it there from the configuration section.

[...]

PR #3040, opened 2020-12-04
Fix default register mode (#3016)

jekkos:
This PR moves the config for default sale register mode to general, as it was under invoice before.

### DIFF
--- a/application/controllers/Config.php
+++ b/application/controllers/Config.php
@@ -817,7 +817,6 @@ class Config extends Secure_Controller
 	{
 		$batch_save_data = array (
 			'invoice_enable' => $this->input->post('invoice_enable') != NULL,
-			'default_register_mode' => $this->input->post('default_register_mode'),
 			'sales_invoice_format' => $this->input->post('sales_invoice_format'),
 			'sales_quote_format' => $this->input->post('sales_quote_format'),
 			'recv_invoice_format' => $this->input->post('recv_invoice_format'),
--- a/application/libraries/Sale_lib.php
+++ b/application/libraries/Sale_lib.php
@@ -591,18 +591,10 @@ class Sale_lib
 
 	public function get_mode()
 	{
-		if(!$this->CI->session->userdata('sales_mode'))
-		{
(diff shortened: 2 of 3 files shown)


SteveIreland (commented):
I downloaded and tested it.  It worked fine for me, except that I did notice a very minor issue that may not be worth the effort to fix.  The issue is that if I enable or disable the invoice feature, the register mode options under General are not refreshed to reflect the changes until after I exit config and return (or simply click on the Config icon to force a refresh).

jekkos:
Code adapted as per the discussion. Will do a quick test to see if this works now.

chunter2:
@jekkos I just tried applying this change and when I process a return the sales module comes back up set to return.  The same thing happens when I click cancel while processing a return.  I'm assuming this is supposed to go back to sale in both cases right?

jekkos:
Yes I would expect it to reset (as the mode is cleared once the sale is completed)

jekkos:
I did a quick test locally and it appears to work fine on my end. The register resets to sale after I complete a return or invoice.

chunter2:
@jekkos I wonder if it has anything to do with me applying it on the 3.3.2 release.  Could there be a change between that release and now that's affecting things?  I tried applying the change without restarting apache and then with restarting.  Or could it be something with my browser cache?

jekkos:
@chunter2 can you check if you have the line in the `clear_all` method to clear the mode? That one is crucial to make it work. I have checked again and can confirm it's fine on my side. Will merge and then let's try on dev server after.

chunter2:
@jekkos I was missing the clear_all.  Sorry about that.  In my switching the code back and forth between your first commit and this one I missed re-adding that line.  It seems to be working as expected now.  Thanks again.

jekkos:
Ok thanks for the testing and confirmation.

lajeeshk:
As per the discussion what I understand, configuration for default register mode is moved to general tab. But I can't see that, can someone give me a screenshot?

odiea:
That option was removed from the Configuration files.

Issue #3044, opened 2020-12-16
Click quickly twice submit button and entries are duplicated


daN4cat:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.3.latest_master
- OSPOS git commit hash is: 
- PHP version is: 7.3
- MySQL or MariaDB version is: 10.3
- OS and version is: CentOs
- WebServer is: Apache
- Selected language is: en_GB
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Scenario:

- add 1 customer with email address
- click the Submit button quickly twice
- dialog box closes
- two pop up notifications
- two customer accounts created with the same email address violating the email uniqueness

Submit button must be deactivate after the first click.

[...]

PR #3045, opened 2020-12-16
Prevent double submit (#3044)

jekkos:
@daN4cat this should fix your issue. Tried it locally and it does seem to prevent a double submit from happening. 
Also verfied the batch insert and it seems fine.
Mind that the minified js needs to be rebuilt.


daN4cat (approved):
Tested and works as expected.

Issue #1980, opened 2018-05-07
Add SameSite cookie flag for CSRF protection


jekkos:
Most modern browsers now support the SameSite cookie flag, which signals that a site only accepts requests coming from the same domain. This might be an effective hardening measure against CSRF, as Ospos is not used in cross domain setups. This constraint will then of course only be enabled for users that have browsers with support for this flag. 

* Need to check on a browser compatibility table for this feature
* Identify possible usecases in which ospos can be used across different domains (none?)
* Check if CI framework offers this protection or add a custom hook to add it to the Cookie headers

jekkos:
@SteveIreland I think in the longer term it would probably relieve us from sending tokens along with the forms. The ajax case is a bit more complicated as these are done by javascript and thus no php support can be leveraged for it. Beside that I tried to make it very generic by hooking the standard jQuery functionality to fetch the token and send it along, but it might not work for every possible scenario in which it's submitting forms (like at login time).

jekkos:
This would also allow us to enable #1498, as if SameSite works, we don't need to use js to read data from the cookies anymore, meaning HttpOnly can be enabled without hiccups.

jekkos:
It seems that this can be enabled by editing the cookiepath in config.php

https://stackoverflow.com/questions/60634341/codeigniter-3-samesite-attribute-for-csrf-protection

$config['cookie_path']      = '/; SameSite=None';

I'm seeing the following message now locally in FF

Cookie “csrf_cookie_ospos_v3” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

jekkos:
Ok this last approach does not work and causes some issues. I have added a method override that now adds the SameSite attribute. Should be harmless for browser that don't support it yet.

objecttothis:
Agreed. We can reopen later if we want to figure out how to close it further.

daN4cat:
Not sure I follow, why was this closed?

objecttothis:
Sorry, perhaps I closed the wrong issue.  Maybe #1498 is the one that needs to be closed.  I was closing it because I thought the issue was resolved with PR #3047 fixing the issue.  Sorry for the confusion.

Issue #1535, opened 2017-08-27
CSRF error at login time


daN4cat:
### Issue / Question / Bug

Before submitting an issue please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] I'm reporting an issue of an unmodified OSPOS installation
- [x] I checked open and closed issues database and no similar issue was already discussed (please make sure you searched!)
- [x] I read the README, WHATS_NEW and UPGRADE
- [x] I read the FAQ (https://github.com/jekkos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP has php-gd, php-intl, sockets and etc. installed)
- [x] I read the wiki
- [x] I ran any database upgrade scripts (e.g. database/2.4_to_3.0.sql), and migrating function
- [x] I'm aware the latest master could be a development version and therefore not stable
- [x] I know the version of OSPOS and git commit hash (check the footer of your OSPOS), the name and version of OS, Web server, PHP and MySQL and will add them to my issue report

### Installation information

- OSPOS version is: 3.1.0
- OSPOS git commit hash is: 
- PHP version is: 7.1
- MySQL or MariaDB version is: MySQL 5.6
- OS and version is: CentOS 6.9
- WebServer is: Apache 2.4
- (If applicable) Installation package for the LAMP/LEMP stack is:  N/A

### Expected behaviour

* Logout from OSPOS on day 1, login from same login page (no refresh) on day 2
* Enter user credentials into login page
* (ReCAPTCHA tick box ok)
* Enter OSPOS home page

### Actual behaviour

* Logout from OSPOS on day 1, login from same login page (no refresh) on day 2
* Enter user credentials into login page
* (ReCAPTCHA tick box ok)
* `403 Forbidden` Error due to invalid csrf token:
![image](https://user-images.githubusercontent.com/636048/29752543-0e4897d6-8b58-11e7-863e-ae16dcd6929d.png)

### Steps to reproduce the issue

* Use a HTTPS connection
* Logout from OSPOS on day 1, login from same login page (no refresh) on day 2
* Enter user credentials into login page
* (ReCAPTCHA tick box ok)

daN4cat:
@jekkos the CSRF error at login time persists. I can systematically reproduce it now.
Probably 1 day is not necessary but try waiting few hours and I bet you can reproduce it.

jekkos:
[...]

Issue #1498, opened 2017-08-22
$config['cookie_httponly']


objecttothis:
### New Feature / Enhancement
During my troubleshooting of CSRF and my server configuration I noticed that `$config['cookie_httponly']` is set to FALSE by default.  As far as I can tell the app is working fine with it enabled.  Is there a reason why it was set to false?  Setting it to true has the added benefit of the CSRF cookies (edited: and all other cookies) being appended with HttpOnly (see below)

`Set-Cookie | csrf_cookie_ospos_v3=EgKo_TvbI…ax-Age=7200; path=/; HttpOnly`

This is a good security measure to enable if we can as it mitigates many XSS attacks.

### Installation information

- OSPOS version is: 3.1.0
- OSPOS git commit hash is: 4f5ad57
- PHP version is:   (e.g. 5.5, 5.6, 7.0, 7.1) 5.6.31
- MySQL or MariaDB version is:   (e.g. MySQL 5.5, MySQL 5.6, MySQL 5.7, MariaDB 10.0, MariaDB 10.1, MariaDB 10.2) MySQL 5.7
- OS and version is:   (e.g. CentOS 6.9, Ubuntu 16.4, Windows 10) FreeBSD 11.0
- WebServer is:   (e.g. Apache 2.2, Apache 2.4, Nginx 1.12, Nginx 1.13) Apache 2.4
- (If applicable) Installation package for the LAMP/LEMP stack is:   (e.g. WAMP, XAMPP) FAMP 

_Sent from my Android XT1575 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jekkos:
Think we'll need to do some regression testing as we rely on reading cookie data in javascript for some types of form submissions.. This might maybe break now so we'll need to verify first. Otherwise then I agree we should turn on this mitigation

objecttothis:
@jekkos ok I wasn't aware that we were reading cookie data with JavaScript. I'm guessing it will break that, but regression testing should be able to quickly test that.

_Sent from my Android XT1575 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

daN4cat:
XSS attack is another story, you can inject any js before reaching the user web browser and execute operations in the background on the user web browser.
See https://www.owasp.org/index.php/Types_of_Cross-Site_Scripting

[...]

PR #3047, opened 2020-12-18
Add SameSite=Strict attribute + enable HTTPOnly for session (#1980)

jekkos:
Added an override of the csrf token generation that includes SameSIte flag. Checked locally and indeed this attribute is being set on the cookie now.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -437,7 +437,7 @@ $config['sess_regenerate_destroy'] = FALSE;
 $config['cookie_prefix'] = '';
 $config['cookie_domain'] = '';
 $config['cookie_path'] = '/';
-$config['cookie_secure'] = FALSE;
+$config['cookie_secure'] = isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on';
 $config['cookie_httponly'] = TRUE;
 
 /*
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -438,7 +438,7 @@ $config['cookie_prefix'] = '';
 $config['cookie_domain'] = '';
 $config['cookie_path'] = '/';
 $config['cookie_secure'] = FALSE;
-$config['cookie_httponly'] = FALSE;
+$config['cookie_httponly'] = TRUE;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/core/MY_Security.php
+++ b/application/core/MY_Security.php
@@ -1,0 +1,35 @@
+<?php
+defined('BASEPATH') or exit('No direct script access allowed');
+
+class MY_Security extends CI_Security
+{
+
+    /**
+     * CSRF Set Cookie with samesite
+     *
+     * @codeCoverageIgnore
+     * @return  CI_Security
+     */
+    public function csrf_set_cookie()
+    {
+        $expire = time() + $this->_csrf_expire;
+        $secure_cookie = (bool)config_item('cookie_secure');
+
+        if ($secure_cookie && !is_https()) {
+            return FALSE;
+        }
+
+        setcookie($this->_csrf_cookie_name,
+            $this->_csrf_hash,
+            ['samesite' => 'Strict',
+                'secure' => config_item('cookie_httponly'),
+                'expires' => $expire,
+                'path' => config_item('cookie_path'),
+                'domain' => config_item('cookie_domain'),
+                'httponly' => config_item('cookie_httponly')]);
+
+        log_message('info', 'CSRF cookie sent');
+
+        return $this;
+    }
+}
--- a/application/core/MY_Security.php
+++ b/application/core/MY_Security.php
@@ -19,14 +19,26 @@ class MY_Security extends CI_Security
             return FALSE;
         }
 
-        setcookie($this->_csrf_cookie_name,
-            $this->_csrf_hash,
-            ['samesite' => 'Strict',
-                'secure' => $secure_cookie,
-                'expires' => $expire,
-                'path' => config_item('cookie_path'),
-                'domain' => config_item('cookie_domain'),
-                'httponly' => FALSE]);
+        if (PHP_VERSION_ID < 70300) {
+            setcookie($this->_csrf_cookie_name,
+                $this->_csrf_hash, $expire,
+                config_item('cookie_path'). '; samesite=strict',
+                config_item('cookie_domain'),
+                $secure_cookie,
+                FALSE);
+        }
+        else
+        {
+            setcookie($this->_csrf_cookie_name,
+                $this->_csrf_hash,
+                ['samesite' => 'Strict',
+                    'secure' => $secure_cookie,
+                    'expires' => $expire,
+                    'path' => config_item('cookie_path'),
+                    'domain' => config_item('cookie_domain'),
+                    'httponly' => FALSE]);
+        }
+
 
         log_message('info', 'CSRF cookie sent');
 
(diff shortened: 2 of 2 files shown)


objecttothis:
Is this related to #1498 and #1980?

application/core/MY_Security.php
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @codeCoverageIgnore
+     * @return  CI_Security
+     */
+    public function csrf_set_cookie()
+    {
+        $expire = time() + $this->_csrf_expire;
+        $secure_cookie = (bool)config_item('cookie_secure');
+
+        if ($secure_cookie && !is_https()) {
```
objecttothis, line 18: I believe we decided on the standard of having curly braces { on their own line.

application/core/MY_Security.php
```diff
@@ -0,0 +1,35 @@
+<?php
+defined('BASEPATH') or exit('No direct script access allowed');
+
+class MY_Security extends CI_Security
+{
+
```
objecttothis, line 6: I would remove this blank line.

objecttothis (changes requested):
Just the two minor formatting changes.  I haven't had a chance to download the build and test it, but I assume that you've done that.

application/core/MY_Security.php
jekkos: fixed

application/core/MY_Security.php
jekkos: done

jekkos:
@objecttothis indeed. I have enabled HTTPOnly flag by default now for the session id. This is a flag that can be set per cookie. I think this should make JS based exploits quite a bit harder as we only leave the csrf cookie value readable by Javascript.

Also the latter now carries SameSite flag, which means that it won't be posted to any external attacker domain in case a XSS exploit is possible.

application/core/MY_Security.php
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        setcookie($this->_csrf_cookie_name,
+            $this->_csrf_hash,
+            ['samesite' => 'Strict',
+                'secure' => $secure_cookie,
+                'expires' => $expire,
+                'path' => config_item('cookie_path'),
+                'domain' => config_item('cookie_domain'),
+                'httponly' => FALSE]);
```
jekkos, line 29: @objecttothis  so in this case we have no choice but to set httpOnly to false (we need to read this value in JS in order to post it back  using AJAX). However we do have SameSite security now (so we won't leak this token to other domains)

objecttothis: Understood.  My understanding is that the attack vector would be limited to malicious code installed on the same domain then.  That is still definitely a possibility but much better than an attack vector open from all domains.  In terms of #1498 we would need to figure out another way to do CSRF with AJAX, so let's keep that ticket open but close #1980 with this one then.

application/config/config.php
```diff
@@ -438,7 +438,7 @@
 $config['cookie_domain'] = '';
 $config['cookie_path'] = '/';
 $config['cookie_secure'] = FALSE;
-$config['cookie_httponly'] = FALSE;
+$config['cookie_httponly'] = TRUE;
```
objecttothis, line 441: Did you mean to push this commit?  Doesn't this break AJAX?

objecttothis (changes requested):

application/config/config.php
jekkos: Yes, this can be set to true now and will only be applied to the php session id

@objecttothis indeed. I have enabled HTTPOnly flag by default now for the session id. This is a flag that can be set per cookie. I think this should make JS based exploits quite a bit harder as we only leave the csrf cookie value readable by Javascript.

objecttothis (approved):

objecttothis:
@daN4cat perhaps this allows you to re-enable CSRF check on the login? #1535

application/config/config.php
objecttothis: @jekkos should we attach #1498 to this and close it then, or leave it as a placeholder reminder to figure out how to get AJAX and CSRF to place nice together?

odiea:
With error on I received the following. I changed config and added my security to the bower update download. Did I miss something?
A PHP Error was encountered
Severity: Warning

Message: setcookie() expects parameter 3 to be integer, array given

Filename: core/MY_Security.php

Line Number: 24
Backtrace:
C:\wamp64\www\bowerup\application\core\MY_Security.php
Line: 24
Function: setcookie
C:\wamp64\www\bowerup\public\index.php
Line: 315
Function: require_once

odiea:
This requires php 7.3 to work as an addon.

odiea:
To get my current branch to work and not show errors with PHP 7.2 and also works in 7.3. As per PHP manual.
```
setcookie($this->_csrf_cookie_name,
            $this->_csrf_hash,(int)
            ['samesite' => 'Strict',
            ['secure' => $secure_cookie,
            ['expires' => $expire,
            ['path' => config_item('cookie_path'),
            ['domain' => config_item('cookie_domain'),
            ['httponly' => FALSE]]]]]]);
```

jekkos:
Indeed it seems that there are two versions of the setcookie() function in PHP. The one we use now is only available in PHP 7.3. Good job spotting that.

Bottom line is that this SameSite will only work with the newer version of this function, as only that one allows you to set cookie attributes using an array. The older version does not offer a boolean flag to set this property, so if we want to support 7.2 we will need to find a workaround here.

I guess we were fine bumping the minimum version to 7.1. Not sure if we can agree to push this to 7.3 for this issue alone.

daN4cat:
7.1 as 5.6 are deprecated.

jekkos:
Ok indeed so let's stick with 7.2 as a minimum then. I have added  a workaround for the issue now. As there is no one size fits all solution a version detection is needed. The 7.4 version does not seem to allow the trick of adding an extra attribute after the path. According to the stackoverflow post it should work for 7.2 but can't test that for now.

application/core/MY_Security.php
```diff
@@ -19,14 +19,26 @@ public function csrf_set_cookie()
...
-        setcookie($this->_csrf_cookie_name,
-            $this->_csrf_hash,
-            ['samesite' => 'Strict',
-                'secure' => $secure_cookie,
-                'expires' => $expire,
-                'path' => config_item('cookie_path'),
-                'domain' => config_item('cookie_domain'),
-                'httponly' => FALSE]);
+        if (PHP_VERSION_ID < 70300) {
+            setcookie($this->_csrf_cookie_name,
```
jekkos, line 23: @odiea we run the old function version in case we're on a version older than 7.3 here

jekkos:
Ok I do see the warning in FF that setting SameSite does not make sense if Secure flag for the cookie is not set. We can thus only set it in case we are on HTTPS. I sugges to set SECURE flag to true in case we are on HTTPS and then together with that the SameSite. Adaptations for this are done in the last commit. I still need to test on HTTPS but basically most of the security improvements should be activated automatically if SSL is setup.

application/core/MY_Security.php
jekkos: If you find some opportunity to test, let me know. Otherwise let's proceed and merge.

application/config/config.php
jekkos: I think indeed that issue can be closed, we enable httpOnly for all cookies except the CSRF token.. I think it's a fairly strong security improvement.

jekkos:
@daN4cat woops I also updated the install instructions here

daN4cat:
Oh well....

PR #3049, opened 2020-12-21
Use jquery.validate to check form submission state (#3044)

jekkos:
This seems to fix the outstanding issue with the form not closing after failed validation.
I have tried and did not succeed in doing a double submit.

I'm reading jquery validation internal state now to cancel a second submit in case it is triggered.


daN4cat (approved):
Tested and it's fine. Thanks.

Issue #3057, opened 2020-12-29
Search box in Customers, Items and etc. not working on mobile phones


daN4cat:
Latest js plugins upgrade caused an issue with ospos on mobile phones, you cannot use the search box anymore

This needs to be fixed before we can release 3.3.3.

@jekkos I don't have that much time next couple of days, can you have a look?

jekkos:
Ok any clue what might be wrong (error message or sth?)

daN4cat:
What happens is that as soon as you press the search box, the keyboard pops up on the phone and disappears immediatly after with the table refreshing.

jekkos:
Ok I have pushed something to dev server, let's check if the issues persists. Basically removed the resize trigger. I think it can of goes into a loop, keyboard pops up, resize is triggered that triggers the refresh, which in turn hides the keyboard again.

daN4cat:
Ok, I'll test it later. Thanks

daN4cat:
This is fixed, closing it.

PR #3060, opened 2020-12-30
Disable resize trigger (#3057)

jekkos:
Trying this fix to see if it works on mobile devices

### DIFF
--- a/application/views/reports/tabular.php
+++ b/application/views/reports/tabular.php
@@ -58,7 +58,6 @@
 				escape: false
 		});
 
-		table_support.init_resize();
 
 	});
 </script>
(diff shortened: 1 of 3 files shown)


jekkos:
Fix on dev seems to solve the issue. However I'm currently not able to test the sticky headers after resize (as I don't have enough data on dev, it's basically an empty db).

daN4cat:
I fixed an issue and tested, works fine and also sticky headers are fine.

I merge and many thanks, I wonder what the table mobile plugin does that we haven't enabled yet.

daN4cat (commented):
Tested and fine.

Issue #3006, opened 2020-11-06
OSPOS goes back to login page when sales items reaches ~120-140


mgmamac:
Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

Installation information

- OSPOS version is: 3.3.2(release)
- OSPOS git commit hash is: bb309c2 (Is this it? I'm not sure.)
- PHP version is: 7.3.21
- MySQL or MariaDB version is: 5.7.31
- OS and version is: Windows 10 64bit 
- WebServer is: Wampserver 3.2.3 64bit
- Selected language is: 1909
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

Issue / Bug  / Question / New Feature

[...]

PR #3068, opened 2021-01-02
Use mediumblob for session data storage (#3006)

jekkos:
Use MEDIUMBLOB type for session data storage. This should increase the max amount of items that can be stored in a cart.

### DIFF
--- a/application/migrations/20210103000000_modify_session_datatype.php
+++ b/application/migrations/20210103000000_modify_session_datatype.php
@@ -1,0 +1,24 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_modify_session_datatype extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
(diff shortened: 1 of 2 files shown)


application/migrations/20200819000000_modify_session_datatype.php
```diff
@@ -0,0 +1,23 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_modify_session_datatype extends CI_Migration
```
daN4cat, line 3: date of the file name is wrong, shouldn't it be 20210102000000....

daN4cat (changes requested):
See my review comment

application/migrations/20200819000000_modify_session_datatype.php
jekkos: Ok filename adapted.

jekkos:
@daN4cat should be good now

PR #3069, opened 2021-01-03
Contribution dates rearranged

WebShells:
Updated & Contribution dates rearranged.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,12 +1,12 @@
 The MIT License (MIT)
 
-Copyright (c) 2021-2013 jekkos
-Copyright (c) 2021-2015 FrancescoUK (aka daN4cat)
-Copyright (c) 2021-2017 Steve Ireland
-Copyright (c) 2021-2017 objecttothis
-Copyright (c) 2021-2017 odiea
-Copyright (c) 2021-2017 WebShells
-Copyright (c) 2021-2020 Andriux1990
+Copyright (c) 2013-2021 jekkos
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,30 +1,30 @@
...
-Copyright (c) 2015 Aamir Shahzad (aka asakpke), RoshanTech.com
-Copyright (c) 2015 Toni Haryanto (aka yllumi)
-Copyright (c) 2016-2017 Ramkrishna Mondal (aka RamkrishnaMondal)
-Copyright (c) 2016 Rinaldy@dbarber (aka rnld26)
-Copyright (c) 2016-2017 Jorge Colmenarez (aka jlctmaster), frontuari.com
-Copyright (c) 2017-2021 Steve Ireland
-Copyright (c) 2017-2021 objecttothis
-Copyright (c) 2017-2021 odiea
-Copyright (c) 2017-2021 WebShells / Shady Sh
+Copyright (c) 2021-2013 jekkos
```
daN4cat, line 3: I'm afraid this is not the correct approach

daN4cat (changes requested):
Please fix it

LICENSE
WebShells: @daN4cat you mean the format "current year - start year" ? I tried sorting them as "start year - current year"  but the display looked mixed up. It's the only way i found to sort them in descending order to go down from 2021 till 2012.

daN4cat: Agreed but that's not correct, also no idea why we need to sort this list. However, I'm not fussy about it.

PR #3080, opened 2021-01-18
Update Specific_supplier.php

printgeek:
If items are grouped by sale_id then it will only show one line for each sale, even if multiple items are sold from supplier. Grouping by item_id will show multiple lines for each sale_id where an item from supplier is sold, thereby showing every item and matching the totals at the bottom of the report.


jekkos (approved):

jekkos:
Thanks for the fix. Merged in master.

Issue #3082, opened 2021-01-18
Cookie secure flag should be automatically configured


jekkos:
### Installation information

- OSPOS version is: 3.3.3
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

On current demo the secure flag is not set properly, which is probably due to the https detection not being triggered. The issue is with setups that use a reverse proxy and thus do not have SSL flag necessarily set to true. We should try to find an alternative way or if there is none, let the user configure the flag himself.

It seems that this impacts chrome users as the browser now ignores the cookie (as it has sameSite flag set but not secure). This might be a security measure that is a bit overly defensive here.

daN4cat:
There is an env variable to force the https with reverse proxy scenarios. Is that considered in the new part of the code?

jekkos:
I can use that but I wonder if it is also set in case you use it outside of docker (with a reverse proxy upfront)? I guess the only way to know is to try it

daN4cat:
That env var is already used in the config file if my memory is right.

jekkos:
Ok I made a small PR for this one. Let's see if it works once deployed on dev.

daN4cat:
BTW, I don't seem to find any PR. Just a commit to a dev branch.

jekkos:
Ok seems to be fixed now on dev. The build is fixed as well, I pinged those jsPdf guys and they tagged their release. So should be good now.

I tested with chrome and am able to post, checked that secure flag is enabled and it's fine.

PR #3083, opened 2021-01-19
Fix cookie secure flag (#3082)

jekkos:
This change extracts and reuses the https_on variable to enable cookie secure flag.


jekkos:
Quite a crucial fix I think..

PR #3090, opened 2021-02-03
Add item_number to item kits module

jekkos:
This is a feature in item kits that an explicit item_number to the table so unique barcodes can be assigned apart from the regular item id.

Initially I built this for a donation from someone that might just have found the build on the server and after I never heard him again. As I just have it lying around here I thought it would still be good if it benefits others (as I won't use it anyway).


application/controllers/Item_kits.php
```diff
@@ -116,6 +117,7 @@ public function view($item_kit_id = -1)
 			$item['item_id'] = $this->xss_clean($item_kit_item['item_id']);
 			$item['quantity'] = $this->xss_clean($item_kit_item['quantity']);
 
+
```
objecttothis, line 120: unneeded extra blank line.

application/controllers/Sales.php
```diff
@@ -401,7 +401,8 @@ public function add()
 		{
 			// Add kit item to order if one is assigned
 			$pieces = explode(' ', $item_id_or_number_or_item_kit_or_receipt);
-			$item_kit_id = $pieces[1];
+
+			$item_kit_id = (count($pieces) > 1)	? $pieces[1] : $item_id_or_number_or_item_kit_or_receipt;
```
objecttothis, line 405: You have a tab left of the ? instead of a single space.

application/models/Item_kit.php
```diff
@@ -31,11 +31,37 @@ public function is_valid_item_kit($item_kit_id)
...
+	*/
+	public function item_number_exists($item_kit_number, $item_kit_id = '')
+	{
+		if($this->config->item('allow_duplicate_barcodes') != FALSE)
+		{
+			return FALSE;
+		}
+
+		$this->db->from('item_kits');
+		$this->db->where('item_kit_number', (string) $item_kit_number);
```
objecttothis, line 54: I haven't been casting variables in my CI QueryBuilder function calls.  Should we be doing this everywhere or is there specific reason to do it here?

application/models/Item_kit.php
```diff
@@ -31,11 +31,37 @@ public function is_valid_item_kit($item_kit_id)
...
+		// check if $item_id is a number and not a string starting with 0
+		// because cases like 00012345 will be seen as a number where it is a barcode
+		if(ctype_digit($item_kit_id) && substr($item_kit_id, 0, 1) != '0')
+		{
+			$this->db->where('item_kit_id !=', (int) $item_kit_id);
```
objecttothis, line 59: Same question about casting here for $item_kit_id.

application/models/Item_kit.php
```diff
@@ -31,11 +31,37 @@ public function is_valid_item_kit($item_kit_id)
...
+		if($this->config->item('allow_duplicate_barcodes') != FALSE)
+		{
+			return FALSE;
+		}
+
+		$this->db->from('item_kits');
+		$this->db->where('item_kit_number', (string) $item_kit_number);
+		// check if $item_id is a number and not a string starting with 0
+		// because cases like 00012345 will be seen as a number where it is a barcode
+		if(ctype_digit($item_kit_id) && substr($item_kit_id, 0, 1) != '0')
```
objecttothis, line 57: For bug prevention you may want to use !== here instead of !=

application/models/Item_kit.php
```diff
@@ -31,11 +31,37 @@ public function is_valid_item_kit($item_kit_id)
...
+		{
+			$this->db->where('item_kit_id !=', (int) $item_kit_id);
+		}
+
+		return ($this->db->get()->num_rows() >= 1);
```
objecttothis, line 62: More of a thought than something that must be changed, but you can delete line 53 and replace this line with `return ($this->db->get('item_kits')->num_rows() >= 1);`

It saves a function call.  The downside is that it could negatively impact code readability.  I had to do it in my optimizations in the other PR because those QueryBuilder function calls were in foreach loops with thousands of calls in them (for CSV imports) and it was making significant improvement in speed.  Here maybe that won't be happening, so I don't feel strongly about it.  thoughts?

application/views/item_kits/form.php
```diff
@@ -123,7 +132,7 @@
 					<?php echo form_radio(array(
 							'name'=>'print_option',
 							'type'=>'radio',
-							'id'=>'print_option',
+							'value'=>1,
```
objecttothis, line 135: Maybe I'm reading the diff wrong, but it looks like this change created a duplicate `'value'=>1,` on lines 135 and 136.  Is that intentional?

application/views/item_kits/form.php
```diff
@@ -263,14 +271,32 @@
...
+				}
+			}
 		},
 
 		messages:
 		{
 			name: "<?php echo $this->lang->line('items_name_required'); ?>",
-			category: "<?php echo $this->lang->line('items_category_required'); ?>"
-		}
+			category: "<?php echo $this->lang->line('items_category_required'); ?>",
```
objecttothis, line 297: I'm confused what happened here.  Is the closing bracket for messages now missing?  Should the category and item_kit_number messages be outside of the messages: block?

objecttothis (changes requested):

application/controllers/Item_kits.php
jekkos: fixed

application/controllers/Sales.php
jekkos: fixed

application/models/Item_kit.php
jekkos: I remember one case that needed an explicit cast. It might change the way the actual query is generated.

jekkos: Not sure if we should do it everywhere though.

application/models/Item_kit.php
jekkos: I cast it to int as the data column type is an int as well. I'm not sure if a correct query would be generated if this variable was a string in php (it might result in `= '123'` instead of `= 123`.

application/models/Item_kit.php
jekkos: I like code that is more concise, so I'll change it as you propose.

application/views/item_kits/form.php
jekkos: Correct. Fixed it.

application/views/item_kits/form.php
jekkos: it's an indentation issue. Fixed now.

jekkos:
@objecttothis I have processed your remarks. I'll do a quick test locally to see if it's good and then merge if you're fine with this.

objecttothis:
Looks like Travis-CI hit the rate limiter again "docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit."

jekkos:
I guess that using these free service is nice until they get too much traction..

application/models/Item_kit.php
```diff
@@ -140,7 +167,7 @@ public function save(&$item_kit_data, $item_kit_id = FALSE)
 	*/
 	public function delete($item_kit_id)
 	{
-		return $this->db->delete('item_kits', array('item_kit_id' => $id));
+		return $this->db->delete('item_kits', array('item_kit_id' => $item_kit_id));
```
jekkos, line 170: @objecttothis I think this was also a bug (item kit delete might have never worked)??

objecttothis (approved):

PR #3098, opened 2021-02-12
add vendor folder to docker container

albjeremias:
To fix this issue:
https://github.com/opensourcepos/opensourcepos/search?q=Your+system+folder+path+does+not+appear+to+be+set+correctly.+Please+open+the+following+file+and+correct+this%3A+index.php&type=issues

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
         volumes:
           - uploads:/app/public/uploads
           - logs:/app/application/logs
+          - ./vendor:/app/vendor:ro
         environment:
           - CI_ENV=${OSPOS_CI_ENV}
           - FORCE_HTTPS=true


albjeremias:
well, the problem actually it was not this.. the problem is that you need to run this commands before you run ` $ docker-compose build`, but this should be automated inside the Dockerfile ? this is only applicable if u want to run directly from source, because u are running this commands in the travis-ci

```
$ docker run --rm -v $(pwd):/app jekkos/composer composer install
$ docker run --rm -v $(pwd):/app jekkos/composer php bin/install.php translations develop
$ sed -i "s/'\(dev\)'/'$rev'/g" application/config/config.php
$ docker run --rm -it -v $(pwd):/app -w /app digitallyseamless/nodejs-bower-grunt
 # sh -c "npm install && bower install && grunt package"
```
https://github.com/opensourcepos/opensourcepos/blob/12d79b73092fd307dc8cbff2e88118d383f21240/.travis.yml#L15

albjeremias:
this make more sense: https://github.com/opensourcepos/opensourcepos/pull/3101/files

PR #3102, opened 2021-02-12
Revert "add vendor folder to docker container"

albjeremias:
Reverts opensourcepos/opensourcepos#3098


jekkos:
@daN4cat @albjeremias The issue is that the docker compose file tried to build the container from the local dir, instead of pulling the image from docker hub. I would revert the fix here and change the docker-compose file to use the image directly.

It's the CI responsability to build the container and provider a ready to use image for the docker-compose.

PR #3106, opened 2021-02-13
Cleanup docker compose files

jekkos:
This PR moves the current compose file (with SSL) to a separate setup.

It also uses the images from docker hub that are built by travis. The original compose file is simplified and completely self contained again. We want users to have an easy way to spin the application with minimal setup / overhead. This can be done by using the standard `docker-compose up` in the project root folder.

The build-assets script is not needed in case you pull the prebuilt container, so this line is removed again from the init scripts.


jekkos:
@daN4cat I'd like to cleanup the SSL config a bit as well.

Do you agree if we offer following options

* a simple minimal setup without reverse proxy, which can be used at will with a proxy around it, or none at all
* the SSL (nginx) setup, with certbot and proper configured certficates

I would like to get rid of the selfcert configuration, as it actually just reverts to the first case I mention. Why have SSL if you don't have properly PKI infrastructure setup on the certificates? It does not make sense from a security point of view. In my opinion it's just the simple case which is a user that wants to spin the containers and try the application locally.

jekkos:
Also Im seeing following warning when running the `docker-compose up`

WARNING: Service "mysql" is using volume "/var/lib/mysql" from the previous container. Host mapping "/home/jeroen/workspace/opensourcepos/database" has no effect.

daN4cat:
The idea of the local HTTPS connection using self signed certificate is to test the setup before putting it live.
The let's encrypt setup with certbot is secure because installs the certificate locally to nginx reverse proxy to use.
I don't understand why you claim it's not secure.

On a Kubernetes cluster that would be the standard ingress setup using a TLS certificate given for free and valid for 3 months with autorenewal.

daN4cat:
BTW, please don't use SSL word anymore, that's long gone. TLS 1.2 is the minimum level for encryption in transit with HTTPS.

jekkos:
> The idea of the local HTTPS connection using self signed certificate is to test the setup before putting it live.
> The let's encrypt setup with certbot is secure because installs the certificate locally to nginx reverse proxy to use.
> I don't understand why you claim it's not secure.

It's because you need two things for a HTTPS [setup to be secure](https://superuser.com/questions/700170/are-ssls-default-snake-oil-certificates-truly-snake-oil-as-opposed-to-being-gen). 

As you say it's good to use it for testing purposes only (using self signed certs in this case).
> 
> On a Kubernetes cluster that would be the standard ingress setup using a TLS certificate given for free and valid for 3 months with autorenewal.

docker/install-nginx.sh
```diff
@@ -10,8 +10,6 @@ fi
 
 . ./.env
 
-/bin/bash ./build_assets.sh
-
-docker-compose -f ../docker-compose.yml build
+docker-compose -f ../docker-compose.nginx.yml build
```
jekkos, line 13: @daN4cat this is the file you said was missing no?

docker/install-nginx.sh
```diff
@@ -10,8 +10,6 @@ fi
...
 
 /bin/bash ./init-letsencrypt.sh
```
daN4cat, line 15: This one

docker/init-selfcert.sh
```diff
@@ -1,43 +0,0 @@
-#!/bin/bash
```
daN4cat, line 1: This file shouldn't be deleted

docker-compose.nginx.yml
```diff
@@ -0,0 +1,102 @@
...
+    logs:
+        driver: local
+
+networks:
+    app_net:
+    db_net:
+
+services:
+    ospos:
+        image: jekkos/opensourcepos:3.3.3
```
albjeremias, line 15: why jekkos? shouldn't it be opensourcepos ?

docker-compose.nginx.yml
```diff
@@ -0,0 +1,102 @@
...
+        image: mariadb:10.5
+        container_name: mysql
+        restart: always
+        expose:
+          - "3306"
+        networks:
+          - db_net
+        volumes:
+          - ./database/database.sql:/docker-entrypoint-initdb.d/database.sql
+          - ./database:/var/lib/mysql:rw
```
albjeremias, line 46: in here its not `./database` it is a name for a volume its 'database'

docker-compose.nginx.yml
jekkos: we don't have an organisation in docker hub yet (but I'll put it on my todo list)

docker-compose.nginx.yml
jekkos: Ok thanks for pointing this out, will revert

daN4cat (approved):

jekkos:
Seems there is still an issue
```
Named volume "database:/var/lib/mysql:rw" is used in service "mysql" but no declaration was found in the volumes section.

The command "docker-compose build" exited with 1.
```

jekkos:
I have added one extra merge constraint which is that checks on the feature branch need to be green. In case we have a failing build, we should investigate why and try to resolve before merge.

albjeremias:
sorry.. i wasn't online.. you should check the CI when merging this kind of stuff! ;) also beware the volume might be stored, so the db wont be loaded everytime the ci runs!

jekkos:
> sorry.. i wasn't online.. you should check the CI when merging this kind of stuff! ;) also beware the volume might be stored, so the db wont be loaded everytime the ci runs!

agree that it what it's for.. It shouldn't happen again. In this case I don't think it will be stored (as it's on travis) as I suppose we'll start with a clean slate every run.

PR #3113, opened 2021-02-20
Fix the discount calculation on reports.

SteveIreland:
The rounding was in the wrong place.


SteveIreland:
Looks like I violated another agreement.

```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

The command "docker-compose -f docker-compose.test.yml up --abort-on-container-exit" exited with 1.
```

daN4cat:
Indeed I wondered why a rounding in the database. I thought we moved to all in code to avoid having different rounding logic. But I wasn't sure.
I merge this.

Issue #3111, opened 2021-02-18
Display Customer Attributes (Tags/Custom fields)  in View 


DamianoSilverhand:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.2
- OSPOS git commit hash is: bb309c
- PHP version is: 7.3.21
- MySQL or MariaDB version is:  5.7.30
- OS and version is: Darwin 19.6.0
- WebServer is: MAMP
- Selected language is: en_US
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.

<strong>Display Customer Tags (Attributes) </strong>

[...]

PR #3118, opened 2021-02-27
Customer tags

DamianoSilverhand:
@jekkos this pull request adds Customer custom fields, hereby named as Tags, with the same functionality as Item Attributes. But as of issue https://github.com/opensourcepos/opensourcepos/issues/3111#, works needs to be done on pulling the tags from the database and displaying them in the customer view.
Major changes include :

1. New Migration to include new Tags module
2. Customer CSV import and file generation refactored to include Tags.


application/controllers/Customers.php
```diff
@@ -346,14 +433,131 @@ public function delete()
...
+	// 					$invalidated = $this->Customer->check_email_exists($email);
+
+	// 					if($account_number != '')
+	// 					{
+	// 						$customer_data['account_number'] = $account_number;
+	// 						$invalidated &= $this->Customer->check_account_number_exists($account_number);
+	// 					}
+	// 				}
+	// 				else
+	// 				{
```
jekkos, line 513: Is this new code or old code commented out?

application/controllers/Customers.php
```diff
@@ -346,14 +433,131 @@ public function delete()
...
+
 	/*
 	Customers import from csv spreadsheet
 	*/
 	public function csv()
 	{
 		$name = 'import_customers.csv';
-		$data = file_get_contents('../' . $name);
-		force_download($name, $data);
+		$allowed_tags = $this->Tag->get_definition_names(FALSE);
```
jekkos, line 558: I like the naming maybe even better then 'attributes'. However currently it's not really clear that attributes are linked to items and tags to customers from the naming only.

application/controllers/Customers.php
jekkos: I guess this is unused so it can be removed?

application/controllers/Customers.php
```diff
@@ -369,49 +573,51 @@ public function do_csv_import()
...
 							'email'			=> $email,
-							'phone_number'	=> $data[5],
-							'address_1'		=> $data[6],
-							'address_2'		=> $data[7],
-							'city'			=> $data[8],
-							'state'			=> $data[9],
-							'zip'			=> $data[10],
-							'country'		=> $data[11],
-							'comments'		=> $data[12]
+							'phone_number'	=> $line['Phone Number'],
```
jekkos, line 601: Ok good to use the key names to identify the fields here

application/controllers/Customers.php
```diff
@@ -422,36 +628,37 @@ public function do_csv_import()
...
 						// save customer to Mailchimp selected list
 						$this->mailchimp_lib->addOrUpdateMember($this->_list_id, $person_data['email'], $person_data['first_name'], '', $person_data['last_name']);
 					}
+				//Insert or update customer failure
 					else
 					{
-						$failCodes[] = $i;
+						$failed_row = $i+1;
+						$failCodes[] = $failed_row;
+						log_message("ERROR","CSV Item import failed on line ". $failed_row .". This customer was not imported.");
```
jekkos, line 649: I guess we show a popup as well in the end? Should be customer import not item

application/controllers/Customers.php
```diff
@@ -422,36 +628,37 @@ public function do_csv_import()
...
 					}
-
-					++$i;
 				}
 
 				if(count($failCodes) > 0)
 				{
 					$message = $this->lang->line('customers_csv_import_partially_failed') . ' (' . count($failCodes) . '): ' . implode(', ', $failCodes);
```
jekkos, line 655: Good idea to do the rollback in case a failure occurs

application/controllers/Customers.php
```diff
@@ -461,5 +668,141 @@ public function do_csv_import()
...
+				$tag_data 	= $this->Tag->get_definition_by_name($definition_name)[0];
+				$tag_type		= $tag_data['definition_type'];
+				$tag_value 	= $line['tag_' . $definition_name];
+
+				if($tag_type == 'DROPDOWN')
+				{
+					$dropdown_values 	= $this->Tag->get_definition_values($tag_data['definition_id']);
+					$dropdown_values[] 	= '';
+
+					if(in_array($tag_value, $dropdown_values) === FALSE && !empty($tag_value))
```
jekkos, line 732: not sure if this code can be extracted / reused (as we might have similar checks now in the item import). Maybe @objecttothis after we'll merge your PR?

application/helpers/importcustomer_helper.php
```diff
@@ -0,0 +1,94 @@
...
+	return $line_array;
+}
+
+/**
+ * Checks the first three characters of a file for the Byte-Order Mark then returns the file position to the first character.
+ *
+ * @param	object $file_handle	File handle to check
+ * @return	bool				Returns TRUE if the BOM exists and FALSE otherwise.
+ */
+function customer_bom_exists(&$file_handle)
```
jekkos, line 78: This helper can probably be shared between item and customer import. I gues the latter will be merged first

application/migrations/sqlscripts/3.3.4_customer_tags.sql
```diff
@@ -0,0 +1,50 @@
...
+ `tag_id` INT NULL,
+ `definition_id` INT NOT NULL,
+ `person_id` INT NULL,
+ KEY `tag_id` (`tag_id`),
+ KEY `definition_id` (`definition_id`),
+ KEY `person_id` (`person_id`),
+ UNIQUE `tag_links_uq1` (`tag_id`, `definition_id`, `person_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
+
+
```
jekkos, line 33: from a db point of view I suppose it's more simple to do it this way.. I would rename the module to sth like Customer Tags and then Attributes can be named 'Item tags' just for consistency

jekkos (changes requested):
Changeset looks good overall, I'll do a more thorough review a bit later on. We'll have to merge the work from @objecttothis first to have the import improvements. Once that is done we can polish this one up a bit and reuse some common logic (like the import validation checks).

Thanks already for the work done here.

objecttothis:
> 
> 
> @jekkos this pull request adds Customer custom fields, hereby named as Tags, with the same functionality as Item Attributes. But as of issue #3111, works needs to be done on pulling the tags from the database and displaying them in the customer view.
> Major changes include :
> 
>     1. New Migration to include new Tags module
> 
>     2. Customer CSV import and file generation refactored to include Tags.

I haven't looked at the code yet, but a couple of thoughts on design.  All of these thoughts are IMO:
- This needs to be implemented at the person level, not just the customer level.  The reason is that you want to be able to add attributes to the Suppliers, Customers and Employees.  All of those are at the People level.  It's OK if we only do customers for now, but this needs to be designed in such a way as to allow the easy expansion to Suppliers and Employees.
- The database needs to use a separate set of definition, attribute and link tables from item attributes.
- I think we need to avoid overcomplication of naming conventions and just stick to "Item Attributes" and "People Attributes".  On the UI, they can just be called attributes and then the business logic needs to differentiate depending on whether we are working with people or items.

application/config/autoload.php
```diff
@@ -159,5 +159,6 @@
 	'Tax',
 	'Tax_category',
 	'Tax_code',
-	'Tax_jurisdiction'
+	'Tax_jurisdiction',
+	'Tag'
```
objecttothis, line 163: We should stick with the current naming and call this person_attribute instead of tag IMO.  @jekkos @daN4cat thoughts?

application/helpers/importcustomer_helper.php
objecttothis: Agreed.  Reduces code duplication.  The item import changes need to be imported before this PR.

application/migrations/20210226124200_customer_tags.php
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	public function up()
+	{
+		execute_script(APPPATH . 'migrations/sqlscripts/3.3.4_customer_tags.sql');
+	}
+
+	public function down()
+	{
+
```
objecttothis, line 17: This down function needs to be completed... I'm guessing it would just be a script that deletes these tables.

application/migrations/sqlscripts/3.3.4_customer_tags.sql
objecttothis: @jekkos I'm fine with changing attributes to tags, which would make the variable names more brief, but I would advocate that this change be implemented on the People level so that tags could later be applied to Suppliers and Employees, even if this PR only implements them for customers ATM.  I think the naming then could be people_tags and then we have a technical debt for someone to refactor attributes to item_tags.  Thoughts?

application/views/tags/item.php
```diff
@@ -0,0 +1,129 @@
...
+
+<div class="form-group form-group-sm">
+	<?php echo form_label($definition_value['definition_name'], $definition_value['definition_name'], array('class' => 'control-label col-xs-3')); ?>
+	<div class='col-xs-8'>
+		<div class="input-group">
+			<?php
+				echo form_hidden("tag_ids[$definition_id]", $definition_value['tag_id']);
+				$tag_value = $definition_value['tag_value'];
+
+				if ($definition_value['definition_type'] == DATE)
```
objecttothis, line 22: IMO these need to be in a switch statement rather than a bunch of else if statements... if not, we need to at minimum modify this to use the === operator.

objecttothis (changes requested):
For clarity, my comments are all IMO and open for discussion on them.  @jekkos and @daN4cat feel free to override them.  I feel strongly about the naming conventions, but then again, I feel strongly about most things ;)

application/controllers/Customers.php
DamianoSilverhand: @jekkos yes the commented code is the original code involved with csv import and is unused in this PR.

application/controllers/Customers.php
DamianoSilverhand: @jekkos  which naming should be implemented, should I keep the tags as they are or change to customer_tags or customer_attributes as suggested by @objecttothis ?

objecttothis: To clarify, since my position changed below after reading some of @jekkos comments.  My recommendation is to change it to person_tags:
- The naming scales well to use this in suppliers and employees.
- It's shorter than person_attributes or customer_attributes.
- We can later refactor attributes to be item_tags instead of attributes.
- The GUI everywhere should be just tags since the context will make it clear about which tags we are talking about... although we will need to think through how we want to define tags.  Once it gets scaled to employees and suppliers then just probably have a multi-select dropdown allowing you to choose which types of "person" the tag will apply to.

application/migrations/sqlscripts/3.3.4_customer_tags.sql
jekkos: If we use the person table to link the tags to then it's just a matter of running the queries in those controllers as well. I think that for this change that's enough.. Link it to the person table and then have it functional in the customer module only. If you get the database model right, you'll be able to rework and factor out the common code to use it in employees as well?

application/config/autoload.php
jekkos: Yes or person_tags in this case? I must say that I like the fact that it's shorter, especially if we have to identify to which module it belongs?

daN4cat: Tags tend to be used to create semantic sets, we risk of mixing up terminology.
What about leaving attribute?

daN4cat: Or person_attribute

application/helpers/importcustomer_helper.php
jekkos: Ok we can leave this open for now, will need to be resolved once @objecttothis changes are in.

application/config/autoload.php
jekkos: Also fine, I mostly prefer shorter naming, but might be less correct as @daN4cat says here.

application/controllers/Customers.php
jekkos: @objecttothis are you fine to make it person_attributes as @daN4cat ? Let's make sure we're on the same page here then.

objecttothis:
Yeah that's fine. I don't have a hangup on person_tags, but person_attributes is fin with me too. We may not even need to refactor attributes to item_attributes if we go with person_attributes.

Get Outlook for Android<https://aka.ms/ghei36>

application/controllers/Customers.php
daN4cat: Yeah, I'm concerned about tags being a semantic set / grouping concept.

jekkos: Ok person_attributes it is.

DamianoSilverhand: Will start working on the recommended changes, meaning I will have to create a new branch `person_attributes` for my next PR

objecttothis: > 
> 
> @objecttothis are you fine to make it person_attributes as @daN4cat ? Let's make sure we're on the same page here then.

No complaint here.  Sorry for not being able to respond earlier.

application/controllers/Customers.php
objecttothis: Yes, in fact it's better to refactor the rest so all of them are using the key names.  This provides a few advantages.

1. If the user moves columns around in their CSV it will still process them properly (as long as they don't change the names of columns.
2. It's easier to read in the code.

application/controllers/Customers.php
```diff
@@ -369,49 +573,51 @@ public function do_csv_import()
...
 				{
-					// XSS file data sanity check
-					$data = $this->xss_clean($data);
+					$invalidated	= FALSE;
+
+					$line = array_combine($keys,$this->xss_clean($line_array[$i]));	//Build a XSS-cleaned associative array with the row to use to assign values
 
-					$consent = $data[3] == '' ? 0 : 1;
+					//check for consent in file upload, y == yes , empty field == no
+					$consent = $line['Consent'] == '' ? 0 : 1;
```
objecttothis, line 591: This functionality wasn't in there before so if you don't want to include it then that's fine, but you might take a look at how I've handled true/false/0/1/NULL columns in item imports from #3029.  I'm not sure, but current implementation here won't be able to handle true/false/TRUE/FALSE/NULL, but instead translating them all to 1.

application/controllers/Customers.php
```diff
@@ -346,14 +433,131 @@ public function delete()
...
 	{
 		$name = 'import_customers.csv';
-		$data = file_get_contents('../' . $name);
-		force_download($name, $data);
+		$allowed_tags = $this->Tag->get_definition_names(FALSE);
+		$data = generate_import_customers_csv($allowed_tags);
+		force_download($name, $data, TRUE);
 	}
 
 	public function csv_import()
```
objecttothis, line 563: with #3029 I have a function in the importfile_helper called get_csv_file().  You'll want to use that here to take advantage of code_reuse, and that I've done extensive speed optimizations on that function to reduce time it takes to injest large files.  It's not as good of a system as an injest queue would be, but we don't currently have that in the code.

DamianoSilverhand:
@jekkos @objecttothis  @daN4cat  I am kindly requesting that this PR be closed in favor of #3161 , branch person attributes, with reworked changes and additions and extension of person attributes to Suppliers and Employees.

Issue #3100, opened 2021-02-12
visiting items/save/{id} (GET request) clears existing values


bricas:
### Background information

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [X] Reporting an issue of an unmodified OSPOS installation
- [X] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [X] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [X] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [X] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [X] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [X] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.3
- OSPOS git commit hash is: 8e52bd
- PHP version is: 7.3.22
- MySQL or MariaDB version is: 5.5.5-10.3.27-MariaDB-1:10.3.27+maria~focal
- OS and version is: Linux 5.8.0-7630-generic
- WebServer is: nginx
- Selected language is: English
- (If applicable) Docker installation: yes, locally developed
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

I was trying to debug an issue when saving an item, and after browsing the network log, i clicked "open in new tab" on /items/save/1 -- which strangely returned a "save successful" json response. I looked back at the item and all of the original data was now blank.

[...]

PR #3124, opened 2021-03-02
Add hook to validate save HTTP method (#3100)

jekkos:
This change adds an extra validation step that will prevent wipeout of any entity if GET is called directly on the resource URL.


jekkos:
@objecttothis @daN4cat maybe check this one so we can merge?

daN4cat (approved):

daN4cat:
Somehow it fails to merge from my mobile app.  Is it blocked by something?

jekkos:
I changed something in the repo config. I'm not sure if we should stick to it. Do you think we should use merge commits or rebase instead when merging to master?

objecttothis:
@jekkos my personal preference is to squash and merge so that you keep commit messages, but that it gets consolidated down to a single commit when being merged into the master.  Would choosing rebase and merge do that?  According to this article (https://rietta.com/blog/github-merge-types/) rebase and merge should be used for things like small hotfixes and patches.

daN4cat:
BTW, largely off topic but try to add 1s to a github.com repo URL like github1s.com and see what happens 😉

Issue #3133, opened 2021-03-17
Rewrite .htacces doesn't work


ffurxio:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.3
- OSPOS git commit hash is: 3b1f459
- PHP version is: 7.4.15
- MySQL or MariaDB version is: MariaDB 10.4.17 
- OS and version is: Windows 
- WebServer is: Apache 2.4.46
- Selected language is: English
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 
All provided by Xampp 7.4.15
### Issue / Bug  / Question / New Feature

Good Day
I'm having this issue where i cant acces to the system because i always been redirect to 
localhost/public 
instead of 
localhost/ospos/public
Yes i have installed and enable the modules for php 
[...]

PR #3139, opened 2021-03-21
Remove unneeded directives (#3133)

jekkos:
Removing some directives that might not be needed

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -8,7 +8,7 @@
 
 # disable directory browsing
 # For security reasons, Option all cannot be overridden.
-Options +ExecCGI +Includes +IncludesNOEXEC +SymLinksIfOwnerMatch -Indexes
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -8,7 +8,7 @@
 
 # disable directory browsing
 # For security reasons, Option all cannot be overridden.
-Options +ExecCGI +Includes +IncludesNOEXEC +SymLinksIfOwnerMatch -Indexes
```
jekkos, line 11: @daN4cat I don't think we need CGI inside of the document root?

I have disabled the first three as I don't think they are needed.

daN4cat: That's fine, sorry for the late reply.

jekkos: Ok thanks for the feedback

Issue #3144, opened 2021-04-04
Sales Controller issues in current version.


odiea:
OSPOS Installation Info: 3.3.3 - 616e8e

Language Code: en-US

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓

User Configuration:
.Browser: Chrome
.Server Software: Apache
.PHP Version: 7.4.13
.DB Version: 5.7.21
.Server Port: 80
.OS: Windows NT 10.0

File Permissions:
» [application/logs:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [public/uploads:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [public/uploads/item_pics:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [import_customers.csv:] - 0666 | Readable ✓ | Vulnerable or Incorrect Permissions ✗

Security Vulnerability Warning
Incorrect permissions leaves this software at risk.

» [application/logs:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [public/uploads:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [public/uploads/item_pics:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [import_customers.csv:] is readable, but the permissions are incorrectly set. Please set it to 640 or 660 and refresh.

Is the controller/sales suppose to be reload() or _reload() it appears to have both also none of the right side buttons work now. Cannot complete a Sale or suspend a sale or add a new customer

jekkos:
Thanks for reporting @odiea .

odiea:
Also in the footer common_copryrights is misspelled.

jekkos:
@objecttothis any idea if it could be something caused by the last merge?

odiea:
It is all caused by the last merge. Try the dev demo.

objecttothis:
I'll take a look

objecttothis:
I submitted a patch, but I'll need to compare the Sales controller with my build that is working properly and see if I can't spot what got messed up.

odiea:
I used the previous release of the Register and now it works properly

odiea:
Also in config I am not able to add either of the 2 taxes.

[...]

PR #3146, opened 2021-04-05
Merge errors patch

objecttothis:
It seems like I was careless in the rebase for the Sales controller and accepted some changes from the master that undid some of my changes to refactoring.  I need to download this build and test it before we merge to see if there are other issues.

Closing #3144

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -39,7 +39,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 |  $autoload['packages'] = array(APPPATH.'third_party', '/usr/local/shared');
 |
 */
-$autoload['packages'] = array();
+$autoload['packages'] = [];
 
 /*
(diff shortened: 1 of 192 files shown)


jekkos:
the changeset is a bit confusing, it looks like there are quite some commits but only two files that were touched?

objecttothis:
@jekkos this is likely because I submitted a single commit to the same branch that I merged from earlier and created a new pr. It shows the commit history of the branch but the changed files are only going to report what is different from the master.

objecttothis:
the reason for it saying that it wants to merge 7 commits is that the original pr I did a squash and merge on, so in the master those appear as one commit, but in the branch they are still separate commits.

jekkos:
Ok I see, maybe we should apply those changes on the squashed commit instead for clarity? If too hard I can try to do it for you.

Issue #3138, opened 2021-03-20
Daily Sales overall Cash total changes when switching pages


chunter2:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Looks like when I have more than 10 sales showing in the daily sales and I choose to switch to the next page the total for Cash changes but the others don't.  These aren't the "Total" row values.  They're the ones below that.  I believe it should be the overall total, not the total for the page.  I think the row named "Total" is for that.  I tested this on dev.opensourcepos.org.  

Thanks

[...]

PR #3149, opened 2021-04-06
Fix cash totals in payment summary (#3138)

jekkos:
This patch should fix the cash totals in the payment summary. 

* Refactored the get_info to reuse create_temp_table(). Also cleaned up some duplicate code


jekkos:
Tested and confirmed working by @chunter2 
Might be good to have some review maybe @SteveIreland you might be more familiar with this part of the code?

SteveIreland:
yeah, I'm going to need to revisit my work after this is merged.  You're changing code that I was also changing ... maybe for the same or similar reason, but regardless I will need to go through it with a fine tooth comb to insure I don't reintroduce an issue that you were fixing.

jekkos:
In this case I'm touching the same files? You might have changed the temporary tables a bit then? I think the functional change is really small here most of the diff is moving the temp table creation to the create_temp_table method.

PR #3167, opened 2021-04-22
Database Optimizations

objecttothis:
- Add indexes to tables to improve query times.
- Delete orphaned attribute values.
- Resolve duplicate attribute values.
- Deleted whitespace after migration which was causing Severity: Warning --> Cannot modify header information - headers already sent by...

This is the first of four PRs that I will break out of the big import_update branch, so that it's easier to test and merge.   Those having issues with large product stocks should see immediate improvement in transactions.

### DIFF
--- a/application/migrations/20210422000000_database_optimizations.php
+++ b/application/migrations/20210422000000_database_optimizations.php
@@ -1,0 +1,123 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_database_optimizations extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		error_log('Migrating database_optimizations');
+		$CI =& get_instance();
+
+		$CI->Attribute->delete_orphaned_values();
+
+		$this->migrate_duplicate_attribute_values(DECIMAL);
+		$this->migrate_duplicate_attribute_values(DATE);
+
+		//Select all attributes that have data in more than one column
+		$this->db->select('attribute_id, attribute_value, attribute_decimal, attribute_date');
+		$this->db->group_start();
+			$this->db->where('attribute_value IS NOT NULL');
+			$this->db->where('attribute_date IS NOT NULL');
+		$this->db->group_end();
+		$this->db->or_group_start();
+			$this->db->where('attribute_value IS NOT NULL');
+			$this->db->where('attribute_decimal IS NOT NULL');
+		$this->db->group_end();
+		$attribute_values = $this->db->get('attribute_values');
+
+		$this->db->trans_start();
+
+		//Clean up Attribute values table where there is an attribute value and an attribute_date/attribute_decimal
+		foreach($attribute_values->result_array() as $attribute_value)
+		{
+			$attribute_links = $this->db->query('SELECT links.definition_id, links.item_id, links.attribute_id, defs.definition_type FROM ospos_attribute_links links JOIN ospos_attribute_definitions defs ON defs.definition_id = links.definition_id where attribute_id = '. $attribute_value['attribute_id']);
+
+			$this->db->where('attribute_id', $attribute_value['attribute_id']);
+			$this->db->delete('attribute_values');
+
+			foreach($attribute_links->result_array() as $attribute_link)
+			{
+				$this->db->where('attribute_id',$attribute_link['attribute_id']);
+				$this->db->where('item_id',$attribute_link['item_id']);
+				$this->db->delete('attribute_links');
+
+				switch($attribute_link['definition_type'])
+				{
+					case DECIMAL:
+						$value = $attribute_value['attribute_decimal'];
+						break;
+					case DATE:
+						$attribute_date	= DateTime::createFromFormat('Y-m-d', $attribute_value['attribute_date']);
+						$value			= $attribute_date->format($CI->Appconfig->get('dateformat'));
+						break;
+					default:
+						$value = $attribute_value['attribute_value'];
+						break;
+				}
+
+				$CI->Attribute->save_value($value, $attribute_link['definition_id'], $attribute_link['item_id'], FALSE, $attribute_link['definition_type']);
+			}
+		}
+		$this->db->trans_complete();
+
+		execute_script(APPPATH . 'migrations/sqlscripts/3.4.0_database_optimizations.sql');
+		error_log('Migrating database_optimizations completed');
+	}
+	/**
+	 * Given the type of attribute, deletes any duplicates it finds in the attribute_values table and reassigns those
+	 */
+	private function migrate_duplicate_attribute_values($attribute_type)
+	{
+	//Remove duplicate attribute values needed to make attribute_decimals and attribute_dates unique
+		$this->db->trans_start();
+
+		$column = 'attribute_' . strtolower($attribute_type);
+
+		$this->db->select("$column, attribute_id");
+		$this->db->group_by($column);
+		$this->db->having("COUNT($column) > 1");
+		$duplicated_values = $this->db->get('attribute_values');
+
+		foreach($duplicated_values->result_array() as $duplicated_value)
+		{
+			$this->db->select('attribute_id');
+			$this->db->where($column, $duplicated_value[$column]);
+			$attribute_ids_to_fix = $this->db->get('attribute_values');
+
+			$this->reassign_duplicate_attribute_values($attribute_ids_to_fix, $duplicated_value);
+		}
+
+		$this->db->trans_complete();
+	}
+
+	/**
+	 * Updates the attribute_id in all attribute_link rows with duplicated attribute_ids then deletes unneeded rows from attribute_values
+	 * @param attribute_ids_to_fix
+	 * @param decimal
+	 */
+	private function reassign_duplicate_attribute_values($attribute_ids_to_fix, $attribute_value)
+	{
+		foreach($attribute_ids_to_fix->result_array() as $attribute_id)
+		{
+			//Update attribute_link with the attribute_id we are keeping
+			$this->db->where('attribute_id', $attribute_id['attribute_id']);
+			$this->db->update('attribute_links', array('attribute_id' => $attribute_value['attribute_id']));
+
+			//Delete the row from attribute_values if it isn't our keeper
+			if($attribute_id['attribute_id'] !== $attribute_value['attribute_id'])
+			{
+				$this->db->where('attribute_id', $attribute_id['attribute_id']);
+				$this->db->delete($this->db->dbprefix('attribute_values'));
+			}
+		}
+	}
+
+	public function down()
+	{
+	}
+}
+?>
--- a/application/migrations/20210422000001_remove_duplicate_links.php
+++ b/application/migrations/20210422000001_remove_duplicate_links.php
@@ -1,0 +1,57 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+
+class Migration_remove_duplicate_links extends CI_Migration
+{
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	public function up()
+	{
+		error_log('Migrating remove_duplicate_links');
+
+
+		$this->migrate_duplicate_attribute_links();
+
+		error_log('Migrating remove_duplicate_links completed');
+	}
+	/**
+	 * Given the type of attribute, deletes any duplicates it finds in the attribute_values table and reassigns those
+	 */
+	private function migrate_duplicate_attribute_links()
+	{
+		$CI =& get_instance();
+	//Remove duplicate attribute links
+		$this->db->trans_start();
+
+		$this->db->where('sale_id');
+		$this->db->where('receiving_id');
+		$this->db->group_by('item_id');
+		$this->db->group_by('definition_id');
+		$this->db->group_by('attribute_id');
+		$this->db->having('COUNT(item_id) > 1');
+		$this->db->having('COUNT(definition_id) > 1');
+		$this->db->having('COUNT(attribute_id) > 1');
+		$duplicated_links = $this->db->get('attribute_links');
+
+		foreach($duplicated_links->result_array() as $duplicated_link)
+		{
+			$this->db->where('sale_id');
+			$this->db->where('receiving_id');
+			$this->db->where('item_id', $duplicated_link['item_id']);
+			$this->db->where('definition_id', $duplicated_link['definition_id']);
+			$this->db->where('item_id', $duplicated_link['item_id']);
+			$this->db->delete('attribute_links');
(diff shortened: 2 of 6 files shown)


application/migrations/sqlscripts/3.4.0_database_optimizations.sql
```diff
@@ -0,0 +1,125 @@
...
+
+#ospos_stock_locations table
+ALTER TABLE `ospos_stock_locations` MODIFY `deleted` tinyint(1) DEFAULT 0 NOT NULL;
+ALTER TABLE `ospos_stock_locations` ADD INDEX(`deleted`);
+
+#ospos_suppliers table
+DROP INDEX `person_id` ON `ospos_suppliers`;
+ALTER TABLE `ospos_suppliers` MODIFY `deleted` tinyint(1) DEFAULT 0 NOT NULL;
+ALTER TABLE `ospos_suppliers` MODIFY `category` tinyint(1) NOT NULL;
+ALTER TABLE `ospos_suppliers` ADD PRIMARY KEY(`person_id`);
```
jekkos, line 105: I'm a bit surprised, didn't we have a pk in this table?

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
```diff
@@ -0,0 +1,125 @@
...
+ALTER TABLE `ospos_suppliers` ADD INDEX(`category`); 
+ALTER TABLE `ospos_suppliers` ADD INDEX(`deleted`);
```
jekkos, line 107: if some columns are always queried together, then it might be better to build a combined index instead. In this case I think that most of the time the table is joined on person_id and deleted flag is also checked? I think that deleted flag index will probably not really give a huge speedup (unless you have many deleted records, which might not be the case)?

application/models/Attribute.php
```diff
@@ -665,6 +665,27 @@ public function delete_definition_list($definition_ids)
...
+	/**
+	 * Deletes any orphaned values that do not have associated links
+	 * @param int $definition_id
+	 * @return boolean TRUE is returned if the delete was successful or FALSE if there were any failures
+	 */
+	public function delete_orphaned_values()
+	{
+		$this->db->distinct();
+		$this->db->select('attribute_id');
+		$attribute_ids = $this->db->get_compiled_select('attribute_links');
```
jekkos, line 677: does the compiled select cache the query? First time I see this

application/migrations/20210422000001_remove_duplicate_links.php
```diff
@@ -0,0 +1,57 @@
...
+		$this->db->having('COUNT(item_id) > 1');
+		$this->db->having('COUNT(definition_id) > 1');
+		$this->db->having('COUNT(attribute_id) > 1');
+		$duplicated_links = $this->db->get('attribute_links');
+
+		foreach($duplicated_links->result_array() as $duplicated_link)
+		{
+			$this->db->where('sale_id');
+			$this->db->where('receiving_id');
+			$this->db->where('item_id', $duplicated_link['item_id']);
```
jekkos, line 42: you have this line in there twice

application/migrations/20210422000001_remove_duplicate_links.php
```diff
@@ -0,0 +1,57 @@
...
+		$this->db->group_by('definition_id');
+		$this->db->group_by('attribute_id');
+		$this->db->having('COUNT(item_id) > 1');
+		$this->db->having('COUNT(definition_id) > 1');
+		$this->db->having('COUNT(attribute_id) > 1');
+		$duplicated_links = $this->db->get('attribute_links');
+
+		foreach($duplicated_links->result_array() as $duplicated_link)
+		{
+			$this->db->where('sale_id');
```
jekkos, line 40: so it only cleans up attributes linked to items? This line says that it will ignore attributes links related with sales and receivings right?

application/migrations/20210422000000_database_optimizations.php
```diff
@@ -0,0 +1,123 @@
...
+		execute_script(APPPATH . 'migrations/sqlscripts/3.4.0_database_optimizations.sql');
+		error_log('Migrating database_optimizations completed');
+	}
+	/**
+	 * Given the type of attribute, deletes any duplicates it finds in the attribute_values table and reassigns those
+	 */
+	private function migrate_duplicate_attribute_values($attribute_type)
+	{
+	//Remove duplicate attribute values needed to make attribute_decimals and attribute_dates unique
+		$this->db->trans_start();
```
jekkos, line 76: I guess this will save us some storage, thanks for that.

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis: > 
> 
> I'm a bit surprised, didn't we have a pk in this table?

It was indexed but not a primary key.  in order to make it a primary key I had to first drop the index.  Making it a primary key guarantees it's unique aand not nullable

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis: > 
> 
> if some columns are always queried together, then it might be better to build a combined index instead. In this case I think that most of the time the table is joined on person_id and deleted flag is also checked? 

I spent a great deal of time trying to track down all the ways that these rows were queried.... it's a mess, but that's just because there are a lot of different queries that need to be made.

My understanding of composite indexes is that if you have index(a, b, c) then the index is only used on queries of (a), (a, b) and (a, b, c).  This means that if the query is (a, c), (c, b, a) or any other permutation, then the index won't be used. That said, there are quite few queries that don't use the same order, and we could correct that in the code.  Further queries using a OR b in the where clause won't use a composite index.

I'm definitely open to using composite indexes, but I think my skill level was limiting knowing exactly which composite indexes will help the most.

> I think that deleted flag index will probably not really give a huge speedup (unless you have many deleted records, which might not be the case)?

in our items table of 19182 rows we have 1315 rows where deleted = 1.  The vast majority of our queries use a where clause that includes deleted = 0.  It may be that it's only worth it to add an index for deleted on ospos_items then, since all the other tables that include a deleted column are not likely to have all that many rows in them.  Thoughts?

application/models/Attribute.php
objecttothis: Yes, this is a CI querybuilder function that caches the select statement but doesn't run the query (https://codeigniter.com/userguide3/database/query_builder.html).  Very handy IMO.

objecttothis: in this case I'm using it to perform a select within a select

application/migrations/20210422000001_remove_duplicate_links.php
objecttothis: fixed it.

application/migrations/20210422000001_remove_duplicate_links.php
objecttothis: That's correct.  To me it's counter-intuitive, but my experience has been that passing just the column name to the where function `$this->db->where('sale_id');` is the equivalent to `WHERE 'sale_id' IS NULL;`  I think the reason is that the overloaded function typically would be called with two parameters with the second one being the right side of the equals sign `where('deleted',1);` but when the second parameter is missing, it's the equivalent of saying `where('deleted', null);`

The tl;dr is that yes, this function doesn't touch anything that has something in the sale_id column or receiving_id column.  I figured doing that would be bad ;)

objecttothis: that said, I think for code clarity, maybe we should start writing `$this->db->where('sale_id', null);` because honestly when I first saw it in other parts of the code I thought it meant the opposite of what it means.

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: Ok just for my info, we probably run different types of queries that use the indexes in different ways, so  we need to find some kind of common denominator in indexing these columns. 
Indeed I would only add an index if we have a significant % of rows deleted in the table. If the deleted is hardly used at all then it will not really gain much.

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: Ok good to know, thanks for fixing this

application/migrations/20210422000001_remove_duplicate_links.php
jekkos: So as you suggest maybe it's better to add it explicitly.

jekkos (approved):
Thanks I think we're good if we resolve the last comments in here.

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis: @jekkos here is an example of what I mean by not really understanding complex indices.  I took the Tax model as an example.
```
		$this->db->select('tax_rate_id');
		$this->db->select('rate_tax_code_id');
		$this->db->select('tax_code');
		$this->db->select('tax_code_name');
		$this->db->select('rate_jurisdiction_id');
		$this->db->select('jurisdiction_name');
		$this->db->select('rate_tax_category_id');
		$this->db->select('tax_category');
		$this->db->select('tax_rate');
		$this->db->select('tax_rounding_code');
		$this->db->from('tax_rates');
		$this->db->join('tax_codes',
			'rate_tax_code_id = tax_code_id', 'LEFT');
		$this->db->join('tax_categories',
			'rate_tax_category_id = tax_category_id', 'LEFT');
		$this->db->join('tax_jurisdictions',
			'rate_jurisdiction_id = jurisdiction_id', 'LEFT');
		$this->db->where('tax_rate_id', $tax_rate_id);

		$query = $this->db->get();
```
in this get_info() function you have a where clause invoking tax_rate_id, so that seems simple enough.  However joins also benefit from indices as well.  This function has three left joins.
- tax_rates.rate_tax_code_id = tax_codes.tax_code
- tax_rates.rate_tax_category_id = tax_categories.tax_category_id
- tax_rates.rate_jurisdiction_id = tax_jurisdictions.jurisdiction_id

If these were a single simple join with no where clause I would say that each side should have a single index.  There are few things that I don't understand:
- Should we instead here have a composite index on rate_tax_code_id, rate_tax_category_id and rate_jurisdiction_id or does it not work the same way on joins?  If they were where statements, we would but I don't know if that's how it works for joins.
- Should the index on the where clause be combined with the index on the join clause?  If so, what order does the index need to be in?  WHERE first or JOINs first?

Many of the functions just in the Tax model have wildly different where columns and join columns.
- exists WHERE(tax_rate_id)
- get_tax_category_usage WHERE(rate_tax_category_id)
- get_info JOIN(rate_tax_code_id, rate_tax_category_id, rate_jurisdiction_id) WHERE(tax_rate_id)
- get_taxes JOIN(rate_tax_code_id, rate_tax_category_id, rate_jurisdiction_id) WHERE(rate_tax_code_id, rate_tax_category_id)
- get_rate_info JOIN(rate_tax_caategory_id) WHERE(rate_tax_code_id, rate_tax_category_id)
- search JOIN(rate_tax_code_id, rate_tax_category_id, rate_jurisdiction_id) WHERE LIKE(rate_tax_code) OR WHERE LIKE(tax_code_name)
- get_tax_category WHERE(tax_category_id)

If you can give me a model for how composite indices work here, then I can apply it to the rest of the tables and we should experience greater speed gains than what I've gotten so far.

objecttothis: @jekkos, sorry, ignore get_tax_category for this conversation.  It's pulling from the tax_category table, not tax_rates

objecttothis:
> 
> 
> Thanks I think we're good if we resolve the last comments in here.

OK, then just the composite indices to figure out.  I've pushed a commit that fixes the other issues.

application/migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: @objecttothis Interesting example that you bring up here. I'm not sure but my intuition would say that in this case individual indexes would be used, one for every join operations. You can easily verify what is most efficient by creating the two cases and then check which one is picked by MySql
* Add individual indexes for the columns we join on
* Add a composite index for the columns we join on
then run this query with `explain` upfront in a mysql client console. 
An easy way to get the full sql is by dumping the `$this->db->last_query()` after the query returns, then copy paste that into the mysql client console.

jekkos: The explain output then shows what indexes were used and how many rows were scanned to find the resultset.

objecttothis:
@jekkos I decided to just do the item model and here is what the better part of a day got me in terms of moving through each function that contained a select query and running MYSQL EXPLAIN against the query that gets generated (thanks for the tip on the $this->db->last_query() function).
image
Now, armed with this data, we need to decide which queries are worth creating a composite index for the whole thing (Item->Search() on the items table for example?) and which ones we can get away with just finding the common denominator and creating the composite index that will encompass most of those queries (The orange indexes are ones that appear more than once in that column... items (PRIMARY, deleted) or item_quantities (PRIMARY, item_id, location_id) for example).  Thoughts?

objecttothis:
Also, if possible, since it takes so long to do, I'd like to limit this round of DB improvements to those affected by the Item and Attribute models.  We can go back through and revamp the indexes on the other tables later if needed.  I think an entire PR could be dedicated just to the Report model tables.

objecttothis:
@jekkos, @opensourcepos/org I removed the deleted indexes on a few of the tables and added a few composite indices (see below) based on the results of the explain.  Can we move on this yet?

image

jekkos:
@objecttothis thanks for the good work here, I didn't have too much time to look in to it yet thoroughly, 

> Also, if possible, since it takes so long to do, I'd like to limit this round of DB improvements to those affected by the Item and Attribute models. We can go back through and revamp the indexes on the other tables later if needed. I think an entire PR could be dedicated just to the Report model tables.

The items / inventory table is enough for now. I think that is typically the table that has most of the data. Also running a report is not something that is used that often, it can be done afterwards as it might be acceptable to be a little bit slower.

jekkos:
Were you able to verify that the queries use the composite index ? I think it would already be a great improvement if an index is able to accelerate the commonly used Item->search() query.

objecttothis:
> Were you able to verify that the queries use the composite index ? I think it would already be a great improvement if an index is able to accelerate the commonly used Item->search() query.

I'll confirm that they are using the indexes.

jekkos:
Ok I have merged this change already. Minor nitpick maybe try to refer to an issue number in your commit message next time, it makes the git changelog better as you can find more info about a changeset under the github issue tracker.

Issue #3129, opened 2021-03-13
Fixed Discount bug on item kits


andriux1990:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.
____________________________________________________
Fixed discount in item kits is not the same in sales module
If we add several items to an item kit, and we choose a fixed discount ..

For example.
[...]

PR #3186, opened 2021-05-16
Fix Discounted Item Kit bug #3129

SteveIreland:
This addresses the reported issue with entering discounts at a kit level.


jekkos:
Thanks for the fixes @SteveIreland .

application/libraries/Sale_lib.php
```diff
@@ -753,40 +753,47 @@ public function add_item(&$item_id, $quantity = 1, $item_location, $discount = 0
...
-			$price = $item_info->unit_price;
-			$cost_price = $item_info->cost_price;
+			$price = $price_override;
 		}
-		elseif($price_mode == PRICE_MODE_KIT)
+
+		if($price_mode == PRICE_MODE_KIT)
 		{
-			if($kit_price_option == PRICE_OPTION_ALL)
+			if(!($kit_price_option == PRICE_OPTION_ALL
```
jekkos, line 770: I've been repeating myself for a while but I think we should really try to add unit tests for new testcases to prevent us from adding regressions. I suppose the fix is fine but we might always overlook other scenarios and introduce regressions.

application/libraries/Sale_lib.php
```diff
@@ -753,40 +753,47 @@ public function add_item(&$item_id, $quantity = 1, $item_location, $discount = 0
...
+			// If price is zero do not include a discount regardless of type
+			if($price == 0.00)
 			{
-				$price = $item_info->unit_price;
-				$cost_price = $item_info->cost_price;
+				$applied_discount = 0.00;
 			}
-			elseif($kit_price_option == PRICE_OPTION_KIT_STOCK && $stock_type == HAS_STOCK)
+			// If fixed discount then apply no more than the item price
+			if($discount_type == FIXED)
```
jekkos, line 783: So if a fixed discount for a kit is added, we just substract it from the items until we have no more discount to apply?

SteveIreland: Correct.  But within the constraint of the kit's items.  If the fixed discount is greater than the  combined price of all of the items in the kit the kit will end up with a price of zero and the remaining discount is ignored even though there may be more discount that could be applied.  There has to be a limit to the insanity ... right?  :-)

application/libraries/Sale_lib.php
SteveIreland: I truly understand the importance of this and I did write a unit test for some earlier work, but I was never really able to make it work according to the way it was advertised to work.  It seemed like it was because I was on a different platform.   I did finally get the tests to run, but I had to jump through hoops to run it and that was quite annoying.   I'll spend a little more time trying to get one to work the way it should so that I can be fully on board.

application/libraries/Sale_lib.php
jekkos: I saw you added the same explanation to the wiki. Thanks for that.

Issue #3110, opened 2021-02-18
BinTray Replacement


objecttothis:
### Issue

When it rains it poors, right?  In addition to needing to find a replacement for Travis-CI (#3063 and #2834) JFrog has announced that they are shutting down BinTray on May 1st (https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).

objecttothis:
yay for DevOps fun!

jekkos:
Yes ... I noticed already. Yesterday I have created a travis-ci.com account. It seems that we should be able to migrate to the new platform and hopefully we'll be able to stick with the free plan.

We're also often hitting rate limitation on docker hub now..

docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.

jekkos:
Maybe github actions can also be used to host build artifacts.. It would be good to have everything integrated in one platform.

jekkos:
https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

jekkos:
Maybe github actions can replace both travis and jfrog. In this case if we can run builds as a workflow and host the resulting artifact on github itself, then that would be great.

daN4cat:
Yes, GitHub actions is quite promoted by Microsoft as the next. They are planning to migrate from Azure DevOps to GitHub.

jekkos:
Bintray will shutdown completely 1st of may.

SteveIreland:
Is that the day you are going to sell OSPOS and retire to a Swiss resort?

jekkos:
Bintray had stopped working as expected

![Screenshot from 2021-05-26 10-20-55](https://user-images.githubusercontent.com/576234/119627290-58955a00-be0c-11eb-9334-cde0ee9867bd.png)

We could deploy to [github releases](https://github.com/travis-ci/dpl) instead using travis (dpl plugin)

jekkos:
Seeing an issue with [OAuth setup now](https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token)

jekkos:
It seems that some of the runs have deployed a zip to releases section here now..

jekkos:
Ok the files do end up in github releases now. I had to find some workaround for the following

* Github release cannot exist without a tag
[...]

PR #3194, opened 2021-05-26
Publish to github releases (#3110)

jekkos:
Let's try and see if we can push a build to github releases from travis


objecttothis:
@jekkos I removed the reference to bintray in the readme.  Do we have anything that we can use in the meantime?

jekkos:
The issue here is that travis is unable to parse the config file maybe due to a wrong format. Currently I think the only way is to build locally

jekkos:
It took me some fiddling to get it to work but I think it works the way I want now.

We will always have a tagged commit in each branch that points to the last build that was produced for that branch. Those build will be available in the releases section as a 'pre-release'.

jekkos:
the final zip files have the same naming as the ones that were pushed to bintray, so this might be recognizable.

PR #3197, opened 2021-06-02
Upgrade phpunit to version 9.5

SteveIreland:
Testing to see if it builds.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
 	},
 	"require-dev": {
 		"mikey179/vfsstream": "1.1.*",
-		"phpunit/phpunit": "7.5.6",
+		"phpunit/phpunit": "^9.5",
 		"kenjis/ci-phpunit-test": "0.18.0"
 	}
 }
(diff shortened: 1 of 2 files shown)


jekkos:
Currently seeing the following error
```

ospos_test | Fatal error: Declaration of CIPHPUnitTestCase::setUpBeforeClass() must be compatible with PHPUnit\Framework\TestCase::setUpBeforeClass(): void in /app/application/tests/_ci_phpunit_test/CIPHPUnitTestCase.php on line 75
```

It might be better to try and run the tests first locally and fix the issues encountered.

jekkos:
From the error message it looks like the function signature for the test setup was changed.

SteveIreland:
Okay, that will be a good learning opportunity for me.

objecttothis:
@SteveIreland phpunit in ci4-upgrade is set to 9.6.3 so this PR can probably just be deleted.

jekkos:
Will be obsolete after CI4 upgrade, will close this one for now.

PR #3190, opened 2021-05-24
Login view and other small changes

BudsieBuds:
Hi everyone,

<details>
<summary>Read this for some context</summary>
Thought I'd provide a little context to what, how and why for who wants to know. So I bumped into this repository while looking for a POS system for a friend's store. That's when I decided to dive into the code, mainly for coding practice. Full disclosure: I'm not a developer, just trying to learn stuff. I've put a lot of time in this pull request and I learned a lot while coding it. It's the first time I create a pull request of this scale, so please be mild. 😉 Probably better to create a pull request per change next time (?). This seems a bit messy, sorry for that. I'll try my best to explain everything below. Also, English is not my native language. Therefore, if anything is unclear, please give me a chance to provide additional info. Thanks!
</details>

## :tv: Login view
* Updated to more modern style and coding standards.
* Login screen fully respects theme colors.
* Fully responsive. Looks good on all displays I've tested. From small screens like iPhone SE to big 4K monitors.
* Uses Bootstrap 5. Should be included in addition to Bootstrap 3.
* Uses Bootswatch 5. Should be included in addition to Bootswatch 3.
  * Fallback for unsupported BS3 themes to Flatly included.
* Shows company logo when available.
  * Fallback to OSPOS logo.
* Added option to `config/general` to change login form style. Two options: floating labels or input groups.
* Added default OSPOS color as `theme-color` for Chrome Android.

## :mouse: Smaller changes
* Capitalize theme names in theme selector.
* Added theme preview link to `config/general` screen.
* Shifted around some translations for easier and more consistent translating.

## :mailbox_with_mail: Suggestions/Questions/Discussion
* Should the version number be removed from the login title and footer for security reasons?
* Is there an SVG version of the OSPOS logo available? I've auto converted it for testing purposes.
* Made the OSPOS logo coloring and rounding to match the selected theme. Don't know if this is something you're ok with? I think it makes the logo look fitting, and good visibility is guaranteed on all themes.
* Can SVG be enabled for company logo upload or will this give errors with logo use in other parts of the POS?
* I'd like the default Bootstrap theming to be added in addition to the Bootswatch themes.
* Should there be an option to add an Apple touch icon? Maybe in an advanced config tab. Could also be moved to a separate `config/design` together with the theme selector and the login form style option.
* Added `noindex, nofollow` to the login view. Is there an argument not too?
[...]

PR #3213, opened 2021-06-14
Login improvements

BudsieBuds:
See #3190 for more info.


molomixx:
Ok

jekkos:
@BudsieBuds is the other PR abandoned in favor of this one now?

Gruntfile.js
```diff
@@ -58,10 +49,89 @@ module.exports = function(grunt) {
 				}
 			}
 		},
+		copy: {
+			themes: {
+				files: [{
+					expand: true,
```
jekkos, line 55: indentation here

BudsieBuds:
@jekkos Sure, this one was added because I created a new branch. The other one was from the forks master. In hindsight, that wasn't ideal in the first place.

jekkos:
Ok should have known that, indeed. I'll give it a last look tonight. Also some testing/feedback from other contributors is welcomed @opensourcepos/org

jekkos:
Will also have a look at this codeQL thingy.. looks cool.

BudsieBuds:
Didn't mean to implement it, that was also a disadvantage of working in the master hehe. But yeah, don't know if I set it up in the optimal way. I think the GitHub actions we should look at are super-linter and PHP composer workflows.

There are a lot of bugs on the dev server right now, but don't know what does and what doesn't have to do with my PR. Most screens don't show normal content, glyphicons seem broken and switching a tab on the config screen sets the user back to home.

jekkos:
It looks like there is an issue in the minification. Maybe this is because of the new bootstrap libs, they might contain code that the grunt minifier is not able to handle.

Screenshot from 2021-06-15 15-54-32

jekkos:
what I did last time to fix this is exclude some libs from the minification, and include the minified versions directly.

jekkos:
@BudsieBuds if you add `?debug=true` you will load a version w/o minified JS. This one seems to work.

BudsieBuds:
@jekkos Yes, the config works with the `?debug=true` tag. Every theme seems to work fine, except for the Docs theme. It doesn't load the correct CSS file.

I didn't load the BS5 JS, because it's not needed for the login screen. I think the problem is in the way I changed the handling of the BS3 JS file.
Let me think out loud here: I've added it to the Gruntfile to copy it to the `public/dist/bootstrap/js/` and `grunt-tags` adds it to the `header.php`. The final difference is the following:
```diff
<!-- bower:js -->
...
- <script src="bower_components/bootstrap/dist/js/bootstrap.js"></script>
...
<!-- endbower -->
<!-- start js template tags -->
...
+ <script type="text/javascript" src="dist/bootstrap/js/bootstrap.min.js"></script>
...
<!-- end js template tags -->
```

Could something go wrong handling this?

I diffed the `opensourcepos.min.js` from the demo and the dev. There are some small differences, but the major diff is that the BS JS seems to be missing. It has this instead of the full file:
```
Package.describe({
  name: "twbs:bootstrap",
  summary:
    "The most popular front-end framework for developing responsive, mobile first projects on the web.",
  version: "5.0.1",
  git: "https://github.com/twbs/bootstrap.git",
}),
Package.onUse((t) => {
  t.versionsFrom("METEOR@1.0"),
    t.addFiles(["dist/css/bootstrap.css", "dist/js/bootstrap.js"], "client");
})(
```

Any suggestions?

jekkos:
> @jekkos Yes, the config works with the `?debug=true` tag. Every theme seems to work fine, except for the Docs theme. It doesn't load the correct CSS file.
> 
> I didn't load the BS5 JS, because it's not needed for the login screen. I think the problem is in the way I changed the handling of the BS3 JS file.
> Let me think out loud here: I've added it to the Gruntfile to copy it to the `public/dist/bootstrap/js/` and `grunt-tags` adds it to the `header.php`. The final difference is the following:
> 
> ```diff
> 
> ...
> - <script src="bower_components/bootstrap/dist/js/bootstrap.js"></script>
> ...
> 
> 
> ...
> + <script type="text/javascript" src="dist/bootstrap/js/bootstrap.min.js"></script>
> ...
> 
> ```
> 
> Could something go wrong handling this?
> 
> I diffed the `opensourcepos.min.js` from the demo and the dev. There are some small differences, but the major diff is that the BS JS seems to be missing. It has this instead of the full file:
> 
> ```
> Package.describe({
>   name: "twbs:bootstrap",
>   summary:
>     "The most popular front-end framework for developing responsive, mobile first projects on the web.",
>   version: "5.0.1",
>   git: "https://github.com/twbs/bootstrap.git",
> }),
> Package.onUse((t) => {
>   t.versionsFrom("METEOR@1.0"),
>     t.addFiles(["dist/css/bootstrap.css", "dist/js/bootstrap.js"], "client");
> })(
> ```
> 
> Any suggestions?
Thanks for the analysis.

That last code snippet seems to be the issue. It was not included before I think and it contains some statements that we can't minify properly. So in some cases it's better to include the minified JS directly instead of adding the original source and try to minify it ourselves (due to some specific statements that might crash the minification).

Gruntfile.js
```diff
@@ -145,7 +215,7 @@ module.exports = function(grunt) {
 					closeTag: '<!-- end js template tags -->',
 					ignorePath: '../../../public/'
 				},
-				src: ['public/js/jquery*', 'public/js/*.js'],
+				src: ['public/dist/bootstrap/js/*.min.js', 'public/js/jquery*', 'public/js/*.js'],
```
jekkos, line 218: @BudsieBuds guessing that the issue might be here. Can you remove the bootstrap from the minification? Let's try to include it directly instead

Gruntfile.js
```diff
@@ -269,7 +341,7 @@ module.exports = function(grunt) {
 	grunt.loadNpmTasks('grunt-apigen');
 	grunt.loadNpmTasks('grunt-contrib-compress');
 
-	grunt.registerTask('default', ['wiredep', 'bower_concat', 'bowercopy', 'concat', 'uglify', 'cssmin', 'tags', 'cachebreaker']);
+	grunt.registerTask('default', ['wiredep', 'bower_concat', 'bowercopy', 'copy', 'concat', 'uglify', 'cssmin', 'tags', 'cachebreaker']);
```
jekkos, line 344: I think the issue is that the copy happens before the concat. The concat will pick up this bootstrap source file and append it to the file that will be minified. Maybe we can switch order here or make sure that the bootstrap original source file is not included in the concat.

Gruntfile.js
```diff
@@ -58,10 +49,89 @@ module.exports = function(grunt) {
...
+					{
+					expand: true,
+					cwd: 'node_modules/bootstrap-5/dist/css',
+					src: ['bootstrap.css', 'bootstrap.min.css'],
+					dest: 'public/dist/bootswatch-5/bootstrap/',
+					filter: 'isFile'},
+					{
+					expand: true,
+					cwd: 'node_modules/bootstrap/dist/js',
+					src: ['bootstrap.js', 'bootstrap.min.js'],
```
jekkos, line 69: What you could try as a first thing is remove the original source file here, so if we concat something, we only concat the minified file.

Gruntfile.js
jekkos: Maybe let's try to remove the bootstrap source file first from the grunt build to see if that resolves the issue

Gruntfile.js
jekkos: Could you try to move copy step in Grunt after uglify?

Gruntfile.js
jekkos: @BudsieBuds If we use files from npm that are already minified, we can just copy them after the others are minified. So then we are certain that Grunt won't pick them up (which is the issue here apparantly).

Gruntfile.js
jekkos: If you want then I can have a look here.

BudsieBuds:
@jekkos It passed the checks. I'll wait until the dev server is updated to test it out.

jekkos:
@BudsieBuds I tested locally and this fix works, so moving the copy after the uglify.

jekkos:
I have pushed the fix.. let's wait for the build.

BudsieBuds:
Should I revert the changes I made in commit a7e2b8f?

jekkos:
It seems to be broken again indeed.
Locally I added my commit on top of the two original ones, so after the `Create codeql-analysis.yml`

I did a git reset --hard to that commit and will rebuild now to make sure it's fine. If so then I can force push to this branch if you want.

jekkos:
Ok seems fine on my local setup with that fix on top of your two commits. Let's start from there.

BudsieBuds:
> It seems to be broken again indeed.
> Locally I added my commit on top of the two original ones, so after the `Create codeql-analysis.yml`
> 
> I did a git reset --hard to that commit and will rebuild now to make sure it's fine. If so then I can force push to this branch if you want.

Yes, please do. If it's working I've got some small changes queud and after that I guess it's ready to be merged to master.

jekkos:
Ok last version deployed still has this issue, I'll have to take another look, maybe somewhere this weekend.

BudsieBuds:
@jekkos The issue with the glyphyicons fonts seems fixed, but the JS issue remains.

BudsieBuds:
Retrying with grunt back to version 0.4.5. It was updated in the package.json for some reason.

jekkos:
@BudsieBuds I don't understand why we have this isssue. The container on dev is the exact last commit you pushed here. Locally it works fine. I run the build with the same command as the one travis is using..

jekkos:
Am trying to remove node_modules and dist/bower_components locally to do a rebuild. Maybe it's related to the buildbox upgrade.

jekkos:
Indeed after this refresh I can now reproduce the issue locally.

jekkos:
Travis currently uses npm 6.14.12, grunt 1.4.0
we might want to try and install grunt-cli using npm instead

jekkos:
The culprit is bower, for some reason it's download bootstrap 5 now and it adds it to the files to be minified. This doesn't work well as it contains some newer ES6 syntax. 
As the idea was to get rid of boostrap via bower, I added an exclude for bootstrap and use the copy file in the concat step.

This however creates a new issue, as bootstrap itself depends on jQuery. I'd say that in this case we should include jQuery as the very first dependency, and to do this we might need to migrate it to npm as well.

jekkos:
@BudsieBuds I pushed my WIP on this.

jekkos:
So the issue here mainly is that once a lib is moved to npm, we should also find the proper order in which to include files  for the minifcation. 

The concat step will add bower managed dependencies in the correct order as it sees how the dependencies are related.
An alternative to moving jquery to npm would be to add back the bootswatch and only copy bs5 using npm for now.

BudsieBuds:
@jekkos Any solution is fine with me at this point. In the future, we can take a look at how to optimize everything. I think the main thing now is to just get it working.

jekkos:
@BudsieBuds I had to revert to option2. Keep the grunt build more or less as is and use npm to download bootstrap 5 for the login page only. So locally the javascript works fine. Will check dev server later on.

jekkos:
Ok I had to revert one more thing in the buildbox, it's on dev now and it looks fine.

jekkos (approved):

jekkos:
I think we are good for this one. Any objections before merge?

BudsieBuds:
@jekkos Very good! I did encounter a bug, try submitting a change in the config screen. It gives the success message in a new screen without markup. Also, the language files should be updated for other languages. Can you do this automagically with Weblate?

jekkos:
@BudsieBuds I had a post receive hook in weblate that would sync these strings automatically. But it might be broken so I'll do it manually later this week.

jekkos:
@BudsieBuds I have rebased this branch with the latest fix on master and the config save works fine on my end.

jekkos:
Language files should also be synced now.

jekkos:
Deployed on dev and looks fine at first glance, would be good if some more folks could do some sanity checks before the merge? @opensourcepos/org

BudsieBuds:
I've played around with it a little, tested out all the themes. Everything seems fine to me.

Issue #3239, opened 2021-07-09
Employee add or update failed OSPOS


kizitox2:
### I installed ospos for managing inventory in a local store.At first everything was working fine until when I tried to add a new employee and I received error message **Employee add or update failed**. I am using ospos 3.3.4

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 3.3.4
- OSPOS git commit hash is: 
- PHP version is: 7.3.29
- MySQL or MariaDB version is: 5.7.23-23
- OS and version is:  Linux 4.19.150-76.ELK.el7.x86_64
- WebServer is: Apache
- Selected language is: English-GB
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature
[...]

PR #3264, opened 2021-08-05
Fix employee update (#3239)

jekkos:
@odiea your fix was applied here

### DIFF
--- a/application/views/employees/form.php
+++ b/application/views/employees/form.php
@@ -209,17 +209,10 @@ $(document).ready(function()
 			last_name: 'required',
 			username:
 			{
-				<?php
-				if($person_info->person_id == "")
-				{
-				?>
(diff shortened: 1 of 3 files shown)


application/views/employees/form.php
```diff
@@ -209,9 +209,16 @@
 			last_name: 'required',
 			username:
 			{
+				<?php
+				if($person_info->person_id == "")
+				{
+				?>
+				remote: "<?php echo site_url($controller_name . '/check_username')?>",
```
objecttothis, line 216: IMO we should use interpolation here since we've done that elsewhere, but I'll leave it up to you: 
`remote: "<?php echo site_url("$controller_name/check_username")?>",`

objecttothis (approved):

PR #3266, opened 2021-08-05
Fixed missing function call

objecttothis:
This removes the function call to a function that doesn't exist anymore.
The replacement does the same job in one line of code.
Added comment to bring clarity to what the code is doing.

This code was producing an error message because custom_array_filter() no longer exists in the Item_lib.  I replaced it with this line which does the exact same function.


jekkos (approved):

jekkos:
Thanks for the fix.

PR #3271, opened 2021-08-10
Upgrade CI to 4.1.3 (Old)

objecttothis:
@jekkos created a new PR with new branch. Don't use this one.  We need the branch to see what needs to be ported which has been merged into the master since this branch was created.

Here are a list of issues that need to be resolved:

- ~~#3451~~
- ~~#3345~~ 
- ~~#3349~~
- #3453 NEEDS OWNER
- #3452 NEEDS OWNER
- ~~#3454~~ 
- #3455 @objecttothis 
- ~~#3456~~
- ~~#3457~~
- ~~#3458~~
- #3468 @jekkos
- #3584 @jekkos 
- #3585 @jekkos 
- #3586 @objecttothis


daN4cat:
Is this all we need to switch to CI 4?

I would have expected more changes than this but I might be wrong.

objecttothis:
> 
> 
> Is this all we need to switch to CI 4?
> 
> I would have expected more changes than this but I might be wrong.

hehe.  No, there is a lot more.  It's a draft pull request right now.  I'm working on the changes now.  The good news is that it's mostly syntax and naming changes.  The bad news is that it's going to touch just about every file, so yay for refactoring!

daN4cat:
Indeed what I was expecting

daN4cat:
Why the My_ files are removed?

objecttothis:
> Why are the My_ files removed?

This is from the upgrade guide

"There is no longer a nested application/core folder, as we have a different mechanism for extending framework components (see below)."

I haven't gotten to the part where the replacement mechanism is explained.  I'm working through it methodically so that I don't miss anything.

jekkos:
@objecttothis Here my first contribution. I have updated the composer buildbox and moved it to opensourcepos organization. Also have added the intl and mbstring dependencies to the buildbox. Let's see if it can install CI using composer now.

I have temporarily removed some of the unix commands for updating the version in the config files (as they were changes as you said).

objecttothis:
> 
> 
> @objecttothis Here my first contribution. I have updated the composer buildbox and moved it to opensourcepos organization. Also have added the intl and mbstring dependencies to the buildbox. Let's see if it can install CI using composer now.
> 
> I have temporarily removed some of the unix commands for updating the version in the config files (as they were changes as you said).

Thanks for doing that.  I'd say let's keep both working in the same branch and I can just fetch commits when I work on it.  I think when/if people take parts of the refactoring work, they can just do the same.

objecttothis:
@daN4cat yeah, the mistakes were caused by using Replace in All Files.  It works great for finding and replacing `array(` but the closing `]` has to be done manually.  Fortunately PHPStorm is pretty good at showing where those are, but I ran out of time and since this is a WIP, I wanted to commit what I had at the end of the day.  I do appreciate you taking a look as I go though, because it can help reduce the workload at the end since, no matter which way we slice it, this conversion is huge.

odiea:
Please let me know when you are ready for me to start testing this. Thanks for all the work you are doing.

objecttothis:
@jekkos Thanks for helping out with fixing the build errors.  I noticed you removed the ext dependencies from composer.json.  I'm just curious why they needed to be removed since they are used in the code. I added them because PHPStorm was yelling at me about them. Like I said in gitter, I'm new to composer, so I'm just trying to learn here.
image

objecttothis:
> @jekkos Thanks for helping out with fixing the build errors. I noticed you removed the ext dependencies from composer.json. I'm just curious why they needed to be removed since they are used in the code. I added them because PHPStorm was yelling at me about them. Like I said in gitter, I'm new to composer, so I'm just trying to learn here. image
> 
> image

@jekkos nevermind, I see your comments in gitter now.  Makes sense.

objecttothis:
The exception handler for CI4 is super sexy. Obviously, that's only helpful for development and we will not want it to display exceptions in all their glory for production, but I suppose that is what the env file is for.
image

app/Config/Filters.php
```diff
@@ -29,12 +29,12 @@ class Filters extends BaseConfig
 	 */
 	public $globals = [
 		'before' => [
-			// 'honeypot',
+			'honeypot',
```
jekkos, line 32: interesting

Issue #2673, opened 2020-01-10
Clarification for Keyboard shortcuts... 


izeblood:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

#Issue 
@daN4cat @SteveIreland @jekkos @odiea any idea to this?
@nvk1990
Tagging from #1630 issue
I try this and it is really works....
$(document).keydown(function(e){ if (e.which == 119) { //focus on customer search $('#customer').focus(); } else if(e.which == 121) { //simulates add payment button click $('#add_payment_form').submit(); } });

[...]

Issue #2618, opened 2019-12-06
Adding support for keyboard shortcuts and Function keys.


DeepRootedCoder:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [] Reporting an issue of an unmodified OSPOS installation
- [] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

- OSPOS version is: 
- OSPOS git commit hash is: 
- PHP version is: 
- MySQL or MariaDB version is: 
- OS and version is: 
- WebServer is: 
- Selected language is: 
- (If applicable) Docker installation: 
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

Please write your issue here. If a bug, please make sure to provide as much information as possible including configuration settings (e.g. Decimals set, Tax mode), language and steps to reproduce the bug.

New Feature Request:
[...]

Issue #1964, opened 2018-04-28
Hotkeys please


kosimoff:
Please add hotkeys to Sales tab

daN4cat:
Please search old issue to read about this topic.

daN4cat:
See #1083 and #425

mckaygerhard:
thanks for the references.. mostly POS systems has shorcut keys specially in the funtions key but this pos system its running inside a browser, so i doupt clients understand that easyle.. 

in any case i describe better the situation (documentation please! not only references)
1. issue #425 at almost end/middle has some example code for functions keys, i'll later paste in my spanish wiki how to improved, but as most developers said, ospos runs inside a browser and can be confused
2. Issue #1083 has code example for use other keys as quick access, using js easy to embebed..

only those two are usefully, the other are only garbage and a example that search issues are no easy , so never will be easy to normal users, how could normal users find the info if are less informed about the application @daN4cat and @WebShells ?

WebShells:
A final stable fix was applied to this issue through PR #4532

Issue #1630, opened 2017-09-20
New Feature / Enhancement - Shortcut Keys in Sales module


nashtm:
### New Feature / Enhancement

Can Shortcut Keys be implemented in the Sales module for Payment Type , Add Payment and Complete buttons to speed up sales processing?  I think it will be a benefit.
thanks Tim

daN4cat:
Search closed issues, this was discussed.

nvk1990:
Tagging old issues so that somebody may find it handy: #425, #1083 

I use the below in script section of register.php

// numeric values available at https://css-tricks.com/snippets/javascript/javascript-keycodes/
$(document).keydown(function(e){
   if (e.which == 119) { //focus on customer search
$('#customer').focus();
}

    else if(e.which == 121) { //simulates add payment button click
      $('#add_payment_form').submit();
    }
});

these are simple ones I use and can be customised based on the usage.

Cheers!

nashtm:
thanks - will add js code as needed.  Tim

guillermohaad:
@nvk1990 thank you

izeblood:
> Tagging old issues so that somebody may find it handy: #425, #1083
> 
> I use the below in script section of register.php
> 
> `$(document).keydown(function(e){ if (e.which == 119) { //focus on customer search $('#customer').focus(); } else if(e.which == 121) { //simulates add payment button click $('#add_payment_form').submit(); } });`
> 
> ** Numeric values available at https://css-tricks.com/snippets/javascript/javascript-keycodes/
> these are simple ones I use and can be customised based on the usage.
> 
> Cheers!
@nvk1990 
how about to complete the sale? $('#finish_sale_form ').submit();  is no effect...

izeblood:
@daN4cat @SteveIreland @jekkos @odiea any idea to this?
@nvk1990  
I try this and it is really works....
$(document).keydown(function(e){ if (e.which == 119) { //focus on customer search $('#customer').focus(); } else if(e.which == 121) { //simulates add payment button click $('#add_payment_form').submit(); } });

But.... how about the quantity? the Complete button? and the Delete amount? please response....

dlwiindi:
still struggling with calling the finish_sale_form submission function. nothing happens.

PR #3312, opened 2021-09-23
Keyboard Shortcuts/Hotkeys

WebShells:
Added Shortcuts/Hotkeys to Register

ESC 	Cancels Current Quote/Invoice/Sale
ALT + 1 	Item Search
ALT + 2 	Customer Search
ALT + 3 	Suspend Current Sale
ALT + 4 	Show Suspended Sales
ALT + 5 	Edit Amount Tendered
ALT + 6 	Add Payment
ALT + 7 	Add Payment and Complete Invoice/Sale
ALT + 8 	Finish Quote/Invoice witdout payment
ALT + 9 	Open Shortcuts Window

#1964 #2673 #1630 #2618 #2554 #1083 #425


application/views/sales/register.php
```diff
@@ -54,14 +54,14 @@
 				<?php
 				}
 				?>
-
+
```
objecttothis, line 57: Whitespace was added into the code here

application/views/sales/register.php
```diff
@@ -54,14 +54,14 @@
...
-
+				
 				<li class="pull-right">
 					<button class='btn btn-default btn-sm modal-dlg' id='show_suspended_sales_button' data-href="<?php echo site_url($controller_name."/suspended"); ?>"
 							title="<?php echo $this->lang->line('sales_suspended_sales'); ?>">
 						<span class="glyphicon glyphicon-align-justify">&nbsp</span><?php echo $this->lang->line('sales_suspended_sales'); ?>
 					</button>
 				</li>
-
+
```
objecttothis, line 64: and here

application/views/sales/register.php
```diff
@@ -574,7 +578,7 @@
 					<div class='btn btn-sm btn-danger pull-right' id='cancel_sale_button'><span class="glyphicon glyphicon-remove">&nbsp</span><?php echo $this->lang->line('sales_cancel_sale'); ?></div>
 				</div>
 			<?php echo form_close(); ?>
-
+
```
objecttothis, line 581: extra whitespace

application/views/sales/register.php
```diff
@@ -396,8 +397,11 @@
 					<button class='btn btn-info btn-sm modal-dlg' data-btn-submit="<?php echo $this->lang->line('common_submit') ?>" data-href="<?php echo site_url("customers/view"); ?>"
 							title="<?php echo $this->lang->line($controller_name. '_new_customer'); ?>">
 						<span class="glyphicon glyphicon-user">&nbsp</span><?php echo $this->lang->line($controller_name. '_new_customer'); ?>
+					</button>					
+					<button class='btn btn-default btn-sm modal-dlg' id='show_keyboard_help' data-href="<?php echo site_url($controller_name."/sales_keyboard_help"); ?>"
```
objecttothis, line 401: Don't use " in php when no string interpolation is happening because it calls the parser unnecessarily.  This php should either be:
`<?php echo site_url($controller_name.'/sales_keyboard_help'); ?>`

or even better, use string interpolation

`<?php echo site_url("$controller_name/sales_keyboard_help"); ?>`

application/views/sales/register.php
```diff
@@ -924,6 +928,53 @@ function check_payment_type()
...
+			$("#finish_invoice_quote_button").click();
+			break;
+		case 57: // Alt + 9 Open Shortcuts Help Modal
+			$("#show_keyboard_help").click();
+			break;
+	}
+	
+	switch(event.keyCode) 
+	{
+		case 27: // ESC Cancel Current Sale
```
objecttothis, line 972: This could be problematic without some kind of a confirmation dialog "Are you sure you want to cancel this sale?"  I don't feel super strongly about it, but I can see one of our staff accidentally hitting escape and losing the entire sale.

application/language/ar-EG/sales_lang.php
```diff
@@ -180,3 +180,16 @@
 $lang["sales_work_order_number_duplicate"] = "رقم طلب العمل يجب ان يكون فريد.";
 $lang["sales_work_order_sent"] = "تم ارسال طلب العمل الى";
 $lang["sales_work_order_unsent"] = "فشل في ارسال طلب العمل الى";
+$lang["sales_key_help"] = "Shortcuts";
```
objecttothis, line 183: In the future you can save yourself some time by just adding these to the default language, which is English (US) and Weblate will pick it up for you.

objecttothis (changes requested):
Great work on this!  Just the few comments.

objecttothis:
@WebShells I didn't have time to test the feature.  Maybe @odiea could give it a spin?

odiea:
Great addition. Looks and works good

application/views/sales/register.php
WebShells: @objecttothis while trying it the browser sent back a confirmation dialog upon pressing ESC. I can easily add it if there is a need for it no problem.

WebShells:
> 
> 
> Great addition. Looks and works good

> 
> 
> Great work on this! Just the few comments.

Thanks for the review guys comments were resolved. 
I think it will help during the sales process if the employee got used to these shortcuts.
Do you think we should add them to db and make them configurable/re-arrange-able or this is enough ?

application/views/sales/register.php
```diff
@@ -924,6 +928,53 @@ function check_payment_type()
...
 	}
 }
+
+// Add Keyboard Shortcuts/Hotkeys to Sale Register
+document.body.onkeyup = function(e)
+{
+	switch(event.altKey && event.keyCode) 
+	{
+        case 49: // Alt + 1 Items Seach
+			$("#item").focus();
```
jekkos, line 938: @WebShells might be nitpicking but I think you can chain these operations so that they are on one line.

application/views/sales/register.php
```diff
@@ -924,6 +928,53 @@ function check_payment_type()
...
+            break;
+        case 50: // Alt + 2 Customers Search
+			$("#customer").focus();
+			$("#customer").select();
+            break;
+		case 51: // Alt + 3 Suspend Current Sale
+			$("#suspend_sale_button").click();
+			break;
+		case 52: // Alt + 4 Check Suspended
+			$("#show_suspended_sales_button").click();
```
jekkos, line 949: As you suggest I think we could make this functionality quite a bit better by loading an associative array from database that contains keycode as key and field id as value? It will make this code a bit more generic as you need to determine the object type you find (focus and select in case of a dropdown, click in case of a button)

application/views/sales/help.php
```diff
@@ -0,0 +1,97 @@
...
+		<table class="table table-striped table-hover">
+		<thead>
+			<tr>
+				<th><?php echo $this->lang->line('sales_key_help'); ?></th>
+				<th><?php echo $this->lang->line('sales_key_function'); ?></th>
+			</tr>
+		</thead>
+			<tbody>
+				<tr>
+					<td><code>F11</code></td>
```
jekkos, line 70: great idea to have this help page here!

application/language/zh-Hant/sales_lang.php
```diff
@@ -180,3 +180,16 @@
 $lang["sales_work_order_number_duplicate"] = "工作編號重複.";
 $lang["sales_work_order_sent"] = "發送工作指示";
 $lang["sales_work_order_unsent"] = "工作指示發送失敗";
+$lang["sales_key_help"] = "Shortcuts";
```
jekkos, line 183: might be nitpicking but we might just leave new keys empty now. It's more clear in weblate that the string needs a translation in that case. So better for statistics.

There is a fallback in place that loads english translations for empty strings.

application/views/sales/register.php
jekkos: So in that case this logic will be encoded in a datastructure instead. I think configuring this from the UI might be out of scope for this PR. But then people will be able to change mappings by running a database query?

jekkos:
If you merge master in this branch, we might see the commit twice in the end

objecttothis (approved):

objecttothis:
Sorry, I think it's kinda dumb that when one person requests changes another can't approve the changes and the person merge... The original person has to go back in and approve it.  I wish they made it an option to merge if at least one approval happens regardless of who requested changes.

jekkos:
> Sorry, I think it's kinda dumb that when one person requests changes another can't approve the changes and the person merge... The original person has to go back in and approve it. I wish they made it an option to merge if at least one approval happens regardless of who requested changes.

But then we would just bypass that commenter's concern no?

objecttothis:
> > Sorry, I think it's kinda dumb that when one person requests changes another can't approve the changes and the person merge... The original person has to go back in and approve it. I wish they made it an option to merge if at least one approval happens regardless of who requested changes.
> 
> But then we would just bypass that commenter's concern no?

I meant only that only another approved "approver" could override rather than just anyone.

daN4cat:
Is this good to merge in? (Provided the merge conflict is solved....)

jekkos:
@WebShells I have tried the branch but the shortcuts don't work for me in Firefox. It seems that ALT combinations are already shortcuts within FF itself, so it's not usable.

I'm quite certain that it will only add value once you allow a user to override the bindings. I haven't tested in chrome but maybe someone else can confirm whether this works?

WebShells:
@odiea holaaa thanks for testing it

jekkos:
@WebShells It might be gnome on linux that has overlapping shortcuts, using alt+number just brings up the n-th tab in my FF?

jekkos:
@opensourcepos/org to merge or not to merge?

WebShells:
My vote is to merge. 
I'll be following it with an update to re-assign keys based on users interest and displaying it on keyboard shortcuts help page for 3.3.7, but i don't want/prefer to postpone this one to future releases.

jekkos:
@WebShells I was thinking that the help button might not be in the best place currently. It also disappears once you add an item to the register. Maybe something we can get some advice on from @BudsieBuds as he seems to be quite UX savvy.

PR #3313, opened 2021-09-23
customer address in receipt.php

pedrobuck:
Hello, I would like to add the customer address on the receipts.php page

I have a cafeteria and need the customer's address for the couriers. How do I do?


daN4cat:
This looks to me like you are suggesting a change done by jekkos but there is no change of receipt.
I guess this PR must be a mistake or partial commit.
Please clarify.

daN4cat (changes requested):
This cannot be merged as it stands.

jekkos:
I guess anyone can signoff using anyone else's credentials on git if you're not using pgp or sth... In this case it might be some commits from the past rebased on his local machine?

Issue #3284, opened 2021-08-26
Trying to get in touch regarding a security issue


zidingz:
Hey there!

I'd like to report a security issue but cannot find contact instructions on your repository.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

odiea:
@jekkos I reinstated your Blind sql injection fixes back into the Items controller. NAME shows an error in the log but it does show the Tables but the Item id does not sore properly. I replaced NAME with $item_id and no errors and the sort does now work. Will using the $item_id create any security issues?

jekkos:
Ok so that means you have both the SQLi fix and a working table? Yes the NAME just refers to the default sorting column in case none is defined, so using item_id instead should be good.

But in this case it should not be the $item_id variable but the column name.

odiea:
It appears that what ever that is changed to the other columns do not sort properly.

jekkos:
I should take a proper look again at this.. currently unable so I'll postpone until October.

PR #3316, opened 2021-09-27
Blind sql injection fixes rework (#3284)

jekkos:
I have reinstated the fix from a month ago. Want to get this out to release the 3.3.6.

### DIFF
--- a/application/controllers/Attributes.php
+++ b/application/controllers/Attributes.php
@@ -24,7 +24,7 @@ class Attributes extends Secure_Controller
 		$search = $this->input->get('search');
 		$limit  = $this->input->get('limit');
 		$offset = $this->input->get('offset');
-		$sort   = $this->input->get('sort');
+		$sort   = $this->Attribute->sort_column($this->input->get('sort'));
 		$order  = $this->input->get('order');
 
 		$attributes = $this->Attribute->search($search, $limit, $offset, $sort, $order);
--- a/application/controllers/Cashups.php
+++ b/application/controllers/Cashups.php
@@ -25,7 +25,7 @@ class Cashups extends Secure_Controller
 		$search   = $this->input->get('search');
 		$limit    = $this->input->get('limit');
 		$offset   = $this->input->get('offset');
-		$sort     = $this->input->get('sort');
+		$sort     = $this->Cashup->sort_column($this->input->get('sort'));
 		$order    = $this->input->get('order');
 		$filters  = array(
 					 'start_date' => $this->input->get('start_date'),
--- a/application/controllers/Customers.php
+++ b/application/controllers/Customers.php
@@ -58,7 +58,7 @@ class Customers extends Persons
 		$search = $this->input->get('search');
 		$limit  = $this->input->get('limit');
 		$offset = $this->input->get('offset');
-		$sort   = $this->input->get('sort');
+		$sort   = $this->Customer->sort_column($this->input->get('sort'));
 		$order  = $this->input->get('order');
 
 		$customers = $this->Customer->search($search, $limit, $offset, $sort, $order);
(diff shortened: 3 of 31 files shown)


odiea:
Applying this to Items the only column that sorts is the Item_id column. All other columns just sort the Item_id.

jekkos:
Thanks for testing @odiea  I have checked your comment. Which column were you trying to sort? If I sort on name or category it does seem to work

**sorting on barcode**
Screenshot from 2021-09-28 09-22-07
Screenshot from 2021-09-28 09-22-01

**sorting on name**

Screenshot from 2021-09-28 09-21-50
Screenshot from 2021-09-28 09-21-56

odiea:
Here is what I see. This was sorting on Name or Wholesale Price.
2021-09-28 02_03_04-Open Source Point of Sale _ Powered by OSPOS 3 3 5
2021-09-28 02_03_31-Open Source Point of Sale _ Powered by OSPOS 3 3 5
2021-09-28 02_04_00-Open Source Point of Sale _ Powered by OSPOS 3 3 5

jekkos:
Ok weird, both cases seem to work fine on my end. I did force push a couple of times. Do you have the latest hash in use for the branch?

odiea:
I downloaded this branch. © 2010 - 2021 · opensourcepos.org · 3.3.5 - 4b12b7.

jekkos:
Ok thanks for the feedback, I have deployed the version to dev now and was checking, the name field does not sort correctly. I think the others do in fact.

odiea:
Here is the code that I am testing out in Items.
$sort = $this->Item->sort_column($this->input->get('sort')); // controller
public function sort_column($field)
	{
		return array_key_exists($field, $this->db->list_fields('items')) ? $field : 'item_id';
	} // model

jekkos:
Think it should be cfe927

jekkos:
Indeed there is still issue, I'll try to debug it, thanks for the feedback

odiea:
I tried changing this and it appears to sort the columns correctly now.
 return array_key_exists($field, $this->db->list_fields('items')) ? 'item_id' : $field;

jekkos:
Did another push and seems better now on dev.

odiea:
The only quirk I see now is on an older established db quantities does not sort? Looks ok on the Demo

jekkos:
@odiea you mean the regular quantities field?

jekkos:
@odiea should be fixed now.

odiea:
Yes it is. Works good now.

odiea:
What is the issue with placing the id before the $field?  I have had to do this to get some of the other tables to sort correctly.

jekkos:
Well what the logic does it it checks if the supplied sorting column exists in the db, then applies it.

If some hacker tries to pass in some SQL for injection, it will be ignored as we only allow to sort on db column names. So that it was the statement does

odiea:
But this does not appear to work correctly. I tried sorting in Customers, Suppliers and Employees and columns do not sort properly. The appear to sort to the id and not the column.

jekkos:
Indeed well spotted, I think those three models need one more adapation

odiea:
Try something on Gift cards and you may be done with this

jekkos:
Ok the columns for giftcards are extended now. I've gotten two more security bug reports through this huntr.dev platform today. Not sure how this works but it seems that some researchers are looking for security bugs in ospos now.. which is good.

jekkos:
The sale table might not work correctly yet.. The current approach will do some extra queries to fetch the database column names, which do not necessarily map to the column headers names.

I had a further look and there might be a better solution. The header names are already defined in the tabular_helper, but they are not in an accessible place. It might be better to extract those header arrays somewhere and then check for the field presence in each array separately.

jekkos:
Links to #3965

objecttothis:
@jekkos can we merge this into ci4-branch instead of master?  This fix will likely need to be applied to all sort and order fields in the various searches to close the vulnerability.  I realize it will likely mean porting the changes over to a new branch based on ci4-branch, but I'd hate to duplicate work.

PR #3318, opened 2021-09-28
Extend method hook validation for deletes

jekkos:
We need to check the HTTP method to prevent CSRF attacks. A logged in user can be tricked to remove items if an attacker embeds a GET call to the delete endpoint in another website.

CI will ignore the CSRF token if the method is changed to a GET. We should disallow this method for REST resource mutations.

This bug was reported through the huntr.dev platform.


jekkos:
Screenshot from 2021-09-29 00-04-59

PR #3323, opened 2021-09-30
[bug]changed the mount file name from database.sql to tables.sql

argoyal:
there is no file named database.sql available in the database folder. I was able to use tables.sql to create my schema and was finally able to run the application after a migration at login. have raised the pull request anyways.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,7 +44,7 @@ services:
         networks:
           - db_net
         volumes:
-          - ./database/database.sql:/docker-entrypoint-initdb.d/database.sql
+          - ./database/tables.sql:/docker-entrypoint-initdb.d/database.sql
           - mysql:/var/lib/mysql:rw
         environment:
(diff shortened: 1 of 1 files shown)


jekkos:
Hi, thanks for the PR. I think you should concat both tables and constraints.sql. If you run grunt at least once then the database.sql file should be created. It doesn't exist because you have probably cloned the repo w/o running a build first.

argoyal:
@jekkos thanks, yes figured that out.. I was running it through Docker and expected it to run out of the box with a `sudo docker-compose up`.  In this case I get 500 because my database was not loaded and therefore I tried searching for it and I could not find database.sql anywhere in a fresh clone, digging deeper I realised how database.sql is created.

I will cancel this PR. I think the Dockerfile needs a lot of improvement, because it expects the npm commands to be run outside the Docker build.

Also, it might be a little difficult to perform development using Docker with this structure, trying to figure out how that can also be improved.

PR #3337, opened 2021-10-05
Fix supplier csrf

jekkos:


daN4cat:
I checked sales register code and not sure how I could solve this given my limited front end dev experience. We would need to store the line number somewhere to be reused in the JS POST code. Nothing springs to mind... @jekkos I leave this to you, sorry for being helpless this time.

PR #3380, opened 2021-12-23
Mobile Reponsive

mogsama:
I'm just a shop owner, I'm not very good in coding. So I think you just take my code and idea as an example to enhance yours. My idea is:

1. Bigger register_wrapper and overall_sale panel on mobile only using min and max width. So one above and one below.
147132283-2fd6fc7c-5582-4d93-b1da-6972178c00cf
2. This is made for us to use in backend. So it should be as simple and minimal as possible. We don't need to be told this is customer's name, that is customer address, etc. As a result I removed quite a lot of language code. Now the button is positioned in one row only and looks nice.
3. I cannot get the person's comments to appear on register so I use Account number instead. I think it is better for the staff to see it on register panel immediately to act accordingly.

I'm quite happy with OSPOS right now. Hope my idea can help you on your project too.

P/S: In 3.3.4 the css styles I added is in /public/public/dist/opensourcepos.min.css but I can't find it in the current master branch.

### DIFF
--- a/application/views/partial/header.php
+++ b/application/views/partial/header.php
@@ -1,6 +1,7 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
+	<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0">
 	<meta http-equiv="content-type" content="text/html; charset=utf-8" />
 	<base href="<?php echo base_url();?>" />
 	<title><?php echo $this->config->item('company') . ' | ' . $this->lang->line('common_powered_by') . ' OSPOS ' . $this->config->item('application_version') ?></title>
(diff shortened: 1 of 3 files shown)


daN4cat:
It looks like some text labels are removed.

0x10101:
I made the changes to css and views but It isn't working for me, what I'm missing,

jekkos:
This does not look too bad @mogsama . But I wonder how the desktop version then looks like. Also the percentage field seems to be a bit packed in this case. I think @BudsieBuds was also playing around here a bit.

mogsama:
> This does not look too bad @mogsama . But I wonder how the desktop version then looks like. Also the percentage field seems to be a bit packed in this case. I think @BudsieBuds was also playing around here a bit.

**My current solution is:**
Narrow the Quantity column by shortening the <th> as only "pcs" or "quant."
Pack the Price and Discount column together.
Change column with in % to suit the situation.
Hope it will be better with Bootstrap V5 as I see there will be no padding & margin gap in the input field.

**Other option:**
Would you please make the quantity, price, discount as a link. This will dramatically help the mobile responsiveness.
When you click the link you will a popup to change the value. With quantity there will be a Up and Down arrow to Increase or Decrease it by one.

BudsieBuds:
> This does not look too bad @mogsama . But I wonder how the desktop version then looks like. Also the percentage field seems to be a bit packed in this case. I think @BudsieBuds was also playing around here a bit.

Yeah. If I get to completing the conversion to BootStrap 5, a responsive layout will be included.

PR #3383, opened 2021-12-28
Fix logout csrf

jekkos:
This PR could fix the logout CSRF issue we currently have (change GET to POST)


odiea:
Appears to work just fine.

jekkos:
@odiea which browser did you test? I tried on chrome and it works consistently. On Firefox I seem to have an occasional hickup (and sometimes I need to click twice)

odiea:
I have now tried it on Edge, Google Chrome, Pale Moon and Firefox. Using Windows 10. Minor hick up on Edge the first time but ok the second time.

odiea:
On Microsoft Edge it jut took a couple seconds to logout the first time I tried it.

daN4cat:
For sure a GET logout was wrong, also noticed that password is GET and not POST. Worth changing that too with this fix?

jekkos:
@daN4cat if the change password is done through GET I think it's more critical then this one, of course a user can always reset again after an attacker changed the pwd.

jekkos:
@daN4cat is it the password change dialog ? I checked and the dialog itself is invoked by a GET while the change itself is a POST

daN4cat:
Yes, it's the dialog box invocation. You are right, GET for that is fine.

Issue #2834, opened 2020-05-20
Replace Travis with GitHub Actions


daN4cat:
This is a placeholder to consider GitHub action as CI/CD pipeline.

daN4cat:
Or Azure Pipelines given there are 10 free pipelines that can be used in GitHub.

jekkos:
Duplicate of  #3110

jekkos:
I might have overseen this, but it seem that travis-ci.com only offers a non renewable set of credits. You basically need to move to their paid plan to have the credits renewed..

I think currently the only option is to move to github actions now. Unless we can find 70$ a month in extra funding to pay for travis plan.

BudsieBuds:
So the [Always Free banner on their website](https://www.travis-ci.com/plans) doesn't apply for this project?

> ### Free for Open Source
> Free for open source. We love the Open Source Community, and to show how much we love it, upon validated request placed with our Support Team you may receive free OSS credits for your public builds.

jekkos:
Indeed it wasn't 100% clear to me how to apply. I emailed them asking for extra credits so let's wait for this first.

jekkos:
I'm not sure but I think the unstable draft releases are only visible to contributors. It's not ideal as it stands, the fact that I need to tag a release in order to get the files in there.

I might look into pushing the zip to another repo so it can be downloaded from there.

jekkos:
Using draft releases for unstable builds does not feel like the best solution.

I think github packages might be a better solution for unstable builds. It's 500MB max for the free plan, but it will support version tagging out of the box.

PR #3396, opened 2022-01-08
Enable npm package uploads for unstable (#2834)

jekkos:
I tried to do some changes in the CI setup some time ago. In this case instead of pushing to github release I try to use npm and publish unstable builds to github npm repo. I think it can only hold 500MB so let's see how far we get with this. I could eventually just spin up a npm repo using docker on a nas at home here to host the unstable builds in case it's not easy enough to remove older builds during deploy


jekkos:
Travis-ci seems to have the option to publish to multiple providers. This works now. Unstable tarballs are pushed to github npm (check the packages tab under opensourcepos org). Users can download them for any branch under that section.

The github releases is then only used for proper 'stable' releases. I should update the docs now to clarify this.

jekkos:
https://github.com/orgs/opensourcepos/packages

PR #3397, opened 2022-01-08
Bump to 3.3.7

jekkos:
Bump can be done if we have no other critical things to fix for the current release.


jekkos:
Holding this back until we know we did not break too much stuff.

daN4cat:
Unless you plan to bump 3.3.6, a new release fixing just two odd bugs will be required.
This time I would not wait too long because the deleted issue is annoying.

jekkos:
Ok the bug is fixed and in master now , will merge this one and then wait a bit until hopefully no more bugs are found :)

minhazulOO7:
> Unless you plan to bump 3.3.6, a new release fixing just two odd bugs will be required. This time I would not wait too long because the deleted issue is annoying.

Bro! Any delete issue with 3.3.6? Just want to get sure because will be updating soon! Please reply when you are free! 😱

jekkos:
In master now

PR #3408, opened 2022-01-24
Change register to show anticipated invoice number.

SteveIreland:


application/controllers/Sales.php
```diff
@@ -627,8 +627,7 @@ public function complete()
 		if($this->sale_lib->is_invoice_mode())
 		{
 			$invoice_format = $this->config->item('sales_invoice_format');
-
-			// generate final invoice number (if using the invoice in sales by receipt mode then the invoice number can be manually entered or altered in some way
+			// generate final invoice numbr (if using the invoice in sales by receipt mode then the invoice number can be manually entered or altered in some way
```
daN4cat, line 630: Any reason for this change to munbr in the comment?

application/controllers/Sales.php
```diff
@@ -1105,14 +1104,15 @@ private function _reload($data = array())
...
 
-		$invoice_number = $this->sale_lib->get_invoice_number();
+		$temp_invoice_number = $this->sale_lib->get_invoice_number();
+		$invoice_format = $this->config->item('sales_invoice_format');
 
-		if ($this->sale_lib->get_invoice_number() == NULL)
+		if ($temp_invoice_number == NULL || $temp_invoice_number == '')
 		{
-			$invoice_number = $this->config->item('sales_invoice_format');
+			$temp_invoice_number = $this->token_lib->render($invoice_format, array(), FALSE);
```
jekkos, line 1112: if I understand well the number will be rendered if nothing is set before, so it will be rendered once and stay until the sale is completed.

application/models/Appconfig.php
```diff
@@ -80,24 +80,33 @@ public function delete_all()
 		return $this->db->empty_table('app_config');
 	}
 
-	public function acquire_save_next_invoice_sequence()
+	public function acquire_next_invoice_sequence($save = TRUE)
 	{
 		$last_used = $this->get('last_used_invoice_number') + 1;
-		$this->save('last_used_invoice_number', $last_used);
+		if($save)
```
jekkos, line 86: is there also a case where we have the save == FALSE?

jekkos: At first sight it feels like we can always save the incremented counter, unless I'm missing something.

jekkos (approved):

application/controllers/Sales.php
SteveIreland: Spelling obsessive compulsive disorder

application/controllers/Sales.php
SteveIreland: It's a bit more complex than that.  The temporary invoice number is returned so that it can be displayed to the user.  However, if the sale is then suspended then the next sale will be assigned that same invoice number (if it is completed).  When the suspended sale is unsuspended then the next invoice number to be used is displayed on the register (not the one that was previously displayed since it was assigned to an earlier completed sale.

PR #3447, opened 2022-03-30
Added meta tag for mobile responsive in header.php file

vishal262:
Mobile responsive is not working properly so I have added meta tag

### DIFF
--- a/application/views/partial/header.php
+++ b/application/views/partial/header.php
@@ -2,6 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
 	<meta http-equiv="content-type" content="text/html; charset=utf-8" />
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
 	<base href="<?php echo base_url();?>" />
 	<title><?php echo $this->config->item('company') . ' | ' . $this->lang->line('common_powered_by') . ' OSPOS ' . $this->config->item('application_version') ?></title>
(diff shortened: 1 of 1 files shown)


jekkos:
@BudsieBuds any comments on this one?

BudsieBuds:
@jekkos Same as the other ones. Unfortunately this doesn't magically make everything responsive. At least not in a ready for release kind of way.

@vishal262 Thanks for the contribution. You've probably seen that only the login page is responsive. Only the login page has been converted to Bootstrap 5, the rest of the app is still on Bootstrap 3 and will be upgraded once @objecttothis is finished with the big undertaking of converting the app from CodeIgniter 3 to version 4.

jekkos:
Thanks for the input @BudsieBuds . I'm not very familiar responsive browser and such. Let's wait for a proper rework instead of merging something half baked. If other opinions please shoot @here

objecttothis:
> Thanks for the input @BudsieBuds . I'm not very familiar responsive browser and such. Let's wait for a proper rework instead of merging something half baked. If other opinions please shoot @here

Making the webapp responsive just means that it scales well to devices of different screen sizes. Bootstrap takes care of all that and it's browser independent. I work with it in Foundation on another project all the time.  I created several issues regarding the remaining CI4 changes that need to be hammered out.  Once those are figured out then it's on to regression testing and debugging.  The more help I get, the faster it will go.

jekkos:
Ok @objecttothis let's wait for a proper solution, thanks for all the hard work already.

Issue #3379, opened 2021-12-22
Reflected XSS in item kits


jekkos:
A bug report that came in through [huntr.dev](https://huntr.dev/bounties/be3ebbe3-be95-4b54-85bd-f0bad9e03211/)

```
    Login to the demo account
    Go to item kits , edit any item and add payload in barcode field and click save
    payload "><iMg SrC="x" oNeRRor="alert(1);">
    poc 1 https://ibb.co/ZJZLKdQ
    poc 2 https://ibb.co/D4x2jSf
```

jekkos:
I had a quick look here and it's not immediately clear why we are able to execute a payload. All input seems to be escaped as follows

```
		foreach(get_object_vars($info) as $property => $value)
		{
			$info->$property = $this->xss_clean($value);
		}
```

This code is reused in other modules so if this trick bypasses the sanitization then I suppose it will also work on other pages. If someone finds the time to test this that would be nice.

jekkos:
https://benhoyt.com/writings/dont-sanitize-do-escape/
@daN4cat @objecttothis

daN4cat:
SQL injection on input, XSS on output

objecttothis:
xss_clean is known to be inadequate at best.  CI4 best practice is to filter inputs and escape outputs.  I'll take a look at the article you posted @jekkos

objecttothis:
I have wondered a few times how not filtering inputs would prevent sql injection, but it seems like what they are advocating to do is (in our case) escape both in the view (outputs) and escape in the models where queries are being run, but not worry about filtering what comes in.  It's not really security, but we should be doing validation (number fields, etc), so that will by nature do some amount of filtering.

jekkos:
The mitigation for this issue might be similar to the fix proposed for #3387 . The payload is in the json encoded array and needs to be escaped. In this case escape can happen in the backend or in the frontend. Bootstrap tables seems to have the escape property that could easily fix the issue for all pages.

jekkos:
If this does not work or has unwanted side effects, then we can try to do it [on the backend](https://stackoverflow.com/questions/12062146/is-json-encode-sufficient-xss-protection) instead, by adding some flags to json_encode.

objecttothis:
I think if we can make the built-in filtering/escaping mechanisms work well then that is preferable.  There are two potential pitfalls that are conceivable

1. Filtering and/or escaping is incredibly slow and it bogs the app down.
[...]

PR #3466, opened 2022-04-09
Enable table data escape (#3379)

jekkos:
Try to mitigate XSS for all modules by enabling bootstrap tables escape.


daN4cat:
If we do this we need to disable in config the XSS filter.

jekkos:
> If we do this we need to disable in config the XSS filter.

Maybe we can try to combine.. as it seems that this specific payload is able to bypass the CI escape?

jekkos:
@daN4cat what this flag does is[ documented here](https://bootstrap-table.com/docs/api/table-options/). So it's not a real xss escape but rather a html entity rendering step.

jekkos:
Thanks for testing @odia. I think this should be enabled per column instead.

PR #3471, opened 2022-04-13
Update nginx template

jekkos:
Set the host headers explicitly for ospos and phpmyadmin.


jekkos:
The fix is not needed as the template is filled in at container build time..

daN4cat:
Indeed.

Issue #3491, opened 2022-04-29
Feature Request: Search FIlter in Summary Reports Items and Suppliers


surprised-potato:
### New Feature
I'm loving OSPOS and I thank you all for your great work!

It would be nice if there was a search textbox inside Summary reports Items. It would be amazing if I could search for the items with their item names and filter by supplier.

chunter2:
I think you just want the last comment of this issue https://github.com/opensourcepos/opensourcepos/issues/3386

@jekkos Any chance this could be added to the code?

surprised-potato:
Thanks that what I was asking for. I got it to work as well.

PR #3494, opened 2022-04-29
Enable search in reports (#3491)

jekkos:
@chunter2 

If you want to propose tiny improvements like this, you can do so even using github directly. No need for a dev environment, as once the PR is made it will be deployed to dev and you can test it over there.

### DIFF
--- a/application/views/reports/tabular.php
+++ b/application/views/reports/tabular.php
@@ -55,7 +55,8 @@
 				data: <?php echo json_encode($data); ?>,
 				iconSize: 'sm',
 				paginationVAlign: 'bottom',
-				escape: false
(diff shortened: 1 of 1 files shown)


chunter2:
@jekkos I don't think I knew that.  I'll do some googling to see how to do it and give it a try next time.  I was just going through the manual changes I've got in my production system to try to make sure they make it in.  Thanks

jekkos:
Yes i think you can edit files on GitHub directly. Then save them and choose to open a PR with the change. If it doesn't work then we can just discard it.

jekkos:
It's on demo and it works fine, I'll merge.

PR #3495, opened 2022-04-29
Add cost price column to item summary report

chunter2:


chunter2:
@jekkos Not sure this is what you were talking about but it's an attempt.  :)  I was hoping I could edit both files and have them in one commit but didn't see how.

jekkos:
@chunter2 exactly small changes like these are easy to check this way..

jekkos:
@chunter2 I think builds are only ran if you create the branch in our repository directly. I have sent you an invite to become an organization member. Normally once accepted you should be able to do that.

jekkos:
See my other PR, it will now build and deploy.

PR #3496, opened 2022-04-29
Add cost price column to item summary report (#3495)

jekkos:


chunter2 (commented):
Just tested on dev.  Looks good to me.  Thanks

PR #3497, opened 2022-04-30
Enable search in detailed reports (#3495)

jekkos:
I think this might only work for the parent table


jekkos:
Tested on dev, seems to work fine. The search only looks in the parent table, not in details.

Issue #3519, opened 2022-06-01
opensourcepos 3.3.7 does not contain the opensourcepos.master.zip file.


odiea:
![2022-06-01 16_30_58-Window](https://user-images.githubusercontent.com/10962177/171512621-bb911f90-843f-492e-ada4-dd89e799d1b1.png)

jekkos:
Indeed I switched to tarballs some time ago, might not be so convenient for most users. I'll try to tweak the build config such that at least the release is a zipfile again.

WebShells:
Solved...

PR #3526, opened 2022-06-15
Add zipfile to releases (#3519)

jekkos:
Adds back the zip file after a release is tagged. Back from holidays btw


odiea:
The only way I could check this out would be after the master is rebuilt.

PR #3541, opened 2022-07-05
Manufacturer added to products

andriux1990:
I think it is very important to have a field for the manufacturer of the item... and to be able to use it in the future... in an easier way

### DIFF
--- a/application/controllers/Items.php
+++ b/application/controllers/Items.php
@@ -516,6 +516,7 @@ class Items extends Secure_Controller
 			'name' => $this->input->post('name'),
 			'description' => $this->input->post('description'),
 			'category' => $this->input->post('category'),
+			'manufacturer' => $this->input->post('manufacturer'),
 			'item_type' => $item_type,
 			'stock_type' => $this->input->post('stock_type') === NULL ? HAS_STOCK : intval($this->input->post('stock_type')),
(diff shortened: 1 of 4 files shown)


daN4cat (commented):
This commit is missing the database table change.
Also, I'm not sure why an attribute cannot do for the case.

daN4cat (changes requested):
Please change.

andriux1990:
> This commit is missing the database table change.
> Also, I'm not sure why an attribute cannot do for the case.

An attribute does it... but it's not the same... for example if we wanted to use the manufacturer's name as a search suggestion in sales... it's very easy... doing it with an attribute, it's very cumbersome

WebShells:
By manufacturer you mean brand ? 
The supplier module already have Agency name and Supplier name, can't we make use of Agency name field as a brand rather than adding additional fields ? 
Thank you

WebShells:
@andriux1990 That's why there is Agency name and Supplier name. So if you are buying Louis Vuitton from a different supplier the Brand is the Company name and the Agency Name is the Reseller ? Correct me if am wrong.

daN4cat:
Maybe we should have brand as separate from company in Suppliers.

Louis Vitton as company owns different brands, agents are a sales entity.

PR #3618, opened 2022-12-24
doc: update readme with proxy parameters

CliffordCheefoon:
Insert FAQ entry to bring visibility to configuration needed for running OSPOS with proxies and load balancers.


jekkos (approved):

jekkos:
Thanks for the tip!

Issue #3692, opened 2023-02-25
Convert all Boolean values from int(1) to tinyint(1)


objecttothis:
Throughout the database we have several boolean values defined in the SQL as int(1).  While MariaDB recognizes this, MySQL ignores the length specifier unless using tinyint(1).  We need to create a migration that will convert all boolean values to tinyint(1)

objecttothis:
https://www.mysqltutorial.org/mysql-boolean/

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
I can send a PR for this if someone else is not working on it already. Thanks.

objecttothis:
It's all yours.  Just create a new migration for it.  It might not get merged until after the ci4-upgrade branch gets updated, so please write the migration for CI4 rather than CI3.  If you aren't familiar with it, you can look at what the migrations in the ci4-upgrade branch look like... they essentially mostly call an SQL script in the up() function.  Also keep in mind that in the migration, no existing data can be lost, so we might have to create temp tables, or columns, transfer over the data to the correct column type then delete the old.... then again, it might convert.  I'm sure you'll figure that out.

PR #3709, opened 2023-03-08
int(1) to tinyint column updates

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
update int(1) columns in `_customers` and `_cash_up` tables to use tinyint finishes issue #3692 

Looking through each table after the migrations are run it looks like a lot of the old int(1) columns have been updated appropriately already.  These were the only remaining that looked like they would be good tinyint candidates.

Thanks.

### DIFF
--- a/app/Database/Migrations/20230307000000_int_to_tinyint.php
+++ b/app/Database/Migrations/20230307000000_int_to_tinyint.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class IntToTinyint extends Migration
+{
+	public function up(): void
+	{
+		$this->db->query('ALTER TABLE ' . $this->db->prefixTable('customers') . ' MODIFY `consent` tinyint NOT NULL DEFAULT 0');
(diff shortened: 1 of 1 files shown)


app/Database/Migrations/20230307000000_int_to_tinyint.php
objecttothis, line 1: Are these the only two columns that are int(1)? For some reason I thought there were more.

jarebear6expepjozn6rakjq5iczi3irqwphcvb: I am pretty sure that was it but will confirm this weekend.

jarebear6expepjozn6rakjq5iczi3irqwphcvb: As far as tables with columns being used as boolean values, yes these are the only remaining two.  

The lions share of them appear to have been altered in the [3.4.0_database_optimizations.sql](https://github.com/opensourcepos/opensourcepos/blob/master/application/migrations/sqlscripts/3.4.0_database_optimizations.sql) migration.

PR #3746, opened 2023-04-22
Update system_info.php

titusito:
add php_Xml dependencies check, php-xml is needed for dompdf to be able to create pdfs in invoice pdf sending.


titusito:
The docker-compose installation has the xml extension installed but local installs might not, adding the xml check will allow users not to have failing invoice sending pdfs.

PR #3781, opened 2023-06-27
Adapt "In dev mode, allow blank root password."

jekkos:
This reverts commit 8db87fd5aba43d47b70cf6e0032ff2074778da30.

@SteveIreland I saw you added a commit to add a blank password, if this is set through the env variable, wil lthat then work for you? I think I should have used the `development.* ` variables instead as those values are picked up from the config.php file, if I'm not mistaken

### DIFF
--- a/app/Config/Database.php
+++ b/app/Config/Database.php
@@ -102,20 +102,9 @@ class Database extends Config
     {
         parent::__construct();
 
-		if(!getenv('database.development.hostname'))
-		{
-			$this->development['hostname'] =  getenv('database.development.hostname');
-		}
-
-		if(!getenv('database.development.hostname'))
-		{
-			$this->development['username'] =  getenv('database.development.username');
-		}
-
-		if(!getenv('database.development.hostname'))
-		{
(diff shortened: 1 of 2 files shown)


SteveIreland:
The code that you reverted it back to will default to "pointofsale" when the root password is blank.  That's why I made the change.  I normally use the default blank password for dev.  However, if you want to block the use of blank root passwords then I'll change my behavior.

jekkos:
No it was not my intention to block that, I just wonder that if you evaluate `getenv('variable`) while variable is an empty string, does it give false? According to the docs it returns false if `if the environment variable name does not exist.`

SteveIreland:
Yes, I read that too.   But apparently the ternary operator assumes that an empty string also equates to false.   At least that was the case in my testing.   I was NOT testing with 8.0, so I don't know if that makes a difference.  I only know I was able to connect when I made that change.  But who knows, I could have been facing south when it needed me to be facing north in order to work

SteveIreland (approved):

PR #3792, opened 2023-08-11
allow unattended scheme mode

dennybaa:
Allow unattended scheme mode disregarding the mode server runs.
This basically means you can simultaneously connect say it to http://localhost:8080 (for example, might be exposed locally via a container) as well as at the same time you are free to connect via https://my.domain.tld. Both will work.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -65,7 +65,7 @@ $config['db_log_enabled'] = FALSE;
 | a PHP script and you can easily do that on your own.
 |
 */
-$config['base_url'] .= '//' . ((isset($_SERVER['HTTP_HOST'])) ? $_SERVER['HTTP_HOST'] : 'localhost') ;
+$config['base_url'] = '//' . ((isset($_SERVER['HTTP_HOST'])) ? $_SERVER['HTTP_HOST'] : 'localhost') ;
 $config['base_url'] .= str_replace(basename($_SERVER['SCRIPT_NAME']), '', $_SERVER['SCRIPT_NAME']);
 
 /*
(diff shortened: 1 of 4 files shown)


application/config/config.php
```diff
@@ -65,9 +65,7 @@
 | a PHP script and you can easily do that on your own.
 |
 */
-$config['https_on'] = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') || (isset($_ENV['FORCE_HTTPS']) && $_ENV['FORCE_HTTPS'] == 'true');
-$config['base_url'] = $config['https_on'] ? 'https' : 'http';
-$config['base_url'] .= '://' . ((isset($_SERVER['HTTP_HOST'])) ? $_SERVER['HTTP_HOST'] : 'localhost') ;
+$config['base_url'] .= '//' . ((isset($_SERVER['HTTP_HOST'])) ? $_SERVER['HTTP_HOST'] : 'localhost') ;
```
jekkos, line 68: So by ommitting the http prefix it will choose what is appropriate for accessing the site? Don't we need to turn the `.=` into a `= `?

dennybaa: Hello @jekkos, yes. relative resolution will be applied, i.e. the necessary protocol will be used depending on the page visited.

dennybaa: Totally we can switch to `=`

daN4cat: By memory, the force https was required if you had a reverse proxy in front.
Test it with the docker compose having nginx and see if it still works.

dennybaa: It's been tested, it works

jekkos (approved):

PR #3808, opened 2023-09-02
Revert config.php changes

daN4cat:
The changes to the config file are causing issues in running OSPOS in a docker container locally.


daN4cat:
The issue is that you can't POST anything, you get error with many forms. I spent quite some time convinced the addition I made was wrong to find out that there was this root cause.
It's a regression and the change needs to be tested better.

PR #3815, opened 2023-09-11
Ci4 enhancement use writable backup

owlbrudder:


app/Helpers/security_helper.php
```diff
@@ -17,12 +18,19 @@ function check_encryption(): bool
...
+		//For security reasons. the web server must not have write
+		//access to the base url, so we must not try to write or update files there.
+		//To make this code work, create the backup directory as below and
+		//ensure it is writeable by the webserver, including setting
+		//any SELinux permissions needed under eg Fedora
+		//We present a message to the user telling them to
+		//copy the .env file into place manually.
 		$config_path = ROOTPATH . '.env';
-		$backup_path = $config_path . '.bak';
+		$backup_path = ROOTPATH . '/writable/backup/.env.bak';
```
objecttothis, line 29: This is incorrect.  It should be `$backup_path = WRITEPATH . '/backup/.env.bak';`

https://codeigniter.com/user_guide/general/common_functions.html#WRITEPATH

app/Helpers/security_helper.php
```diff
@@ -33,18 +41,20 @@ function check_encryption(): bool
 			$config_file = substr_replace($config_file, $old_line, $insertion_point,0);
 		}
 
-		@chmod($config_path, 0770);
+		$new_config_path = ROOTPATH . '/writable/backup/.env' ;
```
objecttothis, line 44: WRITEPATH here rather than ROOTPATH

app/Helpers/security_helper.php
```diff
@@ -33,18 +41,20 @@ function check_encryption(): bool
...
 
-		if(is_writable($config_path))
+		//Changes to .env must only be invoked by Login controller
+		$router = service('router');
+		$controller = $router->controllerName();
+		if(strtolower(class_basename($controller)) == 'login')
```
objecttothis, line 49: Let's add a blank line between the if statement and variable declarations.

app/Helpers/security_helper.php
```diff
@@ -33,18 +41,20 @@ function check_encryption(): bool
...
 		{
-			// Write the new config.php file
-			$handle = @fopen($config_path, 'w+');
+			$handle = @fopen($new_config_path, 'w+') or die("Unable to open " . $new_config_path . " for writing");
```
objecttothis, line 51: Generally, you don't want to use " in php unless you are using string interpolation because it causes php to parse what you have between quotes. It's not a huge deal here, but in a loop with many iterations it adds up.

Let's use string interpolation where possible though, so this should become 
`die("Unable to open $new_config_path for writing");`

app/Helpers/security_helper.php
objecttothis: Also, you don't have to do this, but I would declare $new_config_path up above with $backup_path and $config_path.

app/Helpers/security_helper.php
objecttothis: Rather than killing the process with die() and a message, we need to write this to the OSPOS log using log_message(). https://codeigniter.com/user_guide/general/common_functions.html#log_message. Any messages that we send to the user need to be sent to the view to be gentrified. in the login view you can see this on lines 50-53 with has errors.... though I don't know if you will be able to use CI4 validation or not... look into whether you can add an error programmatically in the code using Services::validation().  If so, all you need to do is that before calling the view.

app/Helpers/security_helper.php
```diff
@@ -33,18 +41,20 @@ function check_encryption(): bool
...
+			@chmod($new_config_path, 0770);
 			fwrite($handle, $config_file) === FALSE;
 			fclose($handle);
-		}
-		else
-		{
-			return false;
+			echo '<script type="text/javascript">';
+			echo ' alert("File ' . $new_config_path . ' has been created. You must move this to ' . ROOTPATH . '")';
```
objecttothis, line 56: Same as above.  This needs to be sent to the view. Take a look at the login view and see how the message regarding needed migrations are done.

This is the reason that we had given write access to Apache to the folder. This method requires the user to manually move the file after from the writepath to baseurl. How about we try a hybrid:
1- backup the file to the writepath backup folder with a copy.
2- modify the .env file in baseurl.
3- Send the message to the view that the backup and modification occurred.

This way Apache doesn't need write permission to the baseurl, just to .env... Potentially this does still have security implications, but regardless, Apache does need to have read permission on that directory regardless.

@jekkos, do you have thoughts on the best way to approach this?

app/Helpers/security_helper.php
```diff
@@ -53,9 +63,9 @@ function check_encryption(): bool
 
 function abort_encryption_conversion()
 {
-	$config_path = ROOTPATH . '.env';
-	$backup_path = $config_path . '.bak';
+	$new_config_path = ROOTPATH . '/writable/backup/.env' ;
```
objecttothis, line 66: use WRITEPATH in these lines too.

objecttothis (changes requested):

app/Helpers/security_helper.php
```diff
@@ -54,8 +81,24 @@ function check_encryption(): bool
...
 
-	unlink($config_path);
-	rename($backup_path, $config_path);
+	//Update the master copy
+	$handle = @fopen($config_path, 'w+');
+	
+	//Cope with failure to open the handle
+	if(empty($handle))
+	{
+		log_message('error', "Unable to open $config_path for updating");
```
objecttothis, line 94: This log message doesn't seem to match what the function is doing.  I think the message should explain that it was unable to abort the encryption conversion or something related to not being able to revert the encryption conversion.

app/Helpers/security_helper.php
```diff
@@ -54,8 +81,24 @@ function check_encryption(): bool
...
+	}
+	else 
+	{	
+		@chmod($config_path, 0770);
+		fwrite($handle, $config_file) === FALSE;
```
objecttothis, line 99: I don't have time to download and test this line, but I think this will return a boolean but there isn't anything done with the boolean.  Should this be in an if statement?

objecttothis (changes requested):
Let's talk if any of my comments are unclear or you think I didn't understand the whole picture... that's possible too.

app/Helpers/security_helper.php
```diff
@@ -9,7 +9,7 @@ function check_encryption(): bool
 {
 	$old_key = config('Encryption')->key;
 
-	if(strlen($old_key) < 64)
+	if((empty($old_key)) or (strlen($old_key) < 64))
```
objecttothis, line 12: I had to look this one up, but when it doesn't matter about using `or` and `||`, lets use `||`.

Taken from the php documentation, the `||` has higher precedence than `or`, so when it matters `or` can do different thing than we are expecting.  In this case, it doesn't matter, unless I'm missing something. If it doesn't make a difference, let's keep the operator usage the same as the rest of the code and use `||`.

app/Helpers/security_helper.php
```diff
@@ -33,19 +40,24 @@ function check_encryption(): bool
...
-			// Write the new config.php file
-			$handle = @fopen($config_path, 'w+');
-			fwrite($handle, $config_file) === FALSE;
-			fclose($handle);
+			log_message('error', "Unable to open $config_path for updating");
+			return false;
 		}
-		else
+		
+		@chmod($config_path, 0770);
```
objecttothis, line 51: Looks like this was already done on line 30.  Did you mean to return the permissions back to 0640? in the end, our .env file should be writable by the owner, readable by Apache (group) and not readable by world. The only downside to doing that is that we wouldn't be able to programmatically modify .env in the future.

app/Helpers/security_helper.php
```diff
@@ -33,19 +40,24 @@ function check_encryption(): bool
...
+		$retval = fwrite($handle, $config_file);
```
objecttothis, line 52: To make the code more readable, don't abbreviate variable names unless it's a word in common usage and use variable names detailing what it contains.  So in this case `$config_file` is fine because config is a word in common usage, but let's refactor `$retval` because neither ret nor val are used commonly and return_value is too generic. 

If instead, you refactor this variable to $write_succeeded then line 55 below becomes `if($write_succeeded === false)`

Even better would be to write this as `$write_failed = !fwrite($handle, $config_file);`
Then write line 55 as `if($write_failed)`

Make sense?

app/Helpers/security_helper.php
```diff
@@ -54,8 +66,26 @@ function check_encryption(): bool
...
+	{
+		log_message('error', "Unable to open $config_path to undo encryption conversion");
+	}
+	else 
+	{	
+		@chmod($config_path, 0770);
+		$retval = fwrite($handle, $config_file);
+		fclose($handle);
+
+		if($retval === false)
```
objecttothis, line 85: same thing here as above regarding $retval

app/Helpers/security_helper.php
```diff
@@ -54,8 +66,26 @@ function check_encryption(): bool
...
+		{
+			log_message('error', "Unable to write to $config_path to undo encryption conversion.");
+		}
+		log_message('info', "File $config_path has been updated to undo encryption conversion");
```
objecttothis, line 89: This is going to fire even if the log message above it gets written but they seem to say conflicting things. The log message on 89 makes sense however if fwrite fails on line 82, I don't think it should get to that point. Rather than put an else statement here, though we could just put a return statement between lines 87 and 88 inside the if statement.

objecttothis (changes requested):

app/Helpers/security_helper.php
objecttothis: Looking back at my comments earlier, I had you remove that 2nd else, but did not intend on you removing the return statement inside the if as well... just add the return statement back into the if.

app/Helpers/security_helper.php
```diff
@@ -2,49 +2,76 @@
...
-		@chmod($config_path, 0440);
+		//Copy to backup
+		if(!copy($config_path, $backup_path))
+		{
+			log_message('error', "Unable to copy $config_path to $backup_path");
+			//we could return false at this point,
+			//but better for business to try the update first
+		}
 
+		//Attempt to make these files writable
```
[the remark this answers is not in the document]
jekkos: You can add some comments in the PR instead

app/Helpers/security_helper.php
```diff
@@ -10,51 +10,68 @@ function check_encryption(): bool
...
+			//we could return false at this point,
+			//but better for business to try the update first
+		}
 
-		copy($config_path, $backup_path);
+		//Attempt to make these files writable
+		//by the webserver user and group
+		//even if the directory is not writable
+		@chmod($config_path, 0770);
 		@chmod($backup_path, 0770);
```
jekkos, line 31: 660 is not enough? Having the exec bit set might be a security issue

app/Helpers/security_helper.php
jekkos: Indeed cosistency would be nice

app/Helpers/security_helper.php
```diff
@@ -2,49 +2,76 @@
...
+		$backup_path = WRITEPATH . '/backup/.env.bak';
 
-		copy($config_path, $backup_path);
-		$config_file = file_get_contents($config_path);
-		@chmod($config_path, 0440);
+		//Copy to backup
+		if(!copy($config_path, $backup_path))
+		{
+			log_message('error', "Unable to copy $config_path to $backup_path");
+			//we could return false at this point,
```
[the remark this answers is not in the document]
jekkos: Indeed comments are seldomly required, only if code cannot explain itself

app/Helpers/security_helper.php
objecttothis: > 660 is not enough? Having the exec bit set might be a security issue

Good point.  @owlbrudder since this is a file and not something that needs to be executed, we should be able to get by with just 0660 permissions.  I think default best practice for files anyway is 640.

app/Helpers/security_helper.php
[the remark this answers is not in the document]
objecttothis: > Indeed comments are seldomly required, only if code cannot explain itself

Yeah, he fixed it already in the code.

app/Helpers/security_helper.php
owlbrudder: Yep, fully agree - I winced when you pointed that out because I normally use ||.

app/Helpers/security_helper.php
owlbrudder: I am developing in my home directory, so I need read/write on everything and apache is my group user so apache needs read on everything plus write on writable/* and on .env, but you are correct - I can't think of any reason to give execute permissions anywhere around the OSPOS landscape. We have no scripts etc needing execute permissions, do we? I will keep .env and the writable tree at 660 if that's okay and the rest of the tree can be 640.

app/Helpers/security_helper.php
owlbrudder: Yes, I like that. retval is a relic from my C programming days when we had to fit an entire application into a 32kb ROM, so we pruned everything to the max and that lead to terse code with equally terse identifiers. No valid reason for it - thanks for the upgrade to my thinking.

app/Helpers/security_helper.php
owlbrudder: Yep - I like it.

app/Helpers/security_helper.php
owlbrudder: Bahaha - mea culpa. I must have removed the return statement I had sitting after the first message for reasons I will blame on sore eyes but for which I have no real excuse!

app/Helpers/security_helper.php
owlbrudder: Agreed - as above.

app/Helpers/security_helper.php
```diff
@@ -18,12 +18,19 @@ function check_encryption(): bool
...
-		$config_file = file_get_contents($config_path);
-		@chmod($config_path, 0440);
+		//Copy to backup
+		if(!copy($config_path, $backup_path))
+		{
+			log_message('error', "Unable to copy $config_path to $backup_path");
+		}
 
+		@chmod($config_path, 0660);
+		@chmod($backup_path, 0660);
```
jekkos, line 31: if we set the writable flag for the .env only for this request, then I guess fine

jekkos:
@owlbrudder It looks like all comments were addressed here? Do you know if we also remove the backup once the encryption key is set? It might be best to do so to avoid any potential leaks of application credentials

jekkos (approved):

owlbrudder:
I agree that the security problem needs to be addressed. As mentioned
elsewhere, the backup of .env only exists to support a single method in
security_helper.php which reverses the attempt to apply an encryption key.
That method is called only by a single migration script at this time. I
suggest adding a 'delete_backup()' method to security_helper and modifying
the migration script to call this method when it has completed wherever
else it is doing.

objecttothis (approved):

jekkos:
I have added this followup ticket: https://github.com/opensourcepos/opensourcepos/issues/3826

PR #3817, opened 2023-09-12
Correct some anomalies around bootstrap_tables_locale.php

owlbrudder:
The lang() strings being processed by this code now function correctly and a typo in Languages/en-US/Attributes.php has been corrected.

### DIFF
--- a/app/Language/en-US/Attributes.php
+++ b/app/Language/en-US/Attributes.php
@@ -19,7 +19,7 @@ return [
 	"definition_unit" => "Measurement Unit",
 	"definition_values" => "Attribute Values",
 	"new" => "New Attribute",
-	"no_attributes_to_display" => "No Items to display",
+	"no_attributes_to_display" => "No Attributes to display",
 	"receipt_visibility" => "Receipt",
 	"show_in_items" => "Show in items",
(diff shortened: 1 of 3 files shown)


objecttothis (approved):

jekkos:
What about other languages? It will only work for American english then?

jekkos:
Maybe better to run a regex replace on all language files to fix it

objecttothis:
> Maybe better to run a regex replace on all language files to fix it

Yes, so the two options are to replace, in the code, push to the ci4-upgrade branch and then have weblate scoop it up or to replace them in weblate.  The first option, I think, is faster but could result in translations in weblate that people have made recently getting overwritten. The second option is slower but potentially retains translations made recently. @jekkos you know weblate better than we do, so just decide and let us know what you want to do.

objecttothis:
@owlbrudder I merged this branch into ci4-upgrades.

owlbrudder:
Thank you for that.

Regarding your conversation with jekkos, I am quite happy to manually
update all the language files for interpolation, because I have done
something similar in my conversion attempt and I am not working to a
deadline, so if jekkos feels weblate is not our best option, just give me
the job and I'll gladly get it done.

Issue #3826, opened 2023-09-19
CI4: Remove backup .env file after migration


jekkos:
For security reasons we need to remove the .env file after the migration is done. Add a remove_backup method in the `security_helper.php` file to accomplish this.

owlbrudder:
I have created a pull request and requested you as reviewer, as you requested the change.

PR #3827, opened 2023-09-21
CI4: Bugfix - add function to remove .env.bak issue #3826

owlbrudder:
Added function remove_backup() to security_helper.php. Added a call to this from the two places that call check_encryption where the backup is created. Added more defensive code to Config.php to ensure the encrypter objectexists before it is called to avoid a crash.

### DIFF
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -299,6 +299,12 @@ class Config extends Secure_Controller
 
 		if(check_encryption())	//TODO: Hungarian notation
 		{
+			if(!isset($this->encrypter))
+			{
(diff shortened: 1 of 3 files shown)


jekkos (approved):

jekkos:
The travis build is failing because of the branch name, it might be better to use only single dashes and no dots.

owlbrudder:
> The travis build is failing because of the branch name, it might be better to use only single dashes and no dots.

I agree, but how do I edit the branch name? I created it by copy and paste from your issue name and did not notice the extra characters being inserted - I will not do that again.

PR #3829, opened 2023-09-22
Ci4 bugfix to remove backup env file after migration

owlbrudder:


jekkos (approved):

app/Database/Migrations/20220127000000_convert_to_ci4.php
```diff
@@ -71,6 +71,11 @@ private function convert_ci3_encrypted_data()
 				abort_encryption_conversion();
 				throw new RedirectException('login');
 			}
+			else 
+			{
+				//Remove any backup of .env created by check_encryption()
```
jekkos, line 76: The comment might not really be needed, I think the code is self explanatory. Same remark for the other line. Beside that it's good to go.

jekkos:
Thanks for the PR! I only have one minor remark. 
A handy tip, if you need to amend changes to a branch only you are working on, you can just make your change, then do a `git commit --amend` and then a `git push -f origin`

This way you keep the history clean as is often requested in opensource projects.

owlbrudder:
> Thanks for the PR! I only have one minor remark. A handy tip, if you need to amend changes to a branch only you are working on, you can just make your change, then do a `git commit --amend` and then a `git push -f origin`
> 
> This way you keep the history clean as is often requested in opensource projects.

Thanks very much for the tip. With GitHub I feel as though I am learning to ride a bicycle and keep falling off.   "8-D

app/Database/Migrations/20220127000000_convert_to_ci4.php
owlbrudder: > The comment might not really be needed, I think the code is self explanatory. Same remark for the other line. Beside that it's good to go.

I am trying to curb my natural instinct to document everything, but in this case I would like to leave the comment. When I was doing my own attempted conversion to CI4, the lack of comments plus lack of design documentation made the process painful. In this particular case, someone reading Config.php would not know that the call to check_encryption() tries to make a backup copy of .env and would be puzzled by the call to delete it.

I fear if we make the code too impenetrable, we will discourage others from trying to help us. Just my personal observation and I will happily go with the decision of the rest of the team. If you disagree with my wish to leave the comment, let me know and I will delete it without complaint.

jekkos: You're right from the method name it's not clear that it's also creating or removing the backup file. What I remember from clean code is that one should try to make a method do one thing and name it accordingly. So if we are doing more then just checking encryption, then we might need to rework the code and have a function separated that does the copy and removal and that is named as such.

Comments are usually a code smell, in this case indicating that the naming is poor or that a function has too many responsibilities.

jekkos: Let's leave the comment in this case then, but we should then try to apply the SRP from clean code next time we think we need comments.

jekkos: Also interesting point that you find the code hard to grasp, it might be an indication that it's not very self explanatory and needs improvement.

jekkos: but in this case the fix could be as follows: check result of check_encryption, if sucess then call remove_backup. Don't call the latter from check_encryption itself.

Issue #3730, opened 2023-04-04
CI4: Missing keys in configuration 


SteveIreland:
I'm going to start working on the configuration piece of this puzzle.  I need something to write the missing configuration values before I can continue work on the sales register.

odiea:
If it is not there then login_form need to be added to appconfig

SteveIreland:
I found a "solution/hack" for the missing config value errors that I'm using for configuration maintenance.   Preceding the $config['xxx'] with an @ symbol tells PHP to ignore the error and I believe a null is returned instead.  It also flags those that we need include in a migration step to add the missing values.

odiea:
@SteveIreland App/views/ Items /form line 227` what is that at the end? Otherwise all is working much better now.
```
<?php endif; ?>`
```

odiea:
@SteveIreland   ) ?>s  views/configs/general line 43. I did not see this earlier.

SteveIreland:
So today's goal was to get Attribute maintenance working again, and to successfully add an Attribute to Item maintenance.

The is working for me now, and the second was successful with a TEXT attribute.  The commits were pushed.

SteveIreland:
Finally got around to adding a migration script to add the config values that didn't exist that were creating issues with loading panels.  Removed the @ hack so that if the config values are missing the panel will resume crashing.  An odd one though was the barcode_content config value.  The issue with that was not that the config value was missing, because it was there.  However, it was being evaluated using the === comparison operator    As soon as I changed it to ==, it started working.  Doesn't make sense to me, I only know that it works.   It should be comparing two string data values.

SteveIreland:
I take back my comment about the barcode_content fix.

I rebuilt my test environment, upgraded the PHP version again, and then retested it.  Shocked to find it, again, fails for the same reason.

odiea:
I finally found it. Line 96 of views barcode_config is missing a bracket. Should look like this. 'checked' =>$config['barcode_content'] == 'id'] missing bracket at end of barcode_content.

SteveIreland:
@odiea Thank you!

jekkos:
@objecttothis do you know if this issue is stil present for other screens? Otherwise I would make a small migration that diffs the config table and adds key entries for missing entries, so that those things don't need to be fixed in the code.

objecttothis:
I'm not sure. I suppose a migration would settle the issue for both new and existing installs unless the issue pops back up in the event of removing a value then adding it back in.

[...]

Issue #3693, opened 2023-02-27
CI4: Stale data in appconfig update of cache


objecttothis:
Currently, when the an app_config key/value pair is updated or inserted into the database, the cash gets invalidated then updated.  There are two problems:
1- Calling the update_settings() function in the save() function causes a lot of calls to invalidating and updating the cache.  Perhaps there is a better place to call this.
2- calling update_settings() invalidates and updates the cache, so any future instances of config('OSPOS') will be current, but existing instances of the config are stale until they are regenerated.  This could cause data to be updated in the database and cache but not in the in-memory values.

For 1, We could call update_settings() in the batch_save() function after the foreach loop.  This will cut down on all but three calls to save() directly.  Then we could call update_settings() after those three calls.  The potential risk is that if we add more calls to save() directly and forget to call update_settings() we will end up with stale data in the cache.

As for 2, we need to create a new instance of config('OSPOS') after update_settings is run. The problem with this method is that it's decentralized.  Perhaps the best option here is to create a wrapper function and pass the config() instance by reference, then call both the save() function, update_settings() and refresh the config instance together.  We may need two functions or to just call batch_save or save from the wrapper based on whether the passed parameters are a single row or multiple.

jekkos:
The more pragmatic solution might be to invalidate the whole cache once something is updated through the model. The crazy thing is that in fact the cache is by default file backed, so instead of reading the values from the db, we read them from a file, which might not be all that faster in the end?

There is no easy in memory solution unless if we enable redis or memcached. But doing this only for this purpose feels a bit overkill.

objecttothis:
Yes, the default is file, but I believe I changed that to be session based since we use sessions in CI3

objecttothis:
Also, we are loading the table all at once currently (if the cache is not populated), storing it in memory cache (after pulling it) and in the global array which all gets repeated if a user makes a change to one of the elements of the array.  As I said in the other PR, I don't think we should get rid of it just for the sake of getting rid of it (since it's working... I think), but only if it serves a purpose, such as making the code empty value resilient without having to add in checks for empty values.

jekkos:
> Yes, the default is file, but I believe I changed that to be session based since we use sessions in CI3

I do believe the appconfig table was fully reloaded in memory at the start of every request in CI3.. I don''t think it was linked to the session at that point. (https://github.com/opensourcepos/opensourcepos/blob/master/application/hooks/load_config.php) . This hook is called pre controller constructor.

[...]

PR #3830, opened 2023-09-23
CI4: Bugfix to login.php

owlbrudder:
This change ensures login.php reacts correctly to an empty or missing company_logo entry in config[]. The file submitted is as it reads before line 29 gets injected by 'npm run build'.

### DIFF
--- a/app/Views/login.php
+++ b/app/Views/login.php
@@ -33,7 +33,7 @@
 		<main class="d-flex justify-content-around align-items-center flex-grow-1">
 			<div class="container-login container-fluid d-flex flex-column flex-md-row bg-body shadow rounded m-3 p-4 p-md-0">
 				<div class="box-logo d-flex flex-column justify-content-center align-items-center border-end px-4 pb-3 p-md-4">
-					<?php if ($config['company_logo']): ?>
+					<?php if (isset($config['company_logo']) && !empty($config['company_logo'])): ?>
 						<img class="logo w-100" src="<?= base_url('images/' . $config['company_logo']) ?>" alt="<?= lang('Common.logo') . '&nbsp;' . $config['company'] ?>">
 					<?php else: ?>
 						<svg class="logo text-primary" role="img" viewBox="0 0 308.57998 308.57997" xmlns="http://www.w3.org/2000/svg">


jekkos:
Thanks for the fix @owlbrudder !
I wonder if `isset` and `empty` are both required here? I checked in the PHP docs and it says the empty would also return false in case the variable is not defined. But maybe in this case because for an array key dereference?

owlbrudder:
> Thanks for the fix @owlbrudder ! I wonder if `isset` and `empty` are both required here? Perhaps the isset() alone is enough? Or is there a reason why both are needed? Do you know if this fixes the logo upload? that would be great.

Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.

I have successfully run with no company_logo record and an empty record and in both cases the built in default logo shows. I have also run in turn with company_logo set to logo.png and emblem.svg with successful rendering of each image.

objecttothis:
> > Thanks for the fix @owlbrudder ! I wonder if `isset` and `empty` are both required here? Perhaps the isset() alone is enough? Or is there a reason why both are needed? Do you know if this fixes the logo upload? that would be great.
> 
> Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.
> 
> I have successfully run with no company_logo record and an empty record and in both cases the built in default logo shows. I have also run in turn with company_logo set to logo.png and emblem.svg with successful rendering of each image.

Good work. In the future when naming a PR, please use a description that tells people what the changeset is for rather than which file. So this one would be something like `CI4: Bugfix account for empty company_logo in database` or something similar.

jekkos:
> Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.

Ok good to know, thanks for the confirmation.

jekkos:
related to #3730 Do we know why there was no empty company_logo present in the database? Maybe for following issues we could just add an entry with empty value, at least that is how it used to be done when features were added.

objecttothis:
> related to #3730 Do we know why there was no empty company_logo present in the database? Maybe for similar issues we could just add an config entry with empty value, at least that is how it used to be done when features were added.

I'd have to look at the current master but I think the issue is that currently we are storing and accessing config values in an associative array which is null sensitive, so we could also potentially think of a way to switch to a means of storing and accessing config data that still plays well with the sessions and is not cumbersome to be peppered throughout the various views and controllers... Doing so would be a good amount of work and really only provide the benefit of not having to pre-populate empty fields.

jekkos:
I also considered this, but where do you then define what is 'default' behavior for a feature? We then add one more case which might make ospos more inconsistent

* Once a new feature is introduced, then it's default config is set by a migration script. 
* If we choose to handle the case of 'no config found' in the code as well, then we actually need to duplicate that default setting in logic, and handle the 'not config' case for every check we do for that key, while it should in fact just have been set in the config.

It feels like we need to add extra redundancy which will not make the logic easier to understand.

owlbrudder:
To add a little colour, I have run a preliminary scan of the code base to discover where we are dereferencing config items compared with what is in the config table. These are the differences I have found at first glance:

In table but not dereferenced:
default_tax_rate
statistics

Dereferenced but not in table:
create_thumb
height
image_library
Login.form
mailchimp_api_key
mailchimp_list_id
maintain_ratio
migration_version
payment_message
source_image
width

owlbrudder:
> To add a little colour, I have run a preliminary scan of the code base to discover where we are dereferencing config items compared with what is in the config table. These are the differences I have found at first glance:
> 
> In table but not dereferenced: default_tax_rate statistics
> 
> Dereferenced but not in table: create_thumb height image_library Login.form mailchimp_api_key mailchimp_list_id maintain_ratio migration_version payment_message source_image width

In my own conversion project (now defunct), I did not use config[] directly, but instead used app/Models/Appconfig.php, which has defensive methods to deal with the ospos_app_config table. Where we use $config['some_key'], I used Appconfig->get_value('some_key') and never had a problem with the crashes. The difference, of course, is that the current version carries config[] in memory whereas I was going to the database each time which is less efficient.

owlbrudder:
> > > Thanks for the fix @owlbrudder ! I wonder if `isset` and `empty` are both required here? Perhaps the isset() alone is enough? Or is there a reason why both are needed? Do you know if this fixes the logo upload? that would be great.
> > 
> > 
> > Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.
> > I have successfully run with no company_logo record and an empty record and in both cases the built in default logo shows. I have also run in turn with company_logo set to logo.png and emblem.svg with successful rendering of each image.
> 
> Good work. In the future when naming a PR, please use a description that tells people what the changeset is for rather than which file. So this one would be something like `CI4: Bugfix account for empty company_logo in database` or something similar.

I see your point. To be honest, I was trying to work with jekkos's desire for brevity, so I chose a short name. I will try to find a middle ground in future.

jekkos:
> > > > Thanks for the fix @owlbrudder ! I wonder if `isset` and `empty` are both required here? Perhaps the isset() alone is enough? Or is there a reason why both are needed? Do you know if this fixes the logo upload? that would be great.
> > > 
> > > 
> > > Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.
> > > I have successfully run with no company_logo record and an empty record and in both cases the built in default logo shows. I have also run in turn with company_logo set to logo.png and emblem.svg with successful rendering of each image.
> > 
> > 
> > Good work. In the future when naming a PR, please use a description that tells people what the changeset is for rather than which file. So this one would be something like `CI4: Bugfix account for empty company_logo in database` or something similar.
> 
> I see your point. To be honest, I was trying to work with jekkos's desire for brevity, so I chose a short name. I will try to find a middle ground in future.

Ah I guess it was the PR name @objecttothis was referring to. That one can be written out fully. The branch name has some limitations as its used as docker tag.

owlbrudder:
> > > > > Thanks for the fix @owlbrudder ! I wonder if `isset` and `empty` are both required here? Perhaps the isset() alone is enough? Or is there a reason why both are needed? Do you know if this fixes the logo upload? that would be great.
> > > > 
> > > > 
> > > > Like you, I thought empty() would be adequate, but it was not. Under test, I found I had to use both conditions. As I have mentioned, I am running PHP 7.4 and it may behave differently under 8.x.
> > > > I have successfully run with no company_logo record and an empty record and in both cases the built in default logo shows. I have also run in turn with company_logo set to logo.png and emblem.svg with successful rendering of each image.
> > > 
> > > 
> > > Good work. In the future when naming a PR, please use a description that tells people what the changeset is for rather than which file. So this one would be something like `CI4: Bugfix account for empty company_logo in database` or something similar.
> > 
> > 
> > I see your point. To be honest, I was trying to work with jekkos's desire for brevity, so I chose a short name. I will try to find a middle ground in future.
> 
> Ah I guess it was the PR name @objecttothis was referring to. That one can be written out fully. The branch name has some limitations as its used as docker tag.

Grin. I learn as I go. Thanks for the clarification: Briefer branch names, descriptive PR names.

jekkos:
> In my own conversion project (now defunct), I did not use config[] directly, but instead used app/Models/Appconfig.php, which has defensive methods to deal with the ospos_app_config table. Where we use $config['some_key'], I used Appconfig->get_value('some_key') and never had a problem with the crashes. The difference, of course, is that the current version carries config[] in memory whereas I was going to the database each time which is less efficient.

We might also need to consider #3693 here

jekkos:
> create_thumb
> height
> image_library
> Login.form
> mailchimp_api_key
> mailchimp_list_id
> maintain_ratio
> migration_version
> payment_message
> source_image
> width

Thanks for the investigation already! One thing we can certainly do is insert the defaultfs for these values in a migration script (as @SteveIreland already did for other keys)

owlbrudder:
> > In my own conversion project (now defunct), I did not use config[] directly, but instead used app/Models/Appconfig.php, which has defensive methods to deal with the ospos_app_config table. Where we use $config['some_key'], I used Appconfig->get_value('some_key') and never had a problem with the crashes. The difference, of course, is that the current version carries config[] in memory whereas I was going to the database each time which is less efficient.
> 
> We might also need to consider #3693 here. I'll have to go through the code to see if this is still an issue. Also wonder if using the cache is then worth the added complexity for the performance gain we get in return, by deviating from the standard CI4 way of working.

Can you clarify for me what you mean by the standard CI4 way of working?

owlbrudder:
> > create_thumb
> > height
> > image_library
> > Login.form
> > mailchimp_api_key
> > mailchimp_list_id
> > maintain_ratio
> > migration_version
> > payment_message
> > source_image
> > width
> 
> Thanks for the investigation already! One thing we can certainly do is insert the defaultfs for these values in a migration script (as @SteveIreland already did for other keys)

Note that the entry 'Login.form' looks suspiciously like a typo to me and probably should not be added to the database. I will investigate.

owlbrudder:
> > > create_thumb
> > > height
> > > image_library
> > > Login.form
> > > mailchimp_api_key
> > > mailchimp_list_id
> > > maintain_ratio
> > > migration_version
> > > payment_message
> > > source_image
> > > width
> > 
> > 
> > Thanks for the investigation already! One thing we can certainly do is insert the defaultfs for these values in a migration script (as @SteveIreland already did for other keys)
> 
> Note that the entry 'Login.form' looks suspiciously like a typo to me and probably should not be added to the database. I will investigate.

Those uses are both in app/Views/login.php and I don't really understand what they are doing - can anyone explain to me?
					<?php if (empty($config['Login.form']) || 'floating_labels'==($config['Login.form'])): ?>
					<?php elseif ('input_groups'==($config['Login.form'])): ?>

objecttothis:
> > In my own conversion project (now defunct), I did not use config[] directly, but instead used app/Models/Appconfig.php, which has defensive methods to deal with the ospos_app_config table. Where we use $config['some_key'], I used Appconfig->get_value('some_key') and never had a problem with the crashes. The difference, of course, is that the current version carries config[] in memory whereas I was going to the database each time which is less efficient.
> 
> We might also need to consider #3693 here. I'll have to go through the code to see if this is still an issue. Also wonder if using the cache is then worth the added complexity for the performance gain we get in return, by deviating from the standard CI4 way of working.

Our current method had two goals. 
1. Increase performance by reducing database calls
2. Simplify notation of reference to app_config variables.

To clarify 2, I mean `$config['foo']` peppered throughout the code is much easier to read than `$this->Appconfig->getConfig('foo')` or whatever the default is.

That said I'm OK with ditching the current cache syncd model for a database call each time as long as we
A. Gain (or re-gain) empty value tolerance in the code
B. It isn't crazy cumbersome to reference in the code since these references appear so often.

A last note is that perhaps we can decouple the caching from the notation so that we don't have to lose one when changing the other. I'm not sure.

owlbrudder:
> Appconfig->get_value('some_key')

I see what you mean. So when I used Appconfig->get_value('some_key'), I was following CI4 standard practice? That makes sense because I was only concerned with upgrading from CI3, not streamlining anything. I would be interested to see what difference there is between the two approaches from a user perspective. I'm sure the caching model is faster, but if we are talking a poofteenth of a sqillisecond difference in screen refresh, is it worth the time and maintenance effort? I have no figures on this, but I was running my CI4 system on an old HP laptop with only an HDD, a dual core processor and 4GB RAM and responses were not noticeably sluggish. Easily fast enough for a small business user. It is much faster on my development machine where the database and web server run from an SSD, but to the average user a blindingly fast response is possibly not as important as it is to us as full-time computer users. This is just my personal observation and is not intended as a design strategy.

jekkos:
> A last note is that perhaps we can decouple the caching from the notation so that we don't have to lose one when changing the other. I'm not sure.

If we want to keep the array notation we can consider this: https://www.php.net/manual/en/class.arrayaccess.php
Also an option would be to specialize the current CI4 AppConfig and add a cache wrapper around it, together with the interface above, so it can be accessed like we used to do in CI3?

jekkos:
Looking at the OSPOS class that might be an option, instead of using the $settings array directly as $config, we can use the object itself and implement the arrayAccess methods. Then we can also handle the missing value case in the logic of the class itself.

jekkos:
Check my comment in https://github.com/opensourcepos/opensourcepos/issues/3693, it might not be worth to keep the cache and the complexity that it introduces after all, as the performance gain might be negligible. Glad to hear your thoughts on this @objecttothis @owlbrudder

PR #3834, opened 2023-09-26
refactor(Customers): Update data customer - id finger

asroboy:
add id finger to data customer to support checking finger during sales transaction


asroboy:
ok

PR #3836, opened 2023-10-04
Ci4 bugfix string interpolation

owlbrudder:
This branch has three commits: 1. the language files updated with CI4 numbered placeholders; 2. the Controllers and Views which send parameters to be interpolated; 3. the Linux bash scripts which updated the Language files and corrected a number of existing typos in a few of those Language files. The bash scripts are in a subdirectory under a new project directory named 'tools'. The scripts are supplied in case they may be useful as templates for future work of a similar nature, but they are not mission-critical and can easily be discarded if they are not thought useful.


objecttothis (approved):

objecttothis:
merged.

Issue #3468, opened 2022-04-12
CI4: Modify Weblate config to work with CI4 style language strings.


objecttothis:
### Issue / Bug  / Question / New Feature
Unlike CI3, CI4 ranges translation phrases into an associative array.  I'm not sure of the ingest/export mechanism of Weblate but we may need to tweak the configuration so that translation files get generated properly.

![image](https://user-images.githubusercontent.com/17935339/162918160-4a101ff8-a7c5-4b4f-9b65-72d623edd072.png)

objecttothis:
Also, currently only en-US language files in #3271 has been ported to the CI4 format.  We need to figure out if we have to convert the rest of them by hand or if weblate can do this for us.

jekkos:
@objecttothis aren't the language files already associative arrays at this point? We have a lang variable (array) with key and assign a value to it. The syntax you show here will probably yield the same result?

objecttothis:
> @objecttothis aren't the language files already associative arrays at this point? We have a lang variable (array) with key and assign a value to it. The syntax you show here will probably yield the same result?

Sorry, you are correct. There are two things to consider though:
1. In order to keep the CI3 styling we need to add `return $lang;`to the bottom of each file. Can we customize Weblate's generator to do this?
2. By not adopting CI4 styling we could face hiccups later. In the end it's not a big deal because both are valid associative arrays, but it won't match CI documentation from here on out.

#1 Is the only "must do" I think. If it's a lot of work to switch the array notation style in weblate then I'm ok with leaving it and adding the return statement. Let me know what you want to do.

jekkos:
@objecttothis I need to take some time to see how the file load mechanism works exactly. I suppose in the past CI was just loading one file and thus overwriting the keys with new values every time a language was selected.

Do we keep the MY_Lang extension that we had in the old install or do we need to modify it? If we keep it then we can probably bootstrap the existing files and then we're also certain it will work with weblate. I wouldn't really worry too much about the CI convention here, as long as the strings go where they should.

I'll be off coming week on a hiking trip, so I might not be able to look into this until after then.

objecttothis:
From https://codeigniter.com/user_guide/installation/upgrade_4xx.html
![image](https://user-images.githubusercontent.com/17935339/163348547-7ce2a6da-f454-4566-8ba6-fda9ba2752e8.png)
[...]

PR #3841, opened 2023-10-14
Sync language files with master (#3468)

jekkos:


jekkos:
@owlbrudder it seems that I also need to run your scripts on the latest master to get the output correct, so this PR is not finished yet. So it was indeed useful to have them in the repository.

jekkos:
Should be fixed now.  I do wonder why git sees the files as different while in fact no changes were made.

jekkos:
The reason there are diffs are due to the line endings. Previously this was dos, now it's LF UNIX. I'd suggest we stick to the latter. The .gitattributes is set to auto. This means that a file line endings should be converted to LF unless it's checked in using CRLF (windows). So if someone edits a file and saves it in CRLF, git should convert it back.

Maybe something went wrong when the files were moved over manually and the line endings were set to CRLF.

Issue #3854, opened 2023-11-03
CI4: Bootstrap-table header broken on scroll


objecttothis:
bootstrap-table handles creating the table header and body. The header is pinned during scroll, which is good, but on scroll it breaks causing the formerly non-visible header columns to appear and scrolling to no longer be fixed to the table.

![image](https://github.com/opensourcepos/opensourcepos/assets/17935339/cf54178b-c38e-483c-8b65-05618d3ac345)

You can see below that the header row is decoupled in horizontal scrolling from the table body.  This does not happen when the header is not trying to pin itself.
![image](https://github.com/opensourcepos/opensourcepos/assets/17935339/d7edd7d1-d659-40bc-9071-32419ded8ba6)

objecttothis:
I upgraded the version of bootstrap-table to 1.22.1, but the issue persists.  I was able to duplicate the issue on the bootstrap-table demo page of their website, so this is an issue specific to the StickyHeaders functionality in bootstrap-table. I created an issue in their repo https://github.com/wenzhixin/bootstrap-table/issues/7090

jekkos:
Thanks for looking into this @objecttothis ! I wonder if there would be a way to make the table more responsive.

objecttothis:
> Thanks for looking into this @objecttothis ! I wonder if there would be a way to make the table more responsive.

It's responsive for large and medium screens but not for small.  It's possible that moving everything to bootstrap 5 will make it easier to make the entire table responsive.  Likely the goal will be to specify how wide the table view is able to get down to and possibly to make the footer sticky on small displays. We may need to replace bootstrap-table if it can't do these things.

I just checked this example on their site (https://examples.bootstrap-table.com/) and it's much more responsive with sticky footers, so it looks like we may be able to make it work... they are using Bootstrap 5 though.

jekkos:
I saw some activity in the bootstrap tables project for this one, will check if we can upgrade to the latest commit.

jekkos:
I managed to bump the bootstrap table, had to upgrade the travis env to ubuntu jammy and now using a native node installation on that one instead of docker.
I will however try to fix the version to a specific commit hash for bstables  
![Screenshot from 2023-12-29 21-02-28](https://github.com/opensourcepos/opensourcepos/assets/576234/41566361-3f11-4a17-a761-ccebf76a60b6)

jekkos:
@objecttothis would the screencap here prove that the issue is fixed?

[...]

Issue #3700, opened 2023-03-03
CI4: Fix internationalization in reports view


objecttothis:
Currently internationalization is broken in the Views/Reports/listing.php view because show_report_is_allowed() function is expecting CI3 style $this->lang->line() but CI4 internationalization is called using lang().

Issue #3698, opened 2023-03-03
CI4: Get Sales MVC working


objecttothis:
Right now there are various bugs and problems preventing the sales views from loading properly.

SteveIreland:
I thought I would dive into this over the weekend and, time permitting, fix as many of the issues that I can find.  Hopefully I don't cause problems for you.  @objecttothis

objecttothis:
> I thought I would dive into this over the weekend and, time permitting, fix as many of the issues that I can find.  Hopefully I don't cause problems for you.  @objecttothis 

No problem. Feel free to take ownership of any tickets that I am assigned. I'm still several weeks out from being able to dive back in due to this remodel.

SteveIreland:
I made changes to the register that got me to the point where I could add, change, and remove items from the sale  I can also add and remove payments, and I can add and remove a customer.  A lot more remains to be debugged, but I'm out of time for this weekend.

One notable point.  The new CI4 method for getting config properties will fail if the property doesn't exist.  So, if for some reason the property doesn't exist, then I am creating the property and adding to the tables.sql so that the property will be added when a new database is created.  For existing databases they will need to be manually added.  Maybe this is an opportunity for a new CI3 to CI4 migration step?

I also noticed that the database was changed while I was working on this, so my test didn't include that database change.  I'll pick that up in the next iteration.

objecttothis:
@SteveIreland tables.sql and constraints.sql is frozen, so we can't add those there. In the CI3 version of OSPOS I believe those rows were being added to ospos_app_config on first use. I may be mistaken but we may just need to check if the row exists and lazy load if not.

objecttothis:
> Yep.  I forgot about that major annoyance.  Thanks for reminding me.

I believe thats the benefit of the save function. When working properly it does a check to see if the key exists and does an insert if no or an update if yes. The CI4 built in Model save function does this so eventually we should be able to get rid of our overridden save functions where we aren't doing something extra.

SteveIreland:
I added a commit that will allow a simple order (item selected, customer selected, and cash amount applied for a POS sale) to be completed.

Issue #3634, opened 2023-01-27
CI4: query() now returns false for failed queries


objecttothis:
https://codeigniter.com/user_guide/installation/upgrade_412.html#baseconnection-query-return-values

As of CI 4.1.2 query() returns false when a query fails.  We may need to go through the code and add a check on the return value.  Formerly I believe it returned an empty object.  This function is used extensively in the models, so we need to think this one through.

PR #3858, opened 2023-11-06
Codeigniter 4 upgrade

jekkos:
Upgrading CodeIgniter to 4.4.3 #842 

Here are a list of issues that need to be resolved:

- [x] #3452 NEEDS OWNER
- [x] #3455 @objecttothis 
- [x] #3610 NEEDS OWNER
- [ ] #3611 NEEDS OWNER
- [ ] #3632 NEEDS OWNER
- [ ] #3633 NEEDS OWNER
- [ ] #3634 @SteveIreland 
- [ ] #3693 NEEDS OWNER
- [x] #3698 @objecttothis 
- [ ] #3700 @edwinbsmith
- [ ] #3854 NEEDS OWNER
- [ ] #3862 NEEDS OWNER


jekkos:
@objecttothis the branch needs to be rebased again after the 3.3.9 release. I h

app/Views/items/form.php
```diff
@@ -573,61 +574,61 @@
...
+			//rules:
+			//{
+			//	name: 'required',
+			//	category: 'required',
+			//	item_number:
+			//	{
+			//		required: false,
+			//		remote:
+			//		{
+			//			url: "<?php //echo esc("$controller_name/checkItemNumber") ?>//",
```
jekkos: wasn't it the checkNumeric that stopped working?

objecttothis: That's what @odiea is saying. If that's the case, the stack trace is of zero help since the db connection dies at instantiating the session. I haven't had a chance to look at checkNumeric other than trying to recreate @odiea 's results by commenting stuff out. We have an event that I am supposed to bring a carload of product to tomorrow. If I get done in time, I'll take more of a look, but anyone else that wants to take a stab at it is welcome to do so.  Text attributes in the item form are failing to update the database. That may be related and it may not be. I think the attributes model needs some work.

jekkos (commented):
this might break again the horizontal scroll issue

objecttothis:
> this might break again the horizontal scroll issue

hmmm.  I tried it with what it was but it was still broken.  Is it working in your build?  I looked at the commit that it references and it's just bumping Cypress (https://github.com/wenzhixin/bootstrap-table/pull/7117) which is a testing plugin.  I'm surprised that this fixed the issue.

jekkos:
the commit hash itself was just latest master at that point in time.  It does not reference the specific fix, but it had it included. 

keeping a reference to master is problematic as every build you include a potential different version

objecttothis:
> the commit hash itself was just latest master at that point in time. It does not reference the specific fix, but it had it included.
> 
> keeping a reference to master is problematic as every build you include a potential different version

I see.  I can roll back the change back to that commit hash.  I'm not sure what I was doing wrong so that it didn't fix in my system.  I'll roll it back and try to do a clean, then install in npm.  Perhaps `npm install` didn't overwrite the version that was installed.

objecttothis:
I pushed a few commits which, I believe, are the last things causing any issues with the Customers section of OSPOS.  Please create an issue if anyone finds any problems in Customers.

andriux1990:
I'm desperate to start adapting OSPOS to CI4. But I encountered problems during installation, all of which are now resolved. Now, I've noticed that with existing databases from previous versions, migration only gets to the login screen. But if the database is new, then it displays correctly.

I also believe it's time to stop providing support for version 3.3.9; we should move on to 3.4.0 and focus on it.

Another issue is that in the case of <link href="<?= "resources/bootswatch5/$theme/bootstrap.min.css" ?>" rel="stylesheet" type="text/css">, when I added a product via SQL in PHPMyAdmin, it doesn't show up in the table.

From what it seems, it appears to be intact. We need to look into it to make it usable.

PR #3917, opened 2023-12-30
ci4 bug fix Sales controller_name and Receivings function name

owlbrudder:
In Views/sales/register.php two button labels did not show the correct language strings, because the variable '$controller_name' was passed to lang(), but this needed to be given literally as 'Sales'. In Views/receivings/receiving.php, the suggested items autocomplete function was called as stock_item_search and needed to be changed to stockItemSearch. There is also an almost identical function itemSearch, which may indicate that one of the two is redundant and should be pruned, but without documentation to guide me I am unwilling to do that at this time.

### DIFF
--- a/app/Views/receivings/receiving.php
+++ b/app/Views/receivings/receiving.php
@@ -497,7 +497,7 @@ $(document).ready(function()
 {
 	$("#item").autocomplete(
 	{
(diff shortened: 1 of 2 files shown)


jekkos (approved):

jekkos:
thanks for the fixes @owlbrudder

PR #3920, opened 2024-01-02
ci4-bugfix ucase first letter of controller name

owlbrudder:
Many labels were not picking up the language stings because the langauge file name was being passes to lang() without an uppercase first letter.


jekkos (approved):

jekkos (approved):

jekkos:
Thanks for the work @owlbrudder ! it seems the travis built was not being triggered due to too many users in the project, so I dsiabled a couple of inactive ones. Hope that it will deploy now to dev.

I'll try to go through the screens once the deployment is- done.

jekkos:
After having a further look it seems that the travis plan was only valid until the 1st of january. I emailed them to extend it for the coming year. Hopefully we get the builds back .

PR #3924, opened 2024-01-09
ci4-bugfix to items customers and attributes

owlbrudder:
Attributes: Noticed log_message() being called with uppercase letters in the level which causes errors in the system; Customers: improved the layout of the stats page in the information dialog issue 3892; Items: got csv import working issue 3896 and bulk edits working - barcode generation does not work yet.


owlbrudder:
Additional fixes now pushed as discussed.

jekkos (approved):

jekkos:
thanks again @owlbrudder !

Issue #3895, opened 2023-12-12
Rework Issue creation templates


objecttothis:
Currently we use a text-based template that we ask people to fill out when creating an issue. We need to instead create a github issue template for each type of issue (https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository)

Doing this will reduce how often we have to ask people to provide more details.  Some of the fields need to be required so that people can not submit bug reports without important information.

The types that I can think of are:
Bug Report
Feature Request
Question

@odiea do you feel like taking this one on?

odiea:
I will look into this after the first of the new year.

objecttothis:
@odiea do you still plan on looking at this? No worries if not. I'm just curious on the status.

odiea:
If there is any coding involved I won't be able to help out. Sorry

objecttothis:
I wouldn't call it coding per se.  It involves writing a YAML markdown "template" (https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). I'd be willing to bet that with a quick google search you'd probably even be able to find pre-made templates that people have already done and just tweak them to meet our needs.  Take a look and then you can decide.  No worries if it feels too much.

objecttothis:
This guy has a repo of them, so maybe you could just take a ready-made one and not really need to do much of anything?

https://github.com/stevemao/github-issue-templates

odiea:
Here is what I have learned so far. The feature request is an option under Issues. See Image. The first image is clicking on issues. The second is selecting the feature request. here is my repository if you would like to try it out.
![2024-02-19 13_17_47-New Issue · odiea_opensourcepos](https://github.com/opensourcepos/opensourcepos/assets/10962177/0127a453-186d-423a-801a-ee3c09c2b865)

![2024-02-19 13_16_55-New Issue · odiea_opensourcepos](https://github.com/opensourcepos/opensourcepos/assets/10962177/2d7bac4c-211e-419f-8bff-885e96172511)

jekkos:
Looks very clean and structured! Maybe one last item for bug reports. 
Also I wonder if we can label the issues automatically depending on the submission type.

odiea:
Ok. I just added the bug report template. Go to issues click issues again then the bug report template should show. The also I don't know or did not find any info on.

odiea:
[...]

PR #3936, opened 2024-02-28
Update issue templates (#3895)

jekkos:
Rebased @odiea work for the templates on master branch so we can use it in our repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug report.yml
+++ b/.github/ISSUE_TEMPLATE/bug report.yml
@@ -1,0 +1,121 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug", "triage"]
+projects: ["ospos/3", "ospos/4"]
+assignees:
+  - none
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Bug reports indicate that something is not working as intended. 
+        Please include as much detail as possible and submit a separate bug report for each problem.
+        Do not include personal identifying information such as email addresses or encryption keys.           
+  - type: dropdown
+    id: ospos-version
+    attributes:
+      label: OpensourcePOS Version
+      description: What version of our software are you running?
+      options:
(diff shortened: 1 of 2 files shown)


odiea (approved):
Looks good to me.

.github/ISSUE_TEMPLATE/bug report.yml
```diff
@@ -0,0 +1,121 @@
...
+        Bug reports indicate that something is not working as intended. 
+        Please include as much detail as possible and submit a separate bug report for each problem.
+        Do not include personal identifying information such as email addresses or encryption keys.           
+  - type: dropdown
+    id: ospos-version
+    attributes:
+      label: OpensourcePOS Version
+      description: What version of our software are you running?
+      options:
+        - opensourcepos 3.3.9
```
objecttothis, line 21: Does it make sense here to list versions back to 3.3.2?

odiea:
My thoughts were if there are any old or new bugs introduced in these versions they should be dealt with. It is up to the user to update their versions to keep up with the current OSPOS versions.
I also would like to see the new CI4 version should start with 4.0.0 to show we are now on CI4.

objecttothis:
> My thoughts were if there are any old or new bugs introduced in these versions they should be dealt with. It is up to the user to update their versions to keep up with the current OSPOS versions. I also would like to see the new CI4 version should start with 4.0.0 to show we are now on CI4.

@jekkos and I talked about this awhile back.  Even though there is a significant amount of work involved, @jekkos had a good point that the update doesn't by itself add any new features, only scaffolding for a lot of really cool features. For this reason, he argued to make it a minor version change 3.4.0 instead of 4.0.0.  Likely someone will want to redo the UI in the latest version of bootstrap and make it more responsive and that would likely warrant a major version change.  Also since we are planning introducing API's and 3rd party integrations, those would also be factors in a major version change.

objecttothis (approved):

PR #3937, opened 2024-02-28
Adding new Bug and Feature Templates

odiea:


jekkos:
Thanks for the PR @odiea but already active on the repo now (as it had to be merged in master). I'll rebase ci4-branch at some point, but these changes are not really needed there now so it can wait.

PR #3949, opened 2024-03-12
Item Kits updates for form and Barcodes to show

odiea:
There may be an issue with the Barcode portion of this update. The Barcode lib version is setting the app config as C128 where the earlier version is setting the app config as Code128. Any ideas would be help full.

### DIFF
--- a/app/Controllers/Item_kits.php
+++ b/app/Controllers/Item_kits.php
@@ -132,6 +132,7 @@ class Item_kits extends Secure_Controller
 			$info->print_option = PRINT_ALL;
 			$info->kit_item_id = 0;
 			$info->item_number = '';
+			$info->kit_discount = 0;
 		}
 
 		foreach(get_object_vars($info) as $property => $value)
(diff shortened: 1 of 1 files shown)


objecttothis:
Yeah, I overlooked that. We need to write a migration script that converts the old value to the new value in the table so that people's current value doesn't get lost

odiea:
I ran into this issue this morning. This happens in both Items and Item kits but not in Sales Receipts. When I go to print the barcodes the barcode itself does not show.
2024-03-17 05_18_15-Window

objecttothis:
> There may be an issue with the Barcode portion of this update. The Barcode lib version is setting the app config as C128 where the earlier version is setting the app config as Code128. Any ideas would be help full.

I pushed a migration script that will onboard everyone to the new codes.  If the script doesn't recognize the value, it will default to C128 since that is the most widely usable.

objecttothis (approved):

objecttothis:
> I ran into this issue this morning. This happens in both Items and Item kits but not in Sales Receipts. When I go to print the barcodes the barcode itself does not show.

@odiea can you try it again and tell me if this is still an issue with the latest commits?  I'm not able replicate.  If it's still an issue, send me a screenshot of your barcode settings in config.

odiea:
Let me know what settings you are using. I just did the latest and still have the same issue. No barcode. I have tried multiple settings with no luck.

2024-03-19 03_23_08-Window
2024-03-19 03_24_18-Window

objecttothis:
OK, I see what you're saying now.  I didn't understand before that it was in the print dialog.  I thought you were saying that the generate barcodes was not working, but it's when you print them.  I'll look into this now.

objecttothis:
The problem was being caused by the barcode generator generating DIVs to render the barcode on the screen so that we could set the width in pixels and each barcode would be the same width.  I created an issue on their github repo (https://github.com/picqer/php-barcode-generator/issues/194).  In the meantime I changed the generation to creating an SVG.  The problem is that their SVG (and other) generation techniques take in a width factor, not a desired width.  This means that the width of the barcode is based on the length of the data. I submitted a feature request to allow generation to be width based, not factor based. (https://github.com/picqer/php-barcode-generator/issues/195) so we will see what they say.

PR #3950, opened 2024-03-19
Update bug report.yml

objecttothis:
Changed order of bug report

This puts the most pertinent information on top
Makes the value text display as placeholder which will automatically disappear when text is entered.


odiea (approved):
Looks good to me. I will check it out after it is merged.

jekkos (approved):

PR #3952, opened 2024-03-19
Update Supplier.php

odiea:
Only way I could get supplier Category to show properly


odiea:
Update Receipt is for the New Barcode to show.

odiea:
After someone commits these I have a fix for Suppliers to sort properly

PR #3953, opened 2024-03-22
Update to  Suppliers and customers

odiea:
Removed post and changed to get for Supplier.

### DIFF
--- a/app/Controllers/Customers.php
+++ b/app/Controllers/Customers.php
@@ -87,11 +87,11 @@ class Customers extends Persons
 	 */
 	public function getSearch(): void
 	{
-		$search = $this->request->getGet('search');
-		$limit  = $this->request->getGet('limit', FILTER_SANITIZE_NUMBER_INT);
+		$search = $this->request->getGet('search', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+		$limit = $this->request->getGet('limit', FILTER_SANITIZE_NUMBER_INT);
 		$offset = $this->request->getGet('offset', FILTER_SANITIZE_NUMBER_INT);
-		$sort   = $this->request->getGet('sort');
-		$order  = $this->request->getGet('order');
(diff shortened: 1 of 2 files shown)


app/Controllers/Suppliers.php
```diff
@@ -44,11 +44,11 @@ public function getRow($row_id): void
 	 **/
 	public function getSearch(): void
 	{
-		$search = $this->request->getPost('search', FILTER_SANITIZE_FULL_SPECIAL_CHARS) ?? '';
-		$limit = $this->request->getPost('limit', FILTER_SANITIZE_NUMBER_INT);
-		$offset = $this->request->getPost('offset', FILTER_SANITIZE_NUMBER_INT);
-		$sort = $this->request->getPost('sort', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
-		$order = $this->request->getPost('order', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+		$search = $this->request->getGet('search', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
```
jekkos, line 47: Do we use get now instead of post? I remember vaguely that we had to use post because of some limitation in CI3 related to pagination. I think it's better to use get as then browser back button should also work. But then we should double check how the link code in the view is generated.

app/Controllers/Customers.php
```diff
@@ -87,11 +87,11 @@ public function getRow(int $row_id): void
 	 */
 	public function getSearch(): void
 	{
-		$search = $this->request->getGet('search');
-		$limit  = $this->request->getGet('limit', FILTER_SANITIZE_NUMBER_INT);
+		$search = $this->request->getGet('search', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+		$limit = $this->request->getGet('limit', FILTER_SANITIZE_NUMBER_INT);
```
jekkos, line 91: seems to be get as well here so I guess that would work then

odiea:
The Supplier table was not sorting or showing the correct count with getPost.

PR #3980, opened 2024-04-28
Fix for attribute addition for items

SpookedByRoaches:
The form_hidden expects a string and the value inputted is always -1. So it should be casted to string


jekkos:
Thanks for the contribution!

SpookedByRoaches:
No problem mate

PR #3986, opened 2024-05-04
Make register background colors use bootswatch

SpookedByRoaches:
The colors in the register are fine when the bootswatch theme is light but are unreadable in the darkly theme. I just deleted the css rules that change the colors and the register looks just fine with theme colors.

### DIFF
--- a/public/css/register.css
+++ b/public/css/register.css
@@ -16,11 +16,6 @@
 	margin: 0;
 }
 
-#mode_form
-{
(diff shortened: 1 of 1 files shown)


jekkos:
Sounds great,  could you check if this works for receivings as well?

SpookedByRoaches:
Yeah the receivings also use the same css rules.

PR #3988, opened 2024-05-06
Fixed csv import when updating items that have a barcode

SpookedByRoaches:
If the item is being updated, the check if there is an existing item with the same barcode does not make sense.

### DIFF
--- a/app/Controllers/Items.php
+++ b/app/Controllers/Items.php
@@ -1103,7 +1103,7 @@ class Items extends Secure_Controller
 						$item_data['is_serialized'] = empty($row['Item has Serial Number'])? '0' : '1';
 					}
 
-					if(!empty($row['Barcode']))
+					if(!empty($row['Barcode']) && !$is_update)
 					{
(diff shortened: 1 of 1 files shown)


jekkos:
@SpookedByRoaches I believe there is an option that allows for duplicate barcodes to exist. We probably can not do the update in case this option is enabled, so this should be an extra condition before the update is performed.

jekkos (changes requested):
Add additional config check for allow duplicate barcodes

SpookedByRoaches:
Why not do the update in that case? The update applies to the items with the existing Id not the Barcode. You would be right in adding the check for the `$is_failed_row` inside that same block. I can add a check there to allow the duplicate barcode if you want. Let me know if I missed something with what you said.

P.S. Apologies for the garbage commits I'm not too experienced with git. I removed those.

jekkos:
Ok indeed the `is_update` is set if `item_id > 0`.

Issue #3708, opened 2023-03-08
Enforce code conventions


objecttothis:
https://www.devbridge.com/articles/coding-best-practices/

Currently there aren't coding convention checks in place to enforce coding conventions for this project other than code reviews.
- .EditorConfig will do some of that voluntarily, but it can be overridden.
- If we end up switching to Github actions, we could kick off scripts to run.
- The URL above suggests using LINT to do the legwork here.

SteveIreland:
Have you been spying on my code again?!?!?

objecttothis:
Lol. No, it was a comment @jekkos made in a PR that @edwinbsmith started.

SteveIreland:
I miss @daN4cat.  I keep expecting him to pop up at any moment to crack the whip on me to keep me in line.

ekardian:
@daN4cat we r expecting

objecttothis:
Moving to 3.5.0 so we can get CI4 out the door with 3.4.0

jekkos:
The config is in the ci4 branch, no sure if sth else is needed to make this work. I tried to go through some of the GitHub actions config but could not find  any option to activate it. Lets check again once we merge to master.

jekkos:
The linter seems to run now in github actions: https://github.com/opensourcepos/opensourcepos/actions/runs/9024670984/job/24798972386

jekkos:
A next step now is to run the linter locally first to fix all indentation issues in our code.

jekkos:
The linter can also be setup in PhpStorm directly: https://www.jetbrains.com/help/phpstorm/using-php-cs-fixer.html#installing-configuring-php-cs-fixer

jekkos:
Once this is integrated then it should be documented. The CI check will then prevent code to not deviate from these rules.

jekkos:
Curious to see what issues it will find: https://github.com/opensourcepos/opensourcepos/actions/runs/12714942531/job/35446147431

objecttothis:
> https://github.com/dbfx/github-phplint
> @objecttothis I believe we need to add something to the CI pipeline that detects obvious code issues. In case of the requisitions, a method call was in place that did not exist. These type of issues can be easily catched with tools like this.

Yeah, PHPStorm complains about those things. I'd say as long as it's not too overzealous, a linter like that could be helpful. There are a handful of todo's in the code for ones I found.

jekkos:
[...]

PR #3989, opened 2024-05-09
[CI4] Login screen update

BudsieBuds:
This pull request updates the login screen to function more like the CI3 version and ensures compatibility with Bootstrap 5.3.

- Updated deprecated Bootstrap 5 classes for a safe update from `5.1` to `5.3`
- Modified error display: triggered errors show in separate alert boxes instead of in one box as an unordered list (`<ul>`)
- Made newly added error message translatable
- Small updates and fixes

I would appreciate feedback on the new and adjusted PHP code, especially around best practices.
For example this code block from `app/Views/login.php`. Is this the best way to do it within CI4?
```php
<?php if ($has_errors): ?>
	<?php foreach ($validation->getErrors() as $error): ?>
		<div class="alert alert-danger mt-3">
			<?= $error ?>
		</div>
	<?php endforeach; ?>
<?php endif; ?>
```

The PHPDoc comment in `app/Views/login.php` should probably be updated to reflect the changes made. I wasn't able to do this with VS Code, but guess it should be straightforward with PhpStorm.


jekkos:
The code style looks good, as far as I can see. It reminds me that I tried to add a linter with CI4 rules in it in this branch and that it does not seem to be executed for now.

jekkos:
I have now setup a linter in the ci4-branch #3708 . First we need to fix issues in the branch and then after that, PRs can be checked automatically for coding style errors.

jekkos:
@BudsieBuds I have pulled your commit and pushed it to a branch on the repo itself, this way it will be deployed on dev server. I'll have a look there and then merge it to ci4-branch.

jekkos:
Merged in ci4-branch.

Issue #3994, opened 2024-05-13
[CI4][Bug]: Language fallback


BudsieBuds:
### Bug Description?

Language fallback doesn't work as it did in the CI3 version. Empty strings in the translation files are returned as an empty string instead of falling back to English. If the string is removed from the language file, the fallback kicks in as it should.

### Steps to Reproduce?

Change the language on a `ci4-branch` build to an incomplete language. For example, Dutch translates `Welcome to OSPOS!` to `Welkom bij !` on the login screen because there's no translation for `OSPOS`. German gives an error on the login screen because of the nested translation and missing translations for both `Login.welcome` and `Common.software_short`.

### Expected Behavior?

This is the expected behavior for CI4, but a fallback to English should be coded. The same happened [here](f43a620bf400514639628c745c4a2ef5149db15f) for the CI3 version by @jekkos. Changing the language from `en-US` to `en` is also important for it to work correctly, as done here #3995.

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox, Chrome

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
@BudsieBuds to avoid a tricky migration I would just drop the language config in the db for 3.4 upgrade. Its only a small annoyance and a user can quickly restore language. so then we do not need to write a lengthy migration that mirrors the renaming you did in the file tree.

PR #3995, opened 2024-05-13
[CI4] Language fallback improvements

BudsieBuds:
Following best practices for language fallback in CI4, I've made the following changes.

- Changed `en-US` to `en` for fallback purposes. [See CI4 docs](https://codeigniter4.github.io/userguide/outgoing/localization.html#language-fallback).
- All other language variants will fall back to English. This is the only language that is consistently and officially updated.
- Changed some names and codes, following the BCP47 recommended by CI4.
- Norwegian and Urdu languages are working again.
- Sort languages by alphabet in config for easier navigation.

**NOTE:** Language fallback to English is currently broken in OSPOS because of empty strings, see #3994.


jekkos:
Great work sorting this out @BudsieBuds . The issue in german with empty strings is not resolved then by renaming the en-US to en? 
I was going through the CI4 code and indeed, if the language line does not yield any content, it falls back to english.

jekkos:
how do we decide whether we use a base language folder like `de` or `es`?  In my local version I see this as a login screen when selecting german

image

BudsieBuds:
> Great work sorting this out @BudsieBuds . The issue in german with empty strings is not resolved then by renaming the en-US to en? I was going through the CI4 code and indeed, if the language line does not yield any content, it falls back to english.

[That's what I thought](https://github.com/codeigniter4/CodeIgniter4/issues/8875), but unfortunately no. It is resolved by removing the keys though. So it's also an option to remove the keys during build or something, but I think coding in that empty values should be ignored would be better and less prone to error.

> how do we decide whether we use a base language folder like `de` or `es`? In my local version I see this as a login screen when selecting german. Also CI4 docs seem to suggest that one should use a base language folder, and then only translate strings that deviate from this in the more specific variant. The idea sounds good as it means less maintenance, if a translation is the same you do not need to duplicate and maintain it.
> 
> But I'm not sure whether weblate would be able to handle this setup.

I thought about this before this PR and think it's better to have everything fall back to English. Mainly because that's the only one being consistently updated, like I mentioned. I think the fact that this is a POS means that seeing the wrong translation (f.e. from `de` when you have selected `de-CH`), could really confuse the end user. I don't know the differences between `de-DE` and `de-CH` for example, but they could be big. Especially with more technical terms for taxes etc. A better example would be `nl-NL` and `nl-BE`, mainly because those are our native languages, and I noticed pretty big differences. How it was set up, everything that isn't translated in `nl-BE` would fall back to `nl` and then to `en` (if renamed and empty string ignored).

The only way I see this being an option is when Weblate can handle showing what it falls back to, in what order, and with multiple languages. But from what I could find in the docs this isn't possible with PHP strings.

jekkos:
@BudsieBuds one last question, why do we convert `de` to `de-DE`, while for `hu-HU` we convert it to `hu` ?

BudsieBuds:
Ask away. I'm not the best at describing my PRs.

Among other resources, I've referred to [this file](https://raw.githubusercontent.com/opensourcepos/opensourcepos/ee89f1971bb368e784d6939b10522897262378b1/application/language/Language%20code%20definition) in the repo. I removed all ISO 3166-1 region codes from languages that are only mentioned once. However, I didn’t do the opposite. For instance, with French, the variant is unknown (`fr-BE`, `fr-CA`, `fr-CH`, `fr-FR`, `fr-LU`, `fr-MC`). It's likely `fr-FR`, but that only really matters when another region is added.

Therefore, I changed `de` to `de-DE` because there are two different German translations available; `de-DE` and `de-CH`. If `de` isn't renamed, `de-CH` will fall back to `de` before `en`, which could lead to confusion. It goes the other way around for `hu-HU`, there's only one Hungarian translation available and no mention of another region in the aforementioned list. Following this logic, I kept `pt-BR` as is for example.

There's only one assumption: changing `es` to `es-ES` assuming it's Spanish (Spain).

Following the [BCP 47](https://www.rfc-editor.org/rfc/rfc5646.txt) as recommended by CI4, I've checked the official resources for the [shortest ISO 639 (language) code and official English name](https://www.loc.gov/standards/iso639-2/php/English_list.php) and the [ISO 3166-1 (region) code](https://www.iso.org/obp/ui/#search/code/). Adjusted them accordingly.

Yes, I went down the rabbit hole. :eyes:

jekkos:
thanks for the clarification and research @BudsieBuds . Looks all fine to me.

BudsieBuds:
@jekkos Should a database migration script be written for updating the languages on upgrade?

jekkos:
selected language might indeed be in the Config table. So indeed this config value needs to be migrated.

jekkos:
@BudsieBuds the fallback is in place by default in CI4 code. It's however only enabled in case a key does not exist in a specific langauge, not if the value is an empty string.

BudsieBuds:
@jekkos can confirm, as mentioned in issue #3994.

PR #3996, opened 2024-05-13
Apply linter fixes (#3708)

jekkos:
Ran this command locally and then pushed the output
```vendor/bin/php-cs-fixer fix --verbose --ansi --config=.php-cs-fixer.no-header.php --using-cache=no --diff  ```

### CODE REVIEW

File: app/Config/App.php [hunk 1]
```diff
@@ -3,56 +3,45 @@
...
 {
-	/**
-	 * This is the code version of the Open Source Point of Sale you're running.
-	 *
-	 * @var string
-	 */
-	public string $application_version = '3.4.0-dev';
+    /**
+     * This is the code version of the Open Source Point of Sale you're running.
+     */
```
objecttothis, line 11: @jekkos I like the idea of the linter, but we may need to make some adjustments.  One area where I can foresee some problems is that when we need to update CI to a new version it involves doing a diff on a lot of these config files because there are changes in them.  If our linter has significantly changed them, it makes the diff super messy and takes longer to differentiate which things are just linter changes and which things are changes by CI4.

app/Config/App.php
```diff
@@ -3,56 +3,45 @@
 namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
```
objecttothis, line 5: This should not have been removed.

objecttothis (changes requested):

objecttothis:
Also, the default rules for this linter change a few standards that we currently have.  Are we changing those?  For example currently the project uses tabs, but the CS Fixer default ruleset uses 4 spaces.

objecttothis:
I looked into it a little and the way to customize the rules would be to create a .php_cs.dist file in the root of the project and configure it the way we want it. We would need to use the same ruleset in the build checks as we use in development though or it will fail checks.

objecttothis:
Also, when I installed cs-fixer it defaulted to PSR-2, but PSR-12 is the current best practice for coding standards rulesets.

objecttothis:
> I looked into it a little and the way to customize the rules would be to create a .php_cs.dist file in the root of the project and configure it the way we want it. We would need to use the same ruleset in the build checks as we use in development though or it will fail checks.

I guess the correct name is `.php-cs-fixer.dist.php` not `.php_cs.dist` which is old.

jplrts:
> Also, the default rules for this linter change a few standards that we currently have. Are we changing those? For example currently the project uses tabs, but the CS Fixer default ruleset uses 4 spaces.

That was not my intention, I made the PR to show the current setup and to get your feedback first ofc.

jplrts:
> Also, when I installed cs-fixer it defaulted to PSR-2, but PSR-12 is the current best practice for coding standards rulesets.

Ok not fully aware on current PHP standards but sounds good

PR #4014, opened 2024-06-30
Correct a constraint that wasn't using the full unique key based on c…

SteveIreland:
Correct a constraint that wasn't using the full unique key based on composite keys.

### DIFF
--- a/app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
+++ b/app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class Migration_fix_keys_for_db_upgrade extends Migration
+{
+	/**
+	 * Perform a migration step.
+	 */
+	public function up(): void
+	{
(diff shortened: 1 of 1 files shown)


jekkos:
perhaps also add the changes for the person_id key in here so we can get rid of those in contstraints.sqL?

SteveIreland:
_"perhaps also add the changes for the person_id key in here so we can get rid of those in contstraints.sqL?"_ - @jekkos 

I'm not clear on what you're asking for here. My understanding is that you would want the foreign key constraints for those tables that are referencing the persons table.

The original tables.sql required changes in order to do a build from scratch in a MySQL 8.x environment.  But after some research I found that a migrated 3.3.9 database already included nearly all of those changes - apparently because they were the right thing to do.  :-)  

A scan of a backup SQL script of my 3.3.9 to 3.4 migrated MySQL 8.4 database shows that everything made it through the migration phase in reasonably good condition.  There are a handful of things that I would clean up - some, but not all involving person_id.  But I'd want to take my time and review those more carefully in order to avoid unintended consequences.

jekkos (approved):

jekkos:
Missed your point here, looks fine then as you explained.

Issue #3965, opened 2024-04-02
Security Strategy


objecttothis:
### Bug Description?

During the conversion to CI4 we need to make sure that our code security strategy is rock solid:

Escaping outputs:
- Many of the built-in CodeIgniter 4 library functions automatically escape outputs. Examples of this are form_input(), form_password(), form_upload(), form_textarea(), form_dropdown(), form_multiselect(), form_checkbox() and form_radio(). anchor() also is automatically escaped, so no need to do anything with those.
- Output outside of this should always be wrapped in esc() and using the correct context as a 2nd parameter.
- If don't correctly, we shouldn't experience problems of data being garbled.

Validation of Inputs:
- CodeIgniter has validation mechanisms that allow us to return errors to the user when invalid data is sent.
- We need to be validating textboxes.

Query Security:
- CodeIgniter4's Query Builder should be used when possible. This is because QueryBuilder has automatic escaping of queries and creates parameterized queries.
- Any time queries are not written or QueryBuilder->rawSql() is used, $this->db->escape() should be wrapped around variables being inserted into the database and $this->db->protect_identifiers() should be wrapped around variables representing table or column names.

Now on to what I am less certain about:
- CSRF is needed, but we need to do everything in such a way that it doesn't hinder APIs since those will be needed for third party integrations.
- General best practice, I'm told, is to store html and tags in the database as they are entered (i.e., not encoded).  Filtering inputs practices though seems to frequently suggestion wrapping $_POST (getPost() in CI4) in htmlspecialchars() since getPost($foo, FILTER_SANITIZE_STRING) is deprecated.  Those functions encode special characters though before they get sent to the model for storing in the database.  So the big question is what does Filtering Inputs need to look like?
- One option is simply stripping unwanted data.  For example with phone numbers we may wrap that in a preg_replace() that filters anything that is not '(', ')', ' ', '-', '+','[0-9]' and maybe 'x' or '/' for indicating extension numbers.  We only have certain kinds of data, so it seems reasonable to have a handful of functions in the security_helper to properly filter these kinds of values without mangling the text. It also makes sense to use the 2nd parameter of getGet() and getPost() where it doesn't mangle text since those are already written.

What's wrong? What's missing?
@jekkos , et. al. thoughts?

### Steps to Reproduce?

See above

### Expected Behavior?

See above

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

[...]

PR #4015, opened 2024-07-01
Feature Add HTMLPurifier filter calls to Controller Methods

objecttothis:
Per a conversation with @jekkos we are going to apply HTMLPurifier to filter data before it gets sent to the view but not filter what is received in the view.  Candidates to be filtered using HTMLPurifier are textboxes and fields where innocuous HTML tags need to be allowed.

Using HTMLPurifier should provide some protection against XSS code injection, but it will not stop SQL injection attacks and is not meant as a replacement for CI4 best practices for XSS prevention.

This PR is a part of the overall Security Strategy for the 3.4.0 release in #3965

- [x] All calls to Services::htmlPurifier()->purify() removed from data received from view.
- [x] Bootstrap and bootswatch bump in package-lock.json
- [x] Add Services::htmlPurifier()->purify() calls to the appropriate fields in the controller just before view() is called.
- [x] Remove calls to FILTER_SANITIZE_FULL_SPECIAL_CHARS on these fields as it will prevent safe HTML from being used.


objecttothis:
@jekkos I responded to you in Element.  Can you take a look when you get time?

objecttothis:
@jekkos with bootstrap-table 1.23.1 installed and `escape` set to false in public\js\manage_tables.js, \views\Reports\tabular.php and \views\Reports\tabular_details.php, I experience the following:

Entering `<script>alert('test');</script>` into the item name field causes the script to execute after clicking submit. 
image

Then bootstrap tables displays the script as entered but does not execute it.
image

When I set escape to true, the behavior was the same, so it's unclear to me what escape is doing.  I would expect it to htmlencode the data but it does not seem to do that.  That said, I wasn't able to definitively rule out bootstrap table because the script executes when I press submit on the modal form.  That said the script does not run when just loading the page with the table, so I don't think it's bootstrap table.  That said, we need to prevent the script from running when the code appears in the form data and is submitted.

jekkos:
In a way it makes sense, bootstrap tables is a javascript library, I believe it cannot actually rewrite any output coming from PHP. The proper solution is to html encode the table data here. Then it will render correctly but not trigger an XSS. Also we can probably just disable the escape functionality. Not sure what it then actually does.

jekkos:
I have rebase the branch in #4041. I was looking for the POC you made that allowed you to render the whole view using HtmlPurifier. I'll check if I can proceed a bit with encoding HTML in the views before it's rendered to the page.

objecttothis:
Sorry. After I realized that it was causing everything to be stripped out I didn't commit it. I thought the answer was to create a post_system event because hooks is the way it used to be done. The problem is that you can't get access to the DOM data that way. The way I did it was using an after filter.

In app/Config/filters.php define a new after filter (e.g. DomPurifier) then in app/Filters/DomPurifier.php that would look something like 
```
<?php
namespace App\Filters;

use CodeIgniter\HTTP\RequestInterface;
use CodeIgniter\HTTP\ResponseInterface;
use CodeIgniter\Filters\FilterInterface;

class DomPurifier implements FilterInterface
{
    public function before(RequestInterface $request, $arguments = null)
    {
  
    }

    public function after(RequestInterface $request, ResponseInterface $response, $arguments = null)
    {
      // Access the DOM (HTML content) here
        $dom = $request->getBody(); 

        $dom = Service::HtmlPurifier->purify($dom);
        $request->setBody($dom);
    }
}
```
Of course if you want to change the config of html purifier then you have to do that too.

objecttothis:
@jekkos I just rebased this on the master after merging your other PR and these are the changes that came out of it.  Can you take a look?  In our last conversation about XSS I think you said you didn't want us to delay this release too long figuring out XSS issues. I think when we talked about the XSS issues, you were going to look into whether we could just run HTMLPurifier against the entire DOM.  In my POC, it was stripping out too much.

This PR represents an implementation that uses HTMLPurifier to remove unsafe HTML and Javascript. HTMLPurifier does not encode anything.  If we can figure out an encoding scheme which does not mangle legitimate characters, then I support that, but every effort I've seen to encode was causing legitimate text (e.g., `O'Malley`) to be encoded and it doesn't look good (`O&#39;Malley`).

jekkos:
I would like to see where the encoding then broke the rendering of these characters. The issue might be that the data rendered in an input field needs to be left untocuhed. There is no chance of xss in of either.

jekkos:
Will pick this up as a next thing

jekkos:
@objecttothis I have rebased this branch again here: https://github.com/opensourcepos/opensourcepos/pull/4066
not many changes remain.

jekkos:
I have double checked the issue and the XSS trigger you found does not come from bootstrap tables. In fact it is triggered through the notification toast, which shows this text: `"You have successfully updated item <script>debugger;</script>"`

jekkos:
Let's move to the other PR. I have added a commit there that fixes the vulnerability that you have found

Issue #3997, opened 2024-05-17
[Bug]: Sales discount in Register not working properly


odiea:
### Bug Description?

Go to Register add an Item then try giving the Item a Discount. The Discount immediately reverts back 

### Steps to Reproduce?

Go to Register add an Item then try giving the Item a Discount. The Discount immediately reverts back 

### Expected Behavior?

Discount should remain the selected with the toggler

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Chrome

### Server Operating System and version

Wamp

### Database Management System and version

Wamp

### Web Server and version

Wamp

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

PR #4024, opened 2024-07-17
Resolving issue #3997 for discount type toggle not staying put.

SteveIreland:


odiea:
@SteveIreland Works as described. Now if you can find the solution to the edit error after removing the discount that would be great. Thanks

odiea (approved):
Works as described.

app/Libraries/Sale_lib.php
```diff
@@ -1309,7 +1309,7 @@ public function edit_item(string $line, string $description, string $serialnumbe
 			$line['quantity'] = $quantity;
 			$line['discount'] = $discount;
 
-			if(!empty($discount_type))
+			if($discount_type != null)
```
jekkos, line 1312: Did behavior of empty change in php8?

SteveIreland: Something changed, because a $discount_type value of 0 was being evaluated as empty.  The manual does say that if the value is zero then it will be treated as false, and false evaluates to empty.  And since the result was that my effort to set the value to zero was being ignored, a change was needed.  :-)   Just checking for null in this scenario seems to be working.  But I need to look at the other error that @odiea found.   I didn't take my testing that far.

odiea:
@SteveIreland currently discount can easily be changed just by placing a 0 in the current discount. I guess I was looking for something to change if I just deleted the current discount where it would automatically update the discount to 0.

SteveIreland:
@odiea there we a a couple of issues that needed to be resolved in order to allow us to simply delete the discount amount.  Hopefully I didn't break something as part of the fix, but I don't have high expectations for myself here.  :-)

I'm still learning the changes introduced with CI4.

@jekkos since a couple of the functions found in local_helper.php are used heavily in the views, I went ahead and changed the ?float to ?string for the functions that were either causing problems in the views, or I thought might possibly cause a problem.  More testing will prove the changes either correct or in need of further enhancement.

app/Libraries/Sale_lib.php
jekkos: just thought that in that case, there might be more bugs like this, because this is used in quite a couple of cases throughout the codebase

odiea:
Works great. No issues in the log have shown.

jekkos (approved):

jekkos (approved):

Issue #4025, opened 2024-07-20
[Bug]: Item add to Sale failed


odiea:
### Bug Description?

Try adding an Item kit or Return on the Register.
Item add to Sale failed

### Steps to Reproduce?

Same as above

### Expected Behavior?

Item add to Sale failed should not fail

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.4

### What browsers are you seeing the problem on?

Firefox, Chrome

### Server Operating System and version

Wamp Server

### Database Management System and version

Mysql

### Web Server and version

Wamp Server

### System Information Report (optional)

This also shows on the Dev Demo

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
Do you get this error `Amount Tendered must be a number. `?

odiea:
CRITICAL - 2024-07-18 04:59:39 --> TypeError: App\Models\Sale::is_valid_receipt(): Argument #1 ($receipt_sale_id) must be of type string, null given, called in C:\newwamp\www\mymaster2\app\Controllers\Sales.php on line 153
[Method: GET, Route: sales/search]
in APPPATH\Models\Sale.php on line 427.
 1 APPPATH\Controllers\Sales.php(153): App\Models\Sale->is_valid_receipt(null)
 2 SYSTEMPATH\CodeIgniter.php(933): App\Controllers\Sales->getSearch()
 3 SYSTEMPATH\CodeIgniter.php(509): CodeIgniter\CodeIgniter->runController(Object(App\Controllers\Sales))
 4 SYSTEMPATH\CodeIgniter.php(355): CodeIgniter\CodeIgniter->handleRequest(null, Object(Config\Cache), false)
 5 SYSTEMPATH\Boot.php(312): CodeIgniter\CodeIgniter->run()
 6 SYSTEMPATH\Boot.php(67): CodeIgniter\Boot::runCodeIgniter(Object(CodeIgniter\CodeIgniter))
 7 FCPATH\index.php(56): CodeIgniter\Boot::bootWeb(Object(Config\Paths))

or with a kit

CRITICAL - 2024-07-19 03:47:37 --> TypeError: App\Models\Item_kit::is_valid_item_kit(): Argument #1 ($item_kit_id) must be of type int, string given, called in C:\newwamp\www\master\app\Controllers\Sales.php on line 560
[Method: POST, Route: sales/add]
[...]

PR #4027, opened 2024-07-21
An attempt to resolve issue #4025.  

SteveIreland:
Since a kit item code is prefixed by "KIT" it's not going to work to always assume that the item id is numeric.  So "int"  needs to be replaced with "string".

I haven't yet fully tested the impact of this change, but I believe these changes were the only ones I made that allowed me to add a kit item to a sale.   But, I'm in a time crunch at the moment, so I'm submitting this as a draft and will do more testing this evening and finalize it then.

### DIFF
--- a/app/Controllers/Sales.php
+++ b/app/Controllers/Sales.php
@@ -547,7 +547,7 @@ class Sales extends Secure_Controller
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


SteveIreland:
I rebuilt my test environment from scratch, deployed the three files that were changed, corrected my test data and the test results gave me a warm a fuzzy feeling, so flagged the pull request as ready for review.

jekkos (approved):

Issue #2992, opened 2020-10-28
Migration errors during clean install


jekkos:
When setting up a docker-compose from scratch, I come across quite a few SQL warnings. This seems to indicate that the changes in the first two migrations were also applied in the database.sql. In fact there are two upgrade paths here

* Either you upgrade from a clean install, that means you run database.sql first, then all the migrations one-by-one
* Either you upgrade from an older version and you need to run all migrations one-by-one

At first I thought that skipping the first two migrations in a clean install might fix these errors, but this does not seem to be a good idea. My idea was to fix it by updating the version number in the migrations table after it is created and set it to a date after the first two first, so they would be skipped.

But this solution actually breaks the second upgrade scenario here, as older versions won't have a migrations table either and then they will also skip these first two. 

The correct solution is to fix the database.sql and remove the changes that are applied in the first two migration files. I thought I'd write it out here, it's not that critical but I'd like to get this right as it seems users with bigger datasets sometimes have a hard time upgrading.

```
ospos_dev | [Wed Oct 28 20:51:48.605705 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: 1062, referer: http://localhost/                                                                                                                                            [30/964]
ospos_dev | [Wed Oct 28 20:51:48.605773 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: Duplicate entry 'receipt_font_size' for key 'PRIMARY', referer: http://localhost/                                                                                                   
ospos_dev | [Wed Oct 28 20:51:48.607183 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: 1061, referer: http://localhost/                                                                                                                                                    
ospos_dev | [Wed Oct 28 20:51:48.607249 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: Duplicate key name 'person_id', referer: http://localhost/                                                                                                                          
ospos_dev | [Wed Oct 28 20:51:48.608701 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: 1061, referer: http://localhost/                                                                                                                                                    
ospos_dev | [Wed Oct 28 20:51:48.608763 2020] [php7:notice] [pid 16] [client 172.18.0.1:41300] error: Duplicate key name 'sale_id', referer: http://localhost/                                                                                                                            
[...]

PR #4032, opened 2024-07-23
Add check to migration to prevent errors

objecttothis:
When the constraint doesn't exist an error is generated when DROP is performed against it.  I've added a check which will only perform the DROP if the constraint exists.


jekkos (approved):
Might be nitpicking, its the same code for both down and up here? Shouldnt we add the constraint back in the down() method? If not we can leave the rollback as a no operation as constraint was probably already dropped.

jekkos:
#2992

PR #4034, opened 2024-07-29
Bugfix : Locale handling of decimals in attribute saves and Migration script fixes

objecttothis:
This resolves two issues:

1. Decimals in attributes were being saved with the locale decimal separator which caused problems with that was a comma.
2. There are several issues in migration scripts when creating new installs.

### DIFF
--- a/app/Database/Migrations/20210422000000_database_optimizations.php
+++ b/app/Database/Migrations/20210422000000_database_optimizations.php
@@ -5,6 +5,7 @@ namespace App\Database\Migrations;
 use CodeIgniter\Database\Migration;
 use CodeIgniter\Database\ResultInterface;
 use App\Models\Attribute;
+use Config\Database;
 use Config\OSPOS;
 use DateTime;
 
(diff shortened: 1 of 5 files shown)


app/Database/Migrations/20210422000000_database_optimizations.php
```diff
@@ -96,15 +115,22 @@ private function migrate_duplicate_attribute_values($attribute_type): void
...
 		$builder = $this->db->table('attribute_values');
-		$builder->select("$column, attribute_id");
+		$builder->select("$column");
 		$builder->groupBy($column);
 		$builder->having("COUNT($column) > 1");
 		$duplicated_values = $builder->get();
 
 		foreach($duplicated_values->getResultArray() as $duplicated_value)
 		{
+			$subquery_builder = $this->db->table('attribute_values');
```
objecttothis, line 125: @SteveIreland sorry I didn't realize that both of us were working on this at the same time.  In your PR, please account for false on failed queries, but they fail for a reason.  For example this one was failing because of the only_on_full_group_by clause, so in order to keep it from failing I had to break apart the query and modify the business logic.  The fix for this particular migration works.  I'd say in our checks for false, we should fail the migration if it does return false. Otherwise we can end up thinking the migration completed but in reality have a schema that is inconsistent.  I'd say that maybe I should change this PR to merge into your branch and then you take over work on this and I will go back to my fun little task of getting filtering/escaping/validation in place...  What do you think?

SteveIreland: Sorry @objecttothis, I knew you were working on it - but my little bit of effort was primarily to get your feedback. Still I didn't want to push anything that wasn't at least usable for an "install a new app" scenario.  So what I pushed at least accomplishes that.

But know that I know what direction you'd like to take it, I'd like to explore other approaches.  If you want to merge the efforts and let me take over the work on this, that is fine with me.

The migration step used to upgrade destination based sales tax to add support for India GST was one that I was familiar with.  It had the potential failure and for that I made the decision to just log the "encountered database issue that should never happen" and allow the migration to complete with a note that the migration failure will need to be cleaned up manually.  As an SQL'r I can normally complete the migration manually.  OR, as I normally do, simply restore the database from a backup, fix the database issues, and restart the migration.

By allowing the migration to finish to the end and reporting any issues that are encountered along the way, then at the end of the migration we have all of the errors that need to be resolved recorded to the log.,

The alternative is to immediately stop the migration at the point that the first severe error is encountered.

I'm not particularly invested in either approach but would prefer the former.  What are your thoughts on this?

objecttothis: @SteveIreland yes, let's merge this branch into your changes and have you take over.  Now, regarding your points above:
- Yes, let's log all errors so that it's clear what the problem is.
- The reasons I don't think we should continue the migration after encountering and logging an error are two-fold:
- 1. Errors may cascade if one migration depends on a previous one successfully completing but we encountered an error yet continued the migration after.
- 2. Many users will not have the experience required to manually fix the data in the database using SQL, so relying on them doing that when an error is incurred is probably not what we want to do.

So for example, if this change, which removes duplicate attribute_values needs to complete so that we can add a unique constraint on the attribute_value (which prevents future duplicates from being created), then that migration which adds the constraint will also fail if we don't stop the migration after encountering an error and logging it.  Does that make sense?

objecttothis:
This PR is no longer needed as it has been merged into another PR.

PR #4035, opened 2024-07-29
This is intended to start resolving #3634. 

SteveIreland:
CI4 query() now returns false for failed queries.  Some of these fixes involve data migration scripts **and, in order to be able to test it, I also needed to resolve the issue with the data optimization migration script that was triggering the migration failure**.

So this is the first draft.  I used it to build a new instance of OSPOS and it completed without failure.  So, it can be committed.  There were a couple of scenarios where it was suggested that someone translate the query to builder and I'd like to try that out.  I can either add to this or submit it as its own build request.

@objecttothis @jekkos Please review to confirm that you are okay with my approach to resolving the false for failed queries scenario.   Thanks.


objecttothis:
@SteveIreland see my comment of the other PR.  I'll take a look at your changes in this one in the morning.

app/Controllers/Sales.php
```diff
@@ -635,7 +635,7 @@ public function postEditItem(string $line): void
 		$rules = [
 			'price' => 'trim|required|numeric',
 			'quantity' => 'trim|required|numeric',
-			'discount' => 'trim|required|numeric',
+			'discount' => 'trim|permit_empty|numeric',
```
objecttothis, line 638: This looks to be a correct change, but don't add any other validation changes to this PR as I will be working on validations eventually and validation is outside the scope of this PR.

app/Database/Migrations/20170502221506_sales_tax_data.php
```diff
@@ -153,7 +153,7 @@ private function get_count_of_unmigrated(): int
 			. $this->db->prefixTable('sales_taxes')
```
objecttothis, line 153: I know this code was not one you modified, but generally we need to move toward using QueryBuilder for as many of these queries as possible to benefit from the security provided by it. I'll leave it up to you as to whether you convert it here or not.

app/Database/Migrations/20170502221506_sales_tax_data.php
```diff
@@ -153,7 +153,7 @@ private function get_count_of_unmigrated(): int
...
 			. ' as ST ON SIT.sale_id = ST.sale_id WHERE ST.sale_id is null GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
-		return $result[0]['COUNT(*)'];
+		return $result[0]['COUNT(*)'] ?: 0;
```
objecttothis, line 156: This will definitely fix the problem of false being returned, but the problem is that false gets returned if there is an error executing the query.  Even if we deem it safe to assume 0 on an error, we should at minimum log the error in the CI error log.

app/Database/Migrations/20200202000000_taxamount.php
```diff
@@ -133,7 +133,7 @@ private function get_count_of_unmigrated(): int
 			. ' as ST ON SIT.sale_id = ST.sale_id GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
 
-		return $result[0]['COUNT(*)'];
+		return $result[0]['COUNT(*)'] ?: 0;
```
objecttothis, line 136: Same here only I'm not so sure it's safe to continue the migration when this errors. it could leave a user with the impression that the old data was inserted when it wasn't.

app/Database/Migrations/20190129212600_indiagst.php
```diff
@@ -154,6 +154,12 @@ private function migrate_tax_rates(): void
 
 		$jurisdiction_id = $this->db->query('SELECT jurisdiction_id FROM ' . $this->db->prefixTable('tax_jurisdictions') . " WHERE jurisdiction_name = 'Jurisdiction1'")->getRow()->jurisdiction_id;
 
+		if(!$jurisdiction_id)
+		{
+			error_log('Tax rate migration failed due to severe errors.  Tax jurisdictions will need to be cleaned up manually');
```
objecttothis, line 159: I'm curious when this fails, why it fails.  You've heard my opinion on whether it's acceptable to complete the migration and ask the user to clean up the data manually. Perhaps @jekkos can share his thoughts on the matter?

app/Database/Migrations/20210422000000_database_optimizations.php
```diff
@@ -54,7 +54,7 @@ public function up(): void
 
 			$builder = $this->db->table('attribute_links');
 
-			foreach($attribute_links->getResultArray() as $attribute_link)
+			foreach($attribute_links->getResultArray() ?: [] as $attribute_link)
```
objecttothis, line 57: personally I would refactor this short ternary operator assignment outside of the foreach loop for code readability.  Something like:
```
$attribute_links = $attribute_links->getResultArray() ?: [];

foreach($attribute_links as $attribute_link)
{
```

app/Helpers/locale_helper.php
```diff
@@ -362,28 +362,28 @@ function to_datetime(int $datetime = DEFAULT_DATETIME): string
 }
 
 /**
- * @param float|null $number
+ * @param string|null $number
  * @return string
  */
-function to_currency(?float $number): string
+function to_currency(?string $number): string
```
objecttothis, line 368: Make sure you test these changes thoroughly.  I think @jekkos would probably advocate for unit tests and I think we are committed to moving toward those. @jekkos maybe once unit tests are working again, you could lead a zoom workshop demonstrating what we are doing and how people need to develop unit tests for their code?

objecttothis (approved):

jekkos:
> @SteveIreland It's not letting me push the commit to resolve the conflicts because I don't have access to the RuleDomain repository. You'll need to resolve the conflicts then squash and merge. I will assign the #4034 PR to you.

Just a hint, you can always pull in a remote branch locally and edit it, then push it back to the organization and work from there.

app/Helpers/locale_helper.php
jekkos: This kind of logic is indeed a perfect candidate for a small unit test. I'm not sure if they are still running in the build. I'll have to check if I can run them first in my IDE, then after that we can run them in travis as well.

app/Database/Migrations/20190129212600_indiagst.php
jekkos: If the error is something they can fix manually afterwards, then fine I think, it's better to proceed than to halt the migration as a lot of users might not know what to do in that case.

If this does not exist then it might mean we have some data inconsistencies? In this case we could just recreate this Jurisdiction1 if it does not exist.

app/Helpers/locale_helper.php
SteveIreland: Oh my.  Something else to learn.  :-)

jekkos:
I haven opened #4040 that contains changes from both @SteveIreland and @objecttothis .

PR #4049, opened 2024-08-26
Ci4 migration fixes

jekkos:
made yet another PR and squashed some commits, to make the actual diffs now more clear. Some changes that were done in 3.4 optimizations were moved to the more recent migration script instead.

### DIFF
--- a/app/Database/Migrations/20170502221506_sales_tax_data.php
+++ b/app/Database/Migrations/20170502221506_sales_tax_data.php
@@ -153,7 +153,14 @@ class Migration_Sales_Tax_Data extends Migration
 			. $this->db->prefixTable('sales_taxes')
 			. ' as ST ON SIT.sale_id = ST.sale_id WHERE ST.sale_id is null GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
-		return $result[0]['COUNT(*)'];
+
+		if(!$result)
+		{
+			error_log('Database error in 20170502221506_sales_tax_data.php related to sales_taxes or sales_items_taxes.');
+			return 0;
+		}
+
+		return $result[0]['COUNT(*)'] ?: 0;
 	}
 
 	/**
--- a/app/Database/Migrations/20190129212600_indiagst.php
+++ b/app/Database/Migrations/20190129212600_indiagst.php
@@ -154,6 +154,12 @@ class Migration_IndiaGST extends Migration
 
 		$jurisdiction_id = $this->db->query('SELECT jurisdiction_id FROM ' . $this->db->prefixTable('tax_jurisdictions') . " WHERE jurisdiction_name = 'Jurisdiction1'")->getRow()->jurisdiction_id;
 
+		if(!$jurisdiction_id)
+		{
+			error_log('Tax rate migration failed due to severe errors.  Tax jurisdictions will need to be cleaned up manually');
+			$jurisdiction_id = 1;
+		}
+
 		// Insert old tax_code rates data into the new tax rates table
 		$this->db->query('INSERT INTO ' . $this->db->prefixTable('tax_rates')
 			. ' (rate_tax_category_id, rate_jurisdiction_id, rate_tax_code_id, tax_rate, tax_rounding_code) '
--- a/app/Database/Migrations/20190129212600_indiagst.php
+++ b/app/Database/Migrations/20190129212600_indiagst.php
@@ -154,12 +154,6 @@ class Migration_IndiaGST extends Migration
 
 		$jurisdiction_id = $this->db->query('SELECT jurisdiction_id FROM ' . $this->db->prefixTable('tax_jurisdictions') . " WHERE jurisdiction_name = 'Jurisdiction1'")->getRow()->jurisdiction_id;
 
-		if(!$jurisdiction_id)
-		{
-			error_log('Tax rate migration failed due to severe errors.  Tax jurisdictions will need to be cleaned up manually');
-			$jurisdiction_id = 1;
-		}
-
 		// Insert old tax_code rates data into the new tax rates table
 		$this->db->query('INSERT INTO ' . $this->db->prefixTable('tax_rates')
 			. ' (rate_tax_category_id, rate_jurisdiction_id, rate_tax_code_id, tax_rate, tax_rounding_code) '
--- a/app/Database/Migrations/20200202000000_taxamount.php
+++ b/app/Database/Migrations/20200202000000_taxamount.php
@@ -133,7 +133,14 @@ class Migration_TaxAmount extends Migration
 			. ' as ST ON SIT.sale_id = ST.sale_id GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
 
-		return $result[0]['COUNT(*)'];
+		if(!$result)
+		{
+			error_log('Database error in 20200202000000_taxamount.php related to sales_taxes or sales_items_taxes.');
+			return 0;
+		}
+
+
+		return $result[0]['COUNT(*)'] ?: 0;
 	}
 
 	/**
--- a/app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
+++ b/app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
@@ -22,8 +22,6 @@ DROP INDEX `person_id` ON `ospos_customers`;
 ALTER TABLE `ospos_customers` MODIFY `taxable` tinyint(1) DEFAULT 1 NOT NULL;
 ALTER TABLE `ospos_customers` MODIFY `deleted` tinyint(1) DEFAULT 0 NOT NULL;
 ALTER TABLE `ospos_customers` MODIFY `discount_type` tinyint(1) DEFAULT 0 NOT NULL;
-ALTER TABLE `ospos_customers` ADD PRIMARY KEY(`person_id`);
-ALTER TABLE `ospos_customers` ADD INDEX(`company_name`);
 
 ALTER TABLE `ospos_customers` ADD CONSTRAINT `ospos_customers_ibfk_1` FOREIGN KEY (`person_id`) REFERENCES `ospos_people`(`person_id`);
 ALTER TABLE `ospos_customers_points` ADD CONSTRAINT `ospos_customers_points_ibfk_1` FOREIGN KEY (`person_id`) REFERENCES `ospos_customers` (`person_id`);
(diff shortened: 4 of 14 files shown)


jekkos:
One last thing to 
* check is the 3.3.9 -> 3.4 upgrade 
* clean db install seems to work
* substitute the error log with throwing an exception

jekkos:
now interesting observation I was testing my last change in the migration which stopped the process, resent the login submission and then found this in the database dump
 
```
  KEY `definition_name` (`definition_name`),
  KEY `definition_type` (`definition_type`),
  KEY `definition_name_2` (`definition_name`),
  KEY `definition_type_2` (`definition_type`),
  KEY `definition_name_3` (`definition_name`),
  KEY `definition_type_3` (`definition_type`),
```

This might explain the attribute key count error I saw some time ago

jekkos:
@objecttothis I have done one last comparison between 3.3.9 upgrade and fresh install, and only found that in the upgraded version we have an extra index in items table on item_id and deleted. This index is not present in the fresh install. I have added the delete statement back to align both schemas as you requested.

I'm not sure why the index was removed but I suppose that adding the deleted flag does not add a lot of performance benefits as typically a user won't have too many deleted items anyway.

Unless someone objects I would merge this and continue.

objecttothis:
No objections. I can do a code review Monday or Tuesday unless someone else gets to it first.

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
```diff
@@ -75,6 +70,7 @@ ALTER TABLE `ospos_expense_categories` ADD INDEX(`category_description`);
 ALTER TABLE `ospos_giftcards` MODIFY `deleted` tinyint(1) DEFAULT 0 NOT NULL;
 
 #ospos_items table
+ALTER TABLE `ospos_items` DROP FOREIGN KEY `ospos_items_ibfk_1`;
```
objecttothis, line 73: This is a problem because it's also being dropped on line 119, so the second time you drop it, it creates an error.

objecttothis: It looks like in the shuffle of merging @SteveIreland 's changes and mine, possibly some duplicates were created.  I thought I had sorted these per table, but perhaps I'm wrong.

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis, line 1: line 11 fails for me on an existing database. The problem is that I can't find anywhere in the code where this constraint is being created besides here.  My theory is that I somehow added it to the existing database and then now I'm trying to run this migration script against it. I do have a pre-3.2.2 database which does not have the constraint.  If my guess here is wrong though, other people will have this line fail on them.  If that's the case then we will need to pull this out of here and move it to the PHP file where we can easily check to see if the constraint exists and only try to add it if it does.  T-SQL has an easy way to wrap SQL in an if statement, but MySQL is jenky and won't let you do it so you have to do it programmatically using QueryBuilder in PHP.

objecttothis (changes requested):

app/Database/Migrations/20190129212600_indiagst.php
```diff
@@ -154,6 +154,12 @@ private function migrate_tax_rates(): void
 
 		$jurisdiction_id = $this->db->query('SELECT jurisdiction_id FROM ' . $this->db->prefixTable('tax_jurisdictions') . " WHERE jurisdiction_name = 'Jurisdiction1'")->getRow()->jurisdiction_id;
 
+		if(!$jurisdiction_id)
+		{
+			error_log('Tax rate migration failed due to severe errors.  Tax jurisdictions will need to be cleaned up manually');
```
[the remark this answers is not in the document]
jekkos: @SteveIreland is this really a severe error? In case the default jurisdiction is not in the db, we might assumed the user is not using any of the specific tax features and we skip the migration? Or we just recreate the Jurisdiction and continue?

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: @objecttothis I tried both clean install and install of 3.3.9 then upgrade to 3.4, without issues. But I must admit that at some point I also saw this attribute links unique error on dev. I have not been able to reproduce since then

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: indeed the chnage was removed then added again , and now removed

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis: @jekkos During my testing of a clean database when I got errors, sometimes clearing the database and starting again made them go away.  This makes me think that perhaps the transactions are not working correctly or are only applying in a certain way.  For example, I noticed that if a script failed, then I made an adjustment and attempted to re-run it, I would get a different error that only happened if part of the previous script had actually been applied.  I wonder if this is how and when we fail the migration will cause this.

app/Database/Migrations/20190129212600_indiagst.php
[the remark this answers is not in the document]
jekkos: @SteveIreland @objecttothis I have searched the codebase for jurisdiction id but do not find any defaults there, so this case might be a perfectly valid one? Can we remove the check and error message?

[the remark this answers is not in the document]
jekkos: Also I wonder if the query returns no results, then getRow()->jurisdiction_id will probably already craash the script so the extra check can only fail if the schema does not contain the column at this point.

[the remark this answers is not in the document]
jekkos: @objecttothis This failure case is limited to what is executed in this method here directly. I believe the case just handles the fact that the insert query that ran before it failed, which I think should not happen. So maybe it's all fine here.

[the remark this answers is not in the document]
jekkos: @objecttothis we can just remove this clause as it will be unlikely that we end up in this situation, unless the table structure is not created as it should. And in that case I agree we should probably stop the migration and fix the bug.

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
jekkos: I guess that indeed the migrations do not run transactionally, so indeed if something goes wrong, you might end up with a half baked database schema which you can then try to migrate by running the migration again.
So I guess most of the time things that were already applied will error out and can be ignored..

jekkos: https://github.com/prisma/prisma/discussions/3774

jekkos: We can just wrap this statement in an exists and then it should be fine no?

jekkos: @objecttothis I see this contraint was already defined in the `3.3.0_attributes.sql`, check line 33

jekkos: The funny thing is that my db dumps contains those two constraints, but they are actually the same thing
```
  UNIQUE KEY `attribute_links_uq1` (`attribute_id`,`definition_id`,`item_id`,`sale_id`,`receiving_id`),
  UNIQUE KEY `attribute_links_uq2` (`item_id`,`sale_id`,`receiving_id`,`definition_id`,`attribute_id`),
```
So perhaps some setting in MySql causes it to error out on this. In my case it did not bother and did just recreate the constraint with the same fields but in different order

jekkos: Did a new push that removes this line

objecttothis (approved):

app/Database/Migrations/20170502221506_sales_tax_data.php
```diff
@@ -153,7 +153,14 @@ private function get_count_of_unmigrated(): int
 			. $this->db->prefixTable('sales_taxes')
 			. ' as ST ON SIT.sale_id = ST.sale_id WHERE ST.sale_id is null GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
-		return $result[0]['COUNT(*)'];
+
+		if(!$result)
```
jekkos, line 157: @SteveIreland was there particular reason to add the false check? 
Do we need to take the only_full_group thing into account here? Perhaps that is the reason the query sometimes fails?

app/Database/Migrations/20200202000000_taxamount.php
```diff
@@ -133,7 +133,14 @@ private function get_count_of_unmigrated(): int
 			. ' as ST ON SIT.sale_id = ST.sale_id GROUP BY SIT.sale_id, ST.sale_id'
 			. ' ORDER BY SIT.sale_id) as US')->getResultArray();
 
-		return $result[0]['COUNT(*)'];
+		if(!$result)
+		{
+			error_log('Database error in 20200202000000_taxamount.php related to sales_taxes or sales_items_taxes.');
```
jekkos, line 138: also here, do we have any expected failures?

app/Database/Migrations/sqlscripts/3.4.0_database_optimizations.sql
objecttothis: > The funny thing is that my db dumps contains those two constraints, but they are actually the same thing
> 
> ```
>   UNIQUE KEY `attribute_links_uq1` (`attribute_id`,`definition_id`,`item_id`,`sale_id`,`receiving_id`),
>   UNIQUE KEY `attribute_links_uq2` (`item_id`,`sale_id`,`receiving_id`,`definition_id`,`attribute_id`),
> ```
> 
> So perhaps some setting in MySql causes it to error out on this. In my case it did not bother and did just recreate the constraint with the same fields but in different order

They are not the same constraint.  They contain the same columns but the order of the constraint matters.  These were created because they are the best indexes for slow running queries in one or more of the models.

objecttothis: > I guess that indeed the migrations do not run transactionally, so if something goes wrong then you will end up with a half baked database schema which you can then try to migrate by running the migration again. So I guess most of the time things that were already applied will error out and can be ignored..

I was kind of surprised to learn this since I thought they did run in a transaction.  IMO we should wrap all the migrations in a transStart() / transComplete() so that they do run transactionally.  As is, it results in inconsistent schemas in the event of a migration failure.  I can't think of anything that would cause this to be a breaking change, can you?

jekkos: that might be a bot of work as there are certainly queries now that we allow to fail and still proceed the migration. All of this will need to be reviewed and handled. But I agree it is probably the more correct way of working and we will get a lot more bug reports related to uograde failures

jekkos: > > The funny thing is that my db dumps contains those two constraints, but they are actually the same thing
> > ```
> >   UNIQUE KEY `attribute_links_uq1` (`attribute_id`,`definition_id`,`item_id`,`sale_id`,`receiving_id`),
> >   UNIQUE KEY `attribute_links_uq2` (`item_id`,`sale_id`,`receiving_id`,`definition_id`,`attribute_id`),
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So perhaps some setting in MySql causes it to error out on this. In my case it did not bother and did just recreate the constraint with the same fields but in different order
> 
> They are not the same constraint. They contain the same columns but the order of the constraint matters. These were created because they are the best indexes for slow running queries in one or more of the models.

I do have the impression that some MySQL versions do not consider these constraints to be different, hence the upgrade failure

jekkos: So the index is indeed where the order matters. For the unique constraint, the order does not matter. We could convert the second constraint to an index then if it is really needed.

objecttothis: > that might be a lot of work as there are certainly queries now that we allow to fail and still proceed the migration. All of this will need to be reviewed and handled. But I agree it is probably the more correct way of working and we will get a lot more bug reports related to uograde failures

I can take a look into this, but perhaps as a feature to be released after 3.4.0. I agree that this may lead to more bug reports, but that's not because the bugs didn't exist before... people just weren't aware of them.

PR #4058, opened 2024-09-16
Add english fallback if no translation (#3995)

jekkos:
@BudsieBuds I have overridden the default CI4 behavior for loading language strings. The fallback was in place but it was not enabled in case the string did already exist in a translation. So I have extended the Language class with one of our own and made this minor modification to do the fallback also in this case.


BudsieBuds (approved):
Tested locally, works great! Closes #3994.

PR #4060, opened 2024-09-19
Blind SQL injection fix (#3284)

jekkos:
superseeds #3316 
sort column is now validated with the list of headers for each table.

### DIFF
--- a/app/Controllers/Secure_Controller.php
+++ b/app/Controllers/Secure_Controller.php
@@ -82,6 +82,11 @@ class Secure_Controller extends BaseController
 		view('viewData', $this->global_view_data);
 	}
 
+	public function sanitizeSortColumn($headers, $field, $default): string
+	{
+		return $field != null && in_array($field, array_keys(array_merge(...$headers))) ? $field : $default;
+	}
+
 	/**
 	 * AJAX function used to confirm whether values sent in the request are numeric
 	 * @return void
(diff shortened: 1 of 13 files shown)


app/Controllers/Secure_Controller.php
```diff
@@ -82,6 +82,11 @@ public function __construct(string $module_id = '', string $submodule_id = null,
 		view('viewData', $this->global_view_data);
 	}
 
+	public function sortColumn($headers, $field, $default)
```
objecttothis, line 85: Two thoughts:
1. I'm unclear based on the name of the function what it's doing.  It's taking in a haystack multidimensional array, string needle and default sort order then returning the needle if it's not null and exists in the flattened haystack or returning the default sort by column if not.  Our whole purpose for the function is, I think, XSS injection prevention, so I think we should choose a name for this function that reflects that.  Something like sanitizeSortByColumn() or something of that nature. 

2. I believe type hinting is the current best practice for function signatures, so I think we should add the types.  PHP has been moving more toward stricter typing and this future proofs the code.  If need be we can declare them as union types or mixed, but I think these ones are straightforward.

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
         image: jekkos/opensourcepos:sql-master
         command: /bin/sh -c 'exit 0'
     ospos:
-        image: jekkos/opensourcepos:master
+        image: jekkos/opensourcepos:3.3.9
```
objecttothis, line 9: Was this intended to be pushed?

objecttothis (approved):
@jekkos I approved this PR but only because I'm in sporadically and do not want to hold up progress if I can't re-review later.  Please take a look at my comments and decide if you want to make the recommended changes.

app/Controllers/Secure_Controller.php
jekkos: Agree will add the type hint before merge.

docker-compose.yml
jekkos: This needs to be reverted indeed

jekkos:
comments were addressed, will merge.

Issue #3652, opened 2023-02-06
Adding image file name that already exists in the folder does not overwrite the file


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code that [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable.

### Installation information

- OSPOS version is: 3.3.2
- OSPOS git commit hash is: 78c8c2
- PHP version is: 7.4
- MySQL or MariaDB version is: 8.0
- OS and version is: FreeBSD 13
- WebServer is: Apache 2.4
- Selected language is: en-US
- (If applicable) Docker installation: N/A
- (If applicable) Installation package for the LAMP/LEMP stack is: 

### Issue / Bug  / Question / New Feature

[...]

PR #4061, opened 2024-09-19
Fix company logo overwrite

jekkos:
Potential fix for #3652

### DIFF
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -386,15 +386,15 @@ class Config extends Secure_Controller
 
 
 		$filename = $file->getClientName();
-		$info = pathinfo($filename);
 
 		$file_info = [
 			'orig_name' => $filename,
-			'raw_name' => $info['filename'],
+			'raw_name' => md5($filename . '_' . date('Y-m-d H:i:s:u')),
 			'file_ext' => $file->guessExtension()
 		];
 
 		$file->move(FCPATH . 'uploads/', $file_info['raw_name'] . '.' . $file_info['file_ext'], true);
+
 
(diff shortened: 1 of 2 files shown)


objecttothis:
Reviewing now.

app/Controllers/Config.php
```diff
@@ -341,6 +341,7 @@ public function postSaveInfo(): void
 
 		if(!empty($upload_data['orig_name']) && $upload_data['raw_name'])
 		{
+			$filename = md5(date('Y-m-d H:i:s:u'));
```
objecttothis, line 344: Looks like you're assigning `$filename` here, but I don't see where you are using it.

app/Controllers/Items.php
```diff
@@ -845,11 +845,10 @@ private function upload_image(): array
 		}
 
 		$filename = $file->getClientName();
-		$info = pathinfo($filename);
 
 		$file_info = [
 			'orig_name' => $filename,
-			'raw_name' => $info['filename'],
+			'raw_name' => md5(date('Y-m-d H:i:s:u')),
```
objecttothis, line 851: Did you decide against using a hash of filename + datestamp? It's not in a foreach/while loops, so I guess the only way for a duplicate is if two different sessions ran this line in the same microsecond. Not a likely scenario, but not impossible. I don't feel strongly about it.

objecttothis (changes requested):

jekkos:
Comments were addressed but I shuold do a short retest

jekkos:
Comments addressed , tested locally and seems fine, will merge.

PR #4065, opened 2024-09-22
Fixed sales item suggestion by location

ri-cky:


.github/workflows/new-build-image.yml
```diff
@@ -0,0 +1,44 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Image version"
+        required: true
+
+jobs:
+  build_and_push:
+    runs-on: ubuntu-latest
```
jekkos, line 12: Cool you are using github actions here to build the project? #2834

application/models/Item.php
```diff
@@ -544,17 +544,32 @@ private function append_label(&$label, $item_field_name, $item_info)
 		}
 	}
 
-	public function get_search_suggestions($search, $filters = array('is_deleted' => FALSE, 'search_custom' => FALSE), $unique = FALSE, $limit = 25)
+	public function get_search_suggestions($search, $filters = array('is_deleted' => FALSE, 'search_custom' => FALSE, 'location' => NULL), $unique = FALSE, $limit = 25)
```
jekkos, line 547: this is probably the change you wanted to submit? but the changeset seems to be made against old 3.3.9. Currently we only accept contributions against master branch which has moved to CI4.

jekkos (changes requested):

application/models/Item.php
jekkos: should not be so hard to port it to the master branch, if you feel up to that, then that would be great!

ri-cky:
@MalikAbiola Please check the comment and conflict in this PR

application/models/Item.php
ri-cky: I will try to do that and make a fresh PR

PR #4066, opened 2024-09-22
Pre-view filtering Items Controller

jekkos:
- Refactored code for clarity
- Created and called sanitization functions.
- Sanitize TEXT type Attributes before being sent to the view.

- Bump bootstrap-table to 1.23.1 in attempt to resolve issue with sticky headers
- Sanitize attribute data in tables
- Sanitize item data with controller function.

Sanitize Item data

- Sanitize category and item_number before display in forms.
- refactor check in pic_filename for empty to be best practices compliant.
- Added TODO

Minor changes

- Refactored for code clarity.
- Removed extra blank lines.
- Minor reformatting.
- Added PHPdocs
- bumped bootstrap-table to 1.23.2

Pre-view filtering Items Controller

- Refactored code for clarity
- Created and called sanitization functions.
- Sanitize TEXT type Attributes before being sent to the view.

- Bump bootstrap-table to 1.23.1 in attempt to resolve issue with sticky headers
- Sanitize attribute data in tables
- Sanitize item data with controller function.

Pre-view filtering Items Controller

- Refactored code for clarity
- Created and called sanitization functions.
- Sanitize TEXT type Attributes before being sent to the view.

Sanitize Item data

- Sanitize category and item_number before display in forms.
- refactor check in pic_filename for empty to be best practices compliant.
- Added TODO

Pre-view filtering Items Controller

- Refactored code for clarity
- Created and called sanitization functions.
- Sanitize TEXT type Attributes before being sent to the view.

Removed unnecessary use statement

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -5,7 +5,6 @@ namespace Config;
 use App\Events\Db_log;
 use App\Events\Load_config;
 use App\Events\Method;
-use App\Events\PurifyOutput;
 use CodeIgniter\Events\Events;
 use CodeIgniter\Exceptions\FrameworkException;
 use CodeIgniter\HotReloader\HotReloader;
--- a/app/Controllers/Customers.php
+++ b/app/Controllers/Customers.php
@@ -114,7 +114,7 @@ class Customers extends Persons
 				$stats->quantity = 0;
 			}
 
-			$data_rows[] = get_customer_data_row($person, $stats);	//TODO: We either need to create a function to sanitize $person here (and for line 77) or we need to sanitize inside of get_customer_data_row().
+			$data_rows[] = get_customer_data_row($person, $stats);
 		}
 
 		echo json_encode (['total' => $total_rows, 'rows' => $data_rows]);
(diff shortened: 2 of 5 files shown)


jekkos:
@objecttothis I have removed calls to HTMLPurifier in items table data fetch, and can now insert script tags without them executing afterwards. 

image

jekkos:
Instead of peppering the code with the purifier, we might just try to enable CSP that will block any injected javascript that did not exist at page render time using a nonce that we add then to the existing script tag
One downside is that we need to externalize any onClick or other event triggers that are currently still inlined. Beside this change it's just a find/replace all.

jekkos:
#3339 #3331

objecttothis:
@jekkos Assuming we can safely sterilize the scripts so they don't run, I think some people are going to want tags stripped out of columns as an asthetic feature.  What do you think about us (me) adding a feature to add a "allow tags in item data" checkbox (default on) to strip tags from item data. The idea being that when item data is submitted, tags would be stripped. This would be less about safety and more about data maintenance.  Perhaps a separate toggle to allow safe tags in descriptions, would be helpful.  Finally with attributes, it would depend on the type of attribute.  For that we could add a checkbox to the attribute definition settings to allow tags, with them defaulted to on for backward compatibility.

jekkos:
If you do not want to allow tags to be entered, then I would just add an input validation rule that prevents a user from entering them.

objecttothis (approved):

PR #4069, opened 2024-09-23
Fix sale edit form

jekkos:


jekkos:
Did some small changes here but then found that a sale is not saved due to the payments object that is embedded in it . This needs to be debugged further.

objecttothis:
Reviewing now.

app/Controllers/Sales.php
```diff
@@ -1590,13 +1590,12 @@ public function save(int $sale_id = NEW_ENTRY): void
...
 
 		$payment_id = NEW_ENTRY;
-		$payment_amount_new = prepare_decimal($this->request->getPost('payment_amount_new'));
-
-		$payment_amount = filter_var($payment_amount_new, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION);
+		$payment_amount_new = $this->request->getPost('payment_amount_new');
 		$payment_type = $this->request->getPost('payment_type_new', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
 
-		if($payment_type != PAYMENT_TYPE_UNASSIGNED && $payment_amount <> 0)
+		if($payment_type != PAYMENT_TYPE_UNASSIGNED && !empty($payment_amount_new))
```
objecttothis, line 1596: @jekkos how do you want to handle the changeover from snake_case to camelCase? We have a mixture going now.  One approach would be to make the refactoring changes as we touch that part of the code. It's a slower process, but less likely to result in big Pull Requests and the risk for bugs is lower with smaller change sets.

objecttothis (dismissed):
Just the one comment.  Take a look.

app/Controllers/Sales.php
jekkos: I wonder if this conversion could be done using AI somehow. Otherwise if the linter supports this, then we might give that a go

objecttothis: Github Copilot like many of them has a token limit that will probably do it, but you'd have to manually go through each file and get it to convert in chunks.  I think locally running the linter with a unified set of rules is the best way to do this.  This way, when we touch a file, then run the linter locally, it reminds us to convert that file.  This way, we are looking at smaller, more manageable Pull Requests at the expense of a period where some files have snake_case and others camelCase.  Although, we could probably interrogate AI since there may be a better way to approach it.

objecttothis: I created a PR to try out PHP_CodeSniffer.  It successfully warns on snake_case.
image

jekkos:
I believe I have fixed the last issue now here. Something was wrong with the payments datastructure in the sale object.Did a quick test and set a different invoice number and seems to store that now.

app/Controllers/Sales.php
jekkos: WIll have a look if I can install this as well.

Issue #4016, opened 2024-07-01
[Bug]: Attributes do not save correctly


objecttothis:
### Bug Description?

In the Item Form view when save is clicked, the save function first removes all attribute_links for that item_id, then saves the attributes as they have been entered from the form. This allows updates to the attributes without creating duplicates. The removal of the attribute_links is completing but the attributes are not being saved properly.

### Steps to Reproduce?

1. Edit an item or create a new item.
2. Add or edit an attribute.
3. Click submit.

### Expected Behavior?

old attribute_links to be removed and attributes as entered in the form to be saved with attribute_links

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox

### Server Operating System and version

Windows 11

### Database Management System and version

MySQL 8.2.0

### Web Server and version

Apache 2.4.58

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
Did a fix for this and tested both DATE and DRODPOWN attributes, they seem to save fine.

objecttothis:
@jekkos I'm copying this from the Element conversation so that it's easy to find later.
There is definitely an issue with attributes that we will need to resolve. On a new database, I ran migrations, created 5 attributes (one of each type). Added TEXT, saved. Added DATE, saved. Added DROPDOWN, saved. The result is that it deleted the DATE and did not save the DROPDOWN value. When I've looked at the code in the past, the code first stores all the attributes (values) associated with the item in memory, then deletes the links. Finally it saves the new links. The problem is that somewhere after deleting the current links and saving the new ones, it fails, so data is lost. I was afraid this was an issue with existing databases, but this one is fresh. I think we need to re-open the issue for this.

objecttothis:
This is fixed in #4080 and awaiting code review.

PR #4080, opened 2024-09-30
Bugfix Attributes not saving

objecttothis:
This PR resolves two issues.

1. Duplicate Attribute links for the same item/attribute definition can be created in the attribute_links table and this shouldn't be allowed. A migration removes the duplicate links already in the database, then creates a generated unique constraint which will allow for duplicate rows when the sale_id or receiving_id is not null, but will not allow duplicate rows for the item_id/definition_id.
2. Attributes are not saving correctly. See #4016

### DIFF
--- a/app/Database/Migrations/20240826000000_fix_duplicate_attributes.php
+++ b/app/Database/Migrations/20240826000000_fix_duplicate_attributes.php
@@ -57,6 +57,12 @@ class fix_duplicate_attributes extends Migration
 		}
 	}
 
+	/**
+	 * Drops the foreign key constraints from the attribute_links table.
+	 * This is required to successfully create the generated unique constraint.
+	 *
+	 * @return void
+	 */
 	private function drop_foreign_key_constraints(): void
 	{
 		$foreignKeys = [
@@ -83,7 +89,6 @@ class fix_duplicate_attributes extends Migration
 
 			if($query->getNumRows() > 0)
 			{
-				log_message('error', 'Dropping foreign key constraint: ' . $fk);
 				$this->db->query("ALTER TABLE `ospos_attribute_links` DROP FOREIGN KEY `$fk`");
 			}
 		}
--- a/app/Database/Migrations/20240826000000_fix_duplicate_attributes.php
+++ b/app/Database/Migrations/20240826000000_fix_duplicate_attributes.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+use App\Models\Attribute;
+use CodeIgniter\Database\ResultInterface;
+
+class fix_duplicate_attributes extends Migration
+{
+	/**
+	 * Perform a migration step.
+	 */
+	public function up(): void
+	{
+		$rows_to_keep = $this->get_all_duplicate_attributes();
+		$this->remove_duplicate_attributes($rows_to_keep);
+
+		helper('migration');
+
+		$this->drop_foreign_key_constraints();
+
+		execute_script(APPPATH . 'Database/Migrations/sqlscripts/3.4.0_attribute_links_unique_constraint.sql');
+	}
+
+	/**
+	 * Retrieves from the database all rows where the item_id and definition_id are the same AND the sale_id/receiving_id is null.
+	 * It also excludes null item_id rows as those are dropdown items.
+	 *
+	 * @return ResultInterface Results containing item_id, definition_id and attribute_id in each row.
+	 */
+	private function get_all_duplicate_attributes(): ResultInterface
+	{
+		$builder = $this->db->table('attribute_links');
+		$builder->select('item_id, definition_id, MIN(attribute_id) as attribute_id');
+		$builder->where('sale_id IS NULL');
+		$builder->where('receiving_id IS NULL');
+		$builder->where('item_id IS NOT NULL');
+		$builder->groupBy('item_id, definition_id');
+		$builder->having('COUNT(attribute_id) > 1');
+		return $builder->get();
+	}
+
+	/**
+	 * Removes the duplicate attributes from the database.
+	 *
+	 * @param ResultInterface $rows_to_keep A multidimensional associative array containing item_id, definition_id and attribute_id in each row which should be kept in the database.
+	 * @return void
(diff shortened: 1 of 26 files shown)


objecttothis:
I think I'm closer to a fix here but I'm out of time for the day.  In the Items controller the saveItemAttributes function is pulling attributeIds for all the attributes submitted in the form:
```$attributeIds = $this->request->getPost('attribute_ids');```

On a new DATE `definitionType` an attributeId of -1 is sent.  I think this means it's a new attribute, but I need to confirm this.  The problem is that in the Attribute model the if statement which checks if it's a new attribute or not is checking for an empty attribute_id (0, null or false).  If, indeed, -1 equals new, then we should also add a check to 
```
		//New Attribute
		if(empty($attribute_id) || empty($item_id))
		{
```
so that it also looks for -1 attribute_ids.  Something like
```
		//New Attribute
		if(empty($attribute_id)
                    || $attribute_id === NEWATTRIBUTE
                    || empty($item_id))
		{
```
The problem is that when I do this, it pukes.  I think the culprit might be Attribute::attributeValueExists().  I need to step through it with a debugger on a date and see what it does because it's what is returning -1 rather than false.

File: app/Database/Migrations/20240826000000_fix_duplicate_attributes.php [hunk 1]
```diff
@@ -0,0 +1,106 @@
...
+	/**
+	 * Retrieves from the database all rows where the item_id and definition_id are the same AND the sale_id/receiving_id is null.
+	 * It also excludes null item_id rows as those are dropdown items.
+	 *
+	 * @return ResultInterface Results containing item_id, definition_id and attribute_id in each row.
+	 */
+	private function get_all_duplicate_attributes(): ResultInterface
+	{
+		$builder = $this->db->table('attribute_links');
+		$builder->select('item_id, definition_id, MIN(attribute_id) as attribute_id');
```
jekkos, line 35: will this query work with the only_full_group_by enabled?

app/Database/Migrations/sqlscripts/3.4.0_attribute_links_unique_constraint.sql
```diff
@@ -0,0 +1,16 @@
+# Prevents duplicate attribute links with the same definition_id and item_id.
+# This accounts for dropdown rows (null item_id) and rows associated with sales or receivings.
+ALTER TABLE `ospos_attribute_links`
+	ADD COLUMN `generated_unique_column` VARCHAR(255) GENERATED ALWAYS AS (
```
jekkos, line 4: At work we recently faced some issues with some of these generated columns.. at first it seems like a nice feature but it hits you back once you want to migrate the database schema.

It might also have an impact on the Mysql versions can be used. From which version on is this generated column supported? Does it also work on MariaDB and yes what is the minimum version there? Documentation might need to be updated for this.

app/Models/Attribute.php
```diff
@@ -843,35 +842,34 @@ public function get_suggestions(int $definition_id, string $term): array
...
 	{
 		$config = config(OSPOS::class)->settings;
-		$locale_date_format = $config['dateformat'];
 
 		$this->db->transStart();
 
+		switch($definition_type)
+		{
+			case DATE:
+				$data_type				= 'date';
```
jekkos, line 854: glad to see this simplification!

jekkos:
Would it be possible to squash the history here? Seems like a lot of old commits were dragged in for some reason

objecttothis:
I think our current process is to do a squash and merge rather than merge the commits without squash. Is that different from what you're asking? I suspect the commits got dragged in when I did a rebase on the master. Is there a better way to do that?

jekkos (changes requested):
I appreciate the work as always! I do have some remarks on the only_full_group_by and unique_constraint for item and definition_id.

app/Database/Migrations/20240826000000_fix_duplicate_attributes.php
objecttothis: Yes, because my dev database has `ONLY_FULL_GROUP_BY` enabled in the sql_mode and it didn't puke during testing.

app/Database/Migrations/sqlscripts/3.4.0_attribute_links_unique_constraint.sql
objecttothis: According to copilot they are supported from MySQL: Starting from version 5.7 and MariaDB: Starting from version 10.2. If there is another way to enforce uniqueness, I'm happy to explore it.

app/Models/Attribute.php
objecttothis: yay for refactoring out duplicate code blocks!

jekkos:
I'll rebase the PR otherwise this weekend.

objecttothis:
@jekkos is the only issue with this PR that the commit history is messy?  It seems that the method I'm using for rebasing on the master is causing the messy commit history.  Within PHPStorm I go to git > fetch all remotes > right-click origin/master > rebase 'X' onto 'origin/master'. I let it replay the commit history. When it comes to a conflict, I select the code changes from my code and continue the rebase. At the end I push the commits generated from resolving conflicts during the rebase operation.  Should I be doing this differently?  If not and it's normal to need to clean up commit history, what is the best way to see the commit history and clean it up. I'm generally not used to using git through the command line, so bonus if there is a GUI method.  

Also, usually when I merge a PR, I squash and merge rather than just merge.  I could clean up the commit messages there as those tend to be concatenated together.

jekkos:
Ah right I just meant squash the history as it seems this has some duplicate commit messages that are already in master. So if you can squash the old commits out at the end then that is fine as well. Typically I do this on the command line before I make the PR, but doing it here can be convenient.

objecttothis:
> Ah right I just meant squash the history as it seems this has some duplicate commit messages that are already in master. So if you can squash the old commits out at the end then that is fine as well. Typically I do this on the command line before I make the PR, but doing it here can be convenient.

Yes, I will do this at the end but it won't allow me to squash and merge unless I get a code review approval.  I can override it but I prefer not to get in that habit.

jekkos (approved):
Sorry I thought I already approved it, didn't mean to block

PR #4089, opened 2024-10-07
Add PHP_CodeSniffer IDE Linter and enforce PSR-12

objecttothis:
- Added Composer Dev dependency
- Added rule file

@jekkos please take a look at my comments on Element. The tl;dr is that I couldn't get PHP CS Fixer to work properly.  PHP_CodeSniffer does have native PHPStorm support and can be setup via Github actions.  

If we decide to go with this, We probably should setup the github action so that it just gives warnings but still allows merging. That is unless we can get it to only check the files which have changes. If it checks the whole project, then it's going to hold Pull Requests up until the whole project is converted.  If warning mode isn't a thing then maybe we should just run it on IDEs first until the codebase is fully converted?

I couldn't find a way to tell PHP_CodeSniffer to require PSR-12 compliance but override some sniffs (such as space indenting).

### DIFF
--- a/.phpcs.xml
+++ b/.phpcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0"?>
+<ruleset name="OSPOSStandard">
+	<description>Custom Coding Standard based on PSR-12 for OpenSourcePOS</description>
+	<arg name="tab-width" value="4"/>
+	<rule ref="PSR12"/>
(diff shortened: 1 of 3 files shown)


jekkos:
I would bundle the linting + formatting changes in this PR instead of in the config adaptions

Issue #3633, opened 2023-01-27
CI4: Possible duplicate import of PSR LoggerInterface


objecttothis:
PHPStorm is complaining about two different instances of PSR\Log\LoggerInterface classes.

One is in `\vendor\psr\log\PSR\log` and the other is in `\vendor\codeigniter4\framework\system\ThirdParty\PSR\log`

![image](https://user-images.githubusercontent.com/17935339/215047869-aa10558a-3cb1-4e42-90a6-e7b49706c169.png)

![image](https://user-images.githubusercontent.com/17935339/215047805-befe5298-443f-48a8-a1b3-41fc0f6010e6.png)

It seems like since PSR\log is now included in CI4, we should remove the duplicate.  This happens during the npm install and I'm seeing the psr/log reference in the composer.lock file as a CodeIgniter dependency. I'm not sure the best way to remedy this because I believe the composer.lock file is automatically generated.

jekkos:
Could be handled as follows: https://stackoverflow.com/questions/34637657/is-it-possible-to-ignore-child-dependencies-in-composer-config

objecttothis:
#4091 fixes this issue.

PR #4091, opened 2024-10-21
Bump Bootstrap-table and minor composer fixes

objecttothis:
It doesn't resolve the issue but keeps the version at the latest.

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -277,7 +277,7 @@ class App extends BaseConfig
 	 * @see http://www.html5rocks.com/en/tutorials/security/content-security-policy/
 	 * @see http://www.w3.org/TR/CSP/
 	 */
-	public bool $CSPEnabled = false;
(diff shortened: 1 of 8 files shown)


objecttothis:
I also pushed a fix for #3633 in this PR.

app/Config/App.php
```diff
@@ -277,7 +277,7 @@ class App extends BaseConfig
 	 * @see http://www.html5rocks.com/en/tutorials/security/content-security-policy/
 	 * @see http://www.w3.org/TR/CSP/
 	 */
-	public bool $CSPEnabled = false;
+	public bool $CSPEnabled = false;	//TODO: Currently CSP3 tags are not supported so enabling this causes problems with script-src-elem, style-src-attr and style-src-elem
```
jekkos, line 280: what needs to be done to enable it?

jekkos (approved):

app/Config/App.php
objecttothis: On our end just change this to true. On the CI end they need to probably add fields representing CSP level 3 tags to ContentSecurityPolicy.php and the back end business logic. I'm guessing it's not a very involved PR.

PR #4106, opened 2024-11-04
Bump CodeIgniter4 to 4.5.5

objecttothis:
- Updated composer.json and composer.lock.
- Ran through steps in https://codeigniter.com/user_guide/installation/upgrade_452.html
- Ran through steps in https://codeigniter.com/user_guide/installation/upgrade_453.html (this bumps several packages)
- Ran through steps in https://codeigniter.com/user_guide/installation/upgrade_454.html
- Ran through steps in https://codeigniter.com/user_guide/installation/upgrade_455.html

### DIFF
--- a/app/Config/ContentSecurityPolicy.php
+++ b/app/Config/ContentSecurityPolicy.php
@@ -76,7 +76,7 @@ class ContentSecurityPolicy extends BaseConfig
 		'https://fonts.googleapis.com',
 	];
 
-	/**
+    /**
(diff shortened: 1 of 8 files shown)


objecttothis:
@jekkos this one needs review also. Let me know when you think you might have time to review them.  I think once we merge these then #4107 might be the only outstanding issue that needs to get resolved before we can release 3.4.0.

jekkos (approved):

Issue #3611, opened 2022-12-10
CI4: Rework file permission and package checker for CI4


objecttothis:
CI4 requires PHP 7.4 or newer, mbstring, intl, php-json, php-mysqlnd, php-xml, libcurl, MySQL 5.1 or newer and 
/writable/ as well as it's subdirectories to be writable by apache.  We need to rework the checker code to make sure it's looking in the right places for the right permissions and packages to match the requirements. @WebShells I think you did the original work on this.  You interested?

#3592

jekkos:
I asked @WebShells if he could have a look at this one.

WebShells:
Hello everyone!
Will do, when is the due for CI4 ? 
@objecttothis just saw this issue am not receiving repos emails anymore... Thanks @jekkos!

objecttothis:
@WebShells sorry you're not getting the notifications.  I'm not sure why, but it isn't anything we've intentionally done.  The CI4 upgrade is moving along at a little quicker pace now that we have a few people working on it as they have time and are healthy, but the more people working on it, the faster we will be able to release 3.4.0.  You should probably try to subscribe to notifications on #3858 since that is the PR for the CI4 branch.  the branch to check out is called `ci4-branch`

WebShells:
@objecttothis am already subscribed to it and to the repo yet receiving nothing. Anyway i'll check that... Thanks

objecttothis:
This is a screenshot of the GitHub app on Android when you click notifications. I only get them when I have participated in an issue or if I am mentioned. Maybe you could look at what is enabled for you.

![Screenshot_2023-11-10-14-50-46-66_320a9a695de7cdce83ed5281148d6f19.jpg](https://github.com/opensourcepos/opensourcepos/assets/17935339/a064f067-67d1-48c5-9a60-ee1d0c5dc20b)

objecttothis:
@WebShells do you still plan on working on this?

WebShells:
@objecttothis Even if am mentioned nothing received.

objecttothis:
> @objecttothis Even if am mentioned nothing received. 

Potentially we could remove all your permissions and readd them. Do you get notifications for other GitHub repos you are participating in? If not, I doubt that will help.

PR #4108, opened 2024-11-04
Adapt Configuration checker for CI4

objecttothis:
- Removed $import variable as it is never used and the code generates the csv file for item imports
- Refactored import_customers.csv to match PSR-12 standard file names
- Refactored variable names to match PSR-12 standard variable names
- Updated .editorconfig to reflect PSR-12 requirement for spaces rather than tab symbols. https://www.php-fig.org/psr/psr-12/#24-indenting
- Replaced tabs with spaces
- Added version number to browser reporting
- Corrected timezone reporting


objecttothis:
Closing #3611

.editorconfig
```diff
@@ -4,19 +4,19 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_size = 4
-indent_style = tab
+indent_style = space
```
jekkos, line 7: if we go from spaces to tabs I would reformat the all files in the whole repo in one commit, otherwise it will be very hard to track any other changes

objecttothis: Good point.  I'll create a separate branch and make that change.

objecttothis: @jekkos both #4108 and #4109 are ready for review.  I will merge the latter into the former once both are approved before merging #4108 into the master.

jekkos: I'd hold back this file from the PR for now. It's functionally not related to the change and belongs to the linter and code formatting improvements that we planned to do. I can tidy this up for you.

jekkos (approved):

PR #4109, opened 2024-11-05
Mass Replace tab characters with spaces

objecttothis:
This is for PSR-12 compliance


jekkos:
Do you want to merge this still before release @objecttothis ? I thought we put the linting tickets on hold until after

objecttothis:
Yeah, I think we could just change the target branch to the other one. I'm fine with waiting to merge it until after the 3.4.0 release.

We had talked about waiting to implement the linter until after 3.4.0 but then we talked about making changes to the code style as we worked on that particular code. I would not have pushed this branch in the 3.4.0 release except in response to changing the one file, you had suggested we do it all at once to prevent difficulty reviewing code.  I think we should just have the one or two files that are different for the 3.4.0 release and push changes for linting to right after 3.4.0.

The order of operations I can see is:
1. Merge final code changes without mass linting changes.
2. Change the target of the linting changes to the CS Fixer branch.
3. Rebase CS Fixer on master.
4. Merge this branch into CS Fixer.
5. Merge CS Fixer into master.

Does that look correct?

jekkos:
Sounds good, I have merged already the code changes without the linting adaptation. So we can point the mass replace to the other branch now.

objecttothis:
Closing as out of date

Issue #4115, opened 2024-11-12
[Bug]: Error Editing Item when changing quantity


objecttothis:
### Bug Description?

![image](https://github.com/user-attachments/assets/173ce763-20e7-4145-8569-e7029e2a4503)

### Steps to Reproduce?

1. Go to sales register.
2. Add an item.
3. Change quantity.
4. Tab.

### Expected Behavior?

Quantity to change without error.

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox

### Server Operating System and version

Windows 11

### Database Management System and version

MySQL 8.2.12

### Web Server and version

Apache 2.4.58

### System Information Report (optional)

OSPOS Installation Info: 3.4.0-dev - dev
Language Code: en

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓
» Xml: Enabled ✓

User Configuration:
.Browser: Firefox 133.0
.Server Software: Apache/2.4.58 (Win64) OpenSSL/3.1.3 mod_fcgid/2.3.10-dev
.PHP Version: 8.2.12
.DB Version: 8.2.0
.Server Port: 80
.OS: Windows NT 10.0

.OS Time Zone: Asia/Baku
.OSPOS Time Zone: Asia/Baku

File Permissions:
» [writeable/logs:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [public/uploads:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [writable/uploads/item_pics:] - 0777 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [importCustomers.csv:] - 0666 | Readable ✓ | Vulnerable or Incorrect Permissions ✗

Security Vulnerability Warning
Incorrect permissions leaves this software at risk.

» [writeable/logs:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [writable/uploads:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [writable/uploads/item_pics:] is writable, but the permissions are incorrectly set. Please set it to 750 and refresh.
» [importCustomers.csv:] is readable, but the permissions are incorrectly set. Please set it to 640 or 660 and refresh. 

[...]

PR #4117, opened 2024-11-12
Use custom rule to account for all locales

objecttothis:
This fixes a problem where validation fails when locales which use a comma (,) as the decimal separator don't pass the numeric validation rule.
closing #4115

### DIFF
--- a/app/Controllers/Sales.php
+++ b/app/Controllers/Sales.php
@@ -633,9 +633,9 @@ class Sales extends Secure_Controller
 		$data = [];
 
 		$rules = [
-			'price' => 'trim|required|numeric',
-			'quantity' => 'trim|required|numeric',
-			'discount' => 'trim|permit_empty|numeric',
+			'price' => 'trim|required|decimal_locale',
+			'quantity' => 'trim|required|decimal_locale',
+			'discount' => 'trim|permit_empty|decimal_locale',
 		];
 
 		if($this->validate($rules))


jekkos:
Won't we have this issue then in other places as well? I mean eg in items we can also input a price in different locale which can be comma or dot. I thought that numeric validation was already taking locale into account? Maybe this part is just broken and it needs fixing?

objecttothis:
> Won't we have this issue then in other places as well? I mean eg in items we can also input a price in different locale which can be comma or dot. I thought that numeric validation was already taking locale into account? Maybe this part is just broken and it needs fixing?

We will have the issue if anyone adds the decimal or numeric validation rule to a number coming from the request object.  numeric and decimal rules do not account for decimals which use a comma as a decimal separator.  You can see from the code below that numeric is just calling the php function is_float() against the value. is_float is expecting a period as a decimal separator so it returns false when the comma is used.  We would need to intercept and modify the request variables before calling the validation function, but that's not good practice. To resolve the issue, we wrote a custom validation rule that takes the locale into account when determining if it's a valid number. Making the replacement I did in the code resolves the problem.  I'm not sure how else it could be broken, but maybe I missed something.

```
    /**
     * Numeric
     *
     * @param array|bool|float|int|object|string|null $str
     */
    public function numeric($str = null): bool
    {
        if (is_int($str) || is_float($str)) {
            $str = (string) $str;
        }

        if (! is_string($str)) {
            return false;
        }

        return $this->nonStrictFormatRules->numeric($str);
    }
```

Issue #4107, opened 2024-11-04
[Bug]: Validation doesn't apply during Items editing


objecttothis:
### Bug Description?

With the locale set to any locale which uses comma as decimal separator when editing the Wholesale price of an item if you use a period, the validation does not catch it and clicking submit causes both wholesale and retail to be erased.

Before editing:
![image](https://github.com/user-attachments/assets/f26f30a0-3413-45e4-a0e9-b0a3ffd605c2)

Edited:
![image](https://github.com/user-attachments/assets/0373be7c-60b4-40f0-bb90-ccb87cc57b15)

Lost Data:
![image](https://github.com/user-attachments/assets/048aa7b0-0189-4593-a295-768947827c06)

NOTE: The validation does trigger and catch the problem if you navigate away from the field (e.g., press tab).  However in this case if you correct the value, the validation either doesn't trigger or doesn't recognize that the value has been fixed which prevents you from submitting the form.

### Steps to Reproduce?

1. Set locale in Config > Localization to `az_AZ` or other locale which uses a comma for the decimal separator.
2. Go to Items
3. Click edit next to an existing item
4. Change the wholesale price and use a period instead of the correct comma.
5. Clicck Submit.

### Expected Behavior?

Three issues need to be resolved here:

1. We need to run validation even when the user doesn't press tab.  I assume this means we need to have Submit OnClick trigger validations to run before actually submitting the data.
2. We need to find and fix whatever is causing both wholesale and retail prices being erased in the event that invalid data is submitted.  I think this means we need to fail the operation and return an error to the user rather than erase the data.
3. We need to re-run validation after it fails and the user fixes the value.  I think this is probably an issue of either validation calls not being triggered again or being triggered but not recognizing that the issue is resolved.

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox

### Server Operating System and version

Windows 11

### Database Management System and version

MySQL 8.2.12

### Web Server and version

Apache 2.4.58

### System Information Report (optional)

OSPOS Installation Info: 3.4.0-dev - dev
Language Code: en

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓
» Xml: Enabled ✓

User Configuration:
.Browser: Firefox
[...]

PR #4121, opened 2024-11-16
Apply validation before submit (#4107)

jekkos:
Have not tested this yet but let's see if it can resolve the validation issue.

### DIFF
--- a/app/Helpers/locale_helper.php
+++ b/app/Helpers/locale_helper.php
@@ -485,17 +485,6 @@ function parse_decimals(string $number, int $decimals = null): mixed
 		return $number;
 	}
 
-	$locale_safe_number = prepare_decimal($number);
-
-	if ($locale_safe_number > MAX_PRECISION)
-	{
-		return false;
-	}
-
-	if($locale_safe_number > 1.e14)
-	{
-		return false;
-	}
 
 	$config = config(OSPOS::class)->settings;
 
@@ -503,10 +492,9 @@ function parse_decimals(string $number, int $decimals = null): mixed
 
 	if(!$decimals)
 	{
-		$decimals = $config['currency_decimals'];
+		$decimals = int($config['currency_decimals']);
+		$fmt->setAttribute(NumberFormatter::FRACTION_DIGITS, $decimals);
 	}
-
-	$fmt->setAttribute(NumberFormatter::FRACTION_DIGITS, $decimals);
 
 	if(empty($config['thousands_separator']))
 	{
@@ -515,7 +503,19 @@ function parse_decimals(string $number, int $decimals = null): mixed
 
 	try
 	{
-		return $fmt->parse($number);
+		$locale_safe_number = $fmt->parse($number);
+
+		if ($locale_safe_number > MAX_PRECISION)
+		{
+			return false;
+		}
+
+		if($locale_safe_number > 1.e14)
+		{
+			return false;
+		}
+
+		return (float) $locale_safe_number;
 	}
 	catch(Exception $e)
 	{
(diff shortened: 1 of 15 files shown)


app/Helpers/locale_helper.php
```diff
@@ -485,37 +485,37 @@ function parse_decimals(string $number, int $decimals = null): mixed
...
 	{
 		$fmt->setTextAttribute(NumberFormatter::GROUPING_SEPARATOR_SYMBOL, '');
 	}
 
 	try
 	{
-		return $fmt->parse($number);
+		$locale_safe_number = $fmt->parse($number);
+
+		if ($locale_safe_number > MAX_PRECISION)
```
objecttothis, line 508: Could these two if statements be combined into
```
if ($locale_safe_number > MAX_PRECISION
    || $locale_safe_number > 1.e14)
{
	return false;
}
```

objecttothis:
@jekkos I have a friend helping with this project for our bookstore. Unless you want otherwise, I'm going to assign this to him to finish off.

jekkos:
@objecttothis Thanks for the offer. I currently do have a bit of limited bandwith due to new family situation, but was planning to pick this one up again at some point. I did fix some smaller bugs and made a master build available through the CI pipeline. This work also had to be done before 3.4 final release.

So in case he's able to work on it this week then fine. I can still pick it up if there is no progress after.

app/Helpers/locale_helper.php
dchristm77: Completed in my latest commit.

dchristm77:
I found a small bug (probably typo) casting a value to int and in the parse function I found that the Numberformatter was correctly returning "false" when the input was invalid according to the set locale but then the code would cast this value to a double before returning it, resulting in returning a zero value instead of False. Fixed this and tested it against various invalid inputs with success.

objecttothis:
> I found a small bug (probably typo) casting a value to int and in the parse function I found that the Numberformatter was correctly returning "false" when the input was invalid according to the set locale but then the code would cast this value to a double before returning it, resulting in returning a zero value instead of False. Fixed this and tested it against various invalid inputs with success.
> 
> Also I have my editor set to format according to the PSR12 standard, meaning it reformatted the locale_helper file on save to remove tabs and fix idents. If you would like these removed let me know. 

I'll take a look at this tomorrow unless someone else wants to review it. The only downside to replacing tabs with spaces automatically is that the diff comes out showing every line changed which makes the review more difficult. I have a branch which mass replaces tabs for spaces. We plan to merge that as soon as we release 3.4.0. order PSR12 changes, like naming conventions, feel free to make.

dchristm77:
Sure, I understand. I'll revert the commit tomorrow and commit again with minimal changes.

jekkos:
> Sure, I understand. I'll revert the commit tomorrow and commit again with minimal changes.

Great thanks! Agree with @objecttothis its hard to review the change this way. Welcome BTW

dchristm77:
Thanks @jekkos and this PR is ready for review again.

app/Helpers/locale_helper.php
```diff
@@ -485,37 +485,36 @@ function parse_decimals(string $number, int $decimals = null): mixed
...
 
 	$config = config(OSPOS::class)->settings;
 
 	$fmt = new NumberFormatter($config['number_locale'], NumberFormatter::DECIMAL);
 
 	if(!$decimals)
 	{
-		$decimals = $config['currency_decimals'];
+		$decimals = intVal($config['currency_decimals']);
+		$fmt->setAttribute(NumberFormatter::FRACTION_DIGITS, $decimals);
```
objecttothis, line 496: This looks like it will be OK since it's using number_locale, but since we don't have working unit tests, make sure to test this against a locale that uses comma as a decimal separator.

objecttothis (commented):
Looks good to me.  @jekkos I believe you made some changes affecting lines 646ff of the Items controller.  Did that branch get merged already or was that this branch?

jekkos:
I need to continue and review and remove occurrences of prepare_decimal to wrap this one up

jekkos:
I have rebase this branch with master and deployed it on the dev server. I can confirm that the decimals are saved now in this case. There's a bit of cleanup remaining but then after we should be good to go.

jekkos:
@objecttothis now that we have fixed parse_decimals I believe that all calls to prepare_decimals are not needed anymore. We can just directly parse the number and then use the result that way.
Also the filter_var might be obsolete when we already have a parsed number. Do you agree if we remove this from the code?

objecttothis:
Sounds good. Simple is better.

jekkos:
Linting for this branch was successful.

jekkos:
Fixed one more bug in the frontend validation that would fail in case plain 0 as a price was used. I believe this is good to go now. I did the cleanup and test with belgian locale that has comma separator for decimals.
Let me know if someone can review @objecttothis

objecttothis (approved):

Issue #4126, opened 2024-11-27
[Bug]: Issue generating a Quote


odiea:
### Bug Description?

![2024-11-27 09_11_00-Window](https://github.com/user-attachments/assets/df7522be-d60a-4462-bf76-7565825d05f9)

### Steps to Reproduce?

Create a Quote .

### Expected Behavior?

Quote number should appear on the quote not the date as shown

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.4

### What browsers are you seeing the problem on?

Chrome

### Server Operating System and version

Windoes 10

### Database Management System and version

Wamp Server

### Web Server and version

Wamp Server

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
Fix in master now.

PR #4129, opened 2024-12-02
Revert IntlFormatter refactor (#4126)

jekkos:
@objecttothis I was checking the bug report from @odiea  and saw some refactoring was done that broke the quote number generation.The strformat was changed with an IntlFormatter->format that now always renders the full date instead of the token format as it is configured in the db.
I did check briefly if I could make it work in the spirit of your refactoring, but did not directly find a solution. THe format string cannot be used with this IntlFormatter so it seemed to be a dead end.

Fine if we revert this or was there a specific reason you wanted to refactor this?

### DIFF
--- a/app/Libraries/Token_lib.php
+++ b/app/Libraries/Token_lib.php
@@ -19,26 +19,25 @@ class Token_lib
 	 */
 	public function render(string $tokened_text, array $tokens = [], $save = true): string
 	{
-		$config = config(OSPOS::class)->settings;
-		$formatter = new IntlDateFormatter(
-			$config['number_locale'], // Locale
-			IntlDateFormatter::FULL, // Date type
-			IntlDateFormatter::FULL // Time type
-		);
-
-		//Apply the transformation for the "%" tokens if any are used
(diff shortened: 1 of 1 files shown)


objecttothis:
Let me take a look real quick.  I'm trying to remember why I had refactored it.  I suspect it had something to do with a problem where dates in ddmmyyyy and mmddyyyy were not behaving the same way.  I can't think of another reason to refactor.  I think as long as we test CSV imports which have ddmmyyyy, dd.mm.yyyy and dd/mm/yyyy formats and they behave themselves correctly, then it's probably fine to revert.

jekkos:
@objecttothis I would not expect the CSV parser to run through token rendering logic? I believe this is only used to render the quote / invoice numbers using some variables so a user can define his own numbering scheme.

objecttothis (approved):

objecttothis:
It's possible I had made a similar refactoring in another place and it was early enough that I could not test its impact on this section of code due to build issues then I never got around to it later.

PR #4131, opened 2024-12-05
To prevent sticky header issue in Reports

odiea:


jekkos:
Thanks for the fix, in master now!

PR #4132, opened 2024-12-06
Now you can search for items in sales in any order. For example, if a…

andriux1990:
…n item is called big red apple, you can search for it as red b  app

### DIFF
--- a/app/Models/Item.php
+++ b/app/Models/Item.php
@@ -695,11 +695,21 @@ class Item extends Model
 		$suggestions = [];
 		$non_kit = [ITEM, ITEM_AMOUNT_ENTRY];
 
+		// Dividir la búsqueda en palabras clave
+	    $keywords = explode(' ', trim($search));
+
+	    // Función para construir condiciones "LIKE" para cada palabra clave
+	    $build_like_query = function($builder, $field, $keywords) {
+	        foreach ($keywords as $word) {
+	            $builder->like($field, $word);
+	        }
+	    };
+
 		$builder = $this->db->table('items');
 		$builder->select($this->get_search_suggestion_format('item_id, name, pack_name'));
 		$builder->where('deleted', $filters['is_deleted']);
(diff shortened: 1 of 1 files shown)


app/Models/Item.php
```diff
@@ -695,11 +695,21 @@ public function get_search_suggestions(string $search, array $filters = ['is_del
 		$suggestions = [];
 		$non_kit = [ITEM, ITEM_AMOUNT_ENTRY];
 
+		// Dividir la búsqueda en palabras clave
+	    $keywords = explode(' ', trim($search));
+
+	    // Función para construir condiciones "LIKE" para cada palabra clave
```
objecttothis, line 701: Comments need to be in English.  Though, from the context, I don't think we need the comments at all.  Same with the above.

app/Models/Item.php
```diff
@@ -695,11 +695,21 @@ public function get_search_suggestions(string $search, array $filters = ['is_del
...
+	    $build_like_query = function($builder, $field, $keywords) {
```
objecttothis, line 702: opening brackets need to be on their own line

objecttothis, line 702: Rather than use an inline function here, I think we should create a private function in the model... alternately this function could potentially be more useful in a helper since we could use it in other models.  For example customer names, sales register searches., etc.

jekkos: Seems like a very generic.function indeed. If we implement this here then why not also I customers?

andriux1990: In fact, I have that same search logic for clients, inventory items, suppliers, and customers in the sales search.  
In OSPOS 3.3, I will migrate them to 3.4 and upload the changes.

jekkos: That would be great @andriux1990

jekkos:
@andriux1990 I did not mean to close this PR as I was not aware you created a full new repository for this. It's better to just push branches to the main repo instead and then create PRs againt master from there. No need to create a new repo for every change.
 Also would it be feasible to implement the change for all controllers? Thanks!

PR #4139, opened 2024-12-28
Receiving Module: Purchasing and requisition development has been com…

andriux1990:
…pleted.

### DIFF
--- a/app/Controllers/Receivings.php
+++ b/app/Controllers/Receivings.php
@@ -415,6 +415,52 @@ class Receivings extends Secure_Controller
 	}
 
 	/**
+	 * Complete a receiving requisition and save to the database.
+	 *
+	 * @throws ReflectionException
+	 * @noinspection PhpUnused
+	 */
+	public function complete(): void
+	{
+		$data = [];
+
+		$data['cart'] = $this->receiving_lib->get_cart();
+		$data['total'] = $this->receiving_lib->get_total();
+		$data['transaction_time'] = to_datetime(time());
+		$data['mode'] = $this->receiving_lib->get_mode();
+		$data['comment'] = $this->receiving_lib->get_comment();
+		$data['reference'] = $this->receiving_lib->get_reference();
+		$data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
+		$data['stock_location_source'] = $this->receiving_lib->get_stock_source();
+		$data['stock_location_destination'] = $this->receiving_lib->get_stock_destination();
+
+		$employee_id = $this->employee->get_logged_in_employee_info()->person_id;
+		$employee_info = $this->employee->get_info($employee_id);
+		$data['employee'] = $employee_info->first_name . ' ' . $employee_info->last_name;
+
+
+		if ($data['stock_location_source'] != $data['stock_location_destination']) {
+			// Save receiving to database
+			$data['receiving_id'] = 'REQ ' . $this->receiving->save_requisition($data['cart'], $employee_id, $data['comment'], $data['reference'], $data['stock_location_source'], $data['stock_location_destination']);
+
+			if ($data['receiving_id'] == 'REQ -1') {
+				$data['error_message'] = lang('Receivings.transaction_failed');
+			} else {
+				$data['barcode'] = $this->barcode_lib->generate_receipt_barcode($data['receiving_id']);
+			}
+
+			$data['print_after_sale'] = $this->receiving_lib->is_print_after_sale();
+
+			echo view("receivings/receipt", $data);
+
+			$this->receiving_lib->clear_all();
+		} else {
+			$data['error'] = lang('Receivings.error_requisition');
+			$this->_reload($data);
+		}
+	}
+
+	/**
 	 * Gets the receipt for a receiving. Used in app/Views/receivings/form.php
 	 *
 	 * @param $receiving_id
--- a/app/Models/Receiving.php
+++ b/app/Models/Receiving.php
@@ -187,7 +187,7 @@ class Receiving extends Model
 
 			$inventory->insert($inv_data, false);
 			$attribute->copy_attribute_links($item_data['item_id'], 'receiving_id', $receiving_id);
-			$supplier = $supplier->get_info($supplier_id);	//TODO: supplier is never used after this.
+		//	$supplier = $supplier->get_info($supplier_id);	//TODO: supplier is never used after this.
 		}
 
 		$this->db->transComplete();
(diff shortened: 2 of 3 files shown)


andriux1990:
I would like you to review the changes and accept the pull request.
I have processed nearly 300 purchases and requisitions, and I can confirm that this part of the system works correctly.

Currently, OSPOS does not handle purchases properly when multiple items are added, nor does it correctly process item transfers between locations.

I have fixed these issues and tested the changes thoroughly. I would like these improvements to be included in the master branch.
If you notice any part of the code that does not look or work as it should, I would appreciate it if you could update it as well.

andriux1990:
I have almost finished implementing AJAX in the receivings view, and it’s amazing not having to deal with page reloads every time an item is edited or added.

In the OSPOS version based on CodeIgniter 3, I implemented something similar to suspended sales, but this time for purchases in receivings. This allows a purchase to be edited multiple times without losing its ID, making it possible to resume and complete it later.
Upon completion, the items are automatically added to the inventory.

I would like you to review these changes because I also want to upload the AJAX implementation and the "suspended receivings" feature.

app/Models/Receiving.php
```diff
@@ -187,14 +187,161 @@ public function save_value(array $items, int $supplier_id, int $employee_id, str
 
 			$inventory->insert($inv_data, false);
 			$attribute->copy_attribute_links($item_data['item_id'], 'receiving_id', $receiving_id);
-			$supplier = $supplier->get_info($supplier_id);	//TODO: supplier is never used after this.
+		//	$supplier = $supplier->get_info($supplier_id);	//TODO: supplier is never used after this.
```
jekkos, line 190: I'd just remove this in case it's unused

app/Controllers/Receivings.php
```diff
@@ -414,6 +414,52 @@ public function postRequisitionComplete(): void
...
 
+	/**
+	 * Complete a receiving requisition and save to the database.
+	 *
+	 * @throws ReflectionException
+	 * @noinspection PhpUnused
+	 */
+	public function complete(): void
+	{
+		$data = [];
```
jekkos, line 425: I believe the `postRequisitionComplete` needs to be reviewed as it contains same logic as you added here. Is this former method now deprecated or should we replace the logic there with the logic now in this function?

jekkos: I see the call to complete() in this former function was just broken is this correct? You did fix this for the CI4 upgrade?

jekkos:
If this is currently broken in master then it needs to be fixed before we release 3.4

andriux1990:
> If this is currently broken in master then it needs to be fixed before we release 3.4. I'll try to deploy your branch to dev to test this functionality a bit further later on. Thanks already for the submission.

I am carrying out all this work in the current version of the repository, which is based on CodeIgniter 4. I no longer use version 3.3.9 due to PHP compatibility issues.

I have thoroughly tested the changes, covering nearly 300 purchases and requisitions. In the database, I reviewed and ensured that the tables ospos_receivings_items, ospos_inventory, and ospos_item_quantities record each piece of data correctly.

I used the public function complete() because it allows me to properly manage the items for requisitions. This is essential since each item must be recorded in two lines: one for the inventory outflow and the other for the destination.

app/Models/Receiving.php
andriux1990: I did not remove this line because of the comment it had, as I assumed it was there for a specific reason. Although I do not know the exact purpose of its presence, I decided not to delete it and simply commented it out. What I do know is that if the line is uncommented, it affects the process of adding more items to receivings.

andriux1990:
> If this is currently broken in master then it needs to be fixed before we release 3.4. I'll try to deploy your branch to dev to test this functionality a bit further later on. Thanks already for the submission.

It is now working correctly because I integrated it into the main branch. It would be ideal if you could review it and, if possible, accept the pull request so that the main branch can function properly. Additionally, I am making further progress with more changes. Currently, I try not to write too much code but rather focus on specific aspects and keep the OSPOS base as pure and clean as possible.

app/Controllers/Receivings.php
andriux1990: I have fixed it. In fact, if I don’t use that function, the behavior of postRequisitionComplete doesn’t work properly. On the other hand, if I replace complete with postRequisitionComplete, although it processes the requisition correctly, it doesn’t detail the outflow in ospos_receivings_items. That’s why I decided to use both functions, and now everything works perfectly.

jekkos:
Looks good at first sight, but I'd like review the business logic a bit closer. Your explanation makes sense and the intention is to merge it soon.

andriux1990:
> Looks good at first sight, but I'd like review the business logic a bit closer. Your explanation makes sense and the intention is to merge it soon.

Thanks @jekkos

I would like you to review this code and approve it, as I'm working on another pull request. I'm implementing the functionality to add payment types in purchases, similar to how it's done in the sales module.

app/Controllers/Receivings.php
jekkos: Might be nitpicking but can we use a better name for this method or just inline the complete method in postRequisitionComplete? Now this last one calls complete which is a very generic function name. So I would.just execute all logic in the postRequistionComplete instead

jekkos:
Your new PR is based on this one? It is quite common to have some.PRs in progress and these can depend on eachother. Maybe we can setup a call to discuss a bit the workflow we can try as it might be annoying to make changes to one PR while you are already working on a next one. I can explain how to leverage git functionality for this

andriux1990:
> Your new PR is based on this one? It is quite common to have some.PRs in progress and these can depend on eachother. Maybe we can setup a call to discuss a bit the workflow we can try as it might be annoying to make changes to one PR while you are already working on a next one. I can explain how to leverage git functionality for this

Sure, here’s the English translation:

It's fine, you can add the PR as soon as this other one is completed.

My new PR is related to the Receivings module, but it does not address completing purchases or performing transfers (requisitions). Currently, Receivings does not have a method to record payments as the Sales module does.

In the ospos_receivings table, we have the columns total, amount_change, and amount_tendered. However, this is not equivalent to recording payments in detail.

I added a table called ospos_receivings_payments following the same logic as ospos_sales_payments, so that payments can be recorded and associated with each receiving by its ID.

This way, the Receivings module will be more complete, as it will allow for an accurate record of payments. Additionally, it would enable the option to incorporate credit payments (when a supplier delivers items, and payment is made later). This functionality could be extended to create an Accounts Payable module, as Receivings would already have the necessary logic to manage payments.
Screenshot 2024-12-29 at 20-45-07 Recepción - NioUI

andriux1990:
It would be great to discuss this over a call. Thank you.
Tonight, my goal is to finalize the Receivings Payments functionality. At the moment, I can already add and remove any type of payment, and the system correctly displays the change.

Next, I need to ensure that when a purchase is completed in Receivings, both the purchase and the associated payment are saved.

For the user interface, I need to properly organize where this functionality will fit best, and I’ll also need to make adjustments to the receipt, as it will now use the payment system.

If I manage to finish this tonight, I’ll dedicate tomorrow to conducting thorough testing.

jekkos:
Perhaps we can setup a call  Friday evening? Do you use gitter we can get in touch that way.

app/Controllers/Receivings.php
odiea: Shouldn't Complete be postComplete?

andriux1990:
https://github.com/user-attachments/assets/ebd07eca-c2bd-4cb1-a659-d62c739a5300

andriux1990:
> simplescreenrecorder-2025-01-01_19.25.50.mp4

In the receiving module, it's taken me longer than expected because the changes are quite extensive. Adding the payment types has required modifying several parts of the code. There are some fields in the receiving table that I think won't be needed anymore, and I'm considering removing them. I also added a field for the invoice expiration date, called 'Terms Days', which indicates the number of days the purchase invoice will be valid.

app/Controllers/Receivings.php
andriux1990: Hello @odiea,
postComplete already exists, and it is for direct purchases. This other complete is for items transfers (requisitions), in postRequisitionComplete.

andriux1990:
In the **ospos_receivings** module, I believe the `amount_change` and `amount_tendered` fields might no longer be necessary since those values are now managed in **ospos_receivings_payments**.
Screenshot 2025-01-01 at 19-42-32 localhost _ localhost _ mobile _ ospos_receivings phpMyAdmin 5 2 1
Screenshot 2025-01-01 at 19-44-27 localhost _ localhost _ mobile _ ospos_receivings_payments phpMyAdmin 5 2 1


However, @jekkos, this is just my opinion. You are the one who can best decide what to do with `amount_change` and `amount_tendered`. What do you think?

jekkos:
If we manage payments as we do in sales then indeed those columns can be dropped. I do wonder how much duplication there is if we set up a separate table only for receiving payments. How do we then configure the payment types for example? Do you reuse what is setup for sales or is it another config?

app/Controllers/Receivings.php
jekkos: @odiea is right, that is how it worked before. So complete was renamed to `postComplete` and then the method in the requisition was not updated, so it should have worked before the 3.4 upgrade.

app/Models/Receiving.php
```diff
@@ -187,14 +187,161 @@ public function save_value(array $items, int $supplier_id, int $employee_id, str
...
+				'trans_location' => $stock_source,
+				'trans_comment' => $recv_remarks . ' OUT',
+				'trans_inventory' => -$quantity,
+			];
+			$existing_source = $inventory->where([
+				'trans_items' => $item_id,
+				'trans_location' => $stock_source,
+				'trans_comment' => $recv_remarks . ' OUT'
+			])->countAllResults();
+			if ($existing_source === 0) {
```
jekkos, line 310: I don't fully grasp this check, we only update inventory in case there is no entry yet for this receiving id? In which situation would that be required?

app/Controllers/Receivings.php
```diff
@@ -414,6 +414,52 @@ public function postRequisitionComplete(): void
...
+
+		$data['cart'] = $this->receiving_lib->get_cart();
+		$data['total'] = $this->receiving_lib->get_total();
+		$data['transaction_time'] = to_datetime(time());
+		$data['mode'] = $this->receiving_lib->get_mode();
+		$data['comment'] = $this->receiving_lib->get_comment();
+		$data['reference'] = $this->receiving_lib->get_reference();
+		$data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
+		$data['stock_location_source'] = $this->receiving_lib->get_stock_source();
+		$data['stock_location_destination'] = $this->receiving_lib->get_stock_destination();
```
jekkos, line 435: It seems that most of this code is also in the `postComplete` method. I do like the fact that you use transactions to save the data to the inventory table.

app/Controllers/Receivings.php
objecttothis: For what it's worth, my understanding is that prepending post, get or other has to do with CodeIgniter 4's autorouting that we are using.  Any functions called directly and maybe AJAX calls I don't think need post or get prepended.

app/Models/Receiving.php
andriux1990: Hello. I noticed how Sales does it. There is some code that I think doesn't have a function, but since Sales has it, I copied it as well. However, I will review that part.

app/Controllers/Receivings.php
jekkos: For now I applied the obvious fix to master. I would like to review how it works currently and compare with these improvements

jekkos:
@andriux1990 I found that indeed the current requisition functionality in master is still broken. We probably need part of this PR to make it work. WIll rebase this on lastest master to try it out

odiea:
Currently in Receivings the function save now needs to be postSave and after that it is just little things that need fixed.

jekkos:
This PR was created from a cloned repo in the organisation. This is not how we typically work. I have ported the patch to a branch of the main repository and created a new PR from that. Also it was rebased on latest master.

I'll test these fixes myself first and will then edit them so they can be merged.

Issue #4144, opened 2025-01-10
Update Copyright Year Dynamically


chetan4921:
### Background information

The current footer contains a static copyright year: All rights reserved © 2010 - 2024

### Issue / Bug  / Question / New Feature

The copyright notice updates automatically each year.

PR #4146, opened 2025-01-15
Add default-timezone to update copyright

yashisrani:
Fixed: #4144 

### 1) Set the Timezone:

- The line date_default_timezone_set('UTC'); ensures the server uses the correct timezone for generating the current year.

### 2) Dynamic Year:

- The date('Y') function will now correctly fetch the year based on the updated timezone.

### DIFF
--- a/app/Views/partial/footer.php
+++ b/app/Views/partial/footer.php
@@ -2,6 +2,9 @@
 
 use Config\OSPOS;
 
+// Set the correct timezone
+date_default_timezone_set('UTC'); // Replace 'UTC' with your desired timezone, e.g., 'America/New_York'
+
 ?>
 </div>
 		</div>


app/Views/partial/footer.php
```diff
@@ -2,6 +2,9 @@
 
 use Config\OSPOS;
 
+// Set the correct timezone
+date_default_timezone_set('UTC'); // Replace 'UTC' with your desired timezone, e.g., 'America/New_York'
```
jekkos, line 6: the timezone needs to be set in OSPOS config directly

objecttothis: @yashisrani, Configuration > Localization > Timezone is the setting @jekkos is referring to.
image

If the footer year isn't set correctly, then the value you need to set for date_default_time_zone() is not a hardcoded zone, but rather the value from the database.  So something like `date_default_timezone_set($config['timezone'];` would be what you want.

objecttothis: Also, I'm not 100% sure that this is the best method to do this.  
I think app/Config/App.php might be better.  It has `$appTimeZone`.  Potentially you could still set it with `$config['timezone']` but you need to populate that array.  See how it's done in the code.
image

yashisrani: @objecttothis I decided to add the below code instead of "public string $appTimezone = 'UTC":

    public $config = [
        'timezone' => 'UTC',
        'locale' => 'en_US',
        'debug' => true,
    ];

    $config = config('App');
    date_default_timezone_set($config->config['timezone']);
    
    is it helpful or not ? please give a suggetions.

objecttothis: The $config[] array gets automatically populated by the application from the database. You need to set those values in settings (I showed you the screenshot), not hardcore them. If you do what you just showed me, the value will be overwritten by the database value.

objecttothis: Take a look at app\Config\OSPOS.php and you'll see the code responsible for populating the $config[] array. So, again, if there is a place relying on the timezone which isn't using the config array value for it then you need to submit a PR which uses that rather than hardcore any values.

Issue #4147, opened 2025-01-15
[Bug]: Requisitions are currently broken


jekkos:
### Bug Description?

Try to do a requisition from one stock location to another

### Steps to Reproduce?

Add an item and one extra stock location. Then create a receiving that moves the item from one location to another. Complete the receiving

### Expected Behavior?

The receipt is generated

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox

### Server Operating System and version

Arch Linux 6.6 LTS

### Database Management System and version

MariaDB 10.5

### Web Server and version

Apache 2.4

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
Probably the last one that needs a fix and testing

jekkos:
@diego-ramos if you would have the time to check this one, that would be great

jekkos:
Tested and fixed in master

PR #4148, opened 2025-01-15
Fix for requisitions (#4147)

jekkos:
@andriux1990 I have moved your work to this branch here and removed the clone repo.

### DIFF
--- a/app/Controllers/Receivings.php
+++ b/app/Controllers/Receivings.php
@@ -412,6 +412,52 @@ class Receivings extends Secure_Controller
 	}
 
 	/**
+	 * Complete a receiving requisition and save to the database.
+	 *
+	 * @throws ReflectionException
+	 * @noinspection PhpUnused
+	 */
+	public function complete(): void
+	{
+		$data = [];
+
+		$data['cart'] = $this->receiving_lib->get_cart();
+		$data['total'] = $this->receiving_lib->get_total();
+		$data['transaction_time'] = to_datetime(time());
+		$data['mode'] = $this->receiving_lib->get_mode();
+		$data['comment'] = $this->receiving_lib->get_comment();
+		$data['reference'] = $this->receiving_lib->get_reference();
+		$data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
+		$data['stock_location_source'] = $this->receiving_lib->get_stock_source();
+		$data['stock_location_destination'] = $this->receiving_lib->get_stock_destination();
+
+		$employee_id = $this->employee->get_logged_in_employee_info()->person_id;
+		$employee_info = $this->employee->get_info($employee_id);
+		$data['employee'] = $employee_info->first_name . ' ' . $employee_info->last_name;
+
+
+		if ($data['stock_location_source'] != $data['stock_location_destination']) {
+			// Save receiving to database
+			$data['receiving_id'] = 'REQ ' . $this->receiving->save_requisition($data['cart'], $employee_id, $data['comment'], $data['reference'], $data['stock_location_source'], $data['stock_location_destination']);
+
+			if ($data['receiving_id'] == 'REQ -1') {
+				$data['error_message'] = lang('Receivings.transaction_failed');
+			} else {
+				$data['barcode'] = $this->barcode_lib->generate_receipt_barcode($data['receiving_id']);
+			}
+
+			$data['print_after_sale'] = $this->receiving_lib->is_print_after_sale();
+
+			echo view("receivings/receipt", $data);
+
+			$this->receiving_lib->clear_all();
+		} else {
+			$data['error'] = lang('Receivings.error_requisition');
+			$this->_reload($data);
+		}
+	}
+
+	/**
 	 * Gets the receipt for a receiving. Used in app/Views/receivings/form.php
 	 *
 	 * @param $receiving_id
--- a/app/Controllers/Receivings.php
+++ b/app/Controllers/Receivings.php
@@ -422,70 +422,58 @@ class Receivings extends Secure_Controller
 	 */
 	public function postRequisitionComplete(): void
 	{
-		if($this->receiving_lib->get_stock_source() != $this->receiving_lib->get_stock_destination())
-		{
-			foreach($this->receiving_lib->get_cart() as $item)
-			{
-				$this->receiving_lib->delete_item($item['line']);
-				$this->receiving_lib->add_item($item['item_id'], $item['quantity'], $this->receiving_lib->get_stock_destination(), $item['discount_type']);
-				$this->receiving_lib->add_item($item['item_id'], -$item['quantity'], $this->receiving_lib->get_stock_source(), $item['discount_type']);
-			}
+	    $data = [];
 
-			$this->postComplete();
-		}
-		else
-		{
-			$data['error'] = lang('Receivings.error_requisition');
+	    // Check if stock locations are different
+	    if ($this->receiving_lib->get_stock_source() != $this->receiving_lib->get_stock_destination()) {
+	        // Process items in the cart
+	        foreach ($this->receiving_lib->get_cart() as $item) {
+	            $this->receiving_lib->delete_item($item['line']);
+	            $this->receiving_lib->add_item($item['item_id'], $item['quantity'], $this->receiving_lib->get_stock_destination(), $item['discount_type']);
+	            $this->receiving_lib->add_item($item['item_id'], -$item['quantity'], $this->receiving_lib->get_stock_source(), $item['discount_type']);
+	        }
 
-			$this->_reload($data);	//TODO: Hungarian notation
-		}
-	}
+	        // Prepare data to complete the requisition
+	        $data['cart'] = $this->receiving_lib->get_cart();
+	        $data['total'] = $this->receiving_lib->get_total();
+	        $data['transaction_time'] = to_datetime(time());
+	        $data['mode'] = $this->receiving_lib->get_mode();
+	        $data['comment'] = $this->receiving_lib->get_comment();
+	        $data['term_days'] = $this->receiving_lib->get_term_days();
+	        $data['reference'] = $this->receiving_lib->get_reference();
+	        $data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
+	        $data['stock_location_source'] = $this->receiving_lib->get_stock_source();
+	        $data['stock_location_destination'] = $this->receiving_lib->get_stock_destination();
 
-	/**
-	 * Complete a receiving requisition and save to the database.
-	 *
-	 * @throws ReflectionException
-	 * @noinspection PhpUnused
-	 */
-	public function complete(): void
-	{
-		$data = [];
+	        $employee_id = $this->employee->get_logged_in_employee_info()->person_id;
+	        $employee_info = $this->employee->get_info($employee_id);
+	        $data['employee'] = $employee_info->first_name . ' ' . $employee_info->last_name;
 
-		$data['cart'] = $this->receiving_lib->get_cart();
-		$data['total'] = $this->receiving_lib->get_total();
-		$data['transaction_time'] = to_datetime(time());
-		$data['mode'] = $this->receiving_lib->get_mode();
-		$data['comment'] = $this->receiving_lib->get_comment();
-		$data['term_days'] = $this->receiving_lib->get_term_days();
-		$data['reference'] = $this->receiving_lib->get_reference();
-		$data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
-		$data['stock_location_source'] = $this->receiving_lib->get_stock_source();
-		$data['stock_location_destination'] = $this->receiving_lib->get_stock_destination();
+	        // Save the requisition in the database
+	        $data['receiving_id'] = 'REQ ' . $this->receiving->save_requisition(
+	            $data['cart'],
+	            $employee_id,
+	            $data['comment'],
+	            $data['reference'],
+	            $data['stock_location_source'],
+	            $data['stock_location_destination']
+	        );
 
-		$employee_id = $this->employee->get_logged_in_employee_info()->person_id;
-		$employee_info = $this->employee->get_info($employee_id);
-		$data['employee'] = $employee_info->first_name . ' ' . $employee_info->last_name;
+	        if ($data['receiving_id'] == 'REQ -1') {
+	            $data['error_message'] = lang('Receivings.transaction_failed');
+	        } else {
+	            $data['barcode'] = $this->barcode_lib->generate_receipt_barcode($data['receiving_id']);
+	        }
 
+	        $data['print_after_sale'] = $this->receiving_lib->is_print_after_sale();
 
-		if ($data['stock_location_source'] != $data['stock_location_destination']) {
-			// Save receiving to database
-			$data['receiving_id'] = 'REQ ' . $this->receiving->save_requisition($data['cart'], $employee_id, $data['comment'], $data['reference'], $data['stock_location_source'], $data['stock_location_destination']);
+	        echo view("receivings/receipt", $data);
 
-			if ($data['receiving_id'] == 'REQ -1') {
-				$data['error_message'] = lang('Receivings.transaction_failed');
-			} else {
-				$data['barcode'] = $this->barcode_lib->generate_receipt_barcode($data['receiving_id']);
-			}
-
-			$data['print_after_sale'] = $this->receiving_lib->is_print_after_sale();
-
-			echo view("receivings/receipt", $data);
-
-			$this->receiving_lib->clear_all();
-		} else {
-			$data['error'] = lang('Receivings.error_requisition');
-			$this->_reload($data);
-		}
+	        $this->receiving_lib->clear_all();
+	    } else {
+	        $data['error'] = lang('Receivings.error_requisition');
+	        $this->_reload($data);
+	    }
 	}
 
 	/**
--- a/app/Controllers/Receivings.php
+++ b/app/Controllers/Receivings.php
@@ -213,10 +213,13 @@ class Receivings extends Secure_Controller
 
 		$price = parse_decimals($this->request->getPost('price'));
 		$quantity = parse_quantity($this->request->getPost('quantity'));
+		$discount = parse_decimals($this->request->getPost('discount'));
 		$raw_receiving_quantity = parse_quantity($this->request->getPost('receiving_quantity'));
 
 		$description = $this->request->getPost('description', FILTER_SANITIZE_FULL_SPECIAL_CHARS);	//TODO: Duplicated code
 		$serialnumber = $this->request->getPost('serialnumber', FILTER_SANITIZE_FULL_SPECIAL_CHARS) ?? '';
+		$price = filter_var($price, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION);
+		$quantity = filter_var($quantity, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION);
 		$discount_type = $this->request->getPost('discount_type', FILTER_SANITIZE_NUMBER_INT);
 		$discount = $discount_type
 			? parse_quantity(filter_var($this->request->getPost('discount'), FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION))
@@ -263,6 +266,7 @@ class Receivings extends Secure_Controller
 		$data['selected_supplier_name'] = !empty($receiving_info['supplier_id']) ? $receiving_info['company_name'] : '';
 		$data['selected_supplier_id'] = $receiving_info['supplier_id'];
 		$data['receiving_info'] = $receiving_info;
+		$balance_due = round($receiving_info['amount_due'] - $receiving_info['amount_tendered'] + $receiving_info['cash_refund'], totals_decimals(), PHP_ROUND_HALF_UP);
 
 		echo view('receivings/form', $data);
 	}
@@ -324,22 +328,26 @@ class Receivings extends Secure_Controller
 	 */
 	public function postComplete(): void
 	{
-
+		$amount_tendered = parse_decimals($this->request->getPost('amount_tendered'));
 		$data = [];
 
 		$data['cart'] = $this->receiving_lib->get_cart();
 		$data['total'] = $this->receiving_lib->get_total();
-		$data['transaction_time'] = to_datetime(time());
+		$data['transaction_time'] = date('Y-m-d H:i:s'); // Fecha y hora actuales
 		$data['mode'] = $this->receiving_lib->get_mode();
 		$data['comment'] = $this->receiving_lib->get_comment();
 		$data['reference'] = $this->receiving_lib->get_reference();
+		$data['term_days'] = $this->receiving_lib->get_term_days();
 		$data['payment_type'] = $this->request->getPost('payment_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+		$data['payments'] = $this->receiving_lib->get_payments(); // Obtener pagos del carrito
 		$data['show_stock_locations'] = $this->stock_location->show_locations('receivings');
 		$data['stock_location'] = $this->receiving_lib->get_stock_source();
-		if($this->request->getPost('amount_tendered') != null)
-		{
-			$data['amount_tendered'] = parse_decimals($this->request->getPost('amount_tendered'));
-			$data['amount_change'] = to_currency($data['amount_tendered'] - $data['total']);
+
+		if ($this->request->getPost('amount_tendered') != null) {
+			$data['amount_tendered'] = filter_var($amount_tendered, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION);
+			$data['amount_change'] = $data['amount_tendered'] - $data['total'];
+		} else {
+			$data['amount_change'] = 0;
 		}
 
 		$employee_id = $this->employee->get_logged_in_employee_info()->person_id;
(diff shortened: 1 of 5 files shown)


app/Models/Receiving.php
```diff
@@ -187,14 +187,161 @@ public function save_value(array $items, int $supplier_id, int $employee_id, str
...
+		}
+
+		// Update the total cost in ospos_receivings
+		$this->db->table('ospos_receivings')
+			->where('receiving_id', $receiving_id)
+			->update(['total' => $total_cost]);
+
+		if (!$this->db->transComplete()) {
+			log_message('error', 'Transaction failed for requisition: ' . print_r($receivings_data, true));
+			return -1;
```
objecttothis, line 340: @jekkos returning something like this is super common in the code so I'm not surprised that this is the way @andriux1990 did it here, but I wonder if we should move toward throwing exceptions and catching them in the calling function. This benefits error handling and logging.

objecttothis (commented):
Sorry, I'm a little slow. Can someone explain to me how requisitions are intended to be used in the app? Maybe it's just because the word is a very specific word that, in English, only gets used in business logistics settings. The dictionary definition is as follows:

1. A formal written request for something needed.
2. A necessity; a requirement.
3. The state or condition of being needed or put into service.

Is that what the code is doing? I see code to transfer quantity from one stock location to another. Is that requisition fulfilment?

andriux1990:
> @andriux1990 I have moved your work to this branch here and removed the clone repo.

Hello @jekkos I'm not an expert in Git; I only know how to use commands like git add, git reset, git commit, git push, and git pull. I plan to clone the branch from the repository and create a pull request with all the changes, including not only those related to item transfers between locations but also improvements to the different payment methods, which I already have prepared.

andriux1990:
-- phpMyAdmin SQL Dump
-- version 5.2.1
-- https://www.phpmyadmin.net/
--
-- Servidor: localhost
-- Tiempo de generación: 17-01-2025 a las 04:05:29
-- Versión del servidor: 10.11.10-MariaDB
-- Versión de PHP: 8.3.15

SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO";
START TRANSACTION;
SET time_zone = "+00:00";


/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
/*!40101 SET NAMES utf8mb4 */;

--
-- Base de datos: `mobile`
--

-- --------------------------------------------------------

--
-- Estructura de tabla para la tabla `ospos_receivings_payments`
--

CREATE TABLE `ospos_receivings_payments` (
  `payment_id` int(11) NOT NULL,
  `payment_type` varchar(40) NOT NULL,
  `payment_amount` decimal(15,2) NOT NULL,
  `cash_refund` decimal(15,2) NOT NULL DEFAULT 0.00,
  `employee_id` int(10) NOT NULL,
  `payment_time` datetime NOT NULL,
  `payment_ref` varchar(100) DEFAULT NULL,
  `original_amount` decimal(15,2) NOT NULL DEFAULT 0.00,
  `original_currency` varchar(10) DEFAULT NULL,
  `receiving_id` int(10) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;

--
-- Índices para tablas volcadas
--

--
-- Indices de la tabla `ospos_receivings_payments`
--
ALTER TABLE `ospos_receivings_payments`
  ADD PRIMARY KEY (`payment_id`,`payment_type`),
  ADD KEY `receiving_id` (`payment_id`);

--
-- AUTO_INCREMENT de las tablas volcadas
--

--
-- AUTO_INCREMENT de la tabla `ospos_receivings_payments`
--
ALTER TABLE `ospos_receivings_payments`
  MODIFY `payment_id` int(11) NOT NULL AUTO_INCREMENT;
COMMIT;

/*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT */;
/*!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS */;
/*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;

andriux1990:
> ## -- phpMyAdmin SQL Dump
> -- version 5.2.1
> -- https://www.phpmyadmin.net/
> 
> -- Servidor: localhost -- Tiempo de generación: 17-01-2025 a las 04:05:29 -- Versión del servidor: 10.11.10-MariaDB -- Versión de PHP: 8.3.15
> 
> SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO"; START TRANSACTION; SET time_zone = "+00:00";
> 
> /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT _/; /_!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS _/; /_!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION _/; /_!40101 SET NAMES utf8mb4 */;
> ## --
> -- Base de datos: `mobile`
> ## --
> -- Estructura de tabla para la tabla `ospos_receivings_payments`
> 
> CREATE TABLE `ospos_receivings_payments` ( `payment_id` int(11) NOT NULL, `payment_type` varchar(40) NOT NULL, `payment_amount` decimal(15,2) NOT NULL, `cash_refund` decimal(15,2) NOT NULL DEFAULT 0.00, `employee_id` int(10) NOT NULL, `payment_time` datetime NOT NULL, `payment_ref` varchar(100) DEFAULT NULL, `original_amount` decimal(15,2) NOT NULL DEFAULT 0.00, `original_currency` varchar(10) DEFAULT NULL, `receiving_id` int(10) NOT NULL ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;
> ## --
> -- Índices para tablas volcadas
> ## --
> -- Indices de la tabla `ospos_receivings_payments`
> 
> ALTER TABLE `ospos_receivings_payments` ADD PRIMARY KEY (`payment_id`,`payment_type`), ADD KEY `receiving_id` (`payment_id`);
> ## --
> -- AUTO_INCREMENT de las tablas volcadas
> ## --
> -- AUTO_INCREMENT de la tabla `ospos_receivings_payments`
> 
> ALTER TABLE `ospos_receivings_payments` MODIFY `payment_id` int(11) NOT NULL AUTO_INCREMENT; COMMIT;
> 
> /*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT _/; /_!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS _/; /_!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;

This is the structure of the payment methods for receivings, similar to the one used in sales.

Additionally, I added the payment_ref field, which will be used to store the reference code when payments are made via transfer and a reference identifier is required.

I also included the original_amount and original_currency fields, which are used to store information in an alternative currency. I added these for my own use, but they can be removed if needed. However, it might be useful to keep them in the code and the database.

I'm not sure if this SQL script should be applied as an additional migration or if it should be included in version 3.4.

andriux1990:
A question, what is this in the code that I see often: //TODO: Hungarian Notation

jekkos:
@andriux1990 better to try and learn a bit how to use git it will make contribution process a lot easier. Also I'm more in favor of splitting up PRs per topic, not throw everything on one pile as it will be hard to review and iterate on that decently

jekkos:
@objecttothis thanks for your comments on the naming. I don't remember how the term was introduced. The functionality might be simpler in the sense that it just transfers a stock quantity from one location to another one, so it might make sense to rename the receiving mode then. If you have any ideas, let me know. I'll try to think about this.

objecttothis:
From the way it was described to me, I would rename requisitions to stock transfer but maybe that's just because I'm not a logistics guy.

jekkos:
@andriux1990 the branch is installed on dev, I did a test and got following error. Any idea what the reason could be?
image

odiea:
Change string to ?string

andriux1990:
Hello, good afternoon.

I have updated the code to the latest master branch and uploaded the changes related to the receivings module. I’d appreciate it if you could review it so that, with your feedback, we can further simplify the code.

Currently, it is now possible to manage payment types and process purchases correctly. However, I’ve noticed an issue with the transfers that I just identified. I’ll work on fixing it and will do another git push.

Regarding payments, I believe we have addressed an important need in receivings, as it was previously not possible to process a purchase properly.

andriux1990:
I understand that this topic is related to transfers. However, implementing different payment types in purchases could also be impacted. Therefore, I believe we should test the payments, merge the code, and, of course, address the issues related to item transfers between locations.

andriux1990:
For payments to work correctly, it is necessary to add the SQL tables I shared in the conversation to the database.

andriux1990:
https://github.com/user-attachments/assets/bc9296fd-bf89-4dcd-b182-6a005a2227ea

jekkos:
@andriux1990 new tables or SQL changes should be integrated in the project using a database migration. This way we do not need to ask users to run SQL manually on an upgrade, the application will do it for them. 
For this release I just wanted to have the stock transfer functionality fixed as now we suddenly have to test a lot more things as new features are introduced. Once the release is done then we can add new payment system, but it would be nice to split it up at this point if that is not too much asked.

jekkos:
Thanks for the video. The functionality in itself looks good but it would be nice to have a consistent experience adding payments in both sales/receivings. So if we revamp the UI in one module, then I'd opt to reuse that in the other, if possible.

andriux1990:
I propose addressing both transfers and payment methods simultaneously since the current issue with transfers is due to the recent integration of payment methods. By tackling both aspects at the same time, we can prevent recurring problems and move forward more quickly. Additionally, we will only need to perform a few tests to ensure everything works correctly.

andriux1990:
I fully understand that a migration can be done, but I uploaded the code because I have been quite busy and thought someone else could take care of the migration while I continue working on the payment methods and transfers in 'receiving'.

andriux1990:
The issue was that, when completing the requisition, the postComplete function from the purchases module was being called instead of the one corresponding to the transfers module.

To resolve this, once I identified the error, I decided to merge the code. This way, the completeRequisition function no longer needs to rely on another function. I combined both functionalities into one, eliminating the need for additional calls.

objecttothis:
> @andriux1990 new tables or SQL changes should be integrated in the project using a database migration. This way we do not need to ask users to run SQL manually on an upgrade, the application will do it for them. 
> For this release I just wanted to have the stock transfer functionality fixed as now we suddenly have to test a lot more things as new features are introduced. Once the release is done then we can add new payment system, but it would be nice to split it up at this point if that is not too much asked.

Since 3.4.0 is feature frozen and the stock transfer fix has multiple feature changes, let's include this in 3.4.1.

objecttothis:
@andriux1990 the master is no longer feature frozen.  Can you rebase on the current master so we can proceed with your changes?

andriux1990:
Hi @objecttothis, alright, I’ll upload the changes. But now I’m facing another issue: I no longer want the backend and frontend of OSPOS to be so tightly coupled. Having everything mixed together prevents me from making the changes I want.

Changing the topic a bit—when there’s an update, we run into the problem of having to update the entire system, which is a lot of work. That’s why I prefer to keep things separated.

I’ve already managed to decouple the items, sales, people, and expenses tables, using the same approach as before.

Currently, OSPOS uses this default API for items:
http://localhost:8080/items/search?search=&offset=0&limit=45&start_date=2010-01-01&end_date=2025-04-02&stock_location=1

The data rendered in the HTML relies on a specific version of Bootstrap and the CSS/JS from OpenSourcePOS. Without those, it doesn't work properly. Personally, I find it very limiting to be tied to those dependencies if I want to use something more modern.

That’s why I developed a new API:
http://localhost:8080/api/items?limit=10&offset=0&search=monitor&stock_location=2

It does pretty much the same thing, but the loading speed is insanely fast. Plus, it doesn’t rely on opensourcepos.js, which allows me to add charts and richer data to the items module.

WebShells:
I agree with @objecttothis about renaming requisition to stock transfer, it is more familiar/heard to users.

jekkos:
We do have rest endpoint already @andriux1990 ? The data for items is returned in json format. Agree that beside table data there might be other things to convert. But at least the table rendering is decoupled from the datamodel

andriux1990:
> We do have rest endpoint already @andriux1990 ? The data for items is returned in json format. Agree that beside table data there might be other things to convert. But at least the table rendering is decoupled from the datamodel

Yes, it's true that we already have REST endpoints returning data in JSON format. However, some of those APIs (like the one for items) include fields like:

"edit": "<\/span><\/a>"

This edit field contains HTML, for over from a button or link. It only includes closing tags like </span></a> without any opening structure or actual content.

Additionally, embedding raw HTML directly in JSON:

Breaks the separation of data and presentation, which goes against RESTful and clean design principles.

Makes it harder to maintain or style from the frontend, especially when the structure needs to change or adapt across views or frameworks.

It would be much cleaner to return only raw data from the API and let the frontend handle the rendering (e.g., generating edit buttons with JavaScript based on item IDs).

We should consider cleaning up that field or removing it entirely from the API response.

andriux1990:
Here I’m using the API without tabular_helper — this is a 100% pure JSON-based API.
It allows more flexibility, since tabular_helper makes it difficult to pass data to the view, and it's tedious to rely on a specific version of jQuery.
OSPOS depends on opensourcepos.js, and tables won’t function properly without that file.

By retrieving the data directly from the API through a pure endpoint, it becomes easier to improve the UI design and work with any version of Bootstrap without limitations.

Attached is a demo video as an example.


https://github.com/user-attachments/assets/32548d8b-0858-4fb9-a673-13f6144005ff

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
+
+		$data['cart'] = $this->receiving_lib->get_cart();
+		$data['total'] = $this->receiving_lib->get_total();
+		$data['transaction_time'] = date('Y-m-d H:i:s'); // Fecha y hora actuales
+		$data['mode'] = $this->receiving_lib->get_mode();
+		$data['comment'] = $this->receiving_lib->get_comment();
+		$data['reference'] = $this->receiving_lib->get_reference();
+		$data['term_days'] = $this->receiving_lib->get_term_days();
+		$data['payment_type'] = $this->request->getPost('payment_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+		$data['payments'] = $this->receiving_lib->get_payments(); // Obtener pagos del carrito
```
objecttothis, line 342: Comments, if they are necessary, need to be in English

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
+			$data['comment'],
+			$data['reference'],
+			$data['term_days'],
+			$data['payment_type']
+		);
+
+		if ($receiving_id == -1) {
+			$data['error_message'] = lang('Receivings.transaction_failed');
+		} else {
+			$data['receiving_id'] = $receiving_id;
```
objecttothis, line 387: Variables in new code need to be PSR-12 compliant naming

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
+
+			// Calcular el cambio (cash_refund)
+			$total_payments = array_sum(array_column($data['payments'], 'payment_amount'));
+			$cash_refund = max(0, $total_payments - $data['total']);
+
+			// Guardar los pagos en la base de datos
```
objecttothis, line 393: Comments need to be in English.  Same above and below

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
-            $supplier_info = $this->supplier->get_info($supplier_id);
-            $data['supplier'] = $supplier_info->company_name;
-            $data['first_name'] = $supplier_info->first_name;
-            $data['last_name'] = $supplier_info->last_name;
-            $data['supplier_email'] = $supplier_info->email;
-            $data['supplier_address'] = $supplier_info->address_1;
-            if (!empty($supplier_info->zip) or !empty($supplier_info->city)) {
-                $data['supplier_location'] = $supplier_info->zip . ' ' . $supplier_info->city;
-            } else {
-                $data['supplier_location'] = '';
```
objecttothis, line 487: If we are going to remove these TODO comments we either need to make sure there is an issue in GitHub for the technical debt or we need to resolve the issue so that it doesn't get forgotten about. Otherwise IMO we should leave the todo there.

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
+		$this->_reload();
+	}
+
+	public function postAddPayment(): void
+	{
+		$data = [];
+		$payment_type = $this->request->getPost('payment_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+	
+		// Validar el tipo de pago
+		if ($payment_type !== lang('Sales.giftcard')) {
```
objecttothis, line 649: This validation section should be refactored into a private method and called in the if statement on line 657 as something like `if (!giftcardNumberValidated($params))`.  This helps maintain SRP

app/Controllers/Receivings.php
```diff
@@ -17,506 +17,690 @@
...
+		}
+	
+		if (!$this->validate($rules, $messages)) {
+			$data['error'] = $payment_type === lang('Sales.giftcard')
+				? lang('Sales.must_enter_numeric_giftcard')
+				: lang('Sales.must_enter_numeric');
+		} else {
+			if ($payment_type === lang('Sales.giftcard')) {
+				$amount_tendered = parse_decimals($this->request->getPost('amount_tendered'));
+				$giftcard_num = filter_var($amount_tendered, FILTER_SANITIZE_FULL_SPECIAL_CHARS, FILTER_FLAG_ALLOW_FRACTION);
```
objecttothis, line 664: All these local variables need to be camelCase for new code

objecttothis, line 664: Sanitizing should be done on output, not input. We should be validating here.  1. This entire block doing validation should be refactored into a private request validation function. It's an SRP violation otherwise. 2. When validating the Giftcard Number we need to simply check to make sure that it's an acceptable value.  We can do this with regex or with ctype_digit()

app/Libraries/Receiving_lib.php
```diff
@@ -19,493 +21,763 @@
...
+	 */
+	public function clear_reference(): void	//TODO: This function verb is inconsistent from the others.  Consider refactoring to remove_reference()
+	{
+		$this->session->remove('recv_reference');
+	}
+
+	/**
+	 * @return string
+	 */
+	public function get_term_days(): string
```
objecttothis, line 204: camelCase function names

app/Libraries/Receiving_lib.php
```diff
@@ -19,493 +21,763 @@
...
+	{
+		return $this->session->get('recv_term_days') ?? '';
+	}
+
+	/**
```
objecttothis, line 209: No useless PHPdocs. If the function name is clear what it's doing and there is no complex return or parameter logic, we should remove these.

app/Libraries/Receiving_lib.php
```diff
@@ -19,493 +21,763 @@
...
+		return $this->get_total() - $this->get_payments_total();
+	}
+
+	public function get_payments_total(): string
+	{
+		$subtotal = '0.0';
+		$cash_mode_eligible = CASH_MODE_TRUE;
+
+		foreach($this->get_payments() as $payments)
+		{
```
objecttothis, line 616: These curly braces need to be put on the same lines as their opening code for PSR compliance

app/Libraries/Receiving_lib.php
```diff
@@ -19,493 +21,763 @@
...
+	 * Adds a new payment to the payments array or updates an existing one.
+	 * It will also disable cash_mode if a non-qualifying payment type is added.
+	 * @param string $payment_id
+	 * @param string $payment_amount
+	 * @param int $cash_adjustment
+	 */
+	public function add_payment(string $payment_id, string $payment_amount, int $cash_adjustment = CASH_ADJUSTMENT_FALSE): void
+	{
+		$payments = $this->get_payments();
+		if(isset($payments[$payment_id]))
```
objecttothis, line 666: Space after `if` and before `(` per PSR rules

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+		$builder = $this->db->table('receivings');
+		$builder->where('receiving_id', $receiving_id);
+
+		return $builder->update($receiving_data);
+	}
+
+	/**
+	 * @throws ReflectionException
+	 */
+	public function save_value(array $items, int $supplier_id, int $employee_id, string $comment, string $reference, string $term_days, string $payment_type, int $receiving_id = NEW_ENTRY): int	//TODO: $receiving_id gets overwritten before it's evaluated. It doesn't make sense to pass this here.
```
objecttothis, line 107: variable names in new code needs to be camelCase.  Might be easier to refactor all the parameter values here.

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+				'trans_items' => $item_data['item_id'],
+				'trans_user' => $employee_id,
+				'trans_location' => $item_data['item_location'],
+				'trans_comment' => $recv_remarks,
+				'trans_inventory' => $items_received
+			];
+
+			$inventory->insert($inv_data, false);
+			$attribute->copy_attribute_links($item_data['item_id'], 'receiving_id', $receiving_id);
+		//	$supplier = $supplier->get_info($supplier_id);	//TODO: supplier is never used after this.
```
objecttothis, line 191: If $supplier isn't used we need to remove the line.  This might be the only use of $supplier_id, so we need to remove the parameter from the function signature and calling functions.

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+		return $this->db->transStatus() ? $receiving_id : -1;
+	}
+
+	/**
+	 * Save a requisition to the `ospos_receivings` and `ospos_receivings_items` tables
+	 * and update stock quantities in the respective locations.
+	 *
+	 * @throws ReflectionException
+	 */
+	public function save_requisition(array $items, int $employee_id, string $comment, string $reference, int $stock_source, int $stock_destination): int
```
objecttothis, line 205: function names and variable names in new code must be camelCase

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+		$this->db->transComplete();
+
+		return $this->db->transStatus() ? $receiving_id : -1;
+	}
+
+	/**
+	 * Save a requisition to the `ospos_receivings` and `ospos_receivings_items` tables
+	 * and update stock quantities in the respective locations.
+	 *
+	 * @throws ReflectionException
```
objecttothis, line 203: This PHPdoc is missing descriptions of the parameters and return value.  It's possible we need to explain the return value especially so those writing calling code know what to do with the value without having to read all the code.

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+	 */
+	public function save_requisition(array $items, int $employee_id, string $comment, string $reference, int $stock_source, int $stock_destination): int
+	{
+		$inventory = model('Inventory');
+		$item_quantity_model = model(Item_quantity::class);
+
+		if (count($items) == 0) {
+			return -1; // No items to save
```
objecttothis, line 211: I think we may have a global constant for this to use instead of -1

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+		}
+
+		// Prepare the receiving data
+		$receivings_data = [
+			'receiving_time' => date('Y-m-d H:i:s'),
+			'employee_id' => $employee_id,
+			'comment' => $comment,
+			'reference' => $reference,
+			'total' => 0, // This will be updated later
```
objecttothis, line 220: Updated later to what?

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+			'amount_change' => 0,
+			'amount_tendered' => 0,
+			'operation_type' => 'Requisition',
+		];
+
+		// Run these queries as a transaction
```
objecttothis, line 226: comment not needed.  The code speaks for itself

app/Models/Receiving.php
```diff
@@ -12,344 +12,509 @@
...
+				'item_location' => $item_data['item_location'] ?? 1,
+				'receiving_quantity' => $item_data['receiving_quantity'] ?? 1,
+			];
+		
+			$builder->insert($receivings_items_data);
+		
+			if (!$this->db->affectedRows()) {
+				log_message('error', 'Failed to insert requisition item: ' . print_r($receivings_items_data, true));
+				$this->db->transRollback();
+				return -1;
```
objecttothis, line 265: @jekkos I think we are doing elsewhere but on error we shouldn't be returning -1 but false because it's clearer that is an error. Alternately creating a global constant `const ERROR = -1` would make the code here clearer, but false is better because it's obvious in the calling code that if they didn't get a primary key id but got `false` instead that it was an error.

app/Views/receivings/receipt.php
```diff
@@ -15,121 +15,156 @@
...
-                    <td>
-                        <div class="total-value"><?= to_currency($amount_tendered) ?></div>
-                    </td>
-                </tr>
+			<?php if(isset($amount_change))
+			{
+			?>
+				<tr>
+					<td colspan="3" style='text-align:right;'><?= lang('Sales.amount_tendered') ?></td>
+					<td><div class="total-value">0.00</div></td>
```
objecttothis, line 146: I think we still need to call to_currency() here.  This hard coded value assumes `.` is the correct decimal separator but not all locales use that.

app/Views/receivings/receiving.php
```diff
@@ -17,528 +17,653 @@
...
+						<th style="width: 20%;"><?= lang(ucfirst($controller_name) .'.payment_amount') ?></th>
+					</tr>
+				</thead>
+
+				<tbody id="payment_contents">
+	              <?php
+	              $payment_types = [];
+
+	              foreach($payments as $payment_id => $payment)
+	              {
```
objecttothis, line 408: This curly brace needs to be on the previous line and a space between foreach and the (

objecttothis (changes requested):

app/Views/receivings/receiving.php
jekkos: @objecttothis personally I would like to avoid having to do review comments like this. I woukd suggest to move on with the linter so that it is automated and violations are detected automatically.

Also i will try to add an agents.md that contains these a reference to these rules so its applied automatically

objecttothis: @jekkos Yes, that's fine.  We can enable the linter.

objecttothis: @jekkos does the linter only fail on changes in code or does it fail if there is non-compliant code anywhere in the file?  If the latter, that will likely mean needing to refactor the entire file when we make a change to one part of the file.  In that case it might be better to update and merge #4281

objecttothis:
@andriux1990 @jekkos if this needs to be released in 3.4.2 then let's get on finishing the fixes so we can release.  We have a significant backlog of features to merge because we are waiting on 3.4.2.

jekkos:
There is still some things that need to be  looked into here. Not sure if @andriux1990 is up to it?

PR #4150, opened 2025-01-17
Fix send email and email receipt

diego-ramos:


app/Views/sales/receipt_email.php
```diff
@@ -21,7 +21,7 @@
 		{
 		?>
 			<div id="company_name">
-				<img id="image" src="data:image/png;base64,<?= base64_encode(file_get_contents('uploads/' . $config['company_logo'])) ?>" alt="company_logo" />
```
jekkos, line 24: can you elaborate a bit on why you changed this to used an external URL? If we embed the image, then doesn't it always render in the email itself or is this blocked by security controls?

diego-ramos: Hi Jekkos, I found that with embeded ( base64 encoded images ) web based emaill clients not render / blocked the images and the email size will be affected by it too while the img tag will not.

jekkos (approved):

app/Views/sales/receipt_email.php
```diff
@@ -21,7 +21,7 @@
...
+				<img id="image" src="<?= base_url('uploads/' . esc($config['company_logo'], 'url')) ?>" alt="company_logo" />
```
jekkos, line 24: @diego-ramos I thought about this a bit further. An installation might not always be running publicly on the internet. So in that case it will be better to send the logo as an embedded image.

Once a user opens the email he has to be able to load the logo from a public URL. I would only use this URL if we know that the server was setup on a public domain, probably when it's using https.

jekkos (changes requested):

app/Views/sales/receipt_email.php
jekkos: I have copied your branch to the organisation's repo and will deploy it now, so I can test it. If it's fine then I'll merge the email fixes. I'll probably hold the image change back for now

diego-ramos: Hi, I get your point  just keep in mind that most of the people ( buyersm that are outside of the company) this days use a web based email, so maybe instead of using the embeedded image you just should remove it if the installation is in a private server, I think is better to use a setting for this, instead of "guessing" if the server is public or not based on the protocol, so you can choose to use the public img, the embeded one or none  in that case

diego-ramos: Or just send the receipt as the invoice as a PDF file, I think this is a better and simplier solutions

diego-ramos:
@jekkos Hi, I just added a new commit with email invoice fixes

app/Views/sales/receipt_email.php
jekkos: Correct the invoice already uses PDF. Let's do that for the receipt then as well? But this is also a separate change I will leave that code as it is for now

app/Views/sales/invoice_email.php
```diff
@@ -208,7 +207,7 @@
 			<?= nl2br($config['return_policy']) ?>
 		</div>
 		<div id='barcode'>
-			<img alt='<?= esc($barcode, 'attr') ?>' src='data:image/png;base64,<?= esc($barcode, 'attr') ?>' /><br>
+			<img alt=<?='Barcode - '.$sale_id ?> src='data:image/svg+xml;base64,<?= base64_encode($barcode) ?>' /><br>
```
jekkos, line 210: I'll remove this prefix as the barcode can otherwise not be scanned in the register for return

jekkos:
Will merge from the other PR I opened.

Issue #4176, opened 2025-02-09
[Bug]: Cannot delete an Attribute from a dropdown list


odiea:
### Bug Description?

Go to attributes. Open an dropdown and try deleting an added attribute. Displays it is deleted but it isn't . The following error shows in the log.
CRITICAL - 2025-02-09 13:28:21 --> [Caused by] CodeIgniter\Database\Exceptions\DatabaseException: Cannot delete or update a parent row: a foreign key constraint fails (`ospos15`.`ospos_attribute_links`, CONSTRAINT `ospos_attribute_links_ibfk_2` FOREIGN KEY (`attribute_id`) REFERENCES `ospos_attribute_values` (`attribute_id`))

### Steps to Reproduce?

Same as above

### Expected Behavior?

Attribute should be removed from a dropdown list.

### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.3

### What browsers are you seeing the problem on?

Google Firefox Dev Demo
### Server Operating System and version

Win 10

### Database Management System and version

Wamp Server

### Web Server and version

Wamp Server

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [x] I agree this copy has not been modified

jekkos:
https://stackoverflow.com/questions/6881905/deleting-from-multiple-tables-with-foreign-constraints

jekkos:
We probably need to define a cascading delete on this foreign key in the link table and then remove the value from the values only.

jekkos:
@objecttothis I see that these keys used to be configured ith cascading delete enabled. In a 3.4 migration script they are dropped and recreated with `ON DELETE RESTRICT' ., which I believe is causing the regression here. Would like to check with you what the reason was behind this change.

objecttothis:
I'd have to look at the code to confirm but my first guess is that it was changed as part of an attempt to fix the issue where duplicate attribute links were being created. When I was messing with constraints that was the main issue at least. I suppose this is another vote for unit tests.

I have to take an oral exam today to renew our permanent residency here so I can't look at it until after that. Our server chaos is also dying down so I should be more available soon. I wonder if the commit messages were preserved when I committed that change. That might give a clue.

jekkos:
@objecttothis good luck with the exam,do you need to speak the native language there then? Quite surprised to hear this, here in Belgium foreigners are even not obliged to learn the official language, although things are changing and becoming more strict.

[...]

PR #4179, opened 2025-02-12
Fix attribute dropdown delete (#4176)

jekkos:
@objecttothis this migration adds back the cascading delete to fix the dropdown delete issue.

### DIFF
--- a/app/Controllers/Attributes.php
+++ b/app/Controllers/Attributes.php
@@ -231,14 +231,15 @@ class Attributes extends Secure_Controller
 	}
 
 	/**
-	 * AJAX called function to delete an attribute value. This is never called in the code. Perhaps it was boiler plate code that just isn't needed?
-	 * @param int $attribute_id
+	 * AJAX called function to delete an attribute value. This is called when a dropdown item is removed.
+	 *
+	 * @param string $attribute_value
 	 * @return bool
 	 * @noinspection PhpUnused
 	 */
-	public function delete_value(int $attribute_id): bool	//TODO: This function appears to never be used in the codebase.  Is it needed?
+	public function delete_value(string $attribute_value): bool
 	{
-		return $this->attribute->delete_value($attribute_id, NO_DEFINITION_ID);
+		return $this->attribute->delete_value($attribute_value, NO_DEFINITION_ID);
 	}
 
 	/**
--- a/app/Database/Migrations/20250213000000_fix_attributes_cascading_delete.php
+++ b/app/Database/Migrations/20250213000000_fix_attributes_cascading_delete.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+use App\Models\Attribute;
+use CodeIgniter\Database\ResultInterface;
+
+class Migration_Attributes_fix_cascading_delete extends Migration
+{
+	/**
+	 * Perform a migration step.
+	 */
+	public function up(): void
+	{
+		helper('migration');
+		drop_foreign_key_constraints(['ospos_attribute_links_ibfk_1', 'ospos_attribute_links_ibfk_2'], 'ospos_attribute_links');
+		$this->db->query("ALTER TABLE `ospos_attribute_links` ADD CONSTRAINT `ospos_attribute_links_ibfk_1` FOREIGN KEY (`definition_id`) REFERENCES `ospos_attribute_definitions` (`definition_id`) ON DELETE CASCADE;");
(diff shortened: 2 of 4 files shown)


app/Controllers/Attributes.php
```diff
@@ -231,14 +231,15 @@ public function getView(int $definition_id = NO_DEFINITION_ID): void
...
-	 * AJAX called function to delete an attribute value. This is never called in the code. Perhaps it was boiler plate code that just isn't needed?
-	 * @param int $attribute_id
+	 * AJAX called function to delete an attribute value. This is called when a dropdown item is removed.
+	 *
+	 * @param string $attribute_value
 	 * @return bool
 	 * @noinspection PhpUnused
 	 */
-	public function delete_value(int $attribute_id): bool	//TODO: This function appears to never be used in the codebase.  Is it needed?
+	public function delete_value(string $attribute_value): bool
```
objecttothis, line 240: Good catch here.  I'm not sure why it was accepting an attribute_id but the model function definitely is looking for a value, not an id.

app/Database/Migrations/20250213000000_fix_attributes_cascading_delete.php
```diff
@@ -0,0 +1,29 @@
...
+use CodeIgniter\Database\ResultInterface;
+
+class Migration_Attributes_fix_cascading_delete extends Migration
+{
+	/**
+	 * Perform a migration step.
+	 */
+	public function up(): void
+	{
+		$this->db->query("ALTER TABLE `ospos_attribute_links` DROP FOREIGN KEY `ospos_attribute_links_ibfk_1`");
```
objecttothis, line 16: The only think about this is that if, for some reason, the foreign key constraint `ospos_attribute_links_ibfk_1` or `ospos_attribute_links_ibfk_2` doesn't exist, then this will throw an error. This shouldn't happen, but the safer way to prevent that AND correct the problem would be to wrap the `DROP` part in an if statement to check to see if the constraint exists before trying to drop it.  Take a look at the drop_foreign_key_constraints() function in https://github.com/opensourcepos/opensourcepos/blob/master/app/Database/Migrations/20240826000000_fix_duplicate_attributes.php for an example of how we are doing that programmatically elsewhere.

objecttothis (approved):
I'm approving it but I fairly strongly recommend we add the if statements around the DROP statements in case someone's database has a problem.

app/Controllers/Attributes.php
jekkos: In the debugger it passed an id so I changed the variable name

app/Database/Migrations/20250213000000_fix_attributes_cascading_delete.php
jekkos: OK will adapt it but indeed, I don't expect that to happen.

jekkos: @objecttothis how would you suggest to reuse the code in the migration? I do not just want to duplicate it. Perhaps a db helper can be created so this functionality can be shared..wdyt

objecttothis: Sure, you could create a helper and refactor the function below into it.  Then pass in a string[] with the foreign keys to drop and the tablle to drop them from.  Then we would need to call that function in your migration. Something like this should do it:
```
	private function drop_foreign_key_constraints(string[] $foreignKeys, string $table): void
	{
		$current_prefix = $this->db->getPrefix();
		$this->db->setPrefix('');
		$database_name = $this->db->database;

		foreach ($foreignKeys as $fk)
		{
			$builder = $this->db->table('INFORMATION_SCHEMA.TABLE_CONSTRAINTS');
			$builder->select('CONSTRAINT_NAME');
			$builder->where('TABLE_SCHEMA', $database_name);
			$builder->where('TABLE_NAME', $table);
			$builder->where('CONSTRAINT_TYPE', 'FOREIGN KEY');
			$builder->where('CONSTRAINT_NAME', $fk);
			$query = $builder->get();

			if($query->getNumRows() > 0)
			{
				$this->db->query("ALTER TABLE `$table` DROP FOREIGN KEY `$fk`");
			}
		}

		$this->db->setPrefix($current_prefix);
	}
```

You could go back and retroactively change the fix_duplicate_attributes migration too but if you want to take a strict "don't touch any past migrations unless it's a breaking change" policy then it can be left as is too.

jekkos:
The constraint drop logic was moved to the migration helper . It's not reused in both migrations.

PR #4195, opened 2025-03-26
fix: added step 6 (dependencies) INSTALL.md

enokby:
sound necessaru to me.


jekkos:
Thanks for the suggestion, but INSTALL is different from BUILD. Dependencies should already be in the zip file when you download it.

enokby:
Ok thank you for the clarirification.

PR #4196, opened 2025-03-28
Feature: PSR-12 Compliant Indentation

objecttothis:
PSR-12 requires the use of 4 spaces rather than tabs for indenting lines of code.  This PR addresses only that by replacing tabs with 4 spaces.  This will allow us to use a PSR-12 linter without workarounds for indentation differences.


BudsieBuds:
I noticed that the current `.editorconfig` settings don't seem to enforce our 4-space indentation standard. Should it be updated to ensure consistency across all contributors?

jekkos (approved):

.env
```diff
@@ -47,7 +47,7 @@ database.tests.port = 3306
 email.SMTPHost = ''
 email.SMTPUser = ''
 email.SMTPPass = ''
-email.SMTPPort = 
+email.SMTPPort =
```
jekkos, line 50: not sure if this should be empty

objecttothis: It's an integer so I think that's why it's not single quoted

objecttothis:
> I noticed that the current `.editorconfig` settings don't seem to enforce our 4-space indentation standard. Should it be updated to ensure consistency across all contributors?

Correct, we should update the .editorconfig to enforce 4 space indentation.  You can include it in a commit in one of your other enhancement branches, include it in its own PR or include it in the CodeIgniter 4.6.0 bump branch.

jekkos:
Any change that goes in now should have the indentation checked. I will merge the github actions for this

PR #4197, opened 2025-03-28
Feature bump ci to 4.6.0

objecttothis:
bump CodeIgniter to 4.6.0


BudsieBuds:
I've added a commit to indicate support for PHP 8.4 after updating to CI 4.6.x. However the linter is failing, maybe just updating the CI version is not enough for 8.4 support?

objecttothis:
> I've added a commit to indicate support for PHP 8.4 after updating to CI 4.6.x. However the linter is failing, maybe just updating the CI version is not enough for 8.4 support?

If you look at the log from the phplint run on PHP 8.4 they are all deprecation warnings, so the linter might be a little aggressive.  I don't know if we can configure it to pass on warnings or not.  That said, in its current configuration it won't pass 8.4 tests unless we also change the code referenced in those deprecation warnings.  I don't want to do it in this PR though.

objecttothis:
@BudsieBuds and @jekkos I think we should turn off the PHP 8.4 linter for this one because it's failing on deprecation warnings.

jekkos:
Maybe it's best to enable the 8.4 linter and fix issues in one go. If that out of scope for this PR then let's disable it again

objecttothis:
I agree that we should make it 8.4 compatible but in a separate PR.

BudsieBuds:
You can delete my commit and merge. It'll remove the linter, and the documentation will still say PHP 8.1 - 8.3.

jekkos:
Some conflicts to resolve here now

objecttothis:
@jekkos I resolved the conflicts and made a minor change.  This PR should be ready for review.  It's now no longer showing so many file changes.

jekkos:
Can you get rid of the merge commits in this branch? I'd like to keep master history clean.

jekkos:
Does this PR now introduce support for 8.4? The linter is not completely happy yet. Also cool feature this copilot review.

objecttothis:
> Can you get rid of the merge commits in this branch? I'd like to keep master history clean.

@jekkos I think you are referring to these two commits in this branch
image

Those were generated by github when I resolved conflicts resulting from merging the two PRs generated by dependabot to bump versions.  It won't let me merge the branch without resolving the conflicts related to that and a few other files, so I think I need some commit in there, though the naming is not helpful.

Our standard procedure is to Squash and Merge.  Before Merging it allows me to edit the commit message and I could replace the wording with more descriptive wording.  Is that what you mean by getting rid of the merge commits to keep commit history clean?

The only other thing I can think of is to delete the commits then rebase, but my understanding is that when you have multiple people making commits to the same branch, you should merge rather than rebase.  Let me know if editing the commit message during the squash is acceptable or not.

objecttothis:
> Does this PR now introduce support for 8.4? The linter is not completely happy yet. Also cool feature this copilot review.

CodeIgniter 4.6.0 is PHP 8.4 compatible (https://codeigniter.com/user_guide/intro/requirements.html) but if you look at the linter, all the errors are deprecation warnings, so the problem is with the code, not CI.  We either need to tell PHP Lint to report but not fail on deprecation warnings or turn off the 8.4 linter until we update the code.  Either way I think upgrading the code to 8.4 compliance should be a separate PR than this one.  I created issue #4200

objecttothis:
@jekkos what do you want to do here?  Can we turn off the PHP 8.4 linter or allow deprecation warnings but not errors?

objecttothis:
OK, for now I'm just going to merge so I can start on the events branch.  We can create another PR to get rid of the deprecation warnings.  I created #4200 for that.

Issue #4209, opened 2025-04-11
[Feature]: Adding Kurdish language


OmerQadirFM:
### Type of Feature

✨ New Feature

### OpensourcePOS Version

opensourcepos 3.3.9

### Description

Hello, my name is Omer Qadir, and I am from Sulaymaniyah city in Iraqi Kurdistan. My native language is Central Kurdish (Sorani Kurdish). I would like to contribute by translating the project (opensourcepos) into my native language. Kindly consider adding it.

Contact me
https://x.com/OmerQadirFM
https://wa.me/9647501943525

### Verify you searched open requests in OpensourcePOS

- [x] I agree I have searched Open Requests

jekkos:
Hi there,you should ve able To add the language in weblate directly. If not then I can have a look

jekkos:
I have added [Kurdish (Central, Iraq)](https://translate.opensourcepos.org/languages/ckb_IQ/) , you should be able to start translating the common file now. Let me know if that works.

OmerQadirFM:
I have completed translating common component
why there is no other component to translate?
 and when we will be able to use OSPOS in Kurdish?

![Image](https://github.com/user-attachments/assets/c07b5cb2-72c6-4cdb-bb38-bb178bfb3ea2)

jekkos:
Dear , you should ve able to add translation for the other components as well. Once the translations are submitted a unstable zip will be generated .

OmerQadirFM:
![Image](https://github.com/user-attachments/assets/84060fcf-966e-4f5c-addc-b182eab5ecf7)

OmerQadirFM:
I do not have a permission to add new translation!

jekkos:
I have added new components yesterday, they should be available now

OmerQadirFM:
Now how can I add Kurdish to the languages list to I will be able use OSPOS in Kurdish?
witch file can I modify? 
I already upload php files in the folder ckb_IQ to the folder app/Language

![Image](https://github.com/user-attachments/assets/9bf05c9b-a731-48d1-a9d0-6e31ce2a83b1)

BudsieBuds:
If #4210 is merged, it will be selectable.

PR #4210, opened 2025-04-13
Add Kurdish language option to UI

BudsieBuds:
Makes the Kurdish language selectable in the UI. As requested here: #4209.

@jekkos I do wonder why this language was added as ckb_IQ instead of the ISO 639-1 standard used for all other languages, in this case being `ku`. If you decide to change this for consistency, you should also change the following:

app/Config/App.php
`ckb_IQ',` to `'ku',`

app/Helpers/locale_helper.php
`'ckb_IQ:centralkurdish' => 'Central Kurdish (Iraq)',` to `'ku:kurdish' => 'Kurdish',`

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -162,6 +162,7 @@ class App extends BaseConfig
         'az',
         'bg',
         'bs',
+        'ckb_IQ',
         'cs',
         'da',
         'de-CH',
--- a/app/Helpers/locale_helper.php
+++ b/app/Helpers/locale_helper.php
@@ -64,6 +64,7 @@ function get_languages(): array
         'az:azerbaijani' => 'Azerbaijani',
         'bg:bulgarian' => 'Bulgarian',
         'bs:bosnian' => 'Bosnian',
+		'ckb_IQ:centralkurdish' => 'Central Kurdish (Iraq)',
         'cs:czech' => 'Czech',
         'da:danish' => 'Danish',
         'de-CH:german' => 'German (Switzerland)',


BudsieBuds:
As is, this errors when the language is selected. Probably because of the underscore in the language name.

jekkos:
Thanks for the PR. I added the language in weblate , so it is probably a setting there. Would need to check if it is able to use what you suggest here instead

BudsieBuds:
Like this it works and follows the ISO naming convention. Changed from `ckb_IQ` to `ku` fixed the error. Had to add the empty keys to solve new errors. Also added some missing language files.

Do you have to change something in Weblate for this to show up? Or does it auto update if merged?

jekkos:
Thanks for the fixes. Now I can pull the code back into weblate. Then I'll need to check if the files end up in the correct place.

BudsieBuds:
@jekkos I overlooked this and assumed it was Kurdish. After some research, I found that adding the country code (ISO 3166) isn’t necessary. Since Central Kurdish lacks a two-letter code under ISO 639-1, I've switched to using the three-letter code from ISO 639-3 instead. The country code may be added in the future if there are more variants. Updated for merging here: #4211. 👍

OmerQadirFM:
Kurds primarily speak three different dialects:  
kmr – Northern Kurdish (it might be worth considering including this if someone requests it),  
ckb – Central Kurdish (this is the dialect I speak),  
sdh – Southern Kurdish. 

As for country codes, they are not necessary since the same dialects are spoken across different countries.

# Conclusion
kmr Kurdish (Northern)
ckb - Kurdish (Central)

OmerQadirFM:
To ensure easier finding, change it from Central Kurdish to Kurdish (Central).

image

BudsieBuds:
I chose the term Central Kurdish because it is used by the Language Coding Agency for ISO 639-3. See here: https://iso639-3.sil.org/code/ckb

OmerQadirFM:
Initially, when switching to Kurdish, users often scroll down to the letter K to find it. It would be much easier to rename it as Kurdish (Central).

BudsieBuds:
I reexamined the naming conventions and have updated several language names to enhance discoverability. Changes will be live if #4218 is merged.

PR #4211, opened 2025-04-15
Convert language ku to ckb

BudsieBuds:
Following #4210.

- convert ku (Kurdish) to ckb (Central Kurdish)
- replaced tabs with spaces
- replace single quotation marks with double


jekkos:
Thanks for the modifications

PR #4223, opened 2025-04-19
Enhance license handling

BudsieBuds:
- Implement license auto-updates to ensure information is always current
- Consolidate all NPM licenses into a single file for easier viewing
- Render license text in a monospace font to enhance clarity and formatting consistency
- Eliminate the outdated Bower license generation code


app/Controllers/Config.php
```diff
@@ -106,17 +107,20 @@ private function _licenses(): array    //TODO: remove hungarian notation.  Super
                     } else {
                         $license[$i]['text'] = $license_text_file . ' file is missing';
                     }
-                } elseif ($fileinfo->getBasename() == 'bower.LICENSES') {
-                    // set a flag to indicate that the JS Plugin bower.LICENSES file is available and needs to be attached at the end
-                    $bower = true;
                 } elseif ($fileinfo->getBasename() == 'composer.LICENSES') {
```
jekkos, line 110: this code could be simplified a bit, quite a couple of branches to check very sim!ilar things.

BudsieBuds: Yeah, my PHP skills are not the greatest. It's worth taking another look at in the future.

PR #4225, opened 2025-04-20
Some bug fixes

BudsieBuds:
1. Replace the login page CSS with the unminified version. Gulp isn’t minifying it, so changes weren’t being applied. Proper minification can be integrated later.
2. Adjust container max width to Bootstrap 5’s container-xxl. Currently the width can exceed the viewport. It looks like the numbers got mixed up.
3. Add RTL CSS to the Bootstrap theme to align with how Bootswatch does it.


jekkos:
Thanks again for the fixes @BudsieBuds . Any idea how the rtl should be enabled now we have the CSS? I guess we need to load it in case an RTL language is configured

BudsieBuds:
@jekkos Yes, I've looked into this. Unfortunately, it's only included with Bootstrap 5 (as an experimental feature). For now, it could only be implemented on the login screen, which doesn't seem worthwhile.

PR #4229, opened 2025-04-21
Fix null error when sending by email a receipt of a sale that has no invoice

diego-ramos:
Fix null error when sending by email a receipt of a sale that has no invoice


jekkos:
@diego-ramos adding the token with an empty string works fine?

diego-ramos:
Yes I tested several times,

jekkos:
Thanks for the fix, merged

PR #4201, opened 2025-04-02
Coding style updates for PSR-12 compliance & improved readability

BudsieBuds:
This PR started as a minor update to the `.editorconfig` but evolved into a refactor to enhance the adherence to PSR-12 standards and more, following #4196. Including:
- An updated .editorconfig to enforce consistent styling across the codebase
- Adjustments to spacing and other formatting elements to align with best practices
- Overall enhancements aimed at improving code readability and maintainability

While I've carefully checked everything, I could use an extra set of eyes to read over the changes.

PR #4230, opened 2025-04-21
Bugfix constraint migration fix

objecttothis:
- Migrate Sessions table to prevent problems with migrating populated databases.
- Refactor a few function names for PSR-12 compliance.
- Fix bug in migration causing failure due to trying to drop indexes with associated foreign key constraints. This bug was causing 3.3.2 databases to fail migration. #4202 
- migration to change constraint back to ON DELETE RESTRICT
20250213000000_fix_attributes_cascading_delete.php reverted a change on the ospos_attribute_links_ibfk_1 and ospos_attribute_links_ibfk_2 constraints on the ospos_attribute_links table to fix a bug with not being able to delete a dropdown value. The problem is that these need to be ON DELETE RESTRICT in order to use generated_unique_column and attribute_links_uq3 to enforce no duplication of attribute links.  The solution to this (in this PR) is to change them back to ON DELETE RESTRICT and programmatically cascade delete before trying to remove an entry.  This way we can enforce uniqueness AND deleting dropdown rows.
- Add indices to ospos_inventory and ospos_attribute_links which were causing the items view queries to time out when attribute_links was large and the hardware wasn't fast.

closing #4237
closing #4251


app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
```diff
@@ -54,33 +54,91 @@ public function down(): void
             . ' REFERENCES ' . $this->db->prefixTable('sales_items') . ' (sale_id)');
     }
 
-    private function create_primary_key(string $table, string $index): void
+    private function createPrimaryKey(string $table, string $index): void
     {
         $result = $this->db->query('SELECT 1 FROM information_schema.columns WHERE table_schema = DATABASE() AND table_name= \'' . $this->db->getPrefix() . "$table' AND column_key = '$index'");
```
jekkos, line 59: this method looks very generic, could it be moved to the db helper?

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
```diff
@@ -54,33 +54,91 @@ public function down(): void
...
+            }
+        }
+
+        return $deletedConstraints;
     }
 
-    private function index_exists(string $table, string $index): bool
+    private function deleteIndex(string $table, string $index): void
+    {
+        if ($this->indexExists($table, $index)) {
```
jekkos, line 121: same with this one, I bet we can reuse those in other migrations

app/Database/Migrations/20250213000000_fix_attributes_cascading_delete.php
```diff
@@ -14,7 +14,12 @@ class Migration_Attributes_fix_cascading_delete extends Migration
     public function up(): void
     {
         helper('migration');
+
+        $this->db->query("ALTER TABLE `ospos_attribute_links` DROP INDEX `attribute_links_uq3`");
```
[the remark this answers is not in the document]
objecttothis: I hear you, but anyone with MySQL would be permanently unable to upgrade their version of OSPOS because this migration fails on those systems without this statement. I suppose the way to make sure that everyone ends up with the same database schema would be another migration which either drops it for everyone or adds it back in for people who migrated through this change.  I am working on a code change which will add the the index back in and revert the constraint changes made here to ON DELETE RESTRICT but without the bug of not being able to delete a dropdown selection item.  The short version of this is that ON DELETE RESTRICT is needed on the constraints in order to enforce uniqueness to attribute_links, so to still get the effects of ON DELETE CASCADE we have to programmatically implement ON DELETE CASCADE.

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
objecttothis: Possibly.  I'll look into it.

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
objecttothis: Sounds good.  I'll look into it.

app/Controllers/Attributes.php
```diff
@@ -230,18 +230,6 @@ public function getView(int $definition_id = NO_DEFINITION_ID): void
         echo view('attributes/form', $data);
     }
 
-    /**
```
objecttothis, line 233: This method wasn't actually called anywhere.

app/Database/Migrations/20210422000001_remove_duplicate_links.php
```diff
@@ -35,6 +35,7 @@ private function migrate_duplicate_attribute_links(): void
         $builder->select('item_id, definition_id, attribute_id, COUNT(*) as count');
         $builder->where('sale_id', null);
         $builder->where('receiving_id', null);
+        $builder->where('item_id IS NOT NULL');
```
objecttothis, line 38: This is a retroactive fix because existing OSPOS MySQL databases can't get past it otherwise.

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
objecttothis: I'm still going to extract this.  The rest of the code is ready for review.

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
objecttothis: I'll do this one also.  I believe we already have a migration helper where it can go.

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_1`;
```
objecttothis, line 1: @jekkos regarding your objection above about retroactively changing the migration to drop attribute_links_uq3 and how MariaDB users will have already run that, the way to fix it is by this migration.
There are two scenarios: MySQL DBMS and MariaDB DBMS.  Those on the MySQL DBMS will not have generated_unique_column when it hits line 6 in this migration and people with MariaDB DBMS will have it.  I haven't written it yet, but I need to write the php up() function that calls this sql to instead programmatically drop the generated_unique_column if it exists so that we don't get an error on MariaDB databases that the column exists.

app/Controllers/Attributes.php
jekkos: Beware it might be in use but called from the frontend. I might have cone across it when I fixed the bug where we were not able to delete an attribute droodown value.

jekkos (changes requested):

.editorconfig
BudsieBuds, line 1: These changes should probably be removed, superseded by #4201.

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
jekkos: That sounds good

app/Database/Migrations/20210422000001_remove_duplicate_links.php
jekkos: Thanks for that

app/Views/attributes/form.php
```diff
@@ -181,7 +181,7 @@
         }
         else
         {
-            $.post('<?= esc("$controller_name/delete_attribute_value/") ?>', {definition_id: definition_id, attribute_value: value});
+            $.post('<?= esc("$controller_name/DeleteDropdownAttributeValue/") ?>', {definition_id: definition_id, attribute_value: value});
```
jekkos, line 184: Does it work with the Delete first char uppercase?

objecttothis: I believe so.  I'll test it.  I was following the capitalization of the function name which is I believe is PostDeleteDropdownAttributeValue().  It's possible the router may not care.  In this case should we just make it all lowercase?

jekkos: I though it was camelcase from the view but I could be wrong

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
objecttothis: Done

app/Controllers/Attributes.php
objecttothis: PHPStorm also doesn't think it's called anywhere but it wouldn't find AJAX calls anyway.  String Searches don't find it.

app/Views/attributes/form.php
objecttothis: PascalCase from the view is working as it is.  It might be case insensitive, but IMO PascalCase matches the actual function name most precisely.

app/Models/Attribute.php
```diff
@@ -41,6 +41,38 @@ class Attribute extends Model
     public const SHOW_IN_SALES = 2;
     public const SHOW_IN_RECEIVINGS = 4;
 
+    // ==============
```
objecttothis, line 44: @jekkos I would like to move toward organizing the models with these section headings.  Right now the models are pretty disorganized.  IDEs like PHPStorm can help but I'd still like to move things around.  C# has `#region foo` but PHP doesn't have anything like that.  What do you think?

app/Models/Attribute.php
```diff
@@ -987,4 +1025,38 @@ public function get_attributes_by_definition(int $definition_id): array
 
         return $builder->get()->getResultArray();
     }
+
+// ==============
```
objecttothis, line 1029: This heading would be for private Model functions.

objecttothis:
@jekkos, I think the commit history is a little rough after rebasing on the master.  When I merge, per usual I'm planning to do a squash and merge with a commit message of this to clean everything up:

```
- Refactored function names for PSR-12 compliance
- Programmatically cascade delete attribute_link rows when a dropdown attribute is deleted but leave attribute_link rows associated with transactions.
- Added `WHERE item_id IS NOT NULL` to migration to prevent failure on MySQL databases during migration
- Retroactive correction of migration to prevent MySQL databases from failing.
- Refactored generic functions to helper
- Reverted attribute_links foreign key to ON DELETE RESTRICT which is required for a unique constraint on this table. Cascading deletes are now handled programmatically.
- Migration Session table to match Code Igniter 4.6
- Add index to attribute_links to prevent query timeout in items view on large databases
- Added overridePrefix() function to the migration_helper. Any time QueryBuilder is adding a prefix to the query when we don't want it to, this query can be used to override the prefix then set it back after you're done.
- Added dropAllForeignKeyConstraints() helper function.
- Added deleteIndex() helper function.
- Added indexExists() helper function.
- Added primaryKeyExists() helper function.
- Added recreateForeignKeyConstraints() helper function.
- Added CRUD section headings to the Attribute model.
- Replaced `==` with `===` to prevent type juggling.
- Removed unused delete_value function.
- Reworked deleteDefinition() and deleteDefinitionList() functions to delete rows from the attribute_links table which are associated.
- Added deleteAttributeLinksByDefinitionId() function
```

This PR is now ready for review.

objecttothis:
@jekkos the PHPLinter is failing on PHP 8.4 deprecations (https://github.com/opensourcepos/opensourcepos/actions/runs/15184079188/job/42700305258).  I do think we need to make these changes to remove deprecated code, but that needs to be a separate PR from this one.

app/Controllers/Attributes.php
jekkos: Ok but let's check if the dropdown delete still works then after

objecttothis: I checked it in my testing and it still worked to delete a drop-down choice. I confirmed also that it did not delete attribute link rows attached to sales or receivings. The cascading delete of items with that drop-down choice also works.

I don't think I tested deleting an entire drop-down attribute so we need to test that.

jekkos:
We can keep the full history in the squashed commit message body. So typically a good commit message has a short first line describing what is is about, and then more details can be added after the first newline. 

I find it very useful to add info to the history this way, it's a bit of an ongoing documentation with every change that is done.

app/Models/Attribute.php
objecttothis: I see your point.  PHPStorm does have a 'structure' window which shows the function signatures and this is very helpful, but as you can see from the sale model below that the list appears in the order that the function appears in the model, which is a hot mess.  This means that if I'm looking to see if we have a specific function, I often have to scan the entire list to find it.  I don't feel extremely strongly about using comments to separate sections, but it really would be useful if our classes all followed a similar ordering. Something like:
Fields
Constructors
Create
Read
Update
Delete
Private
Static
image

Thoughts?

app/Models/Attribute.php
```diff
@@ -842,29 +874,16 @@ public function saveAttributeValue(string $attribute_value, int $definition_id,
         return $attribute_id;
     }
 
-    /**
-     * @param string $attribute_value
-     * @param int $definition_id
-     * @return bool
-     */
-    public function delete_value(string $attribute_value, int $definition_id): bool
```
jekkos, line 850: @objecttothis I double checked the code and found that this method is in use in the controller `Attributes.php:83`

app/Models/Attribute.php
jekkos: I'd try to code to structure things logically.. having to add comments might be a code smell (according to clean code guidelines). It might just indicate naming is not good or classes are not sticking to single resonsibility principle.

app/Views/attributes/form.php
```diff
@@ -160,7 +160,7 @@
             if (is_new) {
                 values.splice($.inArray(value, values), 1);
             } else {
-                $.post('<?= esc("$controller_name/delete_attribute_value/") ?>', {
+                $.post('<?= esc("$controller_name/DeleteDropdownAttributeValue/") ?>', {
```
jekkos, line 163: It's this one that calls the method you removed here

app/Database/Migrations/20250522000000_AttributeLinksUniqueConstraint.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Perform a migration step.
+     */
+    public function up(): void
+    {
+        error_log('Migrating Attribute Links Unique Constraint');
+
+        execute_script(APPPATH . 'Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql');
+
+        error_log('Migrating Attribute Links Unique Constraint');
```
jekkos, line 19: Maybe one message is suffiicient?

app/Models/Attribute.php
jekkos: Ah I see you replaced it with a another one, probably fine

jekkos (dismissed):
I'm not going to block the PR here at this point, I know these things take time to test. But I do have some minor remarks on the usage of comments and eventual removal of a duplicated log line. Beside that I think it looks good.

app/Database/Migrations/20250522000000_AttributeLinksUniqueConstraint.php
```diff
@@ -0,0 +1,28 @@
...
+
+class AttributeLinksUniqueConstraint extends Migration
+{
+    /**
+     * Perform a migration step.
+     */
+    public function up(): void
+    {
+        error_log('Migrating Attribute Links Unique Constraint');
+
```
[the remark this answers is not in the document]
jekkos: Great thanks for resolving this!

objecttothis:
@jekkos in the morning I'll remove the comments and extra log lines. I think I was just mimicking what was in other migrations.

I'm OK with just ordering class functions logically, but I think we should agree on an order and move functions to those locations as we work on the code. Doing it all here would make the commit pretty difficult to parse.

I'll squash and merge after pushing those commits.

app/Models/Attribute.php
jekkos: Agree I would try to stick to an order in which we define method type or fields then instead of using comments. It's more consistent in our case and indeed, we have a lot of methods in a controller class. So better to deal with it this way

objecttothis:
@jekkos I was taught a class order of

Fields
Constructors 
Create
Read
Update
Delete
Private methods

Does that work or a variation of it? If not feel free to propose a different order. I just get tired of wasting time not knowing exactly where a function of a certain time would generally be in a model.

PR #4232, opened 2025-04-22
Fix Codeigniter disallowed characters error with payment types that have accents

diego-ramos:
This should be a temporary fix, as is a better solution to have some kind of enum with ids and descrption, but this will be  a bigger change.


jekkos:
Hi problem with this fix is that existing data for payment types will need to be migrated to base64. So it's not backward compatible

diego-ramos:
Hi @jekkos, sorry for the late response, I just see your comment, but you are not right, if you check my code I'm not storing anything with base64, I'm just taking the stored "id" that is not a real id, is the string of the payment type ( in the language selected, that is what is causing the error ) and then I encoded on the frontend template to use it on the url to send the data without errors to the controller, then I decode the frontend data and continue from there, there is no need to migrate data.

jekkos:
You are right, this is only to store data in the session. Any reason you did not use url encoding?

diego-ramos:
> You are right, this is only to store data in the session. Any reason you did not use url encoding?


Because encoding it throws same exception (the problem is when the url is decoded by codeigniter), I tried several options, and finally this one is the only one I found that works.

jekkos (approved):

PR #4249, opened 2025-05-13
Update INSTALL.md

developersteve:
Minor change to the docs which referenced `/app/Database/database.sql` which does not exist, updated it to `tables.sql`.


jekkos:
I guess you found the database SQL in the zip file in the end?

PR #4250, opened 2025-05-15
Fix item number lookup in sales/receivings (#4212)

jekkos:


objecttothis:
Isn't item_id an int in the database?

jekkos:
That's correct , the code is a bit messy. It uses item_id to store both item number and id. So it can be string or int. I can rename the variable to make that clear

objecttothis:
OK that makes sense. another way, in addition to renaming is to make the data type of the parameter a union type. IMO this makes it clearer that both types are accepted.  e.g., `foo( string|int bar): bool`

objecttothis:
Also I think we said that as we modify the code we are going to refactor function and variable names from snake_case to camel case for PSR-12 compliance. Are you up for that refactor in these few sections of code?

jekkos:
I just checked PSR12 and it recommends to have function names in camelCase, but there is no recommendation for variables.

objecttothis:
PSR12 does not explicitly require camelCase variable names, and when referencing database fields, it's not even possible, but I argue that where possible, we should agree to variable names using camelCase for consistency with the naming convention for methods. A second argument is that this matches CI4 coding conventions so all the files we inherit from CodeIgniter will be camelCase variables. A third argument is a guess that I haven't verified, but I won't be surprised if IDE linters complain about variable names not being camelCase.

I know this isn't the same as a specific declaration, but are there any variable names in the PSR12 OR PSR1 example code that show camelCase or snake_case?

jekkos:
I would like to release a 3.4.1 soon as many fixes made it to master already. would be nice to have the mysql one as well. Then we can do more cleanup after

objecttothis:
That is fine to punt it to another PR.

objecttothis (dismissed):

objecttothis:
> I would like to release a 3.4.1 soon as many fixes made it to master already. would be nice to have the mysql one as well. Then we can do more cleanup after

Without the migrations fixes that I have a PR out for, anyone running MySQL will still not be able to upgrade.  I'm almost done with my PR.  I'll try to get it to a stopping point tomorrow but I could really use someone testing the changes with MariaDB.  The only two changes I have left are to add the migration file to kick off one of the migration SQL's and then test my code changes to make sure we can still delete a dropdown.  I'm traveling June 9 - August 20, so I too want to get this into a good place to release 3.4.1, but I don't think we should release anything that MySQL users can't install.

jekkos:
Agree let's try to get this sorted. If you want just push the changes to a branch here in the repo and it will auto deploy. Then we can call for some testing. We use Mariadb for the dev server so it will be clear if it starts correctly from scratch or not. It runs all the migrations after every deploy so that is a good starting point. Still need to give you acces to it at some point.

 Sounds like you will have some nice time oiff then June onwards, where are you heading?

objecttothis:
> Agree let's try to get this sorted. If you want just push the changes to a branch here in the repo and it will auto deploy.

I changed #4230 from a draft to a ready pull request and pushed changes, but it says it hasn't deployed.  I wonder if because you had requested changes a few weeks back it doesn't deploy.

jekkos:
Ah no the github status is not linked to the dev server deployment, I never looked into that in fact. The deploy is done when travis pushes the final docker image for a commit. It uses a webhook in docker hub if I'm not mistaken.
You can check which commit is deployed by clicking the version string in the footer on the dev server.

jekkos:
https://github.com/opensourcepos/opensourcepos/commit/cae921 Currently this commit is deployed, but it's on MariaDB. The migrations seem to have ran

objecttothis:
> @objecttothis so it's ready for review, if you want.

I'll take a look

app/Libraries/Receiving_lib.php
```diff
@@ -269,20 +269,16 @@ public function clear_stock_destination(): void
      * @param bool $include_deleted
      * @return bool
      */
-    public function add_item(int $item_id, int $quantity = 1, ?int $item_location = null, float $discount = 0, int $discount_type = 0, ?float $price = null, ?string $description = null, ?string $serialnumber = null, ?float $receiving_quantity = null, ?int $receiving_id = null, bool $include_deleted = false): bool
+    public function add_item(string &$item_id, int $quantity = 1, ?int $item_location = null, float $discount = 0, int $discount_type = 0, ?float $price = null, ?string $description = null, ?string $serialnumber = null, ?float $receiving_quantity = null, ?int $receiving_id = null, bool $include_deleted = false): string
```
objecttothis, line 272: All the return statements in this function return a boolean.  Why does it need to return a string?

objecttothis (dismissed):
Just the one comment about changing the return from bool to string.  Unless there is a good reason, I would change it back.  I think we also talked about refactoring the naming of item_id in functions where it could be an item_id or an item_number to a variable name that makes it more clear that it could be either.  Maybe $itemReference or something similar?

jekkos:
It's quite cool, I asked copilot to convert the method to camelCase, gave me the converted version. Also found that VSCode might be able to go one step further and there you can integrate an AI agent that can then even do the update itself. The plugin is called cline.

objecttothis:
> It's quite cool, I asked copilot to convert the method to camelCase, gave me the converted version. Also found that VSCode might be able to go one step further and there you can integrate an AI agent that can then even do the update itself. The plugin is called cline.

Does it only refactor the local variables or does it also refactor function names? I've been nervous to ask AI to do the refactoring out of fear that it won't refactor the function definition or all the usages. That and PHPStorm seems to do a pretty good job at the refactoring... Just not automatic.

jekkos:
Indeed it renamed the function but not the references to it. In this case that was not much of an issue as it was one function only

Issue #4259, opened 2025-06-12
[Bug]: Error when deleting a payment in the Register


odiea:
### Bug Description?

CRITICAL - 2025-06-12 15:07:09 --> CodeIgniter\HTTP\Exceptions\BadRequestException: The URI you submitted has disallowed characters: "Q2FzaA=="
[Method: GET, Route: sales/deletePayment/Q2FzaA==]
in SYSTEMPATH\Router\Router.php on line 739.
 1 SYSTEMPATH\Router\Router.php(207): CodeIgniter\Router\Router->checkDisallowedChars('sales/deletePayment/Q2FzaA==')
 2 SYSTEMPATH\CodeIgniter.php(832): CodeIgniter\Router\Router->handle('sales/deletePayment/Q2FzaA==')
 3 SYSTEMPATH\CodeIgniter.php(455): CodeIgniter\CodeIgniter->tryToRouteIt(Object(CodeIgniter\Router\RouteCollection))
 4 SYSTEMPATH\CodeIgniter.php(354): CodeIgniter\CodeIgniter->handleRequest(null, Object(Config\Cache), false)
 5 SYSTEMPATH\Boot.php(334): CodeIgniter\CodeIgniter->run()
 6 SYSTEMPATH\Boot.php(67): CodeIgniter\Boot::runCodeIgniter(Object(CodeIgniter\CodeIgniter))
 7 FCPATH\index.php(56): CodeIgniter\Boot::bootWeb(Object(Config\Paths))

### Steps to Reproduce?

Using the register add a payment amount then try removing the amount. Payment should be removed.

### Expected Behavior?

Using the register add a payment amount then try removing the amount. Payment should be removed.
This also shows on the online  Demos.
### OpensourcePOS Version

development (unreleased)

### Php version

Php 8.3

### What browsers are you seeing the problem on?

Chrome

### Server Operating System and version

Win 10

### Database Management System and version

Wamp Server

### Web Server and version

Wamp Server

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [x] I agree this copy has not been modified

odiea:
base64_encode($payment_id) in Register and Controller is the issue. After removing  base64_encode the error is not there.

jekkos:
@diego-ramos is this something you have also encountered?

diego-ramos:
[...]

PR #4261, opened 2025-06-14
Create a  Base64 URL-Safe encoding and decoding helper

diego-ramos:
Fix [Bug]: Error when deleting a payment in the Register #4259

### DIFF
--- a/app/Controllers/Sales.php
+++ b/app/Controllers/Sales.php
@@ -466,7 +466,9 @@ class Sales extends Secure_Controller
      */
     public function getDeletePayment(string $payment_id): void
     {
-        $this->sale_lib->delete_payment(base64_decode($payment_id));
+        helper('web');
+    
+        $this->sale_lib->delete_payment(base64url_decode($payment_id));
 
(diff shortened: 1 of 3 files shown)


odiea:
I copied this commit and the remove payment issue has been resolved. Thanks

jekkos:
@diego-ramos perhaps we can just use https://www.php.net/manual/en/function.urlencode.php for url encoding? Or why do you add a custom method for this? Better to use builtin functionailty otherwise we need to maintain this code ourselves

jekkos:
I'll look into it, perhaps a unit test would be nice. Tried this briefly and I was able to add and run one from PhpStorm directly.

diego-ramos:
> I'll look into it, perhaps a unit test would be nice. Tried this briefly and I was able to add and run one from PhpStorm directly.

Sorry for late response about this, I don´t use PhpStorm, and as far I can see there is no php unit tests set on the project, I'm wrong?

jekkos:
Superseeded by #4274

PR #4262, opened 2025-06-14
Fix migration errors

diego-ramos:
Fix Migration 20250522000000_AttributeLinksUniqueConstraint issues

### DIFF
--- a/app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
+++ b/app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
@@ -1,8 +1,10 @@
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_1`;
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_2`;
+ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT IF EXISTS `ospos_attribute_links_ibfk_1`;
+ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT IF EXISTS `ospos_attribute_links_ibfk_2`;
 
 # Prevents duplicate attribute links with the same definition_id and item_id.
 # This accounts for dropdown rows (null item_id) and rows associated with sales or receivings.
+ALTER TABLE `ospos_attribute_links` DROP COLUMN IF EXISTS `generated_unique_column`;
+
 ALTER TABLE `ospos_attribute_links`
     ADD COLUMN `generated_unique_column` VARCHAR(255) GENERATED ALWAYS AS (
         CASE
--- a/app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
+++ b/app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
@@ -1,38 +1,3 @@
--- Drop FK ospos_attribute_links_ibfk_1 if it exists
-SET @fk := (
-  SELECT CONSTRAINT_NAME
-  FROM information_schema.REFERENTIAL_CONSTRAINTS
-  WHERE CONSTRAINT_SCHEMA = DATABASE()
-    AND TABLE_NAME = 'ospos_attribute_links'
-    AND CONSTRAINT_NAME = 'ospos_attribute_links_ibfk_1'
-);
-
-SET @sql := IF(@fk IS NOT NULL,
-  'ALTER TABLE `ospos_attribute_links` DROP FOREIGN KEY `ospos_attribute_links_ibfk_1`',
-  'DO 0'
-);
-PREPARE stmt FROM @sql;
-EXECUTE stmt;
-DEALLOCATE PREPARE stmt;
-
-
--- Drop FK ospos_attribute_links_ibfk_2 if it exists
-SET @fk := (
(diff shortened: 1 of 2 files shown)


jekkos:
Looks good to me . @objecttothis would you understand why we could not have the constraint here in this case?

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
```diff
@@ -1,8 +1,10 @@
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_1`;
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_2`;
+ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT IF EXISTS `ospos_attribute_links_ibfk_1`;
```
objecttothis: This should be tested in MySQL and MariaDB. To make sure it works in both. I think it will work, but we should make sure there are no regressions.

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
```diff
@@ -1,8 +1,10 @@
...
+ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT IF EXISTS `ospos_attribute_links_ibfk_2`;
```
objecttothis: @jekkos the constraints have to be dropped here because they are using ON DELETE CASCADE but need to be ON DELETE RESTRICT in order to prevent duplicate rows. The generated column cannot be created in MySQL when the constraints are ON DELETE CASCADE but it works in MariaDB. To maintain compatibility, we drop, recreate as ON DELETE RESTRICT then programmatically implemented the cascading delete to maintain the functionality we had before. Basically it's because of a difference between MySQL and MariaDB.

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
jekkos: I tried to run this migration both on mariadb and mysql, it works for a clean db.

jekkos:
I'm not sure if it works on MySQL. I have pulled in the docker image for this fix and ran it against Mysql 8.0.39. I'm seeing these errors

```
[Fri Jun 20 23:14:52.120111 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] error: Cannot add foreign key constraint, referer: https://dev.opensourcepos.org/login
[Fri Jun 20 23:14:52.121656 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] error: 1826, referer: https://dev.opensourcepos.org/login
[Fri Jun 20 23:14:52.121921 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] error: Duplicate foreign key constraint name 'ospos_attribute_links_ibfk_1', referer: https://dev.opensourcepos.org/login
[Fri Jun 20 23:14:52.123266 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] error: 1826, referer: https://dev.opensourcepos.org/login
[Fri Jun 20 23:14:52.123424 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] error: Duplicate foreign key constraint name 'ospos_attribute_links_ibfk_2', referer: https://dev.opensourcepos.org/login                                                                      
[Fri Jun 20 23:14:52.123508 2025] [php:notice] [pid 19:tid 19] [client 172.19.0.1:49114] Migrated to constraint, referer: https://dev.opensourcepos.org/login
```

and the generated column does not exist in the attribute_links table.

objecttothis:
> Seems indeed it might not [work on Mysql](https://stackoverflow.com/questions/17161496/drop-foreign-key-only-if-it-exists)

Yeah, IF EXIST might be need to be done programmatically. We already have helper functions in the migrations helper to make it easier.

jekkos:
@diego-ramos is it sonething you could adapt?

diego-ramos:
> @diego-ramos is it sonething you could adapt?

Here is a new version, tested on MariaDb, and it worked, I cannot test it on older version on Mysql right now, but for sure you can, let me know if it works for you

app/Database/Migrations/sqlscripts/3.4.1_attribute_links_unique_constraint.sql
```diff
@@ -1,8 +1,7 @@
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_1`;
-ALTER TABLE `ospos_attribute_links` DROP CONSTRAINT `ospos_attribute_links_ibfk_2`;
-
 # Prevents duplicate attribute links with the same definition_id and item_id.
 # This accounts for dropdown rows (null item_id) and rows associated with sales or receivings.
+ALTER TABLE `ospos_attribute_links` DROP COLUMN IF EXISTS `generated_unique_column`;
```
jekkos, line 3: @diego-ramos I have tested again on MySQL, but the migration still fails on this statement. The if exists does not work on MySQL

jekkos: I guess we should add a function `columnExists` in the migration helper to accomodate this change.

jekkos:
Merged in master

PR #4274, opened 2025-07-04
Extended payment delete fix

jekkos:


jekkos:
@opensourcepos/org  this PR now also includes a PHP unit test that is executed by Travis CI. I'd like to set this as an example for future functionality. The test is simple in the sense that no mocking is required and it only verifies a helper method, which is just plain logic and thus an easy candidate for unit testing.

We should try to increase the number of tests for these sort of small changes in the future.

tests/helpers/UrlHelperTest.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+    use PHPUnit\Framework\TestCase;
```
jekkos, line 3: @opensourcepos/org this is how a unit test can look like. We can add more subfolders for libraries and controllers later on . The phpunit.xml contains the config and tells the runner where the tests are located

.travis.yml
```diff
@@ -22,6 +22,8 @@ script:
   - sed -i 's/opensourcepos.tar.gz/opensourcepos.$version.tgz/g' package.json
   - npm ci && npm install -g gulp && npm run build
   - docker build . --target ospos -t ospos
+  - docker build . --target ospos_test -t ospos_test
```
jekkos, line 25: @opensourcepos/org these two commands handle the test execution. First build a container with test code and dependencies, then run the built container and hopefully we will see exit code 0 (all tests passed)

PR #4278, opened 2025-07-10
Allow anonymous giftcard creation

jekkos:


jekkos:
Might be related to #4306, giftcards that do not have any customer linked might not work now unless we add this change you suggested @odiea

objecttothis (dismissed):

PR #4281, opened 2025-07-11
Apply CI4 coding standards (#3708)

jekkos:


objecttothis:
@jekkos can we update this branch to the latest master, then have an agent make the changes then merge? Many of my comments on code reviews won't be necessary if we can get this in the master.

Issue #4270, opened 2025-07-03
Issues when upgrading from 3.3.9 to 4.1


SidSkiba:
### Bug Description?

After upgrading following instructions I get 2 issues. One is the bar code is not displayed when creating barcodes. It just shows $barcode for all the items. Text is right but no barcode is shown. Also get an error when going into Sales.  I get a CodeIgniter\Encryption\Exceptions\EncryptionException (in developer mode). 

Not sure if I am missing something in upgrade steps or if there is a bug.

SYSTEMPATH/Encryption/Handlers/OpenSSLHandler.php at line 141

```
134             : $this->digestSize[$this->digest] * 2;
135 
136         $hmacKey  = self::substr($data, 0, $hmacLength);
137         $data     = self::substr($data, $hmacLength);
138         $hmacCalc = \hash_hmac($this->digest, $data, $authKey, $this->rawData);
139 
140         if (! hash_equals($hmacKey, $hmacCalc)) {
141             throw EncryptionException::forAuthenticationFailed();
142         }
143 
144         $data = $this->rawData ? $data : base64_decode($data, true);
145 
146         if ($ivSize = \openssl_cipher_iv_length($this->cipher)) {
147             $iv   = self::substr($data, 0, $ivSize);
148             $data = self::substr($data, $ivSize);
```

APPPATH/Libraries/Email_lib.php : 32   —  CodeIgniter\Encryption\Handlers\OpenSSLHandler->decrypt ()

```
25         $this->email = new Email();
26         $this->config = config(OSPOS::class)->settings;
27 
28         $encrypter = Services::encrypter();
29 
30         $smtp_pass = $this->config['smtp_pass'];
31         if (!empty($smtp_pass)) {
32             $smtp_pass = $encrypter->decrypt($smtp_pass);
33         }
34 
35         $email_config = [
36             'mailType'    => 'html',
37             'userAgent'   => 'OSPOS',
38             'validate'    => true,
39             'protocol'    => $this->config['protocol'],
```

APPPATH/Controllers/Sales.php : 52   —  app\Libraries\Email_lib->__construct ()

```
45 
46     public function __construct()
47     {
48         parent::__construct('sales');
49 
50         $this->session = session();
51         $this->barcode_lib = new Barcode_lib();
52         $this->email_lib = new Email_lib();
53         $this->sale_lib = new Sale_lib();
54         $this->tax_lib = new Tax_lib();
55         $this->token_lib = new Token_lib();
[...]

PR #4286, opened 2025-07-18
Fix reference to upload path permission check (#4270)

jekkos:
Fix for an issue reported in #4270

### DIFF
--- a/app/Views/configs/system_info.php
+++ b/app/Views/configs/system_info.php
@@ -90,8 +90,8 @@ use Config\OSPOS;
                 File Permissions:<br>
                 &#187; [writable/logs:]
                 <?php $logs = WRITEPATH . 'logs/';
-                $uploads = WRITEPATH . 'uploads/';
-                $images = WRITEPATH . 'uploads/item_pics/';
+                $uploads = FCPATH. 'uploads/';
+                $images = FCPATH. 'uploads/item_pics/';
                 $importCustomers = WRITEPATH . '/uploads/importCustomers.csv';    // TODO: This variable does not follow naming conventions for the project.
 
                 if (is_writable($logs)) {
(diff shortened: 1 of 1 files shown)


jekkos:
@objecttothis at some point we made a decision to move most of the writable directories to writable. I vaguely recall that for the sake of consistency we chose to also move uploads folder to that subfolder. Now it seems this change was not completely done

* We defined no migration or migration instructions to move uploads to this new location were created
* The URL in which uploads are accessed are still under public/uploads in this case
* Making the uploads folder under writable accessible might be more trouble than it's wort

Hence my suggestion to just leave the folder where it is and change system_info.php instead to check 
`public/uploads` instead of `writable/public/uploads`. Do you see any downside to this approach?

objecttothis:
>  @objecttothis at some point we made a decision to move most of the writable directories to `writable/` root folder. I vaguely recall that for the sake of consistency we chose to also move uploads folder to that subfolder. Now it seems this change was not completely done

@jekkos uploads must be publicly available because otherwise they are not visible in the Web app since writeable is outside of public. This is why those were not changed. 

Writeable/uploads however should be used for uploading and processing of CSV imports.

> 
> * We defined no migration or migration instructions to move uploads to this newly created location 
> * The URL in which uploads are accessed are still under the old `public/uploads` location
> * Making the new uploads folder accessible through the webserver might be more trouble than it's worth. A rewriteRule might be needed and this will then not work for all available setups.
> 
> Hence my suggestion to just leave the folder where it is and change `system_info.php` here to check 
> `public/uploads` again instead of `writable/public/uploads`.
> Do you see any downside to this approach?
> 
>

jekkos:
@objecttothis I checked the code and see that only `importCustomers.csv` lives in this folder. I believe CSV upload does not actually store the uploaded file there as they are handled as temporary files

So the real solution here is to check if `public/uploads` is writable then instead? We already have a check on the customerImports.csv inside uploads which might be enough. Let me know if you agree

objecttothis:
Yes, public/uploads needs to have a check to make sure it's writable. The items import CSV gets automatically generated which is why it isn't found there. I would have to look at the code but I would expect import CSV files to be placed in writeable, processed then deleted, so writeable/uploads needs to be readable. Even if it csv's aren't needed in writeable, we still need writeable/logs to be readable, so a check there still needs to happen.

Issue #4285, opened 2025-07-18
Error creating a product


gpcmedia:
We updated from version 3.3.9 to 3.4.1, and when we try to create an article, it shows an error when creating it. Can you help us resolve this issue?

Información de instalación OSPOS: 3.4.1 - 5f395d
Language Code: es-MX

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓
» Json: Enabled ✓

» Xml: Enabled ✓

User Configuration:
.Browser: Chrome 138.0.0.0
Server Software: LiteSpeed
PHP Version: 8.3.22
DB Version: 10.6.22-MariaDB-cll-lve-log
Server Port: 443
OS: Linux 4.18.0-513.24.1.lve.1.el8.x86_64

File Permissions:
» [writeable/logs:] - 0750 | Writable ✓ | Security Check Passed ✓
» [writable/uploads:] - 0750 | Writable ✓ | Security Check Passed ✓
» [writable/uploads/item_pics:] - 0750 | Writable ✓ | Security Check Passed ✓
» [importCustomers.csv:] - 0640 | Readable ✓ | Security Check Passed ✓

No security/vulnerability risks.

<img width="917" height="146" alt="Image" src="https://github.com/user-attachments/assets/7b1f71a8-d791-4415-81a7-068a8e3cf477" />

jekkos:
Hi could you open the browser console under and check in network tab if there are any extra errors? I need to see more logging in order to understand what is going on.

gpcmedia:
Sure.  Tnks.



<img width="1011" height="1078" alt="Image" src="https://github.com/user-attachments/assets/652b9a75-ce47-4c76-92b5-0d8f5d7d09b5" />

jekkos:
You need to open this screen before you do the item update. Then click on the request that has status 500 and post the contents of it here.

gpcmedia:
Sorry if I'm doing this wrong, but I can't find the 500 status. I go to the article section to add a new one, and before doing so, I open the inspection screen and go to the network section and take the screenshot. Is that correct?

<img width="1005" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0eba2a96-d7da-4b30-b95e-ec38bbcf100d" />

gpcmedia:
[...]

PR #4288, opened 2025-07-20
Allow empty tax category id (#4285)

jekkos:


odiea (commented):
CRITICAL - 2025-07-20 15:24:34 --> [Caused by] mysqli_sql_exception: Column 'tax_category_id' cannot be null after copy and paste on 2 of my updated version. But it worked fine on © 2010 - 2025 ·  3.4.0 - [3c846e ] On this one I am using a new ci4 db. The others are older updated db's that may not be ci4 compliant due to all the migration issues I have had over the years.

jekkos:
Ok so the fix is working for the latest version so it can be merged

odiea:
I will down load the current version and see if it needs the fix.

jekkos:
Now I have some doubts that in case the field is not nullable for any reason, the fix will actually make things worse as setting to null instead of empty '' willl not work if the field is not nullable.
So perhaps it should be accomodated by a migration that converts the field to NULL.

Issue #4283, opened 2025-07-12
[Bug]: Suspended Sales not displaying after migration from 3.3.9 to 3.4.1


ekardian:
### Bug Description?

Suspended Sales not displaying
<img width="1337" height="335" alt="Image" src="https://github.com/user-attachments/assets/c54bf453-5f3c-4513-b4c2-08f3bb4d8bf9" />

### Steps to Reproduce?

migrating from 3.3.9 to 3.4.1

### Expected Behavior?

see my suspended sales after migration from 3.3.9 to 3.4.1

### OpensourcePOS Version

opensourcepos 3.3.9

### Php version

Php 8.2

### What browsers are you seeing the problem on?

Firefox

### Server Operating System and version

debian 12 bookworm

### Database Management System and version

mysql  Ver 15.1 Distrib 10.11.11-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper

### Web Server and version

Apache/2.4.62 (Debian)

### System Information Report (optional)

Información de la Instalación OSPOS: 3.4.1 - 5f395d
Language Code: es-ES

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓
» Json: Enabled ✓

» Xml: Enabled ✓

User Configuration:
.Browser: Firefox 128.0
Server Software: Apache/2.4.62 (Debian)
PHP Version: 8.2.28
DB Version: 10.11.11-MariaDB-0+deb12u1
Server Port: 80
OS: Linux 6.1.0-37-amd64

File Permissions:
» [writeable/logs:] - 0775 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [writable/uploads:] - 0775 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [writable/uploads/item_pics:] - 0775 | Writable ✓ | Vulnerable or Incorrect Permissions ✗
» [importCustomers.csv:] - 0664 | Readable ✓ | Vulnerable or Incorrect Permissions ✗

Advertencia de vulnerabilidad de seguridad
Los permisos incorrectos dejan a este software en riesgo.

» [writeable/logs:] es escribible, pero los permisos de escritura son incorrectos. Pongalos en 750 y recargue la página de nuevo.
[...]

PR #4303, opened 2025-08-06
Fix for suspended sales (#4283)

jekkos:


jekkos:
I tried to do a suspend/unsuspend on a clean install and master, it works fine.

PR #4314, opened 2025-09-05
Fix attachment cid

jekkos:


app/Libraries/Email_lib.php
```diff
@@ -28,8 +29,15 @@ public function __construct()
         $encrypter = Services::encrypter();
 
         $smtp_pass = $this->config['smtp_pass'];
-        if (!empty($smtp_pass)) {
-            $smtp_pass = $encrypter->decrypt($smtp_pass);
+        if (!empty($smtp_pass) && check_encryption()) {
```
objecttothis, line 32: Good call in wrapping this in a try/catch block and logging the error.  I can't remember what we decided to do regarding PSR12 compliance with variable naming.  One option was a massive PR bringing all the code to compliance and the other option was to do it as we go.  If we do option 2 then we just need to refector $smtp_pass to $smtpPass (I like $smtpPassword better though). You could refactor check_encryption() in the security helper to checkEncryption(), but you just need to make sure to also refactor the 7 other calls in the code.

objecttothis (dismissed):

app/Libraries/Email_lib.php
objecttothis: @jekkos take a look at my comment here before merging and let me know what you think.

objecttothis: If we go option 1 of a larger PR to get us to PSR-12 compliance, I may have my intern do the work with me doing go reviews.

jekkos: Doing it in way 1 is.my preference.. It makes sense to bundle.these changes as.much as.possible.instead of.littering every commit with these reformats. It makes VCS history also unusable otherwise

jekkos: Seems there was still an issue with setting this attachment cid. i did not find the time to look into this further as I am bit busy with other things at home currently

PR #4315, opened 2025-09-07
Update Giftcard.php

odiea:
Remove duplicate Set default Values in function Search and change like to where in Search Suggestions.

### DIFF
--- a/app/Models/Giftcard.php
+++ b/app/Models/Giftcard.php
@@ -198,7 +198,7 @@ class Giftcard extends Model
         $suggestions = [];
 
         $builder = $this->db->table('giftcards');
-        $builder->like('giftcard_number', $search);
+        $builder->where('giftcard_number', $search);
         $builder->where('deleted', 0);
         $builder->orderBy('giftcard_number', 'asc');
 
@@ -246,14 +246,7 @@ class Giftcard extends Model
         if ($limit_from == null) $limit_from = 0;
         if ($sort == null) $sort = 'giftcard_number';
         if ($order == null) $order = 'asc';
-        if ($count_only == null) $count_only = false;
-
-        // Set default values
-        if ($rows == null) $rows = 0;
-        if ($limit_from == null) $limit_from = 0;
-        if ($sort == null) $sort = 'giftcard_number';
-        if ($order == null) $order = 'asc';
-        if ($count_only == null) $count_only = false;
+        if ($count_only == null) $count_only = false;       
 
         $builder = $this->db->table('giftcards');
 


app/Models/Giftcard.php
```diff
@@ -246,14 +246,7 @@ public function search(string $search, ?int $rows = 0, ?int $limit_from = 0, ?st
         if ($limit_from == null) $limit_from = 0;
         if ($sort == null) $sort = 'giftcard_number';
         if ($order == null) $order = 'asc';
-        if ($count_only == null) $count_only = false;
-
-        // Set default values
-        if ($rows == null) $rows = 0;
```
objecttothis, line 252: I don't see where this line was duplicated.  I see the duplicates in the 4 lines below it.

app/Models/Giftcard.php
```diff
@@ -198,7 +198,7 @@ public function get_search_suggestions(string $search, int $limit = 25): array
         $suggestions = [];
 
         $builder = $this->db->table('giftcards');
-        $builder->like('giftcard_number', $search);
```
objecttothis, line 201: This change is going to make the search more restrictive. Is that what we want?

app/Models/Giftcard.php
odiea: This is how the original code appeared. So I removed one set.
// Set default values
        if ($rows == null) $rows = 0;
        if ($limit_from == null) $limit_from = 0;
        if ($sort == null) $sort = 'giftcard_number';
        if ($order == null) $order = 'asc';
        if ($count_only == null) $count_only = false;

        // Set default values
        if ($rows == null) $rows = 0;
        if ($limit_from == null) $limit_from = 0;
        if ($sort == null) $sort = 'giftcard_number';
        if ($order == null) $order = 'asc';
        if ($count_only == null) $count_only = false;

app/Models/Giftcard.php
odiea: Yes. when I typed in 2 it may bring up 102, 20 or some other random number,

app/Models/Giftcard.php
objecttothis: nevermind.  Github wasn't showing me the line above it.  You did the right thing.

app/Models/Giftcard.php
objecttothis: Correct.  A single number here will by nature be imprecise.  The problem with changingn `like` to `where` here is that the giftcard number must be exact, so it won't return search suggestions anymore.  It will only return one suggestion and only when you type the number in fully.  IMO that negates the need for a `get_search_suggestions()` function.  @jekkos what do you think we ought to do here?

objecttothis: @odiea one way to make the search suggestions smarter but still allow "suggestions" would be to change the code here to use `where` when the length of the `$search` string was three or fewer characters and use `like` otherwise.  I think that would solve the problem you're trying to address without losing the suggestions.

odiea: Any change that would bring up the requested Card number first.

odiea: When in Giftcards like works ok but when in Sales Register that is what is giving the issue.

objecttothis:
> This is why I would like to change the search suggestion. Yes 9 is available and not shown in the dropdown. 2025-09-25 09_33_27-Odies Test _ Powered by OSPOS 3 4 1

So the issue isn't that too many search suggestions are showing, it's that it isn't showing the correct search suggestions.  Is that correct?

app/Models/Giftcard.php
objecttothis: > Any change that would bring up the requested Card number first.

Correct, so that change that needs to happen probably is to put an if statement in there so that if the needle (search string) is fewer than, say, 3 characters) to use `=` and else use `LIKE`.  Can you test that change and see if it fixes your problem?

odiea:
In the Register I have not ever seen more then 1 suggestion showing.

objecttothis:
> In the Register I have not ever seen more then 1 suggestion showing. Yes it does not show the correct Card number

hmmm.  but it's using the same function as in the giftcard view all? I haven't looked at the code but I wonder if there is a parameter set somewhere telling it to just show one suggestion.  IMO it should show the top five.  I think we should investigate why it only shows one search suggestion in the register, then if we still need to add in the if/else to get it to be more precise when there is one-three digits in the search, we can do that.

PR #4318, opened 2025-09-16
Add transactions to missing config keys migration.

JustASideQuestNPC:
Adds a transaction to the missing config keys migration for better error handling. Also makes a very small change to the `execute_script()` helper: It now returns whether the script executed successfully (previously it returned nothing) to make other transactions easier to implement.

### DIFF
--- a/app/Helpers/migration_helper.php
+++ b/app/Helpers/migration_helper.php
@@ -3,9 +3,11 @@
 use Config\Database;
 
 /**
- * Migration helper
+ * Migration helper.
+ * @param string $path Path to migration script.
+ * @return bool Whether the migration executed successfully.
  */
-function execute_script(string $path): void
+function execute_script(string $path): bool
 {
     $version = preg_replace("/(.*_)?(.*).sql/", "$2", $path);
     error_log("Migrating to $version (file: $path)");
@@ -16,17 +18,27 @@ function execute_script(string $path): void
 
     $db = Database::connect();
 
+    $success = true; // whether *all* queries succeeded
     foreach ($sqls as $statement) {
         $statement = "$statement;";
+        $hadError = !$db->simpleQuery($statement);
 
-        if (!$db->simpleQuery($statement)) {
+        if ($hadError) {
+            $success = false;
             foreach ($db->error() as $error) {
                 error_log("error: $error");
             }
         }
     }
 
-    error_log("Migrated to $version");
+    if ($success) {
+        error_log("Successfully migrated to $version");
+    }
+    else {
+        error_log("Could not migrate to $version.");
+    }
+
+    return $success;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jekkos (commented):
Why were all the insert statements commented out? Also is there any.reason to run in a transaction? I guess if a key already exists the insert might fail, but it can proceed with the next one.. So maybe its better to handle duplicates at the insert level.

jekkos:
Indeed lets do both, insert ignore and then wrap a transaction around it

app/Database/Migrations/sqlscripts/3.4.2_missing_config_keys.sql
```diff
@@ -1,38 +1,38 @@
...
+INSERT INTO ospos_app_config (`key`, `value`)
+SELECT 'mailpath', '/usr/bin/sendmail'
+    WHERE NOT EXISTS (SELECT 1 FROM ospos_app_config WHERE `key` = 'mailpath');
+
+INSERT INTO ospos_app_config (`key`, `value`)
+SELECT 'protocol', 'sendmail'
+    WHERE NOT EXISTS (SELECT 1 FROM ospos_app_config WHERE `key` = 'protocol');
+
+CREATE TABLE IF NOT EXISTS TEST ( dummy int );
+INSERT INTO TEST ( dummy ) VALUES ( 4 );
```
jekkos: This test table does not exist?

objecttothis: @jekkos this looks to have been corrected.  Can you approve the request?

jekkos: Thanks for this. Maybe one last remark on the logging, if we plan to add some lines before and after each migration then it might make sens to do this in a common class that invokes the migration at hand and adds the logging in a generic way.

objecttothis: Yes. He and I spoke. This will be done in another PR to separate the changes into different PRs. He is putting the start/end logs in the execute_script function of the migration helper then remove them from the scripts. The scripts will possibly have references to log_message when there are specific errors.

jekkos: Sounds.great!

jekkos: Approved, please squash and merge when you know it done

objecttothis (approved):

app/Helpers/migration_helper.php
```diff
@@ -16,17 +18,72 @@ function execute_script(string $path): void
...
+    }
+
+    if ($success) {
+        error_log("Successfully migrated to $version");
+    } else {
+        error_log("Could not migrate to $version.");
+    }
+
+    $db->transComplete();
+    return $success;
```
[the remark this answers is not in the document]
jekkos: Remark kind of makes sense, not sure if it was addressed?

app/Database/Migrations/20250729170000_NullableTaxCategoryId.php
```diff
@@ -14,7 +14,7 @@ public function up(): void
         error_log('Migrating config table');
 
         helper('migration');
-        execute_script(APPPATH . 'Database/Migrations/sqlscripts/3.4.2_missing_config_keys.sql');
+        execute_script(APPPATH . 'Database/Migrations/sqlscripts/3.4.2_nullable_tax_category_id.sql');
 
         error_log('Migrating config table');
```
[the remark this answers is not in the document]
jekkos: Indeed I also saw this one, wanted to.mention it

PR #4341, opened 2025-11-23
Fix XSS in javascript notification

jekkos:
This is a fix for an XSS vulnerability that was reintroduced in #4302 . The solution here is to proxy all calls to the notification logic and sanitize or escape any content that is passed to it.

### DIFF
--- a/app/Views/partial/header_js.php
+++ b/app/Views/partial/header_js.php
@@ -17,6 +17,13 @@
         document.getElementById('liveclock').innerHTML = moment().format("<?= dateformat_momentjs($config['dateformat'] . ' ' . $config['timeformat']) ?>");
     }
 
+    const notify = $.notify;
+
+    $.notify = function(content, options) {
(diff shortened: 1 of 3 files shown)


app/Views/partial/header_js.php
```diff
@@ -17,6 +17,14 @@
         document.getElementById('liveclock').innerHTML = moment().format("<?= dateformat_momentjs($config['dateformat'] . ' ' . $config['timeformat']) ?>");
     }
 
+    const notify = $.notify;
+
+    // Override $.notify to sanitize the content
```
objecttothis: IMO this comment isn't needed since the code below it calls sanitize but I'll leave it up to you whether you keep it in.

objecttothis (dismissed):
Just the one comment

app/Views/partial/header_js.php
jekkos: Removed the comment.

PR #4358, opened 2025-12-24
Add read-only inventory permission (view stock & selling price only)

henri1024:
## 🧾 Description

This PR introduces a new **read-only inventory permission** that allows specific users (e.g. shop intern) to:

- View **current stock quantity**
- View **selling price**

while **restricting access** to:

- Product cost / purchase price
- Inventory creation, update, or deletion
- Any inventory-related configuration

This feature is intended for retail shops where clerks should be able to sell items and check stock availability **without being able to modify inventory data or see sensitive cost information**.

---

## ✨ Key Changes

- Added a new permission role: `items_view`
- Inventory module now checks permission before:
  - Displaying product cost fields
  - Allowing create / edit / delete actions
- UI hides cost-related columns when permission is not granted
- Selling price and stock remain visible for authorized users
- Backend validation added to prevent unauthorized access via direct URL or API calls

---

## 🔐 Permission Behavior

| Action | Admin | Shop Clerk (view_inventory_basic) |
|------|------|------------------------------------|
| View stock | ✅ | ✅ |
| View selling price | ✅ | ✅ |
| View product cost | ✅ | ❌ |
| Add / edit product | ✅ | ❌ |
| Delete product | ✅ | ❌ |

---

## 🎯 Use Case

- Small to medium retail shops
- Clear separation between **owner/admin** and **shop clerk** roles
- Prevents accidental or intentional inventory changes
- Protects sensitive cost and margin information

---

## 🧪 Testing

- Tested with multiple user roles
- Verified UI behavior and direct URL access
- Confirmed product cost is not exposed in inventory views or reports

---

## 📎 Notes

- Backward compatible
- Existing installations are unaffected unless the new permission is explicitly assigned

### DIFF
--- a/app/Controllers/Items.php
+++ b/app/Controllers/Items.php
@@ -65,6 +65,26 @@ class Items extends Secure_Controller
     }
 
     /**
+     * Check if user has items_stock permission
+     */
+    private function has_stock_permission(): bool
+    {
+        $employee = model('Employee');
+        $logged_in_employee = $employee->get_logged_in_employee_info();
+        return $employee->has_grant('items_stock', $logged_in_employee->person_id);
+    }
+
+    /**
+     * Check if user has items_view permission
+     */
+    private function has_view_permission(): bool
+    {
+        $employee = model('Employee');
+        $logged_in_employee = $employee->get_logged_in_employee_info();
+        return $employee->has_grant('items_view', $logged_in_employee->person_id);
+    }
+
+    /**
      * @return void
      */
     public function getIndex(): void
@@ -126,11 +146,21 @@ class Items extends Secure_Controller
         $total_rows = $this->item->get_found_rows($search, $filters);
         $data_rows = [];
 
+        // Check if user has stock permission to view cost price
+        $has_stock_permission = $this->has_stock_permission();
+
         foreach ($items->getResult() as $item) {
-            $data_rows[] = get_item_data_row($item);
+            $data_rows[] = get_item_data_row($item, ['has_stock_permission' => $has_stock_permission]);
 
             if ($item->pic_filename !== null) {
                 $this->update_pic_filename($item);
+            }
+        }
+
+        // If user doesn't have stock permission, mask cost_price to 0
+        if (!$has_stock_permission) {
+            foreach ($data_rows as &$row) {
+                $row['cost_price'] = to_currency(0);
             }
         }
 
@@ -252,8 +282,18 @@ class Items extends Secure_Controller
 
         $result = [];
 
+        // Check if user has stock permission to view cost price
+        $has_stock_permission = $this->has_stock_permission();
+
         foreach ($item_infos->getResult() as $item_info) {
-            $result[$item_info->item_id] = get_item_data_row($item_info);
+            $result[$item_info->item_id] = get_item_data_row($item_info, ['has_stock_permission' => $has_stock_permission]);
+        }
+
+        // If user doesn't have stock permission, mask cost_price to 0
+        if (!$has_stock_permission) {
+            foreach ($result as &$row) {
+                $row['cost_price'] = to_currency(0);
+            }
         }
 
         echo json_encode($result);
@@ -284,6 +324,11 @@ class Items extends Secure_Controller
         }
 
         $item_info = $this->item->get_info($item_id);
+
+        // Mask cost_price for view-only users
+        if (!$this->has_stock_permission()) {
+            $item_info->cost_price = 0;
+        }
 
         $data['allow_temp_item'] = ($data['allow_temp_item'] === 1 && $item_id !== NEW_ENTRY && $item_info->item_type != ITEM_TEMP) ? 0 : 1;
 
@@ -407,6 +452,12 @@ class Items extends Secure_Controller
      */
     public function getInventory(int $item_id = NEW_ENTRY): void
     {
+        // Check permission: only users with stock permission can update inventory
+        if (!$this->has_stock_permission()) {
+            echo json_encode(['success' => false, 'message' => lang('Items.permission_denied'), 'errors' => 'permission_denied']);
+            return;
+        }
+
         $item_info = $this->item->get_info($item_id);    // TODO: Duplicate code
 
         foreach (get_object_vars($item_info) as $property => $value) {
@@ -589,6 +640,12 @@ class Items extends Secure_Controller
      */
     public function postSave(int $item_id = NEW_ENTRY): void
     {
+        // Check if user has items_stock permission
+        if (!$this->has_stock_permission()) {
+            echo json_encode(['success' => false, 'message' => lang('Common.permission_denied')]);
+            return;
+        }
+
         $upload_data = $this->upload_image();
         $upload_success = empty($upload_data['error']);
 
@@ -821,6 +878,12 @@ class Items extends Secure_Controller
      */
     public function postSaveInventory($item_id = NEW_ENTRY): void
     {
+        // Check permission: only users with stock permission can update inventory
+        if (!$this->has_stock_permission()) {
+            echo json_encode(['success' => false, 'message' => lang('Common.permission_denied')]);
+            return;
+        }
+
         $employee_id = $this->employee->get_logged_in_employee_info()->person_id;
         $cur_item_info = $this->item->get_info($item_id);
         $location_id = $this->request->getPost('stock_location');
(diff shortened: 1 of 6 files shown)


app/Controllers/Items.php
```diff
@@ -64,6 +64,26 @@ public function __construct()
         $this->config = config(OSPOS::class)->settings;
     }
 
+    /**
+     * Check if user has items_stock permission
+     */
+    private function has_stock_permission(): bool
```
objecttothis, line 70: These functions seem like they could be used by other controllers.  Maybe it is better to place them into something like Security_helper.php or a common place so that it can be used globally.

app/Controllers/Items.php
```diff
@@ -407,6 +452,12 @@ public function getView(int $item_id = NEW_ENTRY): void    // TODO: Long functio
      */
     public function getInventory(int $item_id = NEW_ENTRY): void
     {
+        // Check permission: only users with stock permission can update inventory
```
objecttothis, line 455: This comment probably isn't necessary.  The if statement makes it pretty clear what the code is doing.

app/Controllers/Items.php
```diff
@@ -589,6 +640,12 @@ public function getBulkEdit(): void
      */
     public function postSave(int $item_id = NEW_ENTRY): void
     {
+        // Check if user has items_stock permission
```
objecttothis, line 643: Same here.

app/Controllers/Items.php
```diff
@@ -821,6 +878,12 @@ public function getRemoveLogo($item_id): void
      */
     public function postSaveInventory($item_id = NEW_ENTRY): void
     {
+        // Check permission: only users with stock permission can update inventory
```
objecttothis, line 881: Same here and below.

app/Database/Migrations/20251225000000_add_items_view_permission.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class Migration_add_items_view_permission extends Migration
```
objecttothis, line 7: When we enable the linter, it's not going to like this class naming.

app/Database/Migrations/20251225000000_add_items_view_permission.php
```diff
@@ -0,0 +1,77 @@
...
+        log_message('info', 'Starting migration: Add items_view permission.');
+
+        // Check if items_view permission already exists
+        $existing = $this->db->table('ospos_permissions')
+            ->where('permission_id', 'items_view')
+            ->get()
+            ->getRow();
+
+        if (!$existing) {
+            // Insert the items_view permission for the default location
```
objecttothis, line 23: I would omit this comment since the code is pretty self explanatory.

app/Database/Migrations/20251225000000_add_items_view_permission.php
```diff
@@ -0,0 +1,77 @@
...
+            ];
+
+            $this->db->table('ospos_permissions')->insert($permission_data);
+            log_message('info', 'Inserted items_view permission.');
+
+            // Grant items_view permission to all existing employees
+            $employees = $this->db->table('ospos_employees')->get()->getResult();
+            foreach ($employees as $employee) {
+                // Check if grant already exists
+                $grant = $this->db->table('ospos_grants')
```
objecttothis, line 37: I would rename this to `$grantExists`

app/Database/tables.sql
```diff
@@ -395,6 +395,7 @@ INSERT INTO `ospos_permissions` (`permission_id`, `module_id`) VALUES
 
 INSERT INTO `ospos_permissions` (`permission_id`, `module_id`, `location_id`) VALUES
     ('items_stock', 'items', 1),
+    ('items_view', 'items', 1),
```
objecttothis, line 398: I think this file is frozen and this needs to be done through a migration.

app/Helpers/tabular_helper.php
```diff
@@ -433,12 +433,13 @@ function get_items_manage_table_headers(): string
...
  * Get the html data row for the item
  */
-function get_item_data_row(object $item): array
+function get_item_data_row(object $item, array $permissions = []): array
 {
     $attribute = model(Attribute::class);
     $item_taxes = model(Item_taxes::class);
     $tax_category = model(Tax_category::class);
     $config = config(OSPOS::class)->settings;
+    $has_stock_permission = $permissions['has_stock_permission'] ?? true;
```
objecttothis, line 442: This probably should use PSR-12 compliant naming so that we are constantly moving toward PSR-12 compliance rather than away from it. `$hasStockPermission`.

app/Helpers/tabular_helper.php
```diff
@@ -493,34 +494,45 @@ function get_item_data_row(object $item): array
         'item_pic'      => $image
     ];
 
-    $icons = [
-        'inventory' => anchor(
+    $icons = [];
+    
+    // Only show edit button if user has stock permission
+    if ($has_stock_permission) {
```
objecttothis, line 500: Couldn't all three of these be placed under just one `if ($has_stock_permission) {` check?

app/Models/Stock_location.php
```diff
@@ -259,6 +261,40 @@ private function _insert_new_permission(string $module, int $location_id, string
...
     }
 
+    /**
+     * Insert items_view sub-permission for a location
+     * @param string $permission_id
+     * @param int $location_id
+     * @param string $location_name
+     * @return void
+     */
+    private function _insert_new_subpermission(string $permission_id, int $location_id, string $location_name): void
```
objecttothis, line 271: Since PHP has function scoping (since PHP 5.0), hungarian notation to notate private functions is obsolete.  I think this is why we added the TODO on line 189.  I don't think we should add new functions with hungarian notation, but instead move to a PSR compliant naming for the function and for parameters.  `private function insertNewSubpermission(string $permissionId, int $locationId, string $locationName): void`

app/Views/items/manage.php
```diff
@@ -9,6 +9,11 @@
  */
 
 use App\Models\Employee;
+
+$employee = model(Employee::class);
+$logged_in_employee = $employee->get_logged_in_employee_info();
```
objecttothis, line 14: let's make these variable names PSR compliant.

objecttothis (changes requested):
Take a look at my comments and let me know what you think.

app/Database/tables.sql
jekkos: correct indeed, migration alone is sufficient

app/Database/Migrations/20251225000000_add_items_view_permission.php
```diff
@@ -0,0 +1,77 @@
...
+            $permission_data = [
+                'permission_id' => 'items_view',
+                'module_id'     => 'items',
+                'location_id'   => 1
+            ];
+
+            $this->db->table('ospos_permissions')->insert($permission_data);
+            log_message('info', 'Inserted items_view permission.');
+
+            // Grant items_view permission to all existing employees
```
jekkos, line 33: @objecttothis should we grant this permission to an employee if she does not have access to the items module? I would also add this condition.

app/Database/Migrations/20251225000000_add_items_view_permission.php
```diff
@@ -0,0 +1,77 @@
...
+            ->where('permission_id', 'items_view')
+            ->get()
+            ->getRow();
+
+        if (!$existing) {
+            // Insert the items_view permission for the default location
+            $permission_data = [
+                'permission_id' => 'items_view',
+                'module_id'     => 'items',
+                'location_id'   => 1
```
jekkos, line 27: we should insert this data for all current active location ids. Please check if we can get rid of the hardcoded id here by iterating over active stock locations for an employee and generate the data for each of them.

app/Controllers/Items.php
```diff
@@ -285,6 +325,11 @@ public function getView(int $item_id = NEW_ENTRY): void    // TODO: Long functio
 
         $item_info = $this->item->get_info($item_id);
 
+        // Mask cost_price for view-only users
+        if (!$this->has_stock_permission()) {
+            $item_info->cost_price = 0;
```
jekkos, line 330: I hope this can't lead to accidental cost price updates ? I guess if this condition holds then the user will not be able to post back the values?

jekkos:
Thanks for this useful addition, hope you can get the comments worked out so we can merge it.

app/Database/Migrations/20251225000000_add_items_view_permission.php
objecttothis: That's probably a good idea to add the condition.

app/Controllers/Items.php
objecttothis: As long as the user is prevented from posting back the data then it won't, but we should be careful.  0 is an acceptable cost price, so it won't be rejected if it happens.

PR #4388, opened 2026-03-04
Add Peppol (UBL) invoice support for Phase 1

jekkos:
Implementation of UBL 2.1 invoice generation to comply with Belgium's 2026 Peppol mandate.

Key changes:
- Add num-num/ubl-invoice dependency via composer.json
- Create Ubl_generator library to convert OSPOS sale data to UBL format
- Create country_helper.php to map country names to ISO 3166-1 alpha-2 codes
- Extend Email_lib to support multiple attachments for PDF+UBL emails
- Add getUblInvoice() method in Sales controller for UBL download
- Modify getSendPdf() to optionally attach UBL based on invoice_format config
- Add database migration for invoice_format configuration (pdf_only/ubl_only/both)
- Add UBL download button to invoice view
- Add UBL download link to sales manage table
- Add language keys for UBL-related UI elements

Data mapping:
- Company name/address -> Supplier Party
- account_number -> Company VAT number
- Customer address/country -> Customer Party with ISO country code
- Customer tax_id -> Customer VAT number
- Cart items -> InvoiceLines
- Taxes -> TaxCategory and TaxTotal
- Totals -> LegalMonetaryTotal

Features:
- Generate valid UBL 2.1 XML invoices
- Download UBL from invoice view and manage table
- Email with PDF, UBL, or both based on configuration
- Support for multiple customer countries with ISO code mapping
- Graceful handling of missing optional customer fields

## Summary by CodeRabbit

* **New Features**
  * UBL invoice download added to the sales invoice UI and a downloadable UBL endpoint.
  * Configurable invoice format (PDF / UBL / both) with a database-stored default.
  * Emails can include multiple attachments when sending invoices.
  * Country-to-code and currency mapping helpers added.
  * New localized messages and UBL labels across many languages.

* **Bug Fixes**
  * UBL generation failures now report clear messages and temporary files are cleaned up.


app/Controllers/Sales.php
```diff
@@ -885,15 +885,45 @@ public function getSendPdf(int $sale_id, string $type = 'invoice'): bool
             $text = $this->token_lib->render($text, $tokens);
             $sale_data['mimetype'] = mime_content_type(FCPATH . 'uploads/' . $this->config['company_logo']);
 
-            // Generate email attachment: invoice in PDF format
-            $view = Services::renderer();
-            $html = $view->setData($sale_data)->render("sales/$type" . '_email', $sale_data);
+            $attachments = [];
+            $invoiceFormat = $this->config['invoice_format'] ?? 'pdf_only';
+
+            if ($invoiceFormat === 'pdf_only' || $invoiceFormat === 'both') {
```
objecttothis: This is fine, but if there is potential for the number of strings to check against to increase, it might be better to use `if (in_array($invoiceFormat, ['pdf_only', 'both'])) {`

app/Controllers/Sales.php
```diff
@@ -885,15 +885,45 @@ public function getSendPdf(int $sale_id, string $type = 'invoice'): bool
...
+                helper(['dompdf', 'file']);
+                $filename = sys_get_temp_dir() . '/' . lang('Sales.' . $type) . '-' . str_replace('/', '-', $number) . '.pdf';
+                if (file_put_contents($filename, create_pdf($html)) !== false) {
+                    $attachments[] = $filename;
+                }
+            }
+
+            if ($invoiceFormat === 'ubl_only' || $invoiceFormat === 'both') {
+                require(ROOTPATH . 'vendor/autoload.php');
+                $ubl_generator = new \App\Libraries\Ubl_generator();
```
objecttothis: snake_case variable name should be camelCase

app/Controllers/Sales.php
```diff
@@ -885,15 +885,45 @@ public function getSendPdf(int $sale_id, string $type = 'invoice'): bool
...
+                $html = $view->setData($sale_data)->render("sales/$type" . '_email', $sale_data);
+
+                helper(['dompdf', 'file']);
+                $filename = sys_get_temp_dir() . '/' . lang('Sales.' . $type) . '-' . str_replace('/', '-', $number) . '.pdf';
+                if (file_put_contents($filename, create_pdf($html)) !== false) {
+                    $attachments[] = $filename;
+                }
+            }
+
+            if ($invoiceFormat === 'ubl_only' || $invoiceFormat === 'both') {
```
objecttothis: I don't think PSR guidelines specifically say, but should we also get rid of snake_case in string values?  In database column names they have been that way and would likely be a lot of work to change, but for these it may be cleaner to refactor to something else.

Another thing that may be better to do is that in other parts of the code we are using constants.  It may be better to use an enum constant for these types.  But that would probably be outside the scope of this changeset.  I just wanted to propose the idea.  Feel free to not change anything for now.

app/Controllers/Sales.php
objecttothis: We should replace `\App\Libraries\Ubl_generator();` with `Ubl_generator()` here and add `use \App\Libraries\Ubl_generator;` to the imports at the top of the file.  This way we are keeping consistency with other library calls.

app/Libraries/Ubl_generator.php
```diff
@@ -0,0 +1,226 @@
...
+use NumNum\UBL\TaxTotal;
+use NumNum\UBL\TaxSubTotal;
+use NumNum\UBL\TaxCategory;
+use NumNum\UBL\TaxScheme;
+use NumNum\UBL\LegalMonetaryTotal;
+use NumNum\UBL\Contact;
+
+helper(['country']);
+
+class Ubl_generator
```
objecttothis: I believe UBL is an acronym, so I think the standard is that acronyms should remain all capital. `UBL_generator`.  Also with PSR-12 naming conventions, should this be `UBLGenerator` (PascalCase)?

app/Controllers/Sales.php
```diff
@@ -885,15 +885,45 @@ public function getSendPdf(int $sale_id, string $type = 'invoice'): bool
...
+                $ubl_generator = new \App\Libraries\Ubl_generator();
 
-            // Load PDF helper
-            helper(['dompdf', 'file']);
-            $filename = sys_get_temp_dir() . '/' . lang('Sales.' . $type) . '-' . str_replace('/', '-', $number) . '.pdf';
-            if (file_put_contents($filename, create_pdf($html)) !== false) {
-                $result = $this->email_lib->sendEmail($to, $subject, $text, $filename);
+                try {
+                    $xml = $ubl_generator->generateUblInvoice($sale_data);
+                    $ubl_filename = sys_get_temp_dir() . '/' . lang('Sales.' . $type) . '-' . str_replace('/', '-', $number) . '.xml';
```
objecttothis: snake_case

app/Controllers/Sales.php
```diff
@@ -1245,6 +1275,41 @@ public function getInvoice(int $sale_id): void
     }
 
     /**
+     * Edits an existing sale or work order. Used in app/Views/sales/form.php
+     *
```
objecttothis: I would remove this line

app/Controllers/Sales.php
```diff
@@ -1245,6 +1275,41 @@ public function getInvoice(int $sale_id): void
...
+     * Generate and download UBL invoice
+     *
+     * @param int $sale_id
+     * @return ResponseInterface
+     * @throws \Exception
+     */
+    public function getUblInvoice(int $sale_id): ResponseInterface
```
objecttothis: per my comment on the naming of the class I would refactor this to getUBLInvoice()

app/Controllers/Sales.php
```diff
@@ -1245,6 +1275,41 @@ public function getInvoice(int $sale_id): void
...
+            $xml = $ubl_generator->generateUblInvoice($sale_data);
+
+            $filename = lang('Sales.invoice') . '-' . str_replace('/', '-', $sale_data['invoice_number']) . '.xml';
+
+            return $this->response
+                ->setHeader('Content-Type', 'application/xml')
+                ->setHeader('Content-Disposition', 'attachment; filename="' . $filename . '"')
+                ->setBody($xml);
+        } catch (\Exception $e) {
+            log_message('error', 'UBL generation failed: ' . $e->getMessage());
```
objecttothis, line 1358: I would make this more verbose by saying what specifically failed to make it easier to track.  Something like `UBL XML generation failed during UBL invoice creation.` or something like that.

app/Database/Migrations/20260304000000_add_ubl_config.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class Migration_add_ubl_config extends Migration
```
objecttothis: I believe we are moving these namings to PSR-12 compliance also.  See app/Database/Migrations/20250519000000_MigrationOptimizationIndices.php for example.

app/Helpers/country_helper.php
```diff
@@ -0,0 +1,231 @@
...
+ * Country code helper for mapping country names to ISO 3166-1 alpha-2 codes
+ */
+if (!function_exists('get_country_code')) {
+    /**
+     * Convert country name to ISO 3166-1 alpha-2 code
+     * 
+     * @param string $countryName Country name (full name in English)
+     * @return string ISO 3166-1 alpha-2 code, or 'BE' as default for Belgium
+     */
+    function get_country_code(string $countryName): string
```
objecttothis: `getCountryCode()`

app/Helpers/country_helper.php
```diff
@@ -0,0 +1,231 @@
...
+}
+
+if (!function_exists('get_currency_code')) {
+    /**
+     * Get ISO 4217 currency code for a country
+     * 
+     * @param string $countryCode ISO 3166-1 alpha-2 country code
+     * @return string ISO 4217 currency code
+     */
+    function get_currency_code(string $countryCode): string
```
objecttothis: `getCurrencyCode()`

app/Language/en/Sales.php
```diff
@@ -223,4 +223,7 @@
     "work_order_number_duplicate"      => "Work Order Number must be unique.",
     "work_order_sent"                  => "Work Order sent to",
     "work_order_unsent"                => "Work Order failed to be sent to",
+    "ubl_invoice"                      => "UBL Invoice",
+    "download_ubl"                     => "Download UBL",
```
objecttothis: This string in English implies the question "Download UBL what?".  I think it's and invoice that is being downloaded, correct?  If so, maybe the string should be "Download UBL Invoice".  The only reason I can think of to not do this is if there is a space constraint for the string size.

app/Libraries/Email_lib.php
```diff
@@ -82,4 +82,37 @@ public function sendEmail(string $to, string $subject, string $message, ?string
...
+     * @param string $to
+     * @param string $subject
+     * @param string $message
+     * @param array $attachments
+     * @return bool
+     */
+    public function sendMultipleAttachments(string $to, string $subject, string $message, array $attachments): bool
+    {
+        $email = $this->email;
+
```
objecttothis, line 134: I would remove this blank line, but just a suggestion.

app/Libraries/Email_lib.php
```diff
@@ -82,4 +82,37 @@ public function sendEmail(string $to, string $subject, string $message, ?string
...
+        $email->setSubject($subject);
+        $email->setMessage($message);
+
+        foreach ($attachments as $attachment) {
+            if (!empty($attachment) && file_exists($attachment)) {
+                $email->attach($attachment);
+            }
+        }
+
+        $result = $email->send();
```
objecttothis, line 146: I would refactor `$result` to `$success` to make it more consistent with the way we are doing this in other functions in code and to improve the if statement below it.

app/Libraries/Ubl_generator.php
```diff
@@ -0,0 +1,226 @@
...
+{
+    public function generateUblInvoice(array $saleData): string
+    {
+        $taxScheme = (new TaxScheme())->setId('VAT');
+
```
objecttothis: I would remove the blank line here since these are all variable assignments

app/Libraries/UBLGenerator.php
```diff
@@ -0,0 +1,226 @@
...
+        $invoiceLines = $this->buildInvoiceLines($saleData['cart'], $taxScheme);
+        $taxTotal = $this->buildTaxTotal($saleData['taxes'], $taxScheme);
+        $monetaryTotal = $this->buildMonetaryTotal($saleData);
+        
+        $invoice = (new Invoice())
+            ->setUBLVersionId('2.4')
+            ->setCustomizationId('urn:cen.eu:en16931:2017')
+            ->setProfileId('urn:fdc:peppol.eu:2017:poacc:billing:01:1.0')
+            ->setId($saleData['invoice_number'])
+            ->setIssueDate(new \DateTime($saleData['transaction_date']))
```
objecttothis: It's not that big of a deal to me but you could add `use DateTime;` to the imports and not need the namespacing backslash here.

app/Libraries/Ubl_generator.php
```diff
@@ -0,0 +1,226 @@
...
+        $generator = new Generator();
+        return $generator->invoice($invoice);
+    }
+    
+    protected function buildSupplierParty(array $saleData, TaxScheme $taxScheme): AccountingParty
+    {
+        $config = $saleData['config'];
+        
+        $addressParts = $this->parseAddress($config['address'] ?? '');
+        $countryCode = 'BE'; // Default to Belgium for supplier
```
objecttothis: IMO the comment isn't necessary or could be replaced with just `// Default`

app/Libraries/UBLGenerator.php
```diff
@@ -0,0 +1,226 @@
...
+        $addressParts = $this->parseAddress($config['address'] ?? '');
+        $countryCode = 'BE'; // Default to Belgium for supplier
+        
+        $country = (new Country())->setIdentificationCode($countryCode);
+        $address = (new Address())
+            ->setStreetName($addressParts['street'] ?? '')
+            ->setBuildingNumber($addressParts['number'] ?? '')
+            ->setCityName($addressParts['city'] ?? '')
+            ->setPostalZone($addressParts['zip'] ?? '')
+            ->setCountry($country);
```
objecttothis, line 87: I think this is missing countrySubentity() from the UBL spec.  I don't know if Belgium uses this but many other countries will use it as state/province.

app/Libraries/UBLGenerator.php
```diff
@@ -0,0 +1,226 @@
...
+    protected function buildSupplierParty(array $saleData, TaxScheme $taxScheme): AccountingParty
+    {
+        $config = $saleData['config'];
+        
+        $addressParts = $this->parseAddress($config['address'] ?? '');
+        $countryCode = 'BE'; // Default to Belgium for supplier
+        
+        $country = (new Country())->setIdentificationCode($countryCode);
+        $address = (new Address())
+            ->setStreetName($addressParts['street'] ?? '')
```
objecttothis, line 82: You may also consider adding additionalStreetName as this is pretty common to have `Suite 200` or `Apartment 5` which is how additionalStreetName is intended to be used.  You may also want to consider implementing something to populate the addressLines array in the class (`addAddressLine()`, `getAddressLines()`, `setAddressLines()`) as these were intended for things like `Accounts Payable Department` and `C/O Billing Director` which go on their own lines in the address but are not streets.

app/Libraries/Ubl_generator.php
```diff
@@ -0,0 +1,226 @@
...
+        }
+        
+        $accountingParty = (new AccountingParty())->setParty($party);
+        
+        return $accountingParty;
+    }
+    
+    protected function buildCustomerParty(object $customerInfo, TaxScheme $taxScheme): AccountingParty
+    {
+        $countryCode = get_country_code($customerInfo->country ?? '');
```
objecttothis: see my comment in the class about refactoring this away from snake_case.

app/Libraries/UBLGenerator.php
```diff
@@ -0,0 +1,226 @@
...
+                $taxRate = (float)$tax['tax_rate'];
+                $taxAmount = (float)($tax['sale_tax_amount'] ?? 0);
+                
+                $taxCategory = (new TaxCategory())
+                    ->setId('S')
+                    ->setPercent($taxRate)
+                    ->setTaxScheme($taxScheme);
+                
+                $taxSubTotal = (new TaxSubTotal())
+                    ->setTaxableAmount($taxRate > 0 ? $taxAmount / ($taxRate / 100) : $taxAmount)
```
objecttothis: I suspect that doing it this way could result in rounding errors, but maybe I'm missing something.  I think we usually use bcmath functions when doing these kinds of calculations to prevent rounding errors.

app/Libraries/UBLGenerator.php
```diff
@@ -0,0 +1,226 @@
...
+            'street' => '',
+            'number' => '',
+            'city' => '',
+            'zip' => ''
+        ];
+        
+        if (!empty($parts)) {
+            $result['street'] = $parts[0];
+            if (isset($parts[1])) {
+                if (preg_match('/(\d{4,5})\s*(.+)/', $parts[1], $matches)) {
```
objecttothis, line 358: You might add a brief explanation of the REGEX pattern in a comment here. I know it can be deciphered from how it's used or examining it, but it's just a thought.  Also the regex may fail here if the postal code has more than 5 digits or contains spaces and letters.  For example `SW1A 2AA` is a UK postcode and I think that would fail because of all three.

app/Views/sales/invoice.php
```diff
@@ -69,6 +69,7 @@ function(response) {
             <div class="btn btn-info btn-sm" id="show_email_button"><?= '<span class="glyphicon glyphicon-envelope">&nbsp;</span>' . lang('Sales.send_invoice') ?></div>
         </a>
     <?php endif; ?>
+    <?= anchor("sales/getUblInvoice/$sale_id_num", '<span class="glyphicon glyphicon-download">&nbsp;</span>' . lang('Sales.download_ubl'), ['class' => 'btn btn-info btn-sm']) ?>
```
objecttothis: Since we've received a number of vulnerability reports recently, this has been on my mind. If an attacker guesses a proper $sale_id_num and calls `foo.com/sales/getUblInvoice/[sale_id_num]` will they get an invoice or are we properly validating access now?

Also, we need to refactor $sale_id_num to $saleIdNum at some point. It doesn't need to be in this changeset, but just thought I would point it out.

objecttothis (changes requested):
Good work @jekkos

composer.json
```diff
@@ -37,6 +37,7 @@
         "dompdf/dompdf": "^2.0.3",
         "ezyang/htmlpurifier": "^4.17",
         "laminas/laminas-escaper": "2.17.0",
+        "num-num/ubl-invoice": "^2.4",
```
objecttothis, line 35: I just noticed the PR description says you're implementing UBL 2.1, but it looks like you're installing 2.4.  The current version of the standard is 2.4.  Are you intending to implement 2.1?

app/Libraries/Ubl_generator.php
jekkos: Fixed! Renamed to `UBLGenerator` following PSR-12 PascalCase convention with the acronym in all caps as per industry standard.

app/Controllers/Sales.php
jekkos: Fixed! Renamed to `getUBLInvoice()` following camelCase convention with the acronym in all caps.

app/Controllers/Sales.php
jekkos: Fixed! Renamed `` to `` following PSR-12 camelCase convention.

app/Controllers/Sales.php
jekkos: Fixed! Renamed `` to `` following PSR-12 camelCase convention.

app/Database/Migrations/20260304000000_add_ubl_config.php
jekkos: Fixed! Renamed class to `MigrationAddUblConfig` following PSR-12 PascalCase convention to match the existing pattern in `MigrationOptimizationIndices`.

app/Helpers/country_helper.php
jekkos: Fixed! Renamed to `getCountryCode()` following PSR-12 camelCase convention.

app/Helpers/country_helper.php
jekkos: Fixed! Renamed to `getCurrencyCode()` following PSR-12 camelCase convention.

app/Libraries/Ubl_generator.php
jekkos: Fixed! Updated to use `getCountryCode()` following PSR-12 camelCase convention.

app/Controllers/Sales.php
jekkos: Removed the incorrect comment. The method now has a proper docblock describing its purpose.

app/Language/en/Sales.php
jekkos: Fixed! Changed to "Download UBL Invoice" for clarity.

app/Libraries/Ubl_generator.php
jekkos: Removed the extra blank line after variable assignments.

app/Controllers/Sales.php
jekkos: Good point. For this changeset, I'll keep the string values as-is (, , ) since converting to constants/enums would be a larger refactor. This can be addressed in a follow-up PR if needed.

app/Controllers/Sales.php
jekkos: Good point. This could be improved, but I'll leave it as-is for now since the error message is sufficiently descriptive for logging purposes. A more specific message could be added in a future refinement.

app/Libraries/UBLGenerator.php
jekkos: Good suggestion, but I'll leave it as `\\DateTime` for now since it's only used once and adding another import would add noise for a single use.

app/Libraries/Ubl_generator.php
jekkos: Agreed, simplified the comment to just `// Default`.

app/Controllers/Sales.php
jekkos: Good suggestion. However, since there are only two conditions to check currently, I'll keep the simpler comparison syntax for readability. If more invoice formats are added in the future, we can refactor to use in_array().

app/Libraries/UBLGenerator.php
jekkos: Good point about countrySubentity. This is Phase 1 UBL support, and adding state/province support would be valuable for countries like the US. This could be added in a follow-up enhancement. For now, the implementation focuses on Belgium (Peppol pilot).

app/Libraries/UBLGenerator.php
jekkos: Good suggestions. These address line enhancements (additionalStreetName, addressLines) would make the implementation more robust. These could be added in a future phase. For Phase 1, keeping it simple.

app/Libraries/UBLGenerator.php
jekkos: Good catch on potential rounding errors. For Phase 1, I'm using float arithmetic which is acceptable for most use cases. A future improvement could use bcmath functions for higher precision arithmetic. This could be added in a follow-up PR.

app/Libraries/UBLGenerator.php
jekkos: Good point. The regex currently handles 4-5 digit postal codes. Adding comment documentation and supporting more formats (including UK postcodes with spaces) would be good future enhancements.

app/Views/sales/invoice.php
jekkos: Good security question. The sale_id should be validated against the logged-in user's permissions before returning the invoice. This is worth investigating - is there existing access control in place? If not, this should be addressed before merge. Added to the review todo.

composer.json
jekkos: Good catch! The PR description mentions UBL 2.1 but composer.json installs 2.4 (the current version). The code uses UBL 2.4 correctly. I'll update the PR description to reflect the correct version.

app/Libraries/Email_lib.php
jekkos: This blank line in Email_lib is outside the scope of this changeset. Good suggestion for a follow-up cleanup PR.

app/Libraries/Email_lib.php
jekkos: Refactoring result to success in Email_lib would be a broader change affecting other code. Outside the scope of this PR but noted for future cleanup.

app/Libraries/UBLGenerator.php
jekkos: Fixed! Added CountrySubentity (state/province) support to both supplier and customer addresses.

composer.json
jekkos: Fixed! UBL version is now set to 2.1 as intended.

app/Views/sales/invoice.php
jekkos: Fixed! Added validation to check if the sale exists before generating the UBL invoice. This prevents returning an invoice for a non-existent sale_id.

app/Libraries/UBLGenerator.php
jekkos: Fixed! Added comment explaining the regex pattern handles 4-5 digit postal codes (common European format). Noted that international formats like UK postcodes may need additional handling in future enhancements.

app/Libraries/UBLGenerator.php
jekkos: Fixed! Replaced float arithmetic with bcmath functions (bcadd, bcsub, bcmul, bcdiv, bccomp) for all tax calculations to prevent rounding errors.

app/Controllers/Sales.php
jekkos: Fixed! Added 'use App\Libraries\UBLGenerator;' to imports and replaced the fully qualified class name with the short class name. This follows the convention used for other library imports in the codebase.

app/Controllers/Sales.php
objecttothis: Sounds good to me.

app/Controllers/Sales.php
objecttothis: Sounds good.

app/Libraries/UBLGenerator.php
objecttothis: OK, that's fine.

app/Libraries/UBLGenerator.php
objecttothis: Super.

objecttothis (dismissed):

app/Libraries/Email_lib.php
objecttothis: Oh, OK.  I thought this was new code because it was all green in my view. I see now that other functions in the Email_lib are using the $result verbage as well.  Yeah, it can wait until later.

objecttothis (dismissed):

app/Language/sw-TZ/Sales.php
```diff
@@ -224,4 +224,5 @@
     "work_order_number_duplicate"      => "Nambari ya Agizo la Kazi lazima iwe ya kipekee.",
     "work_order_sent"                  => "Agizo la Kazi limetumwa kwa",
     "work_order_unsent"                => "Imeshindikana kutuma Agizo la Kazi kwa",
+    "sale_not_found"                   => "Mauzo hayapatikani",
```
jekkos: glm5 was even able to translate these strings automatically

objecttothis (dismissed):

app/Language/sw-TZ/Sales.php
objecttothis: I like that!

objecttothis: @jekkos after we push the PSR naming conventions PR to master, maybe we could create an agent task to fill in all the empty or incomplete translations in other phrases.

objecttothis:
@jekkos does this need to go out with 3.4.2 or can it go with the next release?  I would like to get 3.4.2 out so that we can merge some of these features.

jekkos:
I would wait for the next minor release as we already have a couple of things in flight now

PR #4394, opened 2026-03-05
Strip EXIF metadata from uploaded images

### LINKED ISSUE #4010: EXIF Geolocation Data Not Stripped From Uploaded Images
### Bug Description?

When a user uploads an Logo in "https://demo.opensourcepos.org/", the uploaded image’s EXIF Geolocation Data does not gets stripped. As a result, anyone can get sensitive information of "OSPOS" users like their Geolocation, their Device information like Device Name, Version, Software & Software version used etc.

### Steps to Reproduce?

1.     Go to [GitHub](https://github.com/ianare/exif-samples/tree/master/jpg).

2.     There are a lot of images with various resolutions (e.g., 1280 x 720) and different file sizes.

3.     Log in to the application and upload a photo as the Company Logo.

4.     Find the path of the uploaded image (either by right-clicking on the image and selecting "Copy image address" or by right-clicking, inspecting the image, and finding the URL in the inspect element tool).

5.     Open [Jimpl](https://jimpl.com/).

6.      Check if the EXIF data is still displayed.

### Expected Behavior?

All the Exif Metadata should be stripped while uploading image into the server

### OpensourcePOS Version

opensourcepos 3.3.9

### Php version

Php 7.2

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge, Other

### Server Operating System and version

OS: Windows 10

### Database Management System and version

Tested on the demo Website

### Web Server and version

Tested on the demo Website

### System Information Report (optional)

System Information Report

### Unmodified copy of OpensourcePOS

- [X] I agree this copy has not been modified

jekkos:
- Created Image_lib library to handle EXIF stripping for JPEG, PNG, GIF, and WebP images
- Uses GD library to re-encode images without EXIF data
- Added EXIF stripping to both company logo upload (Config controller) and item image upload (Items controller)
- Handles privacy concern by removing geolocation and device info from uploaded images

Fixes #4010

### DIFF
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -3,6 +3,7 @@
 namespace App\Controllers;
 
 use App\Libraries\Barcode_lib;
+use App\Libraries\Image_lib;
 use App\Libraries\Mailchimp_lib;
 use App\Libraries\Receiving_lib;
 use App\Libraries\Sale_lib;
@@ -353,6 +354,9 @@ class Config extends Secure_Controller
         ];
 
         $file->move(FCPATH . 'uploads/', $file_info['raw_name'] . '.' . $file_info['file_ext'], true);
+
+        $image_lib = new Image_lib();
+        $image_lib->strip_exif(FCPATH . 'uploads/' . $file_info['raw_name'] . '.' . $file_info['file_ext']);
 
         return ($file_info);
     }
--- a/app/Libraries/Image_lib.php
+++ b/app/Libraries/Image_lib.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace App\Libraries;
+
+class Image_lib
+{
+    public function strip_exif(string $filepath): bool
+    {
+        if (!file_exists($filepath)) {
+            return false;
+        }
+
+        $mimetype = mime_content_type($filepath);
+        $allowed_types = ['image/jpeg', 'image/jpg', 'image/png', 'image/gif', 'image/webp', 'image/bmp', 'image/tiff'];
+        
+        if (!in_array($mimetype, $allowed_types)) {
+            return false;
+        }
+
+        if ($mimetype === 'image/jpeg' || $mimetype === 'image/jpg') {
+            return $this->strip_exif_jpeg($filepath);
+        }
+        
+        if ($mimetype === 'image/png') {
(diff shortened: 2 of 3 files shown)


app/Libraries/Image_lib.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+namespace App\Libraries;
+
+class Image_lib
+{
+    public function strip_exif(string $filepath): bool
```
objecttothis, line 7: this function name needs to be refactored to stripEXIF() for PSR-12 compliance.

app/Libraries/Image_lib.php
```diff
@@ -0,0 +1,158 @@
...
+            return false;
+        }
+
+        $result = imagewebp($image, $filepath, 100);
+        imagedestroy($image);
+        
+        return $result;
+    }
+
+    private function strip_exif_fallback(string $filepath): bool
```
objecttothis, line 107: From what I can tell, this library will strip all EXIF data, but a good case can be made for not doing this.  For example copyright information probably shouldn't be stripped from an ethical standpoint.  What about using something like PHP Exif Library (https://github.com/FileEye/pel) to selectively strip rather than us implementing it all here?  The EXIF Tag reference (https://exifeditor.io/guides.html) shows you all the EXIF fields. Some of them are quite beneficial to keep.

app/Controllers/Config.php
```diff
@@ -354,6 +355,9 @@ private function upload_logo(): array
 
         $file->move(FCPATH . 'uploads/', $file_info['raw_name'] . '.' . $file_info['file_ext'], true);
 
+        $image_lib = new Image_lib();
+        $image_lib->strip_exif(FCPATH . 'uploads/' . $file_info['raw_name'] . '.' . $file_info['file_ext']);
```
objecttothis, line 359: In terms of Good, Better, Best here are my thoughts:
Best: This is an option to turn on/off in settings and a multi-select dropdown of which EXIF data to remove.
Better: This is an option to turn on/off in settings as some people will not want EXIF data stripped.
Good: This is not provided as an option but some data, light image dimensions, orientation and copyrights are kept.

objecttothis (changes requested):
Let me know what you think.

PR #4404, opened 2026-03-05
feat(plugins): Implement modular plugin system with self-registering events

jekkos:
## Summary

This PR implements a modular plugin system for OSPOS based on the design discussion in PR #4255. The key change is that **each plugin registers its own event listeners**, keeping the core code clean and free of hardcoded plugin dependencies.

### Core Infrastructure
- **PluginInterface** (`app/Libraries/Plugins/PluginInterface.php`) - Standard contract all plugins must implement
- **BasePlugin** (`app/Libraries/Plugins/BasePlugin.php`) - Abstract class with common plugin functionality
- **PluginManager** (`app/Libraries/Plugins/PluginManager.php`) - Discovers, loads, and manages plugins from `app/Plugins/` directory
- **PluginConfigModel** (`app/Models/PluginConfigModel.php`) - Model for plugin settings storage

### Example Implementations
- **ExamplePlugin** (`app/Plugins/ExamplePlugin.php`) - Simple demonstration plugin that logs events
- **MailchimpPlugin** (`app/Plugins/MailchimpPlugin.php`) - Refactored Mailchimp integration as a plugin

### Admin UI
- **Plugin Management Controller** (`app/Controllers/Plugins/Manage.php`) - Enable/disable/configure plugins
- **Plugin Management View** (`app/Views/plugins/manage.php`) - Admin interface for plugins

### Documentation
- **README** (`app/Plugins/README.md`) - Comprehensive documentation for plugin developers

### Event Flow
```
OSPOS Core → Events::trigger('item_sale', $data)
          ↓
PluginManager::registerPluginEvents() (post_system)
          ↓
Each enabled plugin registers listeners via Events::on()
          ↓
Plugin::onItemSale() is called with $data
```

### Key Design Decisions
1. **Self-registering plugins** - Each plugin calls `Events::on()` in its own `registerEvents()` method
2. **No hardcoded dependencies** - Core Events.php has no imports of plugin classes
3. **Modular enable/disable** - Plugins can be toggled via database settings
4. **Standard interface** - All plugins implement the same contract
5. **Plugin discovery** - Automatic scanning of `app/Plugins/` directory

## Creating a Plugin

```php
class MyPlugin extends BasePlugin
{
    public function getPluginId(): string { return 'my_plugin'; }
    public function getPluginName(): string { return 'My Plugin'; }
    
    public function registerEvents(): void
    {
        Events::on('item_sale', [$this, 'onItemSale']);
    }
    
    public function onItemSale(array $saleData): void
    {
        // Plugin logic here
    }
}
```

## Available Events
- `item_sale` - Fired when a sale is completed
- `item_return` - Fired when a return is processed  
- `item_change` - Fired when an item is created/updated/deleted
- `item_inventory` - Fired on inventory changes
- `items_csv_import` - Fired after items CSV import
- `customers_csv_import` - Fired after customers CSV import

## Testing
- Manual testing needed for plugin enable/disable
- Database migration creates `ospos_plugin_config` table
- Example and Mailchimp plugins demonstrate the pattern

## Related
- Based on discussion in #4255
- Milestone: 3.5.0


app/Controllers/Plugins/Manage.php
```diff
@@ -0,0 +1,111 @@
...
+        if (!$plugin) {
+            echo json_encode(['success' => false, 'message' => lang('Plugins.plugin_not_found')]);
+            return $this->response;
+        }
+
+        $settings = $this->request->getPost();
+        unset($settings['_method'], $settings['csrf_token_name']);
+
+        if ($plugin->saveSettings($settings)) {
+            echo json_encode(['success' => true, 'message' => lang('Plugins.settings_saved')]);
```
objecttothis: These probably should be changed from json_encode() calls to $this->response->setJSON() calls

app/Libraries/Receiving_lib.php
```diff
@@ -336,8 +336,8 @@ public function add_item(string $itemId, int $quantity = 1, ?int $itemLocation =
                 'is_serialized'              => $itemInfo->is_serialized,
                 'quantity'                   => $quantity,
                 'discount'                   => $discount,
-                'discount_type'              => $discountType,
-                'in_stock'                   => $this->item_quantity->get_item_quantity((int) $itemId, $itemLocation)->quantity,
+                'discount_type'              => $discount_type,
```
objecttothis, line 339: Was reverting to snake_case intentional here?

app/Plugins/MailchimpPlugin.php
objecttothis, line 1: @jekkos or @rmellmer can you help me understand the purpose of the plugin classes in `\app\Plugins\`? My understanding of the database architecture we landed on was to create a single plugin_config table, which we did in the branch here, but we also talked about making space for plugins to maintain a data table in the database as well. For example CASPOS, the plugin I will be porting into the master needs a data table for the API data that it sends for government cash registers required here in Azerbaijan.  If this is the architecture, then don't we need to maintain MVC design for plugins as well?  I think this was the original design for having the folders
```
\app\Controllers\Plugins\
\app\Models\Plugins\
\app\Libraries\Plugins\
\app\Helpers\Plugins\
```

If that is the case, then I'm fuzzy on which function should go here and which should go into MVC files for the plugins.

app/Models/PluginConfigModel.php
```diff
@@ -0,0 +1,135 @@
...
+namespace App\Models;
+
+use CodeIgniter\Model;
+
+/**
+ * Plugin Configuration Model
+ * 
+ * Manages plugin configuration stored in ospos_plugin_config table.
+ */
+class PluginConfigModel extends Model
```
objecttothis: IMO we should drop the work Model from the name for consistency

app/Views/configs/plugins_config.php
objecttothis, line 1: we might want to name this plugin_config for consistency with the other config views.

PR #4407, opened 2026-03-06
Modular Plugin System with Self-registering Events

### LINKED ISSUE #1: Plugin Infrastructure
### Type of Feature

✨ New Feature

### OpensourcePOS Version

opensourcepos 3.3.9

### Description

Events is the base feature needed for 3rd party integrations and plugins

### Additional Information

_No response_

### Verify you searched open requests in OpensourcePOS

- [x] I agree I have searched Open Requests

### LINKED ISSUE #3789: 3rd party integration standard architecture
With CI4 comes events and with events come the capability to have 3rd party integrations.  For example in my implementation I need two 3rd party integrations that communicate with external applications via RESTful API calls.  I know others have requested WooCommerce integrations, etc.  We need to come up with a standard home and structure for such integrations so that they are easy to distinguish from the base code and do not make the file structure too unruly.

I think we should stick to the MVC structure as much as possible here. Many integrations will not have their own view, unless we gave them a partial view in the config tabs, but some will require their own views.  In terms of Models, we could just create `app\Models\Integrations\` a la `\app\Models\Reports` and require 3rd party integration models to live there.  I don't see that we are doing something similar in `\app\Controllers\` but I imagine we could. Perhaps the same thing in `\app\Libraries`?

The other question is whether the code for API integrations should be done in a Model or Controller. I posted the question here (https://softwareengineering.stackexchange.com/questions/446878/do-i-use-a-model-or-controller-in-an-existing-mvc-application-for-implementing-a) and got good input but someone conflicting. One guy was adamant that IO communication with the external API should go in the model.  What about functions preparing the data to be sent to the API? Do those belong in a controller class?  For example, the government cash register/receipt printer integration will be called up during a sale triggered event and, given the sale_id, look up relevant data about the products and the sale, prepare the JSON, send the API request, process the response then store relevant data.

This brings in another subject about architecture of 3rd party integrations. These integrations are going to need to store data in the database. At minimum, configuration about the integration. That has traditionally been stored in the ospos_config table. Information needing to be stored about items can be stored in item attributes, so that isn't an issue. However, integrations that need to store data regarding a particular sale that aren't needed outside of that integrations is a question mark.  For example this government cash register API will send back a response with information such as a receipt number and UUID to reference that sale in the future.  Creating more columns in the ospos_sales table feels like it's going to clutter up the table when other integrations want to do the same thing.  So that either means we implement attributes tables for customers and sales a la items attributes or we give each integration needing to store data a table than they can use to link data.  For example a table for this government cash register would use the OSPOS sale_id as its index and then store other data it needed in columns of that table.

@jekkos et. al., can you weigh in on these architecture questions?

jekkos:
## Summary

This PR implements a modular plugin system for OSPOS, based on the design discussion in `#4255`. Plugins are self-contained: they register their own event listeners, ship their own MVC components, their own translations, and their own logs — keeping the core free of hardcoded plugin dependencies.

> **Note:** This description supersedes the original PR description, which described an earlier iteration of the design (e.g. `Plugin_config`, `ExamplePlugin`, `Controllers/Plugins/Manage.php`). See the discussion below for the evolution of the design.

## Core Infrastructure

- **`PluginInterface`** (`app/Libraries/Plugins/PluginInterface.php`) — contract every plugin must implement (`getPluginId()`, `getPluginName()`, `getPluginDescription()`, `getVersion()`, `registerEvents()`, `install()`, `uninstall()`, `isEnabled()`, `getConfigView()`, `getSettings()`, `saveSettings()`).
- **`BasePlugin`** (`app/Libraries/Plugins/BasePlugin.php`) — abstract base class providing settings persistence (`getSetting()`/`setSetting()`/`getSettings()`/`saveSettings()`, JSON-encoding arrays/objects) and plugin-scoped logging (`log()`/`logTo()`).
- **`PluginManager`** (`app/Libraries/Plugins/PluginManager.php`) — discovers plugin classes under `app/Plugins/`, validates them with `is_subclass_of(PluginInterface)`, and registers events only for plugins that are enabled. Namespaces are loaded during `pre_system` (`registerAllNamespaces()`) so plugin handlers can register before any event fires.
- **`PluginLogger`** (`app/Libraries/Plugins/PluginLogger.php`) — routes plugin log output to per-plugin log files (`WRITEPATH/logs/plugins/{pluginId}.log`) instead of the main application log, avoiding log spam as more plugins are installed. Exposed via the `log_plugin_message()` helper and the `log()`/`logTo()` methods on `BasePlugin`.
- **`PluginConfig`** (`app/Models/PluginConfig.php`) — model backed by the `ospos_plugin_config` table (`id`, `plugin_id`, `key`, `value`, `is_control`, timestamps, with a `UNIQUE KEY (plugin_id, key)`). Accessors are named `getValue()`/`setValue()` (not `get()`/`set()`) to avoid conflicting with `CodeIgniter\Model`.
- **Migration** — `20250531000000_PluginConfigTableCreate.php` creates the `ospos_plugin_config` table (with a `down()` rollback) and seeds the `plugins` module/permission/grants.

## Example Implementation

- **`MailchimpPlugin`** (`app/Plugins/MailchimpPlugin/`) — the working reference plugin, refactored from the previous core Mailchimp integration. Demonstrates settings storage, a self-contained `Language/{locale}/MailchimpPlugin.php` translation file with a `lang()` helper, and safe logging (customer ID logged instead of email).

## Internationalization

Plugins keep their own translations self-contained under `Plugins/{PluginName}/Language/{locale}/{PluginName}.php`, loaded via a `lang()` helper — no need to modify core `app/Language/` files. See `README.md` in `app/Plugins/` for the full pattern.

## Admin UI

- **`Plugins` controller** (`app/Controllers/Plugins.php`) — lists/searches plugins, and exposes enable/disable/uninstall/configure endpoints.
- **Plugin management view** — lets admins enable, disable, uninstall, and configure plugins from the UI.
- New `plugins` permission/module wired into `ADMIN_MODULES` and the menu.

## Design Notes

- **Event payloads are "thin" by default** — most business events (e.g. `item_saved`, `customer_saved`, `customer_loaded`) pass only the relevant ID(s); plugins query the model for anything else they need. This keeps events from becoming a permanent, ever-growing data contract, and lines up with a future REST API for plugins.
- **`customer_deleted` is a narrow exception** — it must carry `['id' => $customerId, 'email' => $email]` captured before deletion/anonymization, since the record (and the email specifically, when "enforce privacy" is enabled) is not reliably queryable afterward.

## Documentation

- `README.md` (under `app/Plugins/`) documents plugin structure, i18n, examples, and usage.

## Related

- Based on discussion in `#4255`
- Replaces `#4404` (closed due to merge conflicts)
- Milestone: 3.5.0


app/Plugins/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+Plugins are self-contained packages that can be installed by simply dropping the plugin folder into `app/Plugins/`:
+
+```
+app/Plugins/
+├── CasposPlugin/              # Plugin directory (self-contained)
+│   ├── CasposPlugin.php       # Main plugin class (required - must match directory name)
+│   ├── Models/                # Plugin-specific models
+│   │   └── Caspos_data.php
+│   ├── Controllers/           # Plugin-specific controllers
```
objecttothis: This architecture makes sense, but the structure in the directories doesn't seem to match what you have in the readme.  For example, I see `app\Controllers\Plugins\`, so what lives there if Controllers should live in `\app\Plugins\PLUGINNAME\Controllers`?

app/Plugins/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+```php
+<?php
+// app/Plugins/CasposPlugin/Models/Caspos_data.php
+
+namespace App\Plugins\CasposPlugin\Models;
+
+use CodeIgniter\Model;
+
+class Caspos_data extends Model
```
objecttothis: OK, so the plugin-specific tables are not done through a migration but here in code.  I suppose that's the only way to do it if we want plugins to be completely separate from the core code and drop-in installable.

app/Plugins/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+## Overview
+
+The OSPOS Plugin System allows third-party integrations to extend the application's functionality without modifying core code. Plugins can listen to events, add configuration settings, and integrate with external services.
+
+## Installation
+
+### Self-Contained Plugin Packages
+
+Plugins are self-contained packages that can be installed by simply dropping the plugin folder into `app/Plugins/`:
```
objecttothis, line 11: This mimics other webapp plugin interfaces and works well for simplicity.  One question that comes to mind is about code control. CMS like are rife with vulnerable plugins in part because this model makes it more difficult to make sure plugin code is very securely written. One option would be that we create a repo for each plugin that we validate and those repos would live in the opensourcepos organization.  This keeps the separate but maintains the vetting control.  This wouldn't stop people from creating their own plugins hosted privately and potentially even monetized plugins, but it does give us the peace of mind of saying "here are the plugins we have vetted. All other plugins are at your own risk."

app/Language/en/Plugins.php
```diff
@@ -0,0 +1,36 @@
+<?php
+return [
+    // Mailchimp Plugin
```
objecttothis: How are we doing languages with plugins.  Is that part the only part not self-contained?

app/Plugins/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+Plugin developers can package their plugins as zip files:
+
+```
+CasposPlugin-1.0.0.zip
+└── CasposPlugin/
+    ├── CasposPlugin.php
+    ├── Models/
+    ├── Controllers/
+    ├── Views/
```
objecttothis, line 1314: What happens when a plugin needs to inject an element into a core code view?  For example in my current production environment based on OSPOS 3.3.2, in the \views\sales\receipt.php view I have to inject a button into the view allowing the user to re-print the receipt on the governement cash register.  

```
<a href='javascript:void(0);'><div class='btn btn-info btn-sm' id='show_print_caspos_button'><?php echo '<span class="glyphicon glyphicon-print">&nbsp;</span>' . $this->lang->line('common_print_caspos'); ?></div></a>
```

Mailchimp, for example, has a tab in the customer manage view. I think it makes sense to only display these if the plugin is enabled, but I'm not sure how to inject these things but maintain all code in the plugin's folder structure.

Image

objecttothis (commented):
@jekkos I have a much better feel for how you have the architecture setup here and in general I like it, but I have a few questions regarding how multi-language strings, injecting plugin code into core code views and how to vet code security.  See my questions below

app/Plugins/README.md
jekkos: Indeed we will maintain a couple of plugins in this repo or separately, they might be upgraded and tests together with the main application. A version compatibitliy check will be required to see which ones are outdated, and then we can automatically disable them

app/Plugins/README.md
```diff
@@ -100,6 +100,123 @@ OSPOS fires these events that plugins can listen to:
...
+plugin_content_exists(string $section): bool
+```
+
+### Standard Hook Points
+
+Core views should define these standard hook points:
+
+| Hook Name | Location | Usage |
+|-----------|----------|-------|
+| `view:receipt_actions` | Receipt view action buttons | Add receipt-related buttons |
```
jekkos: @objecttothis same idea for the views. I guess we might want to 'register' partial view from the plugins and then inject them in the main components.
The idea is always that dependency goes from the plugin to the core logic. Never the other way around.

app/Plugins/README.md
```diff
@@ -0,0 +1,910 @@
...
+Plugins are self-contained packages that can be installed by simply dropping the plugin folder into `app/Plugins/`:
+
+```
+app/Plugins/
+├── CasposPlugin/              # Plugin directory (self-contained)
+│   ├── CasposPlugin.php       # Main plugin class (required - must match directory name)
+│   ├── Models/                # Plugin-specific models
+│   │   └── Caspos_data.php
+│   ├── Controllers/           # Plugin-specific controllers
+│   │   └── Dashboard.php
```
objecttothis: This folder structure is missing languages folders.  Was that on purpose?

app/Plugins/README.md
```diff
@@ -0,0 +1,910 @@
...
+}
+```
+
+### Complex Plugin (Self-Contained Directory)
+
+For plugins that need database tables, controllers, models, and views:
+
+```
+app/Plugins/
+└── CasposPlugin/                    # Plugin directory
```
objecttothis: Probably should include the languages folder in this structure chart as well for consistency.

app/Plugins/README.md
```diff
@@ -0,0 +1,910 @@
...
+
+```
+app/Plugins/
+└── ExamplePlugin.php              # Just logging - no models/controllers/views needed
+```
+
+### Complex Self-Contained Plugin
+
+```
+app/Plugins/
```
objecttothis: how is this example different from the complex plugin example above?  Maybe I'm missing something but this seems like a more complete version of the example above it.

app/Plugins/README.md
```diff
@@ -0,0 +1,910 @@
...
+            return '';
+        }
+        return view('Plugins/CasposPlugin/Views/customer_tab', $data);
+    }
+    
+    // ... rest of implementation
+}
+```
+
+Core views that want to support plugin hooks:
```
objecttothis: This was what I was looking for.  I think we should move it to just below the section on plugin views so that you see the full implementation of views picture together.

app/Plugins/README.md
```diff
@@ -0,0 +1,910 @@
...
+
+```php
+// app/Views/sales/receipt.php
+<div class="receipt-actions">
+    <?= anchor('sales/reprint/' . $sale_id, 'Reprint', 'class="btn btn-primary"') ?>
+    <?= plugin_content('view:receipt_actions', ['sale' => $sale]) ?>
```
objecttothis: we need to avoid snake case on this function call.

objecttothis (changes requested):
Take a look at my comments on the readme.  Languages is now much clearer to me.  Views is also cleaner this way because it limits plugin code.

app/Config/Events.php
```diff
@@ -8,23 +8,7 @@
 use App\Events\Db_log;
 use App\Events\Load_config;
 use App\Events\Method;
-
-/*
```
objecttothis, line 12: If we remove this it does clean it up, but it also makes the diff more chaotic when upgrading from one CI version to the next. The fewer changes to files in `app/Config/` the better in terms of those upgrades.  I'll let you decide what you want to do here.

app/Config/Events.php
```diff
@@ -39,16 +23,9 @@
         ob_start(static fn ($buffer) => $buffer);
     }
 
-    /*
```
objecttothis, line 46: Same thing with this comments

app/Database/Migrations/sqlscripts/3.4.1_PluginConfigTableCreate.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE `ospos_plugin_config` (
```
objecttothis: I think we should change this to
```
CREATE TABLE IF NOT EXISTS `ospos_plugin_config` (
    `key` varchar(100) NOT NULL,
    `value` text NOT NULL,
    `created_at` timestamp NOT NULL DEFAULT current_timestamp(),
    `updated_at` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
    PRIMARY KEY (`key`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
This will make the table addition defensive and more resilient to failures. The only caveat is that IF NOT EXISTS requires MySQL 8.0.13+ or MariaDB 10.5+.  I think these are already requirements though.  We should update the system requirements if they are not because both those versions are fairly old anyway.

app/Database/Migrations/20250531000000_PluginConfigTableCreate.php
```diff
@@ -0,0 +1,25 @@
...
+use CodeIgniter\Database\Migration;
+
+class PluginConfigTableCreate extends Migration
+{
+    /**
+     * Perform a migration step.
+     */
+    public function up(): void
+    {
+        error_log('Migrating plugin_config table started');
```
objecttothis: All logging should go through log_message() rather than some in error_log() and others in log_message().  This is both for consistency with the other migrations which use log_message() and for consistency with error logging in the rest of the code base.

app/Helpers/plugin_helper.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+use CodeIgniter\Events\Events;
+
+if (!function_exists('plugin_content'))
+{
+    function plugin_content(string $section, array $data = []): string
```
objecttothis: PSR-12 naming

app/Helpers/plugin_helper.php
```diff
@@ -0,0 +1,27 @@
...
+            return implode('', array_filter($results, fn($r) => is_string($r)));
+        }
+        
+        return is_string($results) ? $results : '';
+    }
+}
+
+if (!function_exists('plugin_content_exists'))
+{
+    function plugin_content_exists(string $section): bool
```
objecttothis: PSR-12 function naming

app/Language/en/Plugins.php
objecttothis: @jekkos I think you implemented plugin languages to be self-contained so these mailchimp plugin phrases need to be moved to their corresponding files in the plugin directory structure.

app/Libraries/Plugins/BasePlugin.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Default uninstall implementation.
+     * Override in subclass for custom uninstallation logic.
+     */
+    public function uninstall(): bool
+    {
+        return true;
+    }
+
+    /**
```
objecttothis: Remove useless PHPdocs in this file for consistency with the changes we are making to the rest of code.

app/Libraries/Plugins/PluginInterface.php
```diff
@@ -0,0 +1,96 @@
...
+ */
+interface PluginInterface
+{
+    /**
+     * Get the unique identifier for this plugin.
+     * Should be lowercase with underscores, e.g., 'mailchimp_integration'
+     */
+    public function getPluginId(): string;
+
+    /**
```
objecttothis: Remove useless PHPdocs in this file for consistency with changes we are making in the rest of the code.

app/Libraries/Plugins/PluginInterface.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public function getVersion(): string;
+
+    /**
+     * Register event listeners for this plugin.
+     * 
+     * Use Events::on() to register callbacks for OSPOS events.
+     * This method is called when the plugin is loaded and enabled.
+     * 
+     * Example:
```
objecttothis, line 21: Probably keep PHPdocs like this though that explain important information that can't be understood from the function name or code easily.

objecttothis (changes requested):
I got through 9 of the files.  I'll take a look at the rest later.

jekkos:
I've addressed all the remaining documentation and code style issues:

## Documentation Changes

1. **Consolidated duplicate sections** - Removed redundant "Example Plugin Structure" section since "Complex Plugin (Self-Contained Directory)" already covered the same content

2. **Moved Internationalization section** - Relocated to be right after "Plugin Views" as requested, so all view-related content is together

3. **Removed duplicate "Example: Plugin with View Hooks"** - The "View Hooks" section earlier in the doc already covers this content completely

4. **Fixed Language folder** in all structure examples (lines 20, 306)

5. **Converted snake_case to camelCase** for class names:
   - `Caspos_data` → `CasposData`
   - `Caspos_transaction` → `CasposTransaction`
   - `Api_client` → `ApiClient`

## Code Style Changes (PSR-12)

6. **plugin_helper.php** - Fixed brace placement (opening brace on same line)

7. **PluginInterface.php** - Removed unnecessary PHPDocs that just repeated the function signature, kept meaningful comments about Events registration

8. **BasePlugin.php** - Removed unnecessary PHPDocs, kept only documentation that adds value

9. **Migration file** - Changed `error_log()` to `log_message()` for consistency

10. **SQL migration** - Added `IF NOT EXISTS` to make table creation more resilient:
    ```sql
    CREATE TABLE IF NOT EXISTS `ospos_plugin_config` (
        `key` varchar(100) NOT NULL,
        ...
        PRIMARY KEY (`key`)
    ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
    ```

See commit 519347f for details.

app/Models/PluginConfig.php
objecttothis, line 1: Per PSR-12 rules this class should be named `PluginConfig`.  I understand we have some models using snake case, but those are existing and will be refactored at a later time to match PSR naming rules.

app/Plugins/CasposPlugin/Views/receipt_button.php
```diff
@@ -0,0 +1,4 @@
+<a href="javascript:void(0);" class="btn btn-info btn-sm" id="show_print_caspos_button" onclick="printCasposReceipt(<?= $sale['sale_id'] ?? '' ?>)">
+    <span class="glyphicon glyphicon-print">&nbsp;</span>
+    <?= esc(lang('Plugins.caspos_print_receipt') ?? 'Print Fiscal Receipt') ?>
```
objecttothis: CASPOS will be a plugin but this code is not functioning.  I think we should omit the CASPOS code from this PR and instead use Mailchimp as the example plugin.  I believe there is still space to make mailchimp an example because there is view injection in that too.  Currently mailchimp view code is found in app/Views/customers/form.php as a tab.  We should extract that code out and use it as an example.

```
        <?php if (!empty($mailchimp_info) && !empty($mailchimp_activity)) { ?>
            <li role="presentation">
                <a data-toggle="tab" href="#customer_mailchimp_info"><?= lang('Customers.mailchimp_info') ?></a>
            </li>
        <?php } ?>
```

app/Plugins/CasposPlugin/CasposPlugin.php
objecttothis, line 1: For simplicity lets just remove the CASPOS plugin entirely from this PR so that it's just the plugin framework code, documentation, migrations and the mailchimp plugin. That way we aren't giving a non-working example.

app/Plugins/ExamplePlugin.php
objecttothis, line 1: This is probably fine to keep because it gives an example of a simple single-file plugin.

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+namespace App\Plugins;
+
+use App\Libraries\Plugins\BasePlugin;
+use App\Libraries\Mailchimp_lib;
+use CodeIgniter\Events\Events;
+
+/**
+ * Mailchimp Integration Plugin
+ * 
+ * Example plugin that integrates OSPOS with Mailchimp for customer newsletter subscriptions.
```
objecttothis: While it can serve as an example, this is a functioning plugin, so I think we should reword this to ` * Plugin that integrates OSPOS with Mailchimp for customer newsletter subscriptions.`

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+            return ['success' => false, 'message' => 'API key not configured'];
+        }
+
+        $mailchimp = $this->getMailchimpLib(['api_key' => $apiKey]);
+        $result = $mailchimp->getLists();
+
+        if ($result && isset($result['lists'])) {
+            return [
+                'success' => true, 
+                'message' => 'API key is valid',
```
objecttothis: This should be using our translation string in the mailchimp language file `mailchimp_key_successfully`

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+                'lists' => $result['lists']
+            ];
+        }
+
+        return ['success' => false, 'message' => 'API key is invalid'];
```
objecttothis: This should be using our translation string in the mailchimp language file `mailchimp_key_unsuccessfully`

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * Test the Mailchimp API connection.
+     */
+    public function testConnection(): array
+    {
+        $apiKey = $this->getSetting('api_key');
+        
+        if (empty($apiKey)) {
+            return ['success' => false, 'message' => 'API key not configured'];
```
objecttothis: This needs a language string in the mailchimp plugin language file rather than being monolingual here.

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+        if (!$this->isEnabled() || !$this->shouldSyncOnSave()) {
+            return;
+        }
+
+        $this->log('debug', "Customer saved event received for ID: {$customerData['person_id']}");
+
+        try {
+            $this->subscribeCustomer($customerData);
+        } catch (\Exception $e) {
+            $this->log('error', "Failed to sync customer to Mailchimp: {$e->getMessage()}");
```
objecttothis: I might be wrong but I think we should be using log_message() here instead of log()

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+     * 
+     * @param array $customerData Customer information
+     */
+    public function onCustomerSaved(array $customerData): void
+    {
+        if (!$this->isEnabled() || !$this->shouldSyncOnSave()) {
+            return;
+        }
+
+        $this->log('debug', "Customer saved event received for ID: {$customerData['person_id']}");
```
objecttothis: log_message() instead of log()

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    public function getPluginName(): string
+    {
+        return 'Mailchimp Integration';
+    }
+
+    public function getPluginDescription(): string
+    {
+        return 'Integrate with Mailchimp to sync customers to mailing lists when they are created or updated.';
```
objecttothis: Do we want this to be translatable?  If so, we should create a translation string variable in the mailchimp language file.

app/Plugins/MailchimpPlugin.php
```diff
@@ -0,0 +1,210 @@
...
+    private ?Mailchimp_lib $mailchimpLib = null;
+
+    public function getPluginId(): string
+    {
+        return 'mailchimp';
+    }
+
+    public function getPluginName(): string
+    {
+        return 'Mailchimp Integration';
```
objecttothis: IMO we should just call it `Mailchimp`.  The context tells us it's a plugin or integration, so having the word Integration in the Plugin name seems redundant.

app/Plugins/README.md
objecttothis, line 1: This is really good documentation.  With this we might consider deleting app\Plugins\ExamplePlugin.php.  I don't feel strongly about it, so you decide.

objecttothis (changes requested):

jekkos:
I've addressed all the review comments on the MailchimpPlugin:

**Changes made:**

1. **Reworded docblock** - Changed from "Example plugin that integrates..." to "Plugin that integrates..." since this is a functioning plugin

2. **Renamed plugin** - Changed "Mailchimp Integration" to "Mailchimp" (the context makes it clear it's a plugin)

3. **Made strings translatable** - Used the `lang()` method with the self-contained language file:
   - `getPluginDescription()` now uses `->lang('mailchimp_description')`
   - `testConnection()` now uses `->lang('mailchimp_key_successfully')` and `->lang('mailchimp_key_unsuccessfully')`
   - Added `mailchimp_api_key_required` language string

4. **Used `log_message()` instead of `log()`** - For PSR-12 consistency with CodeIgniter conventions

5. **Added `getPluginDir()` and `lang()` methods** - For loading plugin-specific language strings

6. **Updated language file** - Added missing strings and fixed "MailChimp" to "Mailchimp" branding

Regarding the `plugin_helper.php` comments about PSR-12 naming - the function names `plugin_content` and `plugin_content_exists` use snake_case which is correct for CodeIgniter helpers (following framework convention). PSR-12 doesn't mandate camelCase for function names. The brace placement was fixed in the earlier commit.

See commit 6e99f05 for details.

jekkos:
I've removed the redundant `isEnabled()` checks from plugin callback methods.

## Explanation

Looking at `PluginManager::registerPluginEvents()` - the framework only calls `registerEvents()` for enabled plugins. This means:
- Disabled plugins never have their callbacks registered with `Events::on()`
- The callbacks won't be triggered at all for disabled plugins
- Therefore `isEnabled()` checks in callbacks are unnecessary

## Changes Made

### Documentation (README.md)
- Added note explaining that `isEnabled()` checks are not needed in callbacks
- Removed redundant `isEnabled()` checks from all example code

### Plugin Files
- **MailchimpPlugin.php**: Removed `isEnabled()` checks, kept setting checks (`shouldSyncOnSave`)
- **ExamplePlugin.php**: Removed `isEnabled()` checks, kept setting checks
- **CasposPlugin.php**: Removed `isEnabled()` checks, kept setting checks (`show_receipt_button`)
- Used `log_message()` instead of `log()` for PSR-12 consistency

## When to Keep the Check

The only time you need a check in a callback is when checking **plugin-specific settings**, not enabled status:

```php
public function onCustomerSaved(array $customerData): void
{
    // Checking a FEATURE setting - this is needed
    if (!$this->shouldSyncOnSave()) {
        return;
    }
    // No isEnabled() check needed - framework handles it
}
```

See commit 10a64e7 for details.

objecttothis (dismissed):
The rest of the comments are optional.  Feel free to do them if you want or ignore them.  Either is fine.

jekkos:
I've addressed the remaining review comments:

## Changes Made

### Breaking Changes
1. **Renamed `Plugin_config` to `PluginConfig`** - PSR-12 class naming (camelCase instead of snake_case)

### Removed Non-Functioning Code
2. **Removed `app/Plugins/CasposPlugin/`** - This was a non-functioning example. The `MailchimpPlugin.php` serves as the working example.

3. **Removed `app/Plugins/ExamplePlugin.php`** - Redundant since MailchimpPlugin demonstrates a single-file plugin structure (though it will be moved to a directory to support language files).

### Privacy Fix
4. **Fixed privacy issue in MailchimpPlugin** - Changed logging from customer email to customer ID:
   `log_message('info', "Successfully subscribed customer ID {$customerData['person_id']} to Mailchimp")`

### Code Cleanup
5. **Removed unnecessary PHPDocs** from `PluginManager.php` and `PluginConfig.php` for PSR-12 consistency

### Remaining Work
The README.md still references CasposPlugin in examples. Should I update it to use MailchimpPlugin as the working example, or would you prefer a different approach?

See commit eb264ad for details.

jekkos:
I've addressed the CodeRabbit AI review comments:

## Changes Made

### Events.php - Plugin Discovery Timing
- Moved plugin discovery from `post_system` to `pre_system` so plugin handlers can register before events like `item_sale` fire during request processing

### PluginManager.php - Security & Validation
- Added plugin existence check in `disablePlugin()` before writing to config
- Added `is_subclass_of($className, PluginInterface::class)` check before instantiating discovered classes (prevents running arbitrary constructors from any PHP file)

### BasePlugin.php - Value Serialization
- Fixed `saveSettings()` to JSON-encode arrays/objects (was casting to string, resulting in "Array")

### PluginConfig.php - Prefix Removal
- Fixed `getPluginSettings()` to use `str_starts_with` + `substr` instead of `str_replace` (prevents incorrect key stripping when plugin ID appears elsewhere in key name)

### Migration - Rollback Support
- Added `down()` method to drop the table on rollback

### README.md - Documentation
- Replaced all CasposPlugin examples with MailchimpPlugin as the reference implementation
- Simplified to focus on working example

See commit 896ed87 for details.

objecttothis (dismissed):

app/Config/Events.php
```diff
@@ -48,13 +48,16 @@
...
-        // Hot Reload route - for framework use on the hot reloader.
         if (ENVIRONMENT === 'development') {
             service('routes')->get('__hot-reload', static function (): void {
                 (new HotReloader())->run();
             });
         }
     }
+
+    $pluginManager = new PluginManager();
+    $pluginManager->discoverPlugins();
```
objecttothis: This is causing a problem @jekkos.  The plugin manager is being called before migrations is called which leads to an error message because the plugin_config table doesn't exist.  We need to make sure migrations run before this happens.

jekkos: Correct also saw this yesterday. we probably need to add a check here to see if migration was done. Or plug it in elsewhwere

objecttothis: Agreed. I'll take a look. I think #4384 is done (just needs a code review from you), so this PR is my main priority right now.

objecttothis: Fixed

objecttothis:
Remaining Todo:
- [x] Create UI for plugin config views to be autoloaded as tabbed views under Configuration > Setup & Conf > Integrations
- [x] Rename Integrations to Plugins in config language file for consistency.
- [x] Move strings from all languages Config, Customers and Sales language files to the mailchimp plugin.
- [x] Rename customer_mailchimp_status phrase to mailchimp_customer_status for consistency.
- [x] Resolve issue where the \app\Plugins\MailchimpPlugin\Views\config.php view renders at the top of every page in the webapp.
- [x] Move Customers controller mailchimp logic to plugin. (constructor, addOrUpdateMember function)
- [x] Move mailchimp_lib to the plugin.
- [x] Rework Mailchimp_lib functions to use Plugin resources instead of core code.
- [x] rework the integrations config view by renaming it to plugins_config and edit the content to properly load integrations config files.
- [x] Write partial view to inject the mailchimp customer status into the customer manage view form.
- [x] Write the onsubmit logic for the customer view form to store the customer status with the customer_id.  This means registering the event and the listener in the mailchimp plugin.
- [x] Update language strings in Weblate so they don't keep getting pushed to the repo in the incorrect locations.
- [x] Possibly configure Weblate to handle translations of plugin language files for plugins which we maintain (e.g., Mailchimp)
- [x] Run testing (integration, unit, regression)

objecttothis:
@jekkos I have a design question for you.
- We have app\Views\plugins\manage.php which implies that the intent is that plugins is a module like Items, Customers, Attributes, etc.
- I don't see a migration adding this row to the modules table, so if my understanding is correct, then I need to write that migration script.  I'll probably include it into the existing migration.  Can you clarify that this was the intent?
- If so, we need an SVG in \public\images\menubar\ to represent Plugins.  I don't know how we generated the ones we have, so it may be difficult to get the theme of a plugin svg icon correct.  Did we pull these SVGs from a specific pool?

objecttothis:
> @jekkos I have a design question for you.
> 
>     * We have app\Views\plugins\manage.php which implies that the intent is that plugins is a module like Items, Customers, Attributes, etc.
> 
>     * I don't see a migration adding this row to the modules table, so if my understanding is correct, then I need to write that migration script.  I'll probably include it into the existing migration.  Can you clarify that this was the intent?
> 
>     * If so, we need an SVG in \public\images\menubar\ to represent Plugins.  I don't know how we generated the ones we have, so it may be difficult to get the theme of a plugin svg icon correct.  Did we pull these SVGs from a specific pool?

Nevermind, I found them here: https://github.com/opensourcepos/elegant-circles from the gulpfile

objecttothis:
@jekkos maybe you could take a look at https://commons.wikimedia.org/wiki/Category:Elegant_Themes_Circle_Icons and tell me which one you think works best to represent the Plugins module?  So far I think probably https://commons.wikimedia.org/wiki/File:Circle-icons-plugin.svg.  I would prefer a puzzle piece I think, but I didn't see one.

objecttothis:
@coderabbit review

objecttothis:
@jekkos I think we are ready to merge this.  I've run this through code rabbit and claude for code reviews.  Can you do a code review and approve if there are no problems?

objecttothis (dismissed):

objecttothis:
@jekkos Since you were the one that opened the PR, it won't probably let you approve it.  I approved it, but please let me know if you see any unresolved problems.  Don't be alarmed at the number of files changed.  Most of them are the language string files.  There are only about 40 files of any consequence.  This PR is adding the plugin framework and the first plugin (mailchimp).

objecttothis (dismissed):

jekkos:
Great thanks for pushing this further. I will try to have a look next week when I am back home.

objecttothis:
> Great thanks for pushing this further. I will try to have a look next week when I am back home.

Sounds good. In the meantime I will be working on two plugins in private repos since they are needed and it's not likely there will be any wild changes to the architecture. With the MailChimp plugin I created an issue because changing the email address of a customer in MailChimp it treats it like a new address and adds it instead of updates it. I left the code the way it was because that's the current behavior.

objecttothis:
@jekkos do you still think you can get to a code review this week?

jekkos:
Hi yes, but before we merge i would like to push out a minor release, so was trying to solve the last remaining issues in master first. Ok to hold this one back a bit or wdyt

objecttothis:
> Hi yes, but before we merge i would like to push out a minor release, so was trying to solve the last remaining issues in master first. Ok to hold this one back a bit or wdyt

That's OK to wait to merge until after the minor release.  The reason I'm wanting to be done with the code reviews on this is that I'm already developing against this branch with a plugin that I need, so if there are changes needed to this branch before merging, I want to address those since it could affect my dependent code.  How long do you think it will be before the minor release is ready?

objecttothis:
> > Hi yes, but before we merge i would like to push out a minor release, so was trying to solve the last remaining issues in master first. Ok to hold this one back a bit or wdyt
> 
> That's OK to wait to merge until after the minor release. The reason I'm wanting to be done with the code reviews on this is that I'm already developing against this branch with a plugin that I need, so if there are changes needed to this branch before merging, I want to address those since it could affect my dependent code. How long do you think it will be before the minor release is ready?

@jekkos to clarify what I'm asking in case it isn't clear: can you do the code review soon, then we merge after the minor release is done?

jekkos:
Understood, will have a look today or tomorrow.

app/Controllers/Plugins.php
```diff
@@ -0,0 +1,168 @@
...
+
+    public function postUninstall(string $pluginId): ResponseInterface
+    {
+        if ($this->pluginManager->uninstallPlugin($pluginId)) {
+            return $this->response->setJSON(['success' => true, 'message' => lang('Plugins.uninstalled')]);
+        }
+        return $this->response->setJSON(['success' => false, 'message' => lang('Plugins.uninstall_failed')]);
+    }
+
+    public function getConfig(string $pluginId): ResponseInterface
```
objecttothis, line 121: @jekkos our current design here has a limitation in that other than `$settings`, we can't pass data to the config view.  For example in my CLCDesq plugin which sends data to an API endpoint in CLCDesq (our website CMS). The config view needs to have the user select bindings to bind API variables to attribute_definitions. This means I need an associative array of attribute_definition.definition_id => attribute_definition.definition_name in the view in order to populate the dropdowns of which attribute definitions can be bound to each setting.

Bad: Hard code or do business logic inside the view.
Good: AJAX call in the config view to a plugin controller which fetches the data needed. This is suboptimal because it doesn't run until after the view is in the browser, but it works.
Better: Rework this and add a getConfigData() function. Then each plugin can override getConfigData() to pass whatever data is needed in an array and the view can be rendered in the same convention the rest of the plugin views and CI views render.
Best: Maybe you have a better idea?

objecttothis: I went ahead and implemented the "Better" option and it works pretty well.  If you have another idea, let me know.  I'll push the commit to this branch.

objecttothis:
@jekkos I just pushed the rest of the commits for this PR.  Can you take a look at the code.  Since you authored the PR, I don't know if it allows you to submit a code review on it or not, but if not you can just make comments and I'll address them. We will merge AFTER the next minor update but I'd like to get the code review out of the way so that I can fix anything we can find and have it ready to merge first thing after the minor release.

jekkos:
The branch was now deployed on the dev server but it does not migrate claenly, I saw a couple of issues
first Error: Migration failed: Duplicate column name 'stock_type' 
then another error. Not sure what went wrong but it might still need a bit of testing

objecttothis:
> The branch was now deployed on the dev server but it does not migrate claenly, I saw a couple of issues first Error: Migration failed: Duplicate column name 'stock_type' then another error. Not sure what went wrong but it might still need a bit of testing

hmmm. Is there a place that I can see those logs? I don't get those errors locally.

objecttothis:
> The branch was now deployed on the dev server but it does not migrate claenly, I saw a couple of issues first Error: Migration failed: Duplicate column name 'stock_type' then another error. Not sure what went wrong but it might still need a bit of testing

@jekkos without seeing the build log or how the deployment is happening, It's difficult for me to say exactly what is going on but I'm guessing that since the errors don't occur locally for me, it's a difference in the workflow. Locally I'm testing against MySQL 8.4.3, PHP 8.3.30 and Apache 2.4.62. None of the Plugin code touches stock_type migrations, so I don't think this issue is caused by the PR but rather potentially something merged into the PR from the master.  I know we recently reworked the way the initial schema is loaded.  Is it possible that this isn't working in all workflows?  I'll test this PR locally against MariaDB tomorrow in case the issue is some difference between those two DBMSs.

Claude is telling me that the issue is likely in the 3.0.2 to 3.1.1 SQL migration. It thinks `IF NOT EXISTS` clauses need to be added to the `ADD COLUMN` scripts.  While I think that change needs to happen, and we can make it happen here, I think we need to make sure we are using as unified of a deployment workflow across the board, so I'm still interested in what the differences are between the dev deployment workflow and mine.

jekkos:
It should be easily reproducible on dev, every time you push it will deploy that bracnh on the server. It does a clean setup for that code version for which you can run all migrations. Dev server is currently using mariadb so there are some differences indeed. I'm not sure what the state of the MySQL 8.4 PR is but I believe it was about to introduce a CI job to run migrations against a MySQL / MariaDB version matrix?

objecttothis:
> It should be easily reproducible on dev, every time you push it will deploy that bracnh on the server. It does a clean setup for that code version for which you can run all migrations. Dev server is currently using mariadb so there are some differences indeed. I'm not sure what the state of the MySQL 8.4 PR is but I believe it was about to introduce a CI job to run migrations against a MySQL / MariaDB version matrix?

Do I have access to the logs for the deployment on dev? In the required checks it says three were skipped and everything else passed including the Build and Release Tasks. I'll test locally against MariaDB in the morning, but I should be able to reproduce and troubleshoot if I know where to see the logs showing me the errors.

image

objecttothis:
> > It should be easily reproducible on dev, every time you push it will deploy that bracnh on the server. It does a clean setup for that code version for which you can run all migrations. Dev server is currently using mariadb so there are some differences indeed. I'm not sure what the state of the MySQL 8.4 PR is but I believe it was about to introduce a CI job to run migrations against a MySQL / MariaDB version matrix?
> 
> Do I have access to the logs for the deployment on dev? In the required checks it says three were skipped and everything else passed including the Build and Release Tasks. I'll test locally against MariaDB in the morning, but I should be able to reproduce and troubleshoot if I know where to see the logs showing me the errors.
> image

@jekkos I'm looking into this now.  This is very clear that what broke were changes to the code outside of this PR.  It only is affecting this branch because I merged the latest master into the branch to keep it up to date. I'll fix them but I'm not happy about needing to fix other people's mess just to get this code review and merge after the minor release.

objecttothis:
To clarify, it's not that I've never inadvertently released bugs into the master. It happens to everyone. I'm just frustrated that it's holding up your willingness to review my code. I'll check out a new branch from the master and push commits to fix these problems in a separate PR.

jekkos:
Ok I did not see these issues in master, so as you said it might be due to the database version you are on.. I thought 8.4 was not fully supported yet. 
 Ill have a look today did not have much time as got sick over the weekend, and am back to work now

jekkos:
Maybe the issue was there for a long time but not visible because we were not using ajax? Now error messages are displayed in the Ui it might not have been the case before

objecttothis:
> Maybe the issue was there for a long time but not visible because we were not using ajax? Now error messages are displayed in the Ui it might not have been the case before

It's possible. I'll get started on it now and submit it in a new PR.  I would like to merge that PR right away after code review though.  I'm not a fan of working with multiple dependency branches.

objecttothis:
> Ok I did not see these issues in master, so as you said it might be due to the database version you are on.. I thought 8.4 was not fully supported yet. Ill have a look today did not have much time as got sick over the weekend, and am back to work now

I've been using MySQL 8.4.3 for a long time.  I did see the #4471 for MySQL 8.4 support, and if I remember correctly, they were good changes, but I've been using MySQL without errors for a while.  I'll see if the dev wants to finish of the changes in his PR. If not, I'll probably just make them so we can merge that one.

I'm sorry you weren't feeling well.  I'm glad you're doing better now though.  Do you think you can give this PR a review?

app/Controllers/Customers.php
```diff
@@ -180,77 +172,37 @@ public function getView(int $customer_id = NEW_ENTRY): string
...
-                        $data['mailchimp_activity']['total'] = $total;
-                        $data['mailchimp_activity']['open'] = $open;
-                        $data['mailchimp_activity']['unopen'] = $unopen;
-                        $data['mailchimp_activity']['click'] = $click;
-                        $data['mailchimp_activity']['lastopen'] = $lastopen;
-                    }
-                }
-            }
-        }
+        Events::trigger('customer_loaded', $info);
```
jekkos: We have to think carefully what data we want to pass for each event, as it will be an interface specification that might become expensive to change once defined.

 We might want to keep datastructure backward compatible

objecttothis: Agreed.  Here are my thoughts:
- We need to include as little information as is necessary for the plugins to function.
- Ideally plugin code would not access core code functions directly but we do not currently have an API interface for plugins to query.  This means two choices:

1. We pass the Plugin everything it needs in the Event trigger.
2. We pass minimal data in the event trigger and allow core code function access now and API access later once that is developed.

In this case the trigger is not currently being used by any plugins. We can change this to just passing the $customerId if we want to choose option 2.  What do you think?

jekkos: I would try to limit direct core code dependencies as that will create issues down the line. whenever we need a deoendency we should distinguish internal from external api, in this case the interface is the event trigger. 

We can check case by case what context is available and pass what we have available in the event callback. eg for insert or save we can easily pass the whole object. For delete the id only might be passed.

A plugin should be able to access models to run other queries..not sure how this works in php but perhaps there is a way to flag a method for public use. Then the signature for these methods cannot change easily anymore unless all plugin code is adapted

app/Models/PluginConfig.php
```diff
@@ -0,0 +1,108 @@
...
+        if ($this->exists($key)) {
+            return $builder->update(['value' => $value], ['key' => $key]);
+        }
+
+        return $builder->insert(['key' => $key, 'value' => $value]);
+    }
+
+    public function getPluginSettings(string $pluginId): array
+    {
+        $builder = $this->db->table('plugin_config');
```
jekkos, line 61: Wonder maybe instead of the underscore we can use a column to identify the plugin id?

CLAUDE.md
```diff
@@ -0,0 +1,104 @@
...
+- Can include its own `Views/`, `Models/`, `Controllers/`, and `Language/` subdirectories
+- Configuration is stored in the `ospos_plugin_config` table
+- See `app/Plugins/README.md` and `AGENTS.md` for full plugin development docs
+
+### Frontend Build
+
+`gulpfile.js` (Gulp 5) copies vendor CSS/JS from `node_modules/` into `public/resources/`. Running `npm run build` is required after installing npm packages or changing gulp tasks. Do not manually edit files under `public/resources/`.
+
+## Code Style
+
```
jekkos: can claude also use agents.md? We might have both now maybe we dan try to consolidate

app/Controllers/Customers.php
jekkos: Still some files in build folder that have creeped in

jekkos: Im quite happy with where this is going. The overall design looks good..the only thing that we did not explore in depth is how we will pass the data and where we are putting the trigger..

Eg do we also want to trigger a failed save? In that case we need to trigger the event at the very end of the method, pass the success status and data object.

For now I do not see a use case for that so maybe we should not overcomplicate.

jekkos: Beside the three remarks I will approve if we want to avoid too many merge conflicts then it might be best to merge this one after next minor release. There is still one issue on the install that I would like to sort out but beside that we should be good

CLAUDE.md
objecttothis: Claude reads CLAUDE.md so there are two options.
1- The contents of CLAUDE.md are just "Go read AGENTS.md".  The advantage is seemless use of both and a unified set of instructions to maintain. The disadvantage is that it means tracking both files in the repo.
2- Add CLAUDE.md to the .gitignore and each dev using claude needs to maintain their own CLAUDE.md for the project. The advantage is that it's one less file to track in the repo. The disadvantage is that it's more work for devs using claude and potential inconsistencies introduced in that individual devs may not be using the same instructions regarding the coding standards, etc.

What do you think?

app/Controllers/Customers.php
objecttothis: > Beside the three remarks I will approve if we want to avoid too many merge conflicts then it might be best to merge this one after next minor release. There is still one issue on the install that I would like to sort out but beside that we should be good

Yes, I think we should merge immediately after the next minor release. It avoids conflicts and it gives us a whole development cycle to work out any kinks.

objecttothis: > Still some files in build folder that have creeped in

I added the folder to the .gitignore but it was after it pushed the files.  I'll push a fix.

objecttothis: > Im quite happy with where this is going. The overall design looks good..the only thing that we did not explore in depth is how we will pass the data and where we are putting the trigger..
> 
> Eg do we also want to trigger a failed save? In that case we need to trigger the event at the very end of the method, pass the success status and data object.
> 
> For now I do not see a use case for that so maybe we should not overcomplicate.

Saves need to be triggered only after the internal save is complete.  I currently don't see a use case for triggering on failed internal save, but we can revisit if someone requests the trigger.

objecttothis: > I would try to limit direct core code dependencies as that will create issues down the line. whenever we need a deoendency we should distinguish internal from external api, in this case the interface is the event trigger.
> 
> We can check case by case what context is available and pass what we have available in the event callback. eg for insert or save we can easily pass the whole object. For delete the id only might be passed.
> 
> A plugin should be able to access models to run other queries..not sure how this works in php but perhaps there is a way to flag a method for public use. Then the signature for these methods cannot change easily anymore unless all plugin code is adapted

In my mind, the best solution is to only allow access to external API endpoints and no internal code access, but we don't have external API endpoints yet, so allowing access to public model functions and helpers until then is my initial thought.

Regarding how much information is passed to the trigger, my thought is that when we have external API endpoints setup we only will want to pass the information the plugin needs to retrieve what got changed through the API.  For example when an item is saved, that means just the array of itemIds.  The reason for minimal data being passed is that once we have external APIs, we would likely want to go to passing just the id fields, but like you said earlier, if we take away all the other data we were passing to the event handler, then that is a breaking change.

I suspect there is a wide range in what plugins need. For example, the mailchimp plugin needs very little but the CLCDesq plugin needs item, attribute, quantity, supplier and tax data. It potentially adds a lot of overhead to the core code to add that in before sending it to the event handler.

One argument I can see for passing more data though is if we are already pulling the data for the internal code, then it saves database queries to have plugins pulling the data again.

The biggest argument I can see for NOT sending more than the minimal data though is maintainability and support of the plugin system. If developers are constantly asking for changes to what we pass, that is not super maintainable but if we say "Here is all you need to get the data yourself through the API endpoints (or temporarily model function calls)" then its up to them to maintain their own data and code.

This is also the reason why I think our roadmap should include external API endpoints.  You made a good point that letting plugin devs use model functions means we can't easily change those function signatures without breaking other people's code.  If we provide API endpoints then it just becomes the endpoints themselves that do not change and the underlying code is free to change as long as the API delivers the correct results.

app/Models/PluginConfig.php
objecttothis: @jekkos are you thinking adding a column to the plugin_config table to indicate which pluginId it's associated with?  We can definitely do that. Right now plugin_config has both plugin settings and plugin associated settings (enabled and disabled).  We differentiated the two with two underscores instead of one underscore.  Potentially a better way to do this though is a column to flag whether the setting is a plugin system setting like enabled or disabled or it's a plugin setting.  This would remove the need for underscores all together.  What do you think?

build/.phpunit.cache/test-results
objecttothis, line 1: @jekkos These files are actually in the master now.  My branch is only deleting and untracking them.

CLAUDE.md
objecttothis: @jekkos for now I have unified CLAUDE.md and AGENTS.md and stubbed out CLAUDE.md to make Claude Code read AGENTS.md.  This way we only need to maintain AGENTS.md.  Unfortunately I'm not aware of a way to configure Claude Code to read AGENTS.md and skip CLAUDE.md.  If you absolutely do not want CLAUDE.md in the repo then I could add it to the .gitignore and keep it locally, but I'd prefer to keep the stub in the repo.

app/Models/PluginConfig.php
objecttothis: @jekkos I pushed a refactor that creates `is_control` and `plugin_id` columns and refactors this to use that.  Much less hacky this way.  Let me know if you had something else in mind.

jekkos: That was it thanks for the fix

CLAUDE.md
jekkos: Ok also was that it is resolved like this typically.

objecttothis:
@jekkos regarding the data passing from triggers conversation, here is what Claude thinks.  Maybe @coderabbitai can weigh in too:

Event::trigger() Data Strategy
Context
User debating how much data to pass in Events::trigger() calls across the plugin system. Current state is inconsistent:

item_saved → passes only [$itemId] (user's preferred minimal approach)
customer_saved → passes $personData, $customerData, full $postData (fat approach)
customer_deleted → passes full $customer stdClass (fat approach)
customer_loaded → passes only $customerId (minimal approach)
Future goal: external API endpoints where plugins should NOT access core models/libraries directly.

The Core Tradeoff
Fat events (pass full data objects)
Pro: No extra DB calls — data already in memory
Con: Creates permanent API contract. Every field you pass becomes something plugins depend on. Plugins will request more fields constantly. When you add the REST API layer, you'll have two parallel data-access patterns (event data AND API) that must stay in sync. Harder to refactor core data structures without breaking plugins.

Thin events (pass only IDs)
Pro: Clean separation. Plugins fetch what they need. Natural path to REST API — same endpoints plugins will call later. No contract bloat. Core data structures can change freely.
Con: Extra DB queries per plugin per event. N plugins = N extra queries on hot paths (sales, item saves).

Recommendation: Thin events (IDs only)
User's instinct is correct. Reasoning:

The "more data" requests will never stop. Mailchimp wants name+email. A loyalty plugin wants purchase history. A shipping plugin wants address. You cannot satisfy all plugins without passing everything — at which point you've reinvented a fat DTO that you maintain forever.

The API migration path is clean. When external API endpoints exist, thin-event plugins need zero refactoring — they just swap model->get() calls for http->get() calls. Fat-event plugins would need surgery.

DB cost is real but bounded. A plugin doing one extra customer->get($id) query is cheap. The POS is not a high-concurrency system — it's a single-store app. The query cost is negligible vs. the maintenance cost of a fat event contract.

What to standardize
All business events should pass only the minimum identifier(s):

Event	Pass
customer_saved	$customerId (int)
customer_deleted	$customerId (int)
customer_loaded	$customerId (int) ✅ already correct
item_saved	$itemId or [$itemIds] ✅ already correct
Plugins that need data (like Mailchimp) call the model directly now, call the REST API later.

objecttothis:
@coderabbitai correct. Claude addressed this in another prompt.  When enforce privacy is enabled, this is a problem, so on customer deleted the email address needs to be what is sent.

app/Controllers/Customers.php
jekkos: If we pass whatever we have in memory to the event trigger,  the plugin will get for free. But donwside is that it is harder to version and things might break more easily..

As you suggest using the api will shield us more as it can use versioning schemes.. So in that sense it might be a better decoupling . the minimal approach might become the preferred one

jekkos: Btw i will release tomorrow no remaining issues that I am aware of

objecttothis: > If we pass whatever we have in memory to the event trigger, the plugin will get for free. But donwside is that it is harder to version and things might break more easily..
> 
> As you suggest using the api will shield us more as it can use versioning schemes.. So in that sense it might be a better decoupling . the minimal approach might become the preferred one

OK, let's pass minimal data to the event trigger callbacks for now.  It's far easier to change our mind later and pass more than it is to pass less. Well developed plugins should experience very little performance hit. If we find that there is a large performance hit, then more than likely it's with the database, indices and queries. The CLCDesq plugin I'm working on is very query intensive, so that is a good litmus test.

objecttothis: > Btw i will release tomorrow no remaining issues that I am aware of

I'll merge this after the release.

objecttothis:
@jekkos and @coderabbitai please comment.  In thinking about how Plugins are going to be used, we are going to end up with an immense amount of log spam, so in keeping with the self-contained nature of plugins I committed https://github.com/opensourcepos/opensourcepos/pull/4407/changes/7dde48c0edf47ee31154b4889b1ab44fb36258f5 with an implementation of log_plugin_message() and adjusting log() and logTo() to point at that logger specifically for plugins. 

Usage example:

`log_plugin_message('mailchimp', 'error', 'Mailchimp API Response is invalid due to key: ' . $apiKey, 'api');`
`log_plugin_message(string $pluginId, string $level, string $message, string $logFileName = null);`
Inside a Class which implements BasePlugin you can just call 'log(level, message);` or `logTo(level, message, logFileName);`

objecttothis:
@coderabbitai I committed PluginLogger.php.  It had not been added to git and I missed it. I'll look at your other suggestions tomorrow.

objecttothis:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> 
> `@objecttothis` — `PluginLogger.php` looks solid. The `sanitize()` method covers path traversal (strips everything except alphanumerics and hyphens), `flock()` handles concurrent writes safely, and the fallback to `log_message('warning')` is a good touch. The `log()` parameter order also aligns with `log_plugin_message()`.
> 
> Two remaining items from yesterday's review for when you get a chance:
> 
>     1. **`PluginConfig` schema mismatch** — the model appears to query against a `plugin_id` column (e.g., `$builder->getWhere(['plugin_id' => $pluginId, 'key' => $key])`) but the SQL migration only creates `key` and `value` columns. The migration will need updating to match.
> 
>     2. **`logTo()` parameter order in `BasePlugin`** — currently `logTo(string $logName, string $level, string $message)` puts the log name first, which is the reverse of `log_message($level, $message)` convention. Minor, but worth aligning for consistency.
> 
> 
> No rush — take a look whenever you're ready.

@coderabbit a plugin_id column was added so that we are matching the plugin by a separate column instead of doing string parsing.  That table now has id, plugin_id, key, value, is_control, created_at and updated_at columns.  Parsing the key to determine if it was a control, what plugin it was for and the key was problematic.

I pushed a commit to resort the parameters in log(), logTo() and log_plugin_message() so that they match the general order of log_message() consistently.

objecttothis:
@jekkos where do things stand with the 3.4.2 minor release? I would like to merge this branch soon, but I'm waiting until a bit after 3.4.2 release.

objecttothis (dismissed):

objecttothis:
@jekkos it seems we overlooked plugin-specific table migrations.  Since plugins can't drop migrations into the core folder, I'm implementing a simplified CI3-style migration for plugins.  It will check for plugin migrations on login like our regular migrations system.  Migrations are stored in app/Plugins/[plugin name]/Migrations/ and have the same name format as core migrations.  Let me know if you think we should go a different direction with this.

jekkos:
@objecttothis sounds good, just saw that plugins are deployed on the dev server, looks great so far!

I stll need to verify the changes for encryption setup in docker specifically. After that we can focus on merging new features

jekkos:
Plugin system will make it a lot less intrusive for outsiders to modify ospos. It will be a lot easier to have the core updated and maintain any modifications outside of it.

objecttothis:
@coderabbitai, we need to replace the first post in this PR with an updated detail of this PR.  It has been modified heavily since the initial post and it's out of date.  If you can't do it, please provide a copyable text of the changes and I'll do it.

objecttothis:
@jekkos a shortcoming that I just realized is that while it is possible to use AJAX calls to call plugin Controller functions, it is not possible to pass data to the EventListener function in the plugins from plugin injected  views.  I just pushed a commit that allows this, so now `$pluginData` will get passed through the Events::Trigger() calls and this array will contain variables that plugins want to pass from pluginContent() injected plugin views to the plugin listeners.

objecttothis:
@joshua1234511 because I'm still pushing a few changes here and there to this branch, make sure that if you create a plugin and develop it using the code in this branch, just make sure you periodically update your local copy of the branch.  there won't be any structurally breaking changes, so you're probably safe to develop on it, but pay attention to new additions.  For example, you may need to pass whatsapp data from a plugin view to the listener function in your plugin and in this case you'll need the most recent commits.

objecttothis:
@joshua1234511 and @jekkos I just pushed a commit which adds the infrastructure for creating a module for a plugin with permissions. @joshua1234511 I think your implementation was creating a module, so this should help you.  Take a look at the README.md in plugins and let me know if you get stuck.

@jekkos I'm nearing the end of porting over my plugins and I would like to roll plugins out soon.  How close are  we to releasing 3.4.2 so that we can move forward with staging 3.5.0.  Changes are really starting to stack up.

PR #4419, opened 2026-03-08
Fix Token_lib::render() for PHP 8.4 compatibility

### LINKED ISSUE #4333: Token_lib.render handles some strings incorrectly
I've been testing the existing render function while updating it to PHP 8.4, and most edge cases appear to break it. All of these inputs cause it to give up and return an empty string:
- Any extremely long string.
- Strings with a `%` character that isn't part of a date (`"Discount: 50%"`).
- Strings with an invalid date format (`"%-%-%"`, `"%Y-%q-%bad"`, `"%a%"`, etc.).
Unless there's a good reason for the render function to give up, I'd like the updated PHP 8.4 version to handle all of those edge cases correctly.

jekkos:
## Summary

This PR fixes issue #4333 by updating `Token_lib::render()` to be compatible with PHP 8.4, where `strftime()` was deprecated.

### Changes

- **Replaced `strftime()` with `IntlDateFormatter`**: The deprecated `strftime()` function is replaced with `IntlDateFormatter`, which is the recommended replacement in PHP 8.1+.

- **Added proper edge case handling**: The original implementation had several issues:
  - Returned empty string for very long strings
  - Returned empty string for invalid date formats (e.g., `%-%-%`, `%Y-%q-%bad`)
  - Returned empty string for strings with `%` not in date format (e.g., "Discount: 50%")
  
  These now return sensible output instead of empty strings.

- **Added comprehensive unit tests**: 29 test cases covering:
  - Basic token scanning
  - Date format handling (valid and invalid)
  - Edge cases (long strings, special characters, unicode, etc.)
  - Various date format specifiers (%Y, %m, %d, %B, %A, %z, etc.)

### Testing

All tests pass under PHP 8.4:
```
PHPUnit 11.5.44 by Sebastian Bergmann and contributors.
.............................                                     29 / 29 (100%)

OK, Tests: 29, Assertions: 44, PHPUnit Warnings: 1.
```

Fixes #4333

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage (30+ cases) for token scanning and rendering: percent/date tokens, composite directives, escaped/consecutive percent signs, braces, long/mixed/empty inputs, timezone handling, and many edge cases.

* **Refactor**
  * Reworked token rendering flow to add a dedicated date-format handling path and streamlined token replacement logic for more accurate percent-based date output.


app/Libraries/Token_lib.php
```diff
@@ -129,18 +192,9 @@ public function parse(string $string, string $pattern, array $tokens = []): arra
...
-    /**
-     * @param array $used_tokens
-     * @param array $tokens_to_replace
-     * @param array $token_values
-     * @param array $tokens
-     * @param bool $save
-     * @return array
-     */
-    public function generate(array $used_tokens, array &$tokens_to_replace, array &$token_values, array $tokens, bool $save = true): array    // TODO: $tokens
+    private function generate(array $used_tokens, array &$tokens_to_replace, array &$token_values, bool $save = true): void
```
objecttothis: Did we purposely remove PHPdocs in this file?

tests/Libraries/Token_libTest.php
```diff
@@ -0,0 +1,231 @@
+<?php
+
+require_once __DIR__ . '/../../app/Libraries/Token_lib.php';
+
+use App\Libraries\Token_lib;
+use PHPUnit\Framework\TestCase;
+
+class Token_libTest extends TestCase
```
objecttothis: In other unit tests we are extending CIUnitTestCase but here we are using TestCase.  Do we want two different frameworks?

objecttothis (changes requested):
Just the comments about the PHPdocs and which Unit Test Framework.  Let me know what you think.

app/Libraries/Token_lib.php
jekkos: According to the coding standards, yes ```There SHOULD be no useless PHPDoc annotations.```

objecttothis (approved):

app/Libraries/Token_lib.php
objecttothis: This makes sense.  I do think PHPdocs are needed when it's describing logic for return types that are complicated or complex function behavior.  In many cases if the PHPdoc is needed to explain what the function does, that might mean there's a SRP violation or the function name isn't very good.

objecttothis (dismissed):

PR #4421, opened 2026-03-09
Fix SQL injection in suggestions column configuration

jekkos:
Fixes critical SQL injection vulnerability in suggestions column configuration. See commit for details.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for search suggestion columns so only whitelisted fields are accepted; the primary suggestion always falls back to the item name and secondary/tertiary suggestions are shown only when valid.

* **Tests**
  * Added unit tests covering suggestion column validation: accepts valid fields, allows empty secondary values, and rejects malicious/invalid inputs.

### DIFF
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -386,9 +386,9 @@ class Config extends Secure_Controller
             'gcaptcha_enable'                   => $this->request->getPost('gcaptcha_enable') != null,
             'gcaptcha_secret_key'               => $this->request->getPost('gcaptcha_secret_key'),
             'gcaptcha_site_key'                 => $this->request->getPost('gcaptcha_site_key'),
-            'suggestions_first_column'          => $this->_validate_suggestions_column($this->request->getPost('suggestions_first_column')),
-            'suggestions_second_column'         => $this->_validate_suggestions_column($this->request->getPost('suggestions_second_column')),
-            'suggestions_third_column'          => $this->_validate_suggestions_column($this->request->getPost('suggestions_third_column')),
+            'suggestions_first_column'          => $this->validateSuggestionsColumn($this->request->getPost('suggestions_first_column'), 'first'),
+            'suggestions_second_column'         => $this->validateSuggestionsColumn($this->request->getPost('suggestions_second_column'), 'other'),
+            'suggestions_third_column'          => $this->validateSuggestionsColumn($this->request->getPost('suggestions_third_column'), 'other'),
             'giftcard_number'                   => $this->request->getPost('giftcard_number'),
             'derive_sale_quantity'              => $this->request->getPost('derive_sale_quantity') != null,
             'multi_pack_enabled'                => $this->request->getPost('multi_pack_enabled') != null,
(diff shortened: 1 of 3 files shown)


app/Controllers/Config.php
```diff
@@ -976,4 +977,15 @@ public function postRemoveLogo(): ResponseInterface
...
         return $this->response->setJSON(['success' => $success]);
     }
+
+    /**
+     * Validates suggestions column name to prevent SQL injection.
+     *
+     * @param string|null $column
+     * @return string
+     */
+    private function _validate_suggestions_column(?string $column): string
```
objecttothis: We should not use hungarian notation here and PSR-12 compliance requires this function to be camelCase.  `validateSuggestionsColumn()`

app/Controllers/Config.php
```diff
@@ -976,4 +977,15 @@ public function postRemoveLogo(): ResponseInterface
...
+    {
+        return in_array($column, Item::ALLOWED_SUGGESTIONS_COLUMNS_WITH_EMPTY, true) ? $column : 'name';
```
objecttothis: Using this constant outside of the items model might be an argument for declaring it in \app\Config\Constants.php. In this PR it's only one other class and the Unit tests, so it's not as strong of an argument but if it grows to be used much more then it should probably be globally declared.

app/Models/Item.php
```diff
@@ -544,13 +535,17 @@ public function delete_list(array $item_ids): bool
     public function get_search_suggestion_format(?string $seed = null): string
     {
         $config = config(OSPOS::class)->settings;
-        $seed .= ',' . $config['suggestions_first_column'];
+        
+        $suggestions_first_column = in_array($config['suggestions_first_column'], self::ALLOWED_SUGGESTIONS_COLUMNS, true)
```
objecttothis: new variables need to be camelCase

app/Models/Item.php
```diff
@@ -544,13 +535,17 @@ public function delete_list(array $item_ids): bool
...
+            : 'name';
+        $seed .= ',' . $suggestions_first_column;
 
-        if ($config['suggestions_second_column'] !== '') {
+        if ($config['suggestions_second_column'] !== '' && in_array($config['suggestions_second_column'], self::ALLOWED_SUGGESTIONS_COLUMNS, true)) {
             $seed .= ',' . $config['suggestions_second_column'];
         }
 
-        if ($config['suggestions_third_column'] !== '') {
+        if ($config['suggestions_third_column'] !== '' && in_array($config['suggestions_third_column'], self::ALLOWED_SUGGESTIONS_COLUMNS, true)) {
```
objecttothis: I would refactor out a small private function here
```
private function suggestionColumnIsAllowed(string $columnName): bool
{
    return in_array($columnName, self::ALLOWED_SUGGESTIONS_COLUMNS, true)
}
```
Then you can make this if clause easier to read
```
if ($config['suggestions_third_column'] !== '' && suggestionColumnIsAllowed($config['suggestions_third_column')) {
```
or
```
if ($config['suggestions_third_column'] !== ''
    && suggestionColumnIsAllowed($config['suggestions_third_column')) {
```
You can also call this function on lines 539, 544, 564, 567, and 570

objecttothis (changes requested):

objecttothis (dismissed):

PR #4422, opened 2026-03-09
Fix DECIMAL attribute not respecting locale format

### LINKED ISSUE #2884: DECIMAL attribute type does not respect locale format
### Bug
Currently the attribute types display in the views the way they are stored in the database rather than the way that the locale dictates they should be displayed.

jekkos:
## Summary

Fixes #2884

DECIMAL attribute values were displayed as raw database values instead of being formatted according to the user's locale settings.

## Problem

When displaying DECIMAL attribute values in table views (items table, sales reports, receiving reports), the values showed the raw numeric format from the database (e.g., `1234.56`) instead of respecting the locale-specific format (e.g., European locales showing `1.234,56`).

This was inconsistent with DATE attributes, which already received locale-based formatting.

## Solution

1. **Modified `Attribute::get_definitions_by_flags()`** to optionally return definition types along with names via a new `` parameter

2. **Updated `expand_attribute_values()` in `tabular_helper.php`** to:
   - Detect when a definition is of type `DECIMAL`
   - Apply `to_decimals()` locale formatting to those values
   - Leave other types (TEXT, DROPDOWN, DATE, CHECKBOX) unchanged

3. **Updated callers** in:
   - `get_items_manage_table_headers()` 
   - `get_item_data_row()`
   - Reports (detailed_sales, detailed_receivings)

## Files Changed

- `app/Models/Attribute.php` - Added `` parameter to `get_definitions_by_flags()`
- `app/Helpers/tabular_helper.php` - Added DECIMAL formatting in `expand_attribute_values()` and updated header generation
- `app/Controllers/Reports.php` - Updated detailed_sales and detailed_receivings to pass types

## Testing

Before: DECIMAL values displayed as `1234.56` regardless of locale
After: DECIMAL values use locale format (e.g., `1.234,56` for European locales)

## Release Notes

* **Bug Fixes**
  * Fixed column header formatting in detailed sales and receivings reports to properly display definition names.

* **Improvements**
  * Enhanced numeric attribute formatting with locale-aware handling across reports.
  * Improved attribute value display based on data type, with better formatting for numeric attributes.

### DIFF
--- a/app/Controllers/Reports.php
+++ b/app/Controllers/Reports.php
@@ -1776,7 +1776,7 @@ class Reports extends Secure_Controller
     {
         $this->clearCache();
 
-        $definition_names = $this->attribute->get_definitions_by_flags(attribute::SHOW_IN_SALES);
+        $definition_names = $this->attribute->get_definitions_by_flags(attribute::SHOW_IN_SALES, true);
 
         $inputs = [
             'start_date'     => $start_date,
@@ -1789,7 +1789,12 @@ class Reports extends Secure_Controller
         $this->detailed_sales->create($inputs);
 
         $columns = $this->detailed_sales->getDataColumns();
-        $columns['details'] = array_merge($columns['details'], $definition_names);
+        // Extract just names for column headers
+        $definition_headers = [];
+        foreach ($definition_names as $definition_id => $definition_info) {
+            $definition_headers[$definition_id] = $definition_info['name'];
+        }
+        $columns['details'] = array_merge($columns['details'], $definition_headers);
 
         $headers = $columns;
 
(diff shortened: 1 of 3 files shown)


app/Helpers/tabular_helper.php
```diff
@@ -408,7 +408,7 @@ function get_items_manage_table_headers(): string
 {
     $attribute = model(Attribute::class);
     $config = config(OSPOS::class)->settings;
-    $definition_names = $attribute->get_definitions_by_flags($attribute::SHOW_IN_ITEMS);    // TODO: this should be made into a constant in constants.php
+    $definitions_with_types = $attribute->get_definitions_by_flags($attribute::SHOW_IN_ITEMS, true);
```
objecttothis: Since we are renaming the variable, can we change this to PSR-12 compliant naming (camelCase)?

app/Helpers/tabular_helper.php
```diff
@@ -420,8 +420,8 @@ function get_items_manage_table_headers(): string
 
     $headers[] = ['item_pic' => lang('Items.image'), 'sortable' => false];
 
-    foreach ($definition_names as $definition_id => $definition_name) {
-        $headers[] = [$definition_id => $definition_name, 'sortable' => false];
+    foreach ($definitions_with_types as $definition_id => $definition_info) {
```
objecttothis: same with $definition_info.

app/Helpers/tabular_helper.php
```diff
@@ -651,10 +651,16 @@ function expand_attribute_values(array $definition_names, array $row): array
     }
 
     $attribute_values = [];
-    foreach ($definition_names as $definition_id => $definition_name) {
+    foreach ($definition_names as $definition_id => $definition_info) {
```
objecttothis: Can we do PSR-12 compliant naming for this variable change?

app/Controllers/Reports.php
```diff
@@ -1930,14 +1935,19 @@ public function detailed_receivings(string $start_date, string $end_date, string
...
+        $definition_names = $this->attribute->get_definitions_by_flags(attribute::SHOW_IN_RECEIVINGS, true);
 
         $inputs = ['start_date' => $start_date, 'end_date' => $end_date, 'receiving_type' => $receiving_type, 'location_id' => $location_id, 'definition_ids' => array_keys($definition_names)];
 
         $this->detailed_receivings->create($inputs);
 
         $columns = $this->detailed_receivings->getDataColumns();
-        $columns['details'] = array_merge($columns['details'], $definition_names);
+        // Extract just names for column headers
+        $definition_headers = [];
```
objecttothis: local variables need to be camelCase in new code.

objecttothis (changes requested):
Just the PSR-12 variable naming comments.  The logic looks correct.

objecttothis (approved):

PR #4429, opened 2026-03-12
Fix stored XSS vulnerability in Attribute Definitions (GHSA-rvfg-ww4r-rwqf)

jekkos:
## Summary
Fixes stored XSS vulnerability in Attribute Definitions (GHSA-rvfg-ww4r-rwqf).

## Security Impact
- Authenticated users with attribute management permission can inject XSS payloads in definition names
- Payloads execute when viewing/editing attributes in admin panel
- Can steal session cookies, perform CSRF attacks, or compromise admin operations

## Root Cause
1. Input: `Attributes.php` postSaveDefinition() accepts `definition_name` without sanitization
2. Output: Views echo `definition_name` without proper escaping

## Fix Applied
- **Input sanitization**: Added `FILTER_SANITIZE_FULL_SPECIAL_CHARS` to `definition_name` and `definition_unit` in POST handlers
- **Output escaping**: Added `esc()` wrapper when displaying `definition_name` in views
- **Defense-in-depth**: `htmlspecialchars()` on attribute values saved to database

## Files Changed
- `app/Controllers/Attributes.php` - Sanitize inputs on save
- `app/Views/attributes/form.php` - Escape output on display
- `app/Views/attributes/item.php` - Escape output on display

## Testing
- XSS payloads like `<script>alert(1)</script>` are now sanitized on input
- Any existing malicious definitions are escaped on output
- Does not break legitimate attribute names with special characters

## Summary by CodeRabbit

* **Bug Fixes**
  * Attribute saving now validates referenced groups and returns a localized error when an invalid group is submitted.
  * Attribute forms and lists escape displayed values to prevent unsafe rendering.

* **Localization**
  * Added a new localized message key ("definition_invalid_group") across many languages to surface the invalid-group error.

### DIFF
--- a/app/Controllers/Attributes.php
+++ b/app/Controllers/Attributes.php
@@ -65,7 +65,7 @@ class Attributes extends Secure_Controller
     public function postSaveAttributeValue(): ResponseInterface
     {
         $success = $this->attribute->saveAttributeValue(
-            html_entity_decode($this->request->getPost('attribute_value')),
+            htmlspecialchars(html_entity_decode($this->request->getPost('attribute_value')), ENT_QUOTES, 'UTF-8'),
             $this->request->getPost('definition_id', FILTER_SANITIZE_NUMBER_INT),
             $this->request->getPost('item_id', FILTER_SANITIZE_NUMBER_INT) ?? false,
             $this->request->getPost('attribute_id', FILTER_SANITIZE_NUMBER_INT) ?? false
@@ -108,10 +108,10 @@ class Attributes extends Secure_Controller
 
         // Save definition data
         $definition_data = [
-            'definition_name'  => $this->request->getPost('definition_name'),
-            'definition_unit'  => $this->request->getPost('definition_unit') != '' ? $this->request->getPost('definition_unit') : null,
+            'definition_name'  => $this->request->getPost('definition_name', FILTER_SANITIZE_FULL_SPECIAL_CHARS),
+            'definition_unit'  => $this->request->getPost('definition_unit') != '' ? $this->request->getPost('definition_unit', FILTER_SANITIZE_FULL_SPECIAL_CHARS) : null,
             'definition_flags' => $definition_flags,
-            'definition_fk'    => $this->request->getPost('definition_group') != '' ? $this->request->getPost('definition_group') : null
+            'definition_fk'    => $this->request->getPost('definition_group') != '' ? $this->request->getPost('definition_group', FILTER_SANITIZE_NUMBER_INT) : null
         ];
 
         if ($this->request->getPost('definition_type') != null) {
@@ -126,7 +126,7 @@ class Attributes extends Secure_Controller
                 $definition_values = json_decode(html_entity_decode($this->request->getPost('definition_values')));
 
                 foreach ($definition_values as $definition_value) {
-                    $this->attribute->saveAttributeValue($definition_value, $definition_data['definition_id']);
+                    $this->attribute->saveAttributeValue(htmlspecialchars($definition_value, ENT_QUOTES, 'UTF-8'), $definition_data['definition_id']);
                 }
 
                 return $this->response->setJSON([
--- a/app/Controllers/Attributes.php
+++ b/app/Controllers/Attributes.php
@@ -108,24 +108,14 @@ class Attributes extends Secure_Controller
 
         // Validate definition_group (definition_fk) foreign key
         $definition_group_input = $this->request->getPost('definition_group');
-        $definition_fk = null;
+        $definition_fk = $this->validateDefinitionGroup($definition_group_input);
 
-        if ($definition_group_input !== '' && $definition_group_input !== null) {
-            $definition_group_id = (int) $definition_group_input;
-
-            // Must be a positive integer, exist in attribute_definitions, and be of type GROUP
-            if ($definition_group_id <= 0
-                || !$this->attribute->exists($definition_group_id)
-                || $this->attribute->getAttributeInfo($definition_group_id)->definition_type !== GROUP
-            ) {
-                return $this->response->setJSON([
-                    'success' => false,
-                    'message' => lang('Attributes.definition_invalid_group'),
-                    'id'      => NEW_ENTRY
-                ]);
-            }
-
-            $definition_fk = $definition_group_id;
+        if ($definition_fk === false) {
+            return $this->response->setJSON([
+                'success' => false,
+                'message' => lang('Attributes.definition_invalid_group'),
+                'id'      => NEW_ENTRY
+            ]);
         }
 
         // Save definition data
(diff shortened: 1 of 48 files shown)


app/Controllers/Attributes.php
```diff
@@ -65,7 +65,7 @@ public function getSearch(): ResponseInterface
     public function postSaveAttributeValue(): ResponseInterface
     {
         $success = $this->attribute->saveAttributeValue(
-            html_entity_decode($this->request->getPost('attribute_value')),
+            htmlspecialchars(html_entity_decode($this->request->getPost('attribute_value')), ENT_QUOTES, 'UTF-8'),
```
objecttothis: I think the general rule is to validate input and sanitize output. In this case we are sanitizing the input and I think that's going to result in mangled valid values being stored in the database and increase potential for search failures.  We should be validating for length in the model to prevent truncating strings in the database and we should be sanitizing in the view.  The attack surface for XSS is at the browser, not the database.

app/Controllers/Attributes.php
```diff
@@ -108,10 +108,10 @@ public function postSaveDefinition(int $definition_id = NO_DEFINITION_ID): Respo
 
         // Save definition data
         $definition_data = [
-            'definition_name'  => $this->request->getPost('definition_name'),
-            'definition_unit'  => $this->request->getPost('definition_unit') != '' ? $this->request->getPost('definition_unit') : null,
+            'definition_name'  => $this->request->getPost('definition_name', FILTER_SANITIZE_FULL_SPECIAL_CHARS),
```
objecttothis: Same with these.  We should validate here, but not sanitize.  If definition names should be limited, then we need to use regex to determine if other characters exist and return $success = false if invalid characters are detected. I think it's reasonable to limit definition names to alphanumerics and basic symbols like space and period.

app/Controllers/Attributes.php
```diff
@@ -108,10 +108,10 @@ public function postSaveDefinition(int $definition_id = NO_DEFINITION_ID): Respo
...
+            'definition_unit'  => $this->request->getPost('definition_unit') != '' ? $this->request->getPost('definition_unit', FILTER_SANITIZE_FULL_SPECIAL_CHARS) : null,
             'definition_flags' => $definition_flags,
-            'definition_fk'    => $this->request->getPost('definition_group') != '' ? $this->request->getPost('definition_group') : null
+            'definition_fk'    => $this->request->getPost('definition_group') != '' ? $this->request->getPost('definition_group', FILTER_SANITIZE_NUMBER_INT) : null
```
objecttothis: here we need to check if it's an integer greater than one and return $success = false if not.

app/Views/attributes/item.php
objecttothis: I think we need to sanitize here to prevent stored XSS from running.

objecttothis (changes requested):
I think we need to change our approach to this.  We should be validating input and sanitizing output.  Sanitizing input will only break things.  The attack surface for Stored XSS isn't the database, but the browser.

app/Controllers/Attributes.php
jekkos: Hi agree that stronger protecton is to prevent control characters from entering the database. But then we should implement this validation both in frontend and backend. Its quite some work and currently the quick solution is to escape the output properly.

jekkos: This is to abide by the constraint that user input should not be modified. So a sanitization only in the backend might be problematic as we modify user input after submission and this might cause issues down the line

jekkos: So that means for every form, add validations in jQuery and consistent ones in php using a callback for all forms in the application. Also depending on context decide on which characters to allow and rules to apply. There might be more valuable things to work on imo

objecttothis: > So that means for every form, add validations in jQuery and consistent ones in php using a callback for all forms in the application. Also depending on context decide on which characters to allow and rules to apply. There might be more valuable things to work on imo

I think CI has form validation functionality in the framework and I think we need to take more advantage of that. I suppose the jQuery validations are best to prevent form submission with invalid data and the controller input validation ideally prevents things like SQL injection hitting the database in parallel with the form validation.  The quick fix though is adding the esc() call on the value in the view that I commented on.

app/Views/attributes/item.php
```diff
@@ -23,7 +23,7 @@
 <?php foreach ($definition_values as $definition_id => $definition_value) { ?>
 
     <div class="form-group form-group-sm">
-        <?= form_label($definition_value['definition_name'], $definition_value['definition_name'], ['class' => 'control-label col-xs-3']) ?>
+        <?= form_label(esc($definition_value['definition_name']), esc($definition_value['definition_name']), ['class' => 'control-label col-xs-3']) ?>
```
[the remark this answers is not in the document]
jekkos: @objecttothis so as this bot suggests, combining sanitization with escapie on output can already cause issues. Better to prevent specific chars from entering the system by failing the operation intentionally

app/Controllers/Attributes.php
```diff
@@ -106,12 +106,34 @@ public function postSaveDefinition(int $definition_id = NO_DEFINITION_ID): Respo
             $definition_flags |= $flag;
         }
 
+        // Validate definition_group (definition_fk) foreign key
```
objecttothis, line 109: We should probably refactor this validation logic into it's own private method.  This function is pretty large.

app/Controllers/Attributes.php
```diff
@@ -106,12 +106,34 @@ public function postSaveDefinition(int $definition_id = NO_DEFINITION_ID): Respo
...
+            $definition_fk = $definition_group_id;
+        }
+
         // Save definition data
         $definition_data = [
             'definition_name'  => $this->request->getPost('definition_name'),
             'definition_unit'  => $this->request->getPost('definition_unit') != '' ? $this->request->getPost('definition_unit') : null,
             'definition_flags' => $definition_flags,
-            'definition_fk'    => $this->request->getPost('definition_group') != '' ? $this->request->getPost('definition_group') : null
+            'definition_fk'    => $definition_fk
```
objecttothis, line 126: Then you could just call the function here.

app/Views/attributes/item.php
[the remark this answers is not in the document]
objecttothis: > [@objecttothis](https://github.com/objecttothis) so as this bot suggests, combining sanitization with escapie on output can already cause issues. Better to prevent specific chars from entering the system by failing the operation intentionally

Yes, so escaping here and validating in the controller with a failure if it doesn't contain the right data.  That's my understanding of how we need to be preventing stored XSS injection attacks. Sorry if I used sanitizing interchangeably and cause confusion.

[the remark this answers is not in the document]
objecttothis: The first layer of jQuery validation would even add on top of that.  CI4 has a very versitile custom rule defining capability  that I think we should take advantage of and we are already using it.  See \app\Config\Validation\OSPOSRules and it's uses (e.g., index() in \app\Controllers\Login line 50).  It really lightens the load of validation by 1- having a lot of built-in rules to use and 2- allowing the defining of custom rules for more complicated validations.  The biggest candidates are validations we need to do frequently.

objecttothis (changes requested):

objecttothis (approved):

objecttothis (changes requested):
Sorry, I approved too soon.  see my other comments in the attributes controller.

Issue #2819, opened 2020-05-06
Item search with "Attribute search" enabled produces pagination problems


objecttothis:
### Background information

__IMPORTANT: If you choose to ignore this issue report template, your issue will be closed as we cannot help without the requested information.__

Please make sure you tick (add an x between the square brackets with no spaces) the following check boxes:

- [x] Reporting an issue of an unmodified OSPOS installation
- [x] Checked [open and closed issues](https://github.com/opensourcepos/opensourcepos/issues?utf8=%E2%9C%93&q=is%3Aissue) and no similar issue was already reported (please make sure you searched!)
- [x] Read [README](https://github.com/opensourcepos/opensourcepos/blob/master/README.md), [WHATS_NEW](https://github.com/opensourcepos/opensourcepos/blob/master/WHATS_NEW.txt), [INSTALL.md](https://github.com/opensourcepos/opensourcepos/blob/master/INSTALL.md) and [UPGRADE](https://github.com/opensourcepos/opensourcepos/blob/master/UPGRADE.txt)
- [x] Read the [FAQ](https://github.com/opensourcepos/opensourcepos#faq) for any known install and/or upgrade gotchas (in specific PHP extensions installed)
- [x] Read the [wiki](https://github.com/opensourcepos/opensourcepos/wiki)
- [x] Executed any database upgrade scripts if an upgrade pre 3.0.0 (e.g. database/2.4_to_3.0.sql)
- [x] Aware the installation code is in [bintray](https://bintray.com/jekkos/opensourcepos/opensourcepos/view/files?sort=updated&order=asc#files) (see README), and [GitHub master](https://github.com/opensourcepos/opensourcepos/tree/master) is for [developers only](https://github.com/opensourcepos/opensourcepos/wiki/Development-setup) and therefore not complete nor stable

### Installation information

OSPOS Installation Info: 3.3.2 - 9d2fd1

Language Code: en-US

Extensions & Modules:
» GD: Enabled ✓
» BC Math: Enabled ✓
» INTL: Enabled ✓
» OpenSSL: Enabled ✓
» MBString: Enabled ✓
» Curl: Enabled ✓

User Configuration:
.Browser: Firefox
.Server Software: Server
.PHP Version: 7.4.4
.DB Version: 5.7.29-log
.Server Port: 80
.OS: FreeBSD 11.3-RELEASE-p7

File Permissions:
» [application/logs:] - 0770 | Writable ✓
[...]

PR #4442, opened 2026-03-16
Fix/attribute search refactor

jekkos:
## Summary by CodeRabbit

* **New Features**
  * Mark attributes as searchable so items can be found by attribute values even if those attributes aren’t displayed elsewhere.
  * Search accepts attribute-specific queries and combines attribute + text filtering.

* **Improvements**
  * Attribute values can be used for sorting; attribute-based filters are prepopulated for more relevant results.
  * Search visibility is now considered by default where applicable.

* **Localization**
  * Added English strings for the new "Show in search" visibility option.


app/Helpers/tabular_helper.php
```diff
@@ -401,6 +401,25 @@ function item_headers(): array
     ];
 }
 
+/**
+ * Get all sortable column keys for items table, including dynamic attribute columns.
+ *
+ * @return array Array of column keys that are valid for sorting
+ */
+function item_sort_columns(): array
```
objecttothis, line 410: @jekkos when do you have in mind to turn on the linters so that it catches PSR naming violations like this?

app/Models/Item.php
```diff
@@ -131,31 +131,166 @@ public function get_found_rows(string $search, array $filters): int
...
     /**
-     * Perform a search on items
+     * Parse search string for attribute-specific queries
+     * Supports syntax like "color: blue size: large" or "color:blue AND size:large"
+     *
+     * @param string $search The raw search string
+     * @return array{terms: array, attributes: array} Parsed terms and attribute queries
      */
-    public function search(string $search, array $filters, ?int $rows = 0, ?int $limit_from = 0, ?string $sort = 'items.name', ?string $order = 'asc', ?bool $count_only = false)
+    public function parse_attribute_search(string $search): array
```
objecttothis: I support refactoring the function to better describe it, but let's not use snake case.  It feels like going away from PSR compliance and not toward it.

app/Models/Item.php
```diff
@@ -131,31 +131,166 @@ public function get_found_rows(string $search, array $filters): int
...
+            'attributes' => []
+        ];
+
+        if (empty($search)) {
+            return $result;
         }
-        if ($limit_from == null) {
-            $limit_from = 0;
+
+        $pattern = '/(\w+)\s*:\s*([^\s,]+)(?:\s+(?:AND|OR)\s+)?/i';
```
objecttothis: I might be wrong but I think this is going to break searches containing a colon.  For example, if we have a DROPDOWN attribute called Aspect Ratio and someone wants to search for items with `16:9` in an attribute, this creates a problem.  I think it would be better to pass a keyed array through or use a different delimiter.  Maybe `=`.  It is nice to be able to use boolean logic in the search though.

objecttothis (changes requested):
Mostly just issues with new code with variables which are not PSR-12 compliant. Then also the one discussion about using `:` as a delimeter.

objecttothis (dismissed):
These changes resolve my concerns.  Probably see if anything coderabbitai says is worth changing, then merge.

app/Models/Item.php
```diff
@@ -228,16 +402,43 @@ public function search(string $search, array $filters, ?int $rows = 0, ?int $lim
...
+            $builder->join("attribute_links AS {$sortAlias}", "{$sortAlias}.item_id = items.item_id AND {$sortAlias}.definition_id = {$sortDefinitionId} AND {$sortAlias}.sale_id IS NULL AND {$sortAlias}.receiving_id IS NULL", 'left');
+            $builder->join("attribute_values AS {$sortAlias}_val", "{$sortAlias}_val.attribute_id = {$sortAlias}.attribute_id", 'left');
+            
+            // Determine the correct column to sort by based on attribute type
+            $attribute = model(Attribute::class);
+            $definitionInfo = $attribute->get_definitions_by_flags(Attribute::SHOW_IN_ITEMS, true);
+            $sortColumn = "{$sortAlias}_val.attribute_value"; // default to text
+            
+            if (isset($definitionInfo[$sortDefinitionId])) {
+                $defType = is_array($definitionInfo[$sortDefinitionId]) ? ($definitionInfo[$sortDefinitionId]['type'] ?? TEXT) : TEXT;
```
objecttothis, line 431: I think the fallback prevents a problem here, but it appears this doesn't account for DROPDOWN attribute types, which are also attribute_value.

app/Helpers/tabular_helper.php
jekkos: I believe it might be best not to have too many changes in flight as it might cause a lot of trouble and conflicts with open pull requests. Rebasing those will be tricky. 
So I would suggest to merge most of the open work first and do a feature freeze after. Then convert the whole codebase.

objecttothis: > I believe it might be best not to have too many changes in flight as it might cause a lot of trouble and conflicts with open pull requests. Rebasing those will be tricky. So I would suggest to merge most of the open work first and do a feature freeze after. Then convert the whole codebase.

I agree, but shouldn't we do a freeze now on opening new PRs while we finish and merge most of the open ones?

jekkos: Agree we should try to focus on getting things validated now. Now with many security fixes I might draft a new minor releaae.wdyt

objecttothis: Yes, a new minor release is a good idea as long as we are sure there are no regressions currently in the master.

jekkos:
## Performance Analysis: Pre-query + WHERE IN vs GROUP_CONCAT + HAVING

Given the performance regression reported in #4519, I did a formal analysis comparing the two search approaches to validate this PR's architecture.

### The Two Approaches

| | Master (GROUP_CONCAT + HAVING) | This PR (Pre-query + WHERE IN) |
|---|---|---|
| **Mechanism** | Single query. LEFT JOINs attribute tables, aggregates with `GROUP_CONCAT`, filters with `HAVING LIKE` on concatenated strings | Separate pre-query to find matching item_ids, then main query with `WHERE IN` for filtering |
| **Queries per search** | 1 | N+1 (N = number of attribute predicates, typically 1-5) |
| **Index usage** | :x: None for HAVING (computed columns, leading-wildcard LIKE) | :white_check_mark: Full on subqueries (indexed `definition_id`, `item_id`) |
| **Temp tables** | :x: Full dataset materialized before HAVING can filter | :white_check_mark: Minimal - only matching item_ids |
| **Pagination accuracy** | :x: **Broken** - `COUNT(DISTINCT)` computed before HAVING filter | :white_check_mark: Correct - WHERE IN filters before GROUP BY |

### Why This PR is Faster

**Common case** (simple text search with attributes enabled):

Master execution path:
1. JOIN items x attribute_links x attribute_values -> 100K-1M row temp table
2. GROUP BY items.item_id (still ALL items, not just matches)
3. Compute GROUP_CONCAT for EVERY item (expensive string concat)
4. HAVING LIKE on concatenated strings (full scan, no index possible)
5. LIMIT 20 (applied AFTER all above work)

**Total work: O(items x avg_attributes)** - materialize, group, concat, then filter

This PR execution path:
1. Pre-query: `SELECT DISTINCT item_id FROM attribute_links JOIN attribute_values WHERE value LIKE '%term%' AND definition_id IN (...)` -> returns ~50-500 matching item_ids (indexed, fast)
2. Main query: `SELECT ... FROM items WHERE item_id IN (50-500 IDs)` LEFT JOIN attribute_links/values (for display GROUP_CONCAT only) GROUP BY items.item_id LIMIT 20

**Total work: O(matching_items)** - filter early, then minimal display work

**Estimated speedup: 3-10x** on datasets with 10K+ items, scaling with dataset size.

**Worst case** (multi-attribute AND search: "color:blue AND size:large AND brand:nike"):

Master: Same single query with compound HAVING conditions. Must still materialize ALL items with ALL attributes, then filter. AND logic on concatenated strings is fragile (pattern matching on aggregated text).

This PR: 3 subqueries + 1 main = 4 total queries. Each subquery is indexed and returns a small set of item_ids. `array_intersect()` in PHP combines them (O(n), negligible). Main query operates on the intersection (~10-100 items).

### "N+1" Concern Analysis

This is **not classic N+1**. N is the number of attribute predicates (typically 2-5), NOT the number of items (10K+). Even 10 subqueries = 10 round-trips = ~50ms, versus 2-10 seconds for the GROUP_CONCAT approach on large datasets. The query count is bounded and small.

### GROUP_CONCAT HAVING Has Unfixable Correctness Bugs

Beyond performance, the current master approach has correctness issues that **cannot be fixed without this architectural change**:

1. **Broken pagination** (#2819): `COUNT(DISTINCT items.item_id)` is computed **before** HAVING filter - pagination totals are wrong
2. **GROUP_CONCAT truncation**: Default `group_concat_max_len=1024` clips long attribute lists; matches past the truncation point are silently missed
3. **No multi-attribute AND/OR logic**: Cannot express "color=blue AND size=large" - HAVING LIKE on a concatenated string gives false positives
4. **No attribute-type-aware sorting**: All attributes sort by text column, causing "10" before "2" for numeric attributes

### Optimization Opportunities (Non-Blocking)

Two improvements that would make this PR even faster:

1. **Batch predicate queries**: Replace the per-attribute-value loop in `searchByAttributes()` with a single `UNION ALL` + `GROUP BY item_id HAVING COUNT(*) = predicate_count` for AND logic - reduces N queries to 1
2. **Cache definition lookups**: `get_definitions_by_flags()` is called independently in `searchByAttributes()` (line 193) and the sort block (line 427). Call it once in `search()` and pass the result down

### Verdict

This PR is **provably faster** for all practical OSPOS workloads and fixes correctness issues that are architecturally unfixable in the GROUP_CONCAT approach. The rebased branch also includes the `orWhere` scoping fix from #4520 which addresses the separate items performance regression from #4519.

objecttothis:
This work is completed in #4652 and the work in this branch has been harmonized into that PR.

PR #4444, opened 2026-03-17
Fix customer reward points consistency

jekkos:
## Summary by CodeRabbit

* **New Features**
  * Centralized reward-point management for sales, including automatic calculation and localized reward-aware payment detection.
  * Automatic reward reconciliation when saving sales or when a sale's customer changes.

* **Bug Fixes**
  * Restores reward points when deleting non-canceled sales.
  * Ensures correct reward transfer and delta adjustments on customer reassignment, with insufficient-point handling.

* **Tests**
  * Added unit tests covering calculations, transfers, payment aggregation, customer-change scenarios, and edge cases.

### DIFF
--- a/app/Enums/RewardOperation.php
+++ b/app/Enums/RewardOperation.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace App\Enums;
+
+/**
+ * Reward operation types for customer points adjustments.
+ * 
+ * Used by Reward_lib to perform type-safe reward point operations.
+ */
+enum RewardOperation: string
+{
+    case Deduct = 'deduct';
+    case Restore = 'restore';
+    case Adjust = 'adjust';
+}
--- a/app/Libraries/Reward_lib.php
+++ b/app/Libraries/Reward_lib.php
@@ -34,11 +34,12 @@ class Reward_lib
     /**
      * Adjusts customer reward points for a sale transaction.
      * Handles new sales, sale updates, and sale cancellations.
+     * Prevents negative balances by validating sufficient points before deduction.
      *
      * @param int|null $customerId Customer ID (null for walk-in customers)
      * @param float $rewardAmount Amount to deduct from points (for new/updated sales)
      * @param string $operation Operation type: 'deduct', 'restore', 'adjust'
-     * @return bool Success status
+     * @return bool Success status (false if insufficient points for deduct/adjust)
      */
     public function adjustRewardPoints(?int $customerId, float $rewardAmount, string $operation): bool
     {
@@ -51,6 +52,14 @@ class Reward_lib
         switch ($operation) {
             case 'deduct':
             case 'adjust':
+                if ($currentPoints < $rewardAmount) {
+                    log_message(
+                        'warning',
+                        'Reward_lib::adjustRewardPoints insufficient points customer_id=' . $customerId
+                        . ' current=' . $currentPoints . ' requested=' . $rewardAmount
+                    );
+                    return false;
+                }
                 $this->customer->update_reward_points_value($customerId, $currentPoints - $rewardAmount);
                 return true;
             case 'restore':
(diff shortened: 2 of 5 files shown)


app/Libraries/Reward_lib.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Adjusts customer reward points for a sale transaction.
+     * Handles new sales, sale updates, and sale cancellations.
+     *
+     * @param int|null $customerId Customer ID (null for walk-in customers)
+     * @param float $rewardAmount Amount to deduct from points (for new/updated sales)
+     * @param string $operation Operation type: 'deduct', 'restore', 'adjust'
+     * @return bool Success status
+     */
+    public function adjustRewardPoints(?int $customerId, float $rewardAmount, string $operation): bool
```
objecttothis: using a string to dictate the operation runs the risk of typos in code breaking behavior and you lose code completion and IDE level checks.  Better to instead define an enum above this.  Something like:
```
enum RewardOperation: string
{
    case Deduct = 'deduct';
    case Restore = 'restore';
    case Adjust = 'adjust';
}
```

Then your method signature becomes `public function adjustRewardPoints(?int $customerId, float $rewardAmount, RewardOperation $operation): bool` and your switch becomes

```
switch ($operation) {
    case RewardOperation::Deduct:
...
```
If you wanted to take it next level, you could put the logic in the enum and that makes it a lot cleaner.
```
enum RewardOperation: string
{
    case Deduct = 'deduct';
    case Restore = 'restore';
    case Adjust = 'adjust';

    public function apply(float $current, float $amount): float
    {
        return match ($this) {
            self::Deduct  => $current - $amount,
            self::Restore => $current + $amount,
            self::Adjust  => $amount,
        };
    }
}
```

Then down below after line 49 you could just do 
```
$newPoints = $operation->apply($currentPoints, $rewardAmount);
return $this->customer->update_reward_points_value($customerId, $newPoints);
```

Note that another advantage of an enum for the operation is that you cannot pass an invalid type.  It will error.  My last recommendation does require that we update the update_reward_points_value() function in the customer model to something like this so that it actually returns the result of the QueryBuilder update call:
```
    /**
     * Updates reward points value
     */
    public function update_reward_points_value(int $customer_id, int $value): bool
    {
        return $this->db->table('customers')
           ->where('person_id', $customer_id)
           ->update(['points' => $value]);
    }
```

app/Libraries/Reward_lib.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Validates if a customer has sufficient reward points for a purchase.
+     *
+     * @param int $customerId Customer ID
+     * @param float $requiredPoints Points required for purchase
+     * @return bool True if customer has sufficient points
+     */
+    public function hasSufficientPoints(int $customerId, float $requiredPoints): bool
```
objecttothis, line 155: I wonder if some people may intend on rewards points to be used the way Amazon does.  For example, if a customer has 800 reward points.  It appears our current use is that reward points have a conversion value in the local currency and they can use those reward points if they have enough to purchase the item. The Amazon method is to say that reward points have a conversion value in the local currency and you can use them as a payment method to pay for part of your purchase.  Having this option might change how we do calculations to see if they have sufficient points. This might be a feature set for a different PR but having the option to use one system or the other might be nice.

objecttothis (changes requested):

app/Enums/RewardOperation.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
```
objecttothis, line 2: I wasn't thinking that we needed to create a folder structure for a separate file for each enum.  Yet removing the clutter from the classes with constants and enums but not clogging up the app\Config\Constants.php file is not a bad idea.  Here is what I recommend.

Instead of `\app\Enums\SpecificEnum.php` we create `\app\Constants\MVCEntityConstants.php`.  For example this would become `\app\Constants\RewardsConstants.php` Then this enum and other Rewards specific constants go here.  Alternately we can just put this enum in the Reward_lib class.

app/Models/Sale.php
```diff
@@ -1391,6 +1508,94 @@ private function save_customer_rewards(int $customer_id, int $sale_id, float $to
         }
     }
 
+    /**
```
objecttothis, line 1511: I think you said we are removing PHPdocs which don't add information of value that can't be gotten from the function name or return logic easily.

app/Models/Sale.php
```diff
@@ -1391,6 +1508,94 @@ private function save_customer_rewards(int $customer_id, int $sale_id, float $to
...
+            if ($payment_type === $label) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Returns unique localized labels for the rewards payment type.
```
objecttothis, line 1530: Same here

objecttothis (changes requested):

PR #4467, opened 2026-04-03
fix: Handle empty database on fresh install

jekkos:
## Problem

When starting OSPOS on a completely empty database (fresh MySQL/MariaDB instance), the application crashes with multiple errors:

1. **Session error**: `Table 'ospos.ospos_sessions' doesn't exist`
2. **Config error**: `Table 'ospos.ospos_app_config' doesn't exist`
3. **Language error**: `Undefined array key "language_code"`

This prevents users from initializing a fresh installation.

## Root Cause

PR #4455 fixed the file-based sessions issue for databases with existing migrations, but didn't handle the **completely empty database** case where:
- No migrations have run yet
- No tables exist at all
- CSRF filter tries to start session before checking database state
- Config tries to load from non-existent `ospos_app_config` table

## Solution

This PR adds handling for the completely empty database state:

### 1. Session.php - Switch to file-based sessions when database not ready
```php
public function __construct()
{
    parent::__construct();
    
    if ($this->driver === DatabaseHandler::class) {
        try {
            $db = \Config\Database::connect();
            if (!$db->tableExists('migrations')) {
                $this->driver = FileHandler::class;
                $this->savePath = WRITEPATH . 'session';
            }
        } catch (\Throwable $e) {
            $this->driver = FileHandler::class;
            $this->savePath = WRITEPATH . 'session';
        }
    }
}
```

Switches to file-based sessions when:
- Database connection fails
- `migrations` table doesn't exist

### 2. OSPOS.php - Handle missing config table gracefully
```php
public function __construct()
{
    parent::__construct();
    $this->cache = Services::cache();
    $this->set_settings();
    
    // Set defaults if database not initialized
    if (empty($this->settings)) {
        $this->settings = [
            'company' => 'OSPOS',
            'language' => 'english',
            'language_code' => 'en',
            // ... other defaults
        ];
    }
}

public function set_settings(): void
{
    // ... cache check ...
    try {
        $appconfig = model(Appconfig::class);
        // ... load from database ...
    } catch (\CodeIgniter\Database\Exceptions\DatabaseException $e) {
        // Database table doesn't exist yet
        $this->settings = [];
    }
}
```

### 3. MY_Migration.php - Handle database connection failures
```php
public static function get_current_version(): int
{
    try {
        $db = Database::connect();
        if ($db->tableExists('migrations')) {
            // ... check migration version ...
        }
    } catch (\Throwable $e) {
        // Database doesn't exist yet
        return 0;
    }
    return 0;
}
```

### 4. Load_config.php - Set default language before checking
```php
// Set default language if settings are empty (migrations not run yet)
if (empty($config->settings) || !isset($config->settings['language_code'])) {
    $config->settings['language'] = 'english';
    $config->settings['language_code'] = 'en';
}
```

## Testing

Tested locally with Docker on completely fresh MySQL database:

**Before fix:**
```
❌ HTTP 500 Error
❌ "Table 'ospos.ospos_sessions' doesn't exist"
❌ Application crashes
```

**After fix:**
```
✅ HTTP 200 OK
✅ "Database Migration Required" page displays
✅ File-based sessions work (no sessions table needed)
✅ Config loads with defaults (no app_config table needed)
✅ Migration UI works correctly
```

## Related

- Builds on PR #4455 (file-based sessions for migration state)
- Handles the edge case of completely empty database
- Tested with Docker MySQL from scratch

## Checklist

- [x] Tested locally with Docker
- [x] No breaking changes to existing functionality
- [x] Follows CodeIgniter 4 patterns
- [x] Proper exception handling for database errors

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Improved startup resilience when the database is unavailable or partially migrated.
  * Session storage now automatically falls back to file-based storage if the configured database session store is inaccessible.
  * Migration and version checks tolerate missing or unreachable database resources and avoid startup failures.
  * Default language settings are initialized earlier when configuration is empty or incomplete.


app/Config/OSPOS.php
```diff
@@ -34,11 +35,21 @@ public function set_settings(): void
...
-            foreach ($appconfig->get_all()->getResult() as $app_config) {
-                $this->settings[$app_config->key] = $app_config->value;
+            try {
+                $appconfig = model(Appconfig::class);
+                foreach ($appconfig->get_all()->getResult() as $app_config) {
+                    $this->settings[$app_config->key] = $app_config->value;
+                }
+                $this->cache->save('settings', encode_array($this->settings));
+            } catch (DatabaseException $e) {
+                // Database table doesn't exist yet (migrations haven't run)
```
objecttothis, line 45: If we are looking to reduce how much work the user needs to do on the database side, we could create a small database seeder run through the command line which checks for the database and  creates it if it doesn't exist.
```
php spark db:seed initialDatabaseCreation
```
The only thing is that the username in the .env has to exist and have database creation rights for it to work. It's fine to let the database creation remain completely manual too since we aren't going backward if we do.

app/Config/Session.php
```diff
@@ -124,4 +125,35 @@ class Session extends BaseConfig
...
+    public function __construct()
+    {
+        parent::__construct();
+        
+        // Only switch to file handler if database is configured but not migrated
+        if ($this->driver === DatabaseHandler::class) {
+            try {
+                $db = \Config\Database::connect();
+                
+                // Check if the sessions table exists (the actual table we need)
```
objecttothis: These comments probably aren't needed since the code is self-explanatory

app/Events/Load_config.php
```diff
@@ -40,44 +40,33 @@ public function load_config(): void
...
+        $language = $config->settings['language'] ?? null;
+        
+        if (empty($config->settings) || $language_code === null) {
+            $config->settings['language'] = 'english';
+            $config->settings['language_code'] = 'en';
+            $language_code = 'en';
+            $language = 'english';
+        }
+        
+        $language_exists = file_exists('../app/Language/' . $language_code);
```
objecttothis: 43-55 should be refactored into a helper function `public function languageExists(): bool` rather than placed here.

app/Events/Load_config.php
```diff
@@ -40,44 +40,33 @@ public function load_config(): void
...
 
-        if (current_language_code() == null || current_language() == null || !$language_exists) {    // TODO: current_language() is undefined
+        if (!$language_exists) {
```
objecttothis: Please use PSR compliant variable and function naming for all new code.

app/Libraries/MY_Migration.php
```diff
@@ -35,11 +36,18 @@ public function get_latest_migration(): int
...
+        try {
+            $db = Database::connect();
+            if ($db->tableExists('migrations')) {
+                $builder = $db->table('migrations');
+                $builder->select('version')->orderBy('version', 'DESC')->limit(1);
+                $result = $builder->get()->getRow();
+                return $result ? $result->version : 0;
+            }
+        } catch (DatabaseException $e) {
+            // Database doesn't exist yet or connection failed
```
objecttothis: We can probably remove these comments as the code speaks for itself.

app/Libraries/MY_Migration.php
```diff
@@ -143,4 +156,5 @@ public function convert_table(): void
 
         $this->ensureTable();
     }
+
```
objecttothis, line 157: I don't think this blank line should be here.

app/Views/login.php
```diff
@@ -46,60 +47,93 @@
...
-                    <div class="alert alert-warning mt-3">
-                        <strong><?= lang('Login.migration_auth_message', [$latest_version]) ?></strong>
-                    </div>
-                <?php else: ?>
-                    <h3 class="text-center m-0"><?= lang('Login.welcome', [lang('Common.software_short')]) ?></h3>
-                <?php endif; ?>
+                <?= form_open('login', ['id' => 'login-form']) ?>
+                
+                <h3 id="form-heading" class="text-center m-0">
+                    <?php if (!$is_latest || $is_new_install): ?>
```
objecttothis, line 53: PSR Compliant variable naming on new code

objecttothis (changes requested):
It looks like overall this will resolve the issue.  take a look at my review comments.  After those are addressed/discarded I will check out this branch and regression test it locally.

objecttothis (dismissed):

PR #4471, opened 2026-04-06
feat: Add MySQL 8.4 support (#4048)

### LINKED ISSUE #4048: [Feature]: MySQL 8.4 support
### Type of Feature

✨ New Feature

### OpensourcePOS Version

opensourcepos 3.3.9

### Description

@SteveIreland has reported issues related to MySQL 8.4 and I think @odiea may be having issues with the GiftCard search due to it.  That still needs to be confirmed in testing, but the question is this: Can we release 3.4.0 without MySQL 8.4 support and then add it in a future release or do we need to include it in 3.4.0? @jekkos @odiea @SteveIreland @daN4cat .  

My vote is to release without, but that's just because this release has been dragging.

### Additional Information

_No response_

### Verify you searched open requests in OpensourcePOS

- [X] I agree I have searched Open Requests

Helios531:
## Summary
- Add MySQL 8.4 (LTS) compatibility alongside existing MariaDB 10.x support
- Convert all database tables and migration scripts from utf8/utf8_general_ci to utf8mb4/utf8mb4_general_ci
- Fix ONLY_FULL_GROUP_BY violations by qualifying ambiguous column names in GROUP BY clauses
- Fix index key length overflow for utf8mb4 by using prefix indexes on composite varchar indexes
- Add try-catch in migration helper to handle strict error mode in MySQL 8.4
- Fix deprecated named PRIMARY KEY syntax in initial schema
- Add MySQL 8.4 to CI/CD test matrix alongside MariaDB 10.5
- Update Docker configs to default to MySQL 8.4 (configurable via OSPOS_DB_IMAGE env var)
- Update documentation (INSTALL.md, .env.example) to list MySQL 8.4 as supported

## Changes
- **Schema/Migrations:** utf8 → utf8mb4 across all SQL scripts (initial schema + 10 migration scripts)
- **Models:** Qualified GROUP BY columns in Sale.php, Summary_report.php, Summary_payments.php
- **Migration helper:** Added exception handling for MySQL 8.4 strict mode
- **New migration:** 20260405000000_Mysql84Utf8mb4Conversion for existing installations
- **CI/CD:** Test matrix now covers PHP 8.1-8.4 × MariaDB 10.5 + MySQL 8.4
- **Docker:** Default DB image changed to mysql:8.4 (override with OSPOS_DB_IMAGE=mariadb:10.5)

## Test plan
- [x] Fresh install on MySQL 8.4 — migrations complete successfully
- [x] Login and session handling works
- [x] Items CRUD with utf8mb4 characters (emoji, accented chars)
- [x] Sales register and payment flow
- [x] Summary reports (Transactions, Payments) load without GROUP BY errors
- [x] Customers, Suppliers, Employees tables display correctly
- [x] Gift card search returns correct results
- [x] Customer search in sales register works

Closes #4048

## Summary by CodeRabbit

- **New Features**
  - Added support for MySQL 8.4 and documented supported MySQL and MariaDB versions.
  - Added an automatic database migration to convert supported tables to `utf8mb4`.

- **Bug Fixes**
  - Improved report and sales aggregation accuracy by correcting grouped data handling.
  - Preserved foreign-key relationships during the database conversion process.

- **Tests**
  - Expanded automated testing across MariaDB 10.5, MySQL 8.0, and MySQL 8.4.


.github/workflows/phpunit.yml
```diff
@@ -38,6 +38,9 @@ jobs:
           - '8.2'
           - '8.3'
           - '8.4'
+        db-image:
+          - 'mariadb:10.5'
```
jekkos, line 41: @objecttothis looks great, this would run the tests + migrations both with mariadb and mysql

docker/docker-mysql.yml
```diff
@@ -12,7 +12,7 @@ networks:
 
 services:
     mysql:
-        image: mariadb:10.5
+        image: ${OSPOS_DB_IMAGE:-mysql:8.4}
```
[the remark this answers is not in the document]
jekkos: we might prefer mariadb still by default

.github/workflows/phpunit.yml
```diff
@@ -38,6 +38,9 @@ jobs:
...
+          - 'mysql:8.4'
```
jekkos, line 43: can we add older versions of MySQL here? I would go from 8.0 -> 8.4

app/Database/Migrations/sqlscripts/3.0.2_to_3.1.1.sql
```diff
@@ -82,7 +82,7 @@ CREATE TABLE IF NOT EXISTS `ospos_customers_packages` (
   `points_percent` float NOT NULL DEFAULT '0',
   `deleted` int(1) NOT NULL DEFAULT '0',
   PRIMARY KEY (`package_id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8 AUTO_INCREMENT=1;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci AUTO_INCREMENT=1;
```
jekkos: are these changes required? Or can't we create the table in utf8 on mysql 8.4 at all? 

Also I wonder if this will also work on older versions (hence the request to extend version range in the test matrix)

jekkos (changes requested):
Thanks for this PR, looks like a good start.

app/Database/Migrations/sqlscripts/3.0.2_to_3.1.1.sql
Helios531: utf8 still works on MySQL 8.4 but it's deprecated and may be removed later. utf8mb4 is backward compatible with all older versions so nothing breaks. And added MySQL 8.0 to CI as suggested.

app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql
objecttothis, line 1: We need to make sure this changeset  gets tested on multiple versions of MySQL and MariaDB since we need to maintain compatibility.

Helios531: Hi, @objecttothis 
I added MySQL 8.0 and 8.4 to the CI matrix and also MariaDB 10.5, so all versions are tested now.

objecttothis: I'm going to download this and test it migrating against an existing 3.3.2 database.  Sometimes these kinds of changes reveal issues with existing database upgrades that don't happen with new databases.

app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql
```diff
@@ -0,0 +1,35 @@
+-- Convert all tables to utf8mb4 for MySQL 8.4 compatibility
+ALTER TABLE ospos_app_config CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
```
objecttothis, line 3: @jekkos maybe you can weigh in on this too, but utf8mb4 is legacy and does not handle unicode characters consistently with sorting. I think the current recommended collation is utf8mb4_0900_ai_ci as it handles non-latin character sorting much more accurately. It used to be that general was 1-3% faster, but on modern systems it's closer to 0-2% for sorting, 0-1% for comparisons and identical for index lookups. All this with the advantage of better multilingual support.

.github/workflows/phpunit.yml
```diff
@@ -38,6 +38,10 @@ jobs:
           - '8.2'
           - '8.3'
           - '8.4'
+        db-image:
+          - 'mariadb:10.5'
+          - 'mysql:8.0'
```
objecttothis, line 42: @jekkos our minimum system requirements list MySQL 5.7 as compatible... I don't think it is anymore, so we might want to update it there. If it is then maybe add 5.7 here.

app/Database/Migrations/sqlscripts/pre-3.0.2/2.2.2_to_2.3.sql
```diff
@@ -1,9 +1,9 @@
 CREATE TABLE IF NOT EXISTS `ospos_stock_locations` (
   `location_id` int(11) NOT NULL AUTO_INCREMENT,
-  `location_name` varchar(255) CHARACTER SET utf8 DEFAULT NULL,
+  `location_name` varchar(255) CHARACTER SET utf8mb4 DEFAULT NULL,
```
objecttothis: Unless it's a breaking change where migrations won't run, all old migrations must be frozen and unchanged.  If the migration will still run through all the changes, then we need to leave these migrations the same. The reason for this is onboarding for existing databases.  For example if I migrate from 3.3.2 to 3.4.2 then some of these migrations would have already run and will not run again during the upgrade. This means that we could potentially end up with incongruent schemas in the user base and that creates a rat's nest for compatibility.

app/Database/Migrations/sqlscripts/initial_schema.sql
objecttothis, line 1: Same as with the old migrations.  This initial schema which used to be database.sql which was strictly locked, so unless it's a breaking change and it won't run, we need to leave it as it was and let the migrations do the schema changes.

app/Helpers/migration_helper.php
```diff
@@ -18,10 +18,15 @@ function execute_script(string $path): bool
 
     $db = Database::connect();
 
-    $success = true; // whether *all* queries succeeded
+    $success = true;
     foreach ($sqls as $statement) {
         $statement = "$statement;";
-        $hadError = !$db->simpleQuery($statement);
+        try {
```
objecttothis: No harm in this but my research tells me that simpleQuery() doesn't throw any exceptions.  I might be wrong though, so I would leave this.

.github/workflows/phpunit.yml
```diff
@@ -38,6 +38,10 @@ jobs:
           - '8.2'
```
objecttothis: In the wiki and install.md 8.2 and 8.3 are not listed as supported.  @jekkos should we remove them here for consistency?  Also 5.7 is listed, but I'm not so sure about compatibility.

objecttothis (changes requested):
Take a look at my review comments.  I'll work on checking out this code and making sure it handles a database upgrade properly.

.github/workflows/phpunit.yml
jekkos: I would say that 8.1 and 8.2 should also be supported. If possible then I would add them to the matrix. Glad we made the move to github actions as it will allow to check all these things easily

objecttothis:
> **Actionable comments posted: 1**
> ♻️ Duplicate comments (1)
> 
> > app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql (1)> `19-23`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Add suspended-sales tables to the utf8mb4 conversion list.**
> > > Line 19–Line 23 converts active sales tables, but suspended-sales tables are still missing. Incremental upgrades can keep those tables on legacy collation and still trigger charset/collation mismatches.
> > > 
> > > 🛠️ Proposed patch
> > > ```diff
> > >  ALTER TABLE ospos_sales CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_items CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_items_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_payments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_items CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_items_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_payments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql`
> > > around lines 19 - 23, The migration converts active sales tables to utf8mb4 but
> > > omits the suspended-sales equivalents, risking collation mismatches; add the
> > > same ALTER TABLE ... CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci
> > > statements for the suspended tables (ospos_suspended_sales,
> > > ospos_suspended_sales_items, ospos_suspended_sales_items_taxes,
> > > ospos_suspended_sales_payments, ospos_suspended_sales_taxes) so they undergo the
> > > same charset/collation conversion as ospos_sales and its related tables.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> 🪄 Autofix (Beta)
> 
> 
> ℹ️ Review info



> **Actionable comments posted: 1**
> ♻️ Duplicate comments (1)
> 
> > app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql (1)> `19-23`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Add suspended-sales tables to the utf8mb4 conversion list.**
> > > Line 19–Line 23 converts active sales tables, but suspended-sales tables are still missing. Incremental upgrades can keep those tables on legacy collation and still trigger charset/collation mismatches.
> > > 
> > > 🛠️ Proposed patch
> > > ```diff
> > >  ALTER TABLE ospos_sales CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_items CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_items_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_payments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_items CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_items_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > +ALTER TABLE ospos_sales_suspended_payments CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > >  ALTER TABLE ospos_sales_taxes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql`
> > > around lines 19 - 23, The migration converts active sales tables to utf8mb4 but
> > > omits the suspended-sales equivalents, risking collation mismatches; add the
> > > same ALTER TABLE ... CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci
> > > statements for the suspended tables (ospos_suspended_sales,
> > > ospos_suspended_sales_items, ospos_suspended_sales_items_taxes,
> > > ospos_suspended_sales_payments, ospos_suspended_sales_taxes) so they undergo the
> > > same charset/collation conversion as ospos_sales and its related tables.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> 🪄 Autofix (Beta)
> 
> 
> ℹ️ Review info

You're hallucinating @coderabbitai. Those tables aren't a thing. Suspended sales tables were replaced a long time ago in lieu of a sale_status column in the ospos_sales table.

.github/workflows/phpunit.yml
jekkos: CI4 should still be compatible according to the docs, but they are not testing on 5.7 anymore. I believe it's already end-of-life. We could give it a try and add it to the list here, wdyt

jekkos:
> @jekkos I wonder if it's possible for us to create example populated databases from various versions so that we can better run automated testing. I think some migrations succeed when they are run against new installs but fail when run against upgrade databases.



> @jekkos I wonder if it's possible for us to create example populated databases from various versions so that we can better run automated testing. I think some migrations succeed when they are run against new installs but fail when run against upgrade databases.

I believe the PR here might include this already. If it adds a testmatrix with different mysql versions then we  might be running the tests on all these different versions.

So at least the schema migration is covered. 
Maybe your idea is to test a real data migration as well. Like for a specific language?

.github/workflows/phpunit.yml
objecttothis: #4485 is about to make anything below 8.2 incompatible, so let's remove PHP 8.1 and 8.0.  Yes to 8.3 and up.

objecttothis:
> > @jekkos I wonder if it's possible for us to create example populated databases from various versions so that we can better run automated testing. I think some migrations succeed when they are run against new installs but fail when run against upgrade databases.
> 
> > @jekkos I wonder if it's possible for us to create example populated databases from various versions so that we can better run automated testing. I think some migrations succeed when they are run against new installs but fail when run against upgrade databases.
> 
> I believe the PR here might include this already. If it adds a testmatrix with different mysql versions then we might be running the tests on all these different versions.
> 
> So at least the schema migration is covered. Maybe your idea is to test a real data migration as well. Like for a specific language?

Yes, having real data populated in the tables to simulate an upgrade path.  When we just simulation migrations on clean installs we catch many things, but sometimes migrations fail in upgrades but not in new installs.

app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql
objecttothis: @aliworksx08 thanks for pushing those commits. I will check this out and test it against an existing database on Monday.  I'd like to have @jekkos weigh in on my recommendation to use `utf8mb4_0900_ai_ci` here though given that the webapp is multilingual and general is not as recommended as it once was.  I can be convinced though if someone has an opinion.

objecttothis: **Short answer:**  
Choose **utf8mb4_0900_ai_ci** for any modern, multilingual CodeIgniter app running on MySQL 8+. It’s more accurate, more Unicode‑complete, and handles multilingual sorting/searching far better than **utf8mb4_general_ci**.

---

## 🧠 Core takeaway  
**utf8mb4_0900_ai_ci is the superior collation** for multilingual applications because it uses MySQL 8’s improved Unicode 9.0 rules, giving you better sorting, comparisons, and linguistic correctness across languages.

---

## 🔍 What actually differs between them

### 1. **Unicode version**
- **utf8mb4_0900_ai_ci** → Unicode 9.0 rules  
- **utf8mb4_general_ci** → pre‑Unicode 4.0 era, extremely outdated

This affects:
- how characters compare  
- how accents are handled  
- how languages sort  
- how symbols and emojis behave  

If you care about correct ordering in languages like German, French, Turkish, Arabic, Chinese, etc., the 0900 collations are dramatically better.

---

### 2. **Accuracy vs. speed**
- **utf8mb4_general_ci** was designed to be *fast but imprecise*.  
  - It treats many different characters as equal.  
  - It ignores many linguistic rules.  
  - It’s basically a “quick and dirty” collation.

- **utf8mb4_0900_ai_ci** is *more accurate* and still fast enough for modern hardware.  
  - Better accent handling  
  - Better case folding  
  - Better multilingual sorting  

Unless you’re running a 2005‑era server, the performance difference is negligible.

---

### 3. **Multilingual correctness**
If your app supports:
- English  
- European languages  
- Middle Eastern languages  
- Asian languages  
- Emojis  
- Mixed‑script user input  

…then **utf8mb4_0900_ai_ci** is the right choice.

**utf8mb4_general_ci** will mis-sort or mis-compare many characters.  
Examples:
- “ß” vs “ss”  
- Turkish dotted/dotless “i”  
- Chinese radicals  
- Accented characters in French or Spanish  

---

### 4. **MySQL version compatibility**
- **utf8mb4_0900_ai_ci** → **MySQL 8+ only**  
- **utf8mb4_general_ci** → works on older versions

If you’re on MySQL 5.7 or earlier, you *cannot* use the 0900 collations.

---

## 🧩 CodeIgniter considerations
CodeIgniter doesn’t care which collation you use as long as:
- your database connection charset is set to `utf8mb4`
- your tables/columns use the same charset/collation
- your PHP files are UTF‑8 encoded

In `app/Config/Database.php`:

```php
'charset'   => 'utf8mb4',
'DBCollat'  => 'utf8mb4_0900_ai_ci',
```

---

## 🏁 Recommendation
If you’re on **MySQL 8+**, use:

**utf8mb4_0900_ai_ci**  
or, if you need accent‑sensitive comparisons:  
**utf8mb4_0900_as_ci**

If you’re stuck on MySQL 5.7 or earlier, use:

**utf8mb4_unicode_ci** (better)  
instead of  
**utf8mb4_general_ci** (worst)

.github/workflows/phpunit.yml
objecttothis: If we choose `utf8mb4_0900_ai_ci` in the migration (see the conversation below), that will make it incompatible with MySQL 5.7 and therefore we would want to remove MySQL 5.7 from the wiki and README.md compatibility charts. I would question the wisdom in using MySQL 5.7 anyway, though.

app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql
objecttothis: @aliworksx08 let's change these collations to `utf8mb4_unicode_520_ci` instead of `utf8mb4_general_ci`

objecttothis (changes requested):

objecttothis:
@Matsx901 do you want to make these last requested changes so that we can merge this?  The PR needs to have the latest changes from the master merged into it since several migration changes have been made recently to resolve issues with MariaDB migrations.

app/Database/Migrations/sqlscripts/3.5.0_mysql84_utf8mb4_conversion.sql
objecttothis: @aliworksx08 correction.  I just learned that utf8mb4_0900_ai_ci is not compatible with MariaDB and we don't intend to drop support for MariaDB. Instead the best cross-compatible version to use is `utf8mb4_unicode_520_ci`.  @coderabbitai, please confirm this assessment.

PR #4484, opened 2026-04-08
Merge branch 'master' of https://github.com/opensourcepos/opensourcepos

rafaelpoli:
## Summary by CodeRabbit

* **Chores**
  * Refactored asset loading strategy to explicitly use bundled and minified CSS/JavaScript files in production environments.
  * Expanded development environment with comprehensive external library references supporting UI components, charting, and utility features.


jekkos:
## Build Analysis

I tested the build process to understand the implications of this PR.

### Key Finding: This approach will NOT work as intended

**Why:**

1. **Resources folder is gitignored**: `public/resources/` is in `.gitignore`, so the actual CSS/JS files won't be committed to the repository. The injected references in header.php would point to files that don't exist in the repo.

2. **Content-based hashes differ between environments**: The gulp-rev plugin generates hashes based on file content. I compared the hashes:

   | File | PR hash | My build hash |
   |------|---------|---------------|
   | bootstrap.autocomplete.css | 292fc0ad3b | 4875cf7b0d |
   | invoice.css | 1eae5e39b9 | a99a4dfac3 |
   | ospos.css | 9ead7561a7 | d0b91fdf8f |
   | ospos_print.css | 2ba645b044 | bf10c1438b |
   | ... | ... | ... |

   Different dependency versions, npm versions, or even timestamps cause different hashes. Your injected content would reference different files than what gets built on other machines.

3. **gulp-inject REPLACES content**: When gulp runs, it will replace any existing content between the inject tags. So even if you committed this, the next build would overwrite it with fresh (different) hashes.

### Conclusion

The current workflow is correct:
- header.php contains only empty inject tags in the repo
- CI/CD or developers run `npm run build` to generate resources and inject references
- Resources are built per-environment with correct hashes

Committing pre-injected content would break the build for anyone else.

jekkos:
## Resolution (confirming decision to close)

For anyone following this PR, the proposed approach would not work as intended.

**The current workflow is correct:**
- Keep inject tags empty in the repository
- Developers run `npm run build` after cloning to generate resources
- CI/CD builds resources during deployment

**Why committing injected content would fail:**
1. `public/resources/` is gitignored, so referenced files wouldn't exist
2. Content-based hashes differ between environments (different npm versions, dependency versions, etc.)
3. `gulp-inject` replaces content on each build anyway

This is standard practice for asset pipelines - similar to not committing webpack/vite output. The empty inject tags are intentional and prevent merge conflicts from differing build hashes.

PR #4498, opened 2026-04-15
feat: Update translations for multiple languages

jekkos:
## Summary
- Added Calendar.php translations for multiple languages (ar-EG, de-DE, es-ES, es-MX, fr, he, it, km, lo, ml, nb, nl-NL, pt-BR, ta, tl, ur)
- Fixed translation issues in Common.php, Employees.php, Items.php, Reports.php, Sales.php for de-DE, es-ES, fr, it, nl-NL, pt-BR
- Removed English fallbacks from non-English translations to activate proper CI4 language fallback mechanism

## Changes
- 45 files changed across 16+ locales
- Added new Calendar translations where missing
- Corrected existing translations with proper context

## Test Plan
- [ ] Verify PHP syntax is valid (no parse errors)
- [ ] Language files load correctly in the application
- [ ] Empty values fallback to English as expected

## Summary by CodeRabbit

* **Documentation**
  * Corrected calendar month key mappings across multiple language files (Georgian, Lao, Malayalam, Norwegian).
  * Improved Spanish error message wording for stock location validation clarity.


objecttothis (dismissed):
Looks good.  Two things to note

- Refactor all these arrays to stop using `"` and use `'` instead.  In PHP when `"` is used the parser is called for string interpolation, but I believe the way we use string interpolation does not require the parser to be called.
- We need to push these up to weblate so that it doesn't overwrite the translations.

PR #4500, opened 2026-04-15
fix: Add null/empty validation for $item_id in Receivings::postEditItem()

### LINKED ISSUE #4486: Add null/empty validation for $item_id in Receivings::postEditItem()
## Summary

The `postEditItem(int|string|null $item_id)` method in `app/Controllers/Receivings.php` currently accepts `null` or an empty string without any explicit guard. This allows edit requests with a missing item identifier to reach `Receiving_lib::edit_item()` and silently fail rather than returning a clear error response to the caller.

## Suggested Fix

Add an explicit guard at the start of `postEditItem` to reject `null` or empty values and return an appropriate error before proceeding with any lookup or database calls:

```php
public function postEditItem(int|string|null $item_id): string
{
    $data = [];
    if ($item_id === null || trim((string)$item_id) === '') {
        $data['error'] = lang('Receivings.error_editing_item');
        return $this->_reload($data);
    }
    // ... rest of method
}
```

## Context

- File: `app/Controllers/Receivings.php`
- Method: `postEditItem`
- Identified in: #4485 (comment: https://github.com/opensourcepos/opensourcepos/pull/4485#discussion_r3058744619)
- Requested by: @objecttothis

## Acceptance Criteria

- [ ] Guard clause added to reject `null` or empty `$item_id` at the start of `postEditItem`
- [ ] Returns a meaningful error message (via `_reload` with `error` key) rather than silently proceeding
- [ ] All user inputs are sanitized/cast before use in any lookup or DB calls

jekkos:
## Summary

- Adds guard clause to `postEditItem()` method to reject null or empty `$item_id` values
- Returns meaningful error message via `_reload()` before any database operations
- Prevents silent failures in `Receiving_lib::edit_item()`

## Changes

- Added validation check at the start of `postEditItem()` in `app/Controllers/Receivings.php`
- Uses existing error message pattern `lang('Receivings.error_editing_item')`

## Related

Closes #4486

### DIFF
--- a/app/Controllers/Receivings.php
+++ b/app/Controllers/Receivings.php
@@ -198,6 +198,12 @@ class Receivings extends Secure_Controller
     {
         $data = [];
 
+        // Validate item_id to prevent null/empty values from reaching edit_item()
+        if ($item_id === null || $item_id === '') {
+            $data['error'] = lang('Receivings.error_editing_item');
+            return $this->_reload($data);
+        }
+
         $validation_rule = [
             'price'    => 'trim|required|decimal_locale',
             'quantity' => 'trim|required|decimal_locale',


jekkos:
## Implementation Note

After investigating whether the CodeIgniter 4 validation system could be used instead of the guard clause, I confirmed that **CI4's `$this->validate()` is designed for form/POST data validation, not for URL segment route parameters**.

The `$item_id` parameter comes from the URL route (`receivings/editItem/{item_id}`), not from POST form data, so the CI4 validation rules cannot be applied to it.

### Route Parameters vs POST Data

| Type | Source | Validation Method |
|------|--------|-------------------|
| **Route parameter** | URL segment | Guard clause in controller |
| **POST data** | Form body | `$this->validate($rules)` |

The guard clause is the **correct, idiomatic CI4 approach** for validating route parameters and follows the existing patterns in this codebase.

---

Closing this PR as the fix has been merged. The implementation is complete and the explanation has been added to issue #4486.

jekkos:
As discussed in issue #4486, the CI4 validation system cannot be used for route parameters. The guard clause implementation is the correct approach. Closing this PR.

app/Controllers/Receivings.php
```diff
@@ -198,6 +198,12 @@ public function postEditItem(int|string|null $item_id): string
     {
         $data = [];
 
+        // Validate item_id to prevent null/empty values from reaching edit_item()
+        if ($item_id === null || $item_id === '') {
```
objecttothis, line 202: Good = Leave this as is.
Better = Refactor $item_id to $itemId in this function for PSR compliance.
Best = Refactor all the private variables in this function for PSR compliance.

objecttothis (commented):
See my comments.

PR #4507, opened 2026-04-15
fix: Capture CSV import failures in save_tax_data and save_inventory_quantities

jekkos:
## Summary

- Change `save_tax_data()` return type from `void` to `bool`
- Change `save_inventory_quantities()` return type from `void` to `bool` 
- Accumulate failure status with `&=` operator in `save_inventory_quantities()`
- Update `postImportCsvFile()` to capture return values and set `isFailedRow`
- Properly propagate failures to `failCodes` array

## Problem

In `postImportCsvFile()`, `save_tax_data()` and `save_inventory_quantities()` were called fire-and-forget. A write failure in either helper never propagated to the row's `failCodes` array, so the import could report a row as successfully imported even when tax or stock-quantity data was not persisted.

## Solution

1. Changed both helper functions to return `bool` indicating success/failure
2. Captured return values in `postImportCsvFile()` and set `isFailedRow = true` on failure
3. Changed `return;` to `continue;` in `save_inventory_quantities()` to allow loop completion
4. Updated log message to reflect all failure cases

Fixes #4475

Ultraworked with [Sisyphus](https://github.com/code-yeongyu/oh-my-openagent)

Co-authored-by: Sisyphus <clio-agent@sisyphuslabs.ai>

## Summary by CodeRabbit

* **Bug Fixes**
  * CSV import now treats tax, inventory, and attribute save failures as definitive row failures, preventing partially imported rows from being marked complete.
  * Partial-import failure messaging clarified to report errors more accurately as occurring "while saving item."
  * Imports no longer abort when per-location quantity cells are empty; missing tax cells are handled gracefully so other row data continues to process.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opensourcepos/opensourcepos/pull/4507?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/Controllers/Items.php
+++ b/app/Controllers/Items.php
@@ -1055,14 +1055,20 @@ class Items extends Secure_Controller
                         });
 
                         if (!$isFailedRow && $this->item->save_value($itemData, $itemId)) {
-                            $this->save_tax_data($row, $itemData);
-                            $this->save_inventory_quantities($row, $itemData, $allowedStockLocations, $employeeId);
+                            if (!$this->save_tax_data($row, $itemData)) {
+                                $isFailedRow = true;
+                            }
+                            if (!$this->save_inventory_quantities($row, $itemData, $allowedStockLocations, $employeeId)) {
+                                $isFailedRow = true;
+                            }
                             $csvAttributeValues = $this->extractAttributeData($row);
-                            $isFailedRow = !$this->attribute->saveCSVRowAttributeData($csvAttributeValues, $itemData, $attributeData);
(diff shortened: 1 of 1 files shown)


objecttothis (approved):
Looks good.  I'll leave it up to you whether you take CodeRabbit's suggestion to refactor the naming on functions you touched or not.  I know you are in favor of just changing all the code at once.  I'm OK with this, but we need to do it if that's what we are going to do.  As is, the code is in this funny hybrid state.  We keep starting new branches like that rather than updating the code, so unless we are going to change it as we go, then it won't get done unless we open a PR and just do it all at once.

PR #4513, opened 2026-04-16
Fix allowed hostnames with external reverse proxy

jekkos:
## Summary by CodeRabbit

* **Documentation**
  * Clarified reverse-proxy guidance, documented proxy IP formats (wildcard, CIDR, comma-separated) and requirement to whitelist derived hostnames; fixed markdown formatting in install docs.
* **New Features**
  * Optional proxy configuration to trust X-Forwarded-Host from specified IPs/CIDR/wildcard; hostname validation still enforced against allowed hostnames.
* **Tests**
  * Added tests covering proxy trust rules, CIDR/wildcard behavior, env parsing, validation and fallback scenarios.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,23 @@ CI_ENVIRONMENT = production
 app.allowedHostnames = ''
 
 #--------------------------------------------------------------------
+# REVERSE PROXY CONFIGURATION
+#--------------------------------------------------------------------
+# Configure trusted proxy IPs for X-Forwarded-Host header support.
+#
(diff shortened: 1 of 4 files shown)


jekkos:
Dev server is currently broken and this might be the fix. The HOST header is not sent if you have a setup behind a reverse proxy like nginx, so it fallsback to localhost instead. We should also look at the forwarded-for header in this case.

jekkos:
Closing this one as it makes setup even more complex (and does not solver current issue in dev)

PR #3161, opened 2021-04-16
Person attributes

DamianoSilverhand:
This PR adds person attributes on Customers, Suppliers and Employees. Also includes person attributes in customer import using csv.

PR #4521, opened 2026-04-20
[Feature]: Add person attributes support with unified attribute system

jekkos:
## Summary

This PR implements person attributes using the existing unified attribute system, addressing PR #3161 with a complete rework based on the latest master.

## Architectural Approach

**Using Unified Attribute System** (instead of separate person_attribute tables):
- Add `person_id` column to existing `attribute_links` table
- Add `person_attribute` column to `attribute_definitions` to distinguish item vs person attributes
- Add visibility flags: `SHOW_IN_CUSTOMERS = 16`, `SHOW_IN_EMPLOYEES = 32`, `SHOW_IN_SUPPLIERS = 64`
- Reuse existing `attribute_definitions`, `attribute_values`, and `attribute_links` tables
- Shared logic in base `Persons` controller for Customers, Employees, Suppliers

### Database Migrations
- `20260402000000_AddPersonToAttributeLinks.php` - Adds `person_id` to `attribute_links`
- `20260403000000_AddPersonAttributeFlag.php` - Adds `person_attribute` to `attribute_definitions`

### Model Changes
- `Attribute.php` - Added person attribute methods and `getDefinitionsByType()` for filtering

### Controller Changes
- `Persons.php` - Added shared `getPersonAttributes()` and `savePersonAttributes()` methods
- `Customers.php`, `Employees.php`, `Suppliers.php` - Integrated person attribute handling
- `Attributes.php` - Added person visibility flags

### View Changes
- `attributes/person.php` - New view for rendering person attributes
- Customer/Employee/Supplier forms - Added person attributes container

### Language Changes  
- `Attributes.php` - Added person visibility labels

## Constants (Bit Flags)
```php
SHOW_IN_ITEMS = 1
SHOW_IN_SALES = 2
SHOW_IN_RECEIVINGS = 4
SHOW_IN_SEARCH = 8
SHOW_IN_CUSTOMERS = 16
SHOW_IN_EMPLOYEES = 32
SHOW_IN_SUPPLIERS = 64
```

## Benefits
1. **Consistency** - Same attribute system for items, sales, receivings, and persons
2. **Less code duplication** - Reuse existing attribute infrastructure  
3. **Easier maintenance** - Single source of truth for attribute logic
4. **CI4 native** - Built on current codebase

## Testing
1. Run database migrations
2. Test attribute definition visibility settings
3. Test person attribute forms for customers, employees, suppliers

Supersedes #3161

## Release Notes

**New Features**
- Custom attributes can now be applied to customers, employees, and suppliers, enabling you to extend these entity types with configurable fields for your business needs.

### DIFF
--- a/app/Controllers/Customers.php
+++ b/app/Controllers/Customers.php
@@ -3,7 +3,7 @@
 namespace App\Controllers;
 
 use App\Libraries\Mailchimp_lib;
-
+use App\Models\Attribute;
 use App\Models\Customer;
 use App\Models\Customer_rewards;
 use App\Models\Tax_code;
@@ -15,34 +15,31 @@ use stdClass;
 
 class Customers extends Persons
 {
-    private string $_list_id;
-    private Mailchimp_lib $mailchimp_lib;
-    private Customer_rewards $customer_rewards;
+    private string $listId;
+    private Mailchimp_lib $mailchimpLib;
+    private Customer_rewards $customerRewards;
     private Customer $customer;
-    private Tax_code $tax_code;
-    private array $config;
+    private Tax_code $taxCode;
+    private array $appConfig;
 
     public function __construct()
     {
         parent::__construct('customers');
-        $this->mailchimp_lib = new Mailchimp_lib();
-        $this->customer_rewards = model(Customer_rewards::class);
+        $this->mailchimpLib = new Mailchimp_lib();
+        $this->customerRewards = model(Customer_rewards::class);
         $this->customer = model(Customer::class);
-        $this->tax_code = model(Tax_code::class);
-        $this->config = config(OSPOS::class)->settings;
+        $this->taxCode = model(Tax_code::class);
+        $this->appConfig = config(OSPOS::class)->settings;
 
         $encrypter = Services::encrypter();
 
-        if (!empty($this->config['mailchimp_list_id'])) {
-            $this->_list_id = $encrypter->decrypt($this->config['mailchimp_list_id']);
+        if (!empty($this->appConfig['mailchimp_list_id'])) {
+            $this->listId = $encrypter->decrypt($this->appConfig['mailchimp_list_id']);
         } else {
-            $this->_list_id = '';
+            $this->listId = '';
         }
     }
 
-    /**
-     * @return string
-     */
     public function getIndex(): string
     {
         $data['table_headers'] = get_customer_manage_table_headers();
@@ -50,19 +47,13 @@ class Customers extends Persons
         return view('people/manage', $data);
     }
 
-    /**
-     * Gets one row for a customer manage table. This is called using AJAX to update one row.
-     * @return ResponseInterface
-     */
-    public function getRow(int $row_id): ResponseInterface
+    public function getRow(int $rowId): ResponseInterface
     {
-        $person = $this->customer->get_info($row_id);
+        $person = $this->customer->get_info($rowId);
 
-        // Retrieve the total amount the customer spent so far together with min, max and average values
-        $stats = $this->customer->get_stats($person->person_id);    // TODO: This and the next 11 lines are duplicated in search().  Extract a method.
+        $stats = $this->customer->get_stats($person->person_id);
 
         if (empty($stats)) {
-            // Create object with empty properties.
             $stats = new stdClass();
             $stats->total = 0;
             $stats->min = 0;
@@ -72,17 +63,11 @@ class Customers extends Persons
             $stats->quantity = 0;
         }
 
-        $data_row = get_customer_data_row($person, $stats);
+        $dataRow = get_customer_data_row($person, $stats);
 
-        return $this->response->setJSON($data_row);
+        return $this->response->setJSON($dataRow);
     }
 
-
-    /**
-     * Returns customer table data rows. This will be called with AJAX.
-     *
-     * @return void
-     */
     public function getSearch(): ResponseInterface
     {
         $search = $this->request->getGet('search');
@@ -92,15 +77,13 @@ class Customers extends Persons
         $order = $this->request->getGet('order', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
 
         $customers = $this->customer->search($search, $limit, $offset, $sort, $order);
-        $total_rows = $this->customer->get_found_rows($search);
+        $totalRows = $this->customer->get_found_rows($search);
 
-        $data_rows = [];
+        $dataRows = [];
 
         foreach ($customers->getResult() as $person) {
-            // Retrieve the total amount the customer spent so far together with min, max and average values
-            $stats = $this->customer->get_stats($person->person_id);    // TODO: duplicated... see above
+            $stats = $this->customer->get_stats($person->person_id);
             if (empty($stats)) {
-                // Create object with empty properties.
                 $stats = new stdClass();
                 $stats->total = 0;
                 $stats->min = 0;
@@ -110,16 +93,12 @@ class Customers extends Persons
                 $stats->quantity = 0;
             }
 
-            $data_rows[] = get_customer_data_row($person, $stats);
+            $dataRows[] = get_customer_data_row($person, $stats);
         }
 
-        return $this->response->setJSON(['total' => $total_rows, 'rows' => $data_rows]);
+        return $this->response->setJSON(['total' => $totalRows, 'rows' => $dataRows]);
     }
 
-    /**
-     * Gives search suggestions based on what is being searched for
-     * @return ResponseInterface
-     */
     public function getSuggest(): ResponseInterface
     {
         $search = $this->request->getGet('term');
@@ -128,10 +107,7 @@ class Customers extends Persons
         return $this->response->setJSON($suggestions);
     }
 
-    /**
-     * @return ResponseInterface
-     */
-    public function suggest_search(): ResponseInterface
+    public function suggestSearch(): ResponseInterface
     {
         $search = $this->request->getGet('term');
         $suggestions = $this->customer->get_search_suggestions($search, 25, false);
@@ -139,16 +115,11 @@ class Customers extends Persons
         return $this->response->setJSON($suggestions);
     }
 
-    /**
-     * Loads the customer edit form
-     * @return string
-     */
-    public function getView(int $customer_id = NEW_ENTRY): string
+    public function getView(int $customerId = NEW_ENTRY): string
     {
-        // Set default values
-        if ($customer_id == null) $customer_id = NEW_ENTRY;
+        if ($customerId == null) $customerId = NEW_ENTRY;
 
-        $info = $this->customer->get_info($customer_id);
+        $info = $this->customer->get_info($customerId);
         foreach (get_object_vars($info) as $property => $value) {
             $info->$property = $value;
         }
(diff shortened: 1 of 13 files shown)


app/Controllers/Customers.php
```diff
@@ -373,43 +327,24 @@ public function postDelete(): ResponseInterface
...
     public function getCsvImport(): string
     {
         return view('customers/form_csv_import');
     }
 
-    /**
-     * Imports a CSV file containing customers. Used in app/Views/customers/form_csv_import.php
-     *
-     * @return ResponseInterface
-     * @noinspection PhpUnused
```
objecttothis, line 404: removing `@noinspection PhpUnused` causes IDE's to complain that this is an unused function even though it's called by AJAX.  I would prefer that we keep these.

app/Controllers/Employees.php
```diff
@@ -152,108 +132,98 @@ public function postSave(int $employee_id = NEW_ENTRY): ResponseInterface
             'comments'     => $this->request->getPost('comments', FILTER_SANITIZE_FULL_SPECIAL_CHARS)
         ];
 
-        $grants_array = [];
-        $isAdmin = $this->employee->isAdmin($current_user->person_id);
+        $grantsArray = [];
```
objecttothis, line 135: This variable name needs to be changed.  We shouldn't be listing the type of variable in the name.  Instead the variable name needs to better explain what it's doing.  I recommend $grantedPermissions or $permissionGrants.  Also, not required here but it would be better to refactor the code block below into a function and call it from here.  Something like `$grantedPermissions = $this->getPermissionGrants($currentUser);` should do the trick. As is, this function violates SRP.  I'm not suggesting a complete rewrite in this PR, but this small improvement could help.

app/Controllers/Employees.php
```diff
@@ -152,108 +132,98 @@ public function postSave(int $employee_id = NEW_ENTRY): ResponseInterface
...
-                if ($this->employee->isAdmin((int)$emp_id)) {
+        if (!$this->employee->isAdmin($currentUser->person_id)) {
+            foreach ($employeesToDelete as $empId) {
+                if ($this->employee->isAdmin((int)$empId)) {
                     return $this->response->setJSON(['success' => false, 'message' => lang('Employees.error_deleting_admin')]);
                 }
             }
         }
 
-        if ($this->employee->delete_list($employees_to_delete)) {    // TODO: this is passing a string, but delete_list expects an array
```
objecttothis, line 237: If we are going to delete this todo, we should create an issue in GitHub to resolve it so it doesn't get forgotten.

app/Controllers/Persons.php
```diff
@@ -43,34 +39,88 @@ public function getSuggest(): ResponseInterface
...
-     * i.e. <?php echo nameize("john o'grady-smith"); ?>
-     *
-     * returns John O'Grady-Smith
-     */
     protected function nameize(string $input): string
     {
-        $adjusted_name = str_name_case($input);
+        $adjustedName = str_name_case($input);
 
-        // TODO: Use preg_replace to match HTML entities and convert them to lowercase. This is a workaround for https://github.com/tamtamchik/namecase/issues/20
```
objecttothis, line 71: If we are going to delete this TODO then we need a GitHub issue to resolve it.

app/Controllers/Persons.php
```diff
@@ -43,34 +39,88 @@ public function getSuggest(): ResponseInterface
...
-    /**
-     * Capitalize segments of a name, and put the rest into lower case.
-     * You can pass the characters you want to use as delimiters as exceptions.
-     * The function supports UTF-8 strings
-     *
-     * Example:
-     * i.e. <?php echo nameize("john o'grady-smith"); ?>
-     *
-     * returns John O'Grady-Smith
-     */
```
objecttothis, line 66: We either need to keep the PHPdoc block or we need to rename this function to better describe what it does. If we remove the doc block without refactoring, it's not incredibly clear what the function is doing.

app/Controllers/Suppliers.php
```diff
@@ -17,33 +18,21 @@ public function __construct()
...
         return view('people/manage', $data);
     }
 
-    /**
-     * Gets one row for a supplier manage table. This is called using AJAX to update one row.
-     * @param $row_id
-     * @return ResponseInterface
-     */
-    public function getRow($row_id): ResponseInterface
+    public function getRow($rowId): ResponseInterface
```
objecttothis, line 28: If we are going to remove the PHPdoc block we should probably rename this function to getSupplier() or getSupplierById() to better describe what the function is doing. If it's fetching a table row then getSupplierTableRow() would be better IMO.

app/Models/Attribute.php
```diff
@@ -41,7 +43,12 @@ class Attribute extends Model
     public const SHOW_IN_ITEMS = 1;    // TODO: These need to be moved to constants.php
     public const SHOW_IN_SALES = 2;
     public const SHOW_IN_RECEIVINGS = 4;
-    public function deleteDropdownAttributeValue(string $attribute_value, int $definition_id): bool
+    public const SHOW_IN_SEARCH = 8;
+    public const SHOW_IN_CUSTOMERS = 16;
+    public const SHOW_IN_EMPLOYEES = 32;
+    public const SHOW_IN_SUPPLIERS = 64;
```
objecttothis, line 49: Problem: 8 is the limit number of flags we can use a tinyint to store a bitmask.  A better solution would be a flags static table and a attribute definition flags mapping table.  This makes adding future flags easier and more human readable in the database.  It also means that an ORM using Entities is more feasible.

Something like this would be better.  I think we should create a github issue to implement this.  Best would be instead of adding more flags using our current method, but it would be fine to do it this way then later refactor it.

```
CREATE TABLE `flags` (
    `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
    `name` VARCHAR(100) NOT NULL,
    PRIMARY KEY (`id`),
    UNIQUE KEY `uq_flags_name` (`name`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

CREATE TABLE `attribute_definition_flags` (
    `attribute_definition_id` INT UNSIGNED NOT NULL,
    `flag_id` INT UNSIGNED NOT NULL,

    PRIMARY KEY (`attribute_definition_id`, `flag_id`),

    CONSTRAINT `fk_adf_attribute_definition`
        FOREIGN KEY (`attribute_definition_id`)
        REFERENCES `attribute_definitions` (`id`)
        ON DELETE CASCADE
        ON UPDATE CASCADE,

    CONSTRAINT `fk_adf_flag`
        FOREIGN KEY (`flag_id`)
        REFERENCES `flags` (`id`)
        ON DELETE CASCADE
        ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

app/Models/Attribute.php
```diff
@@ -1227,4 +1253,227 @@ private function updateAttributeValue(int $attributeId, string $dataType, mixed
             $itemsBuilder->update();
         }
     }
+
+    /**
+     * Gets all attributes connected to a person given the person_id
+     *
+     * @param int $personId Person to retrieve attributes for.
```
objecttothis, line 1260: I think this doc block needs to be removed to keep it in line with the standard of not having useless docblocks.

app/Models/Attribute.php
```diff
@@ -1227,4 +1253,227 @@ private function updateAttributeValue(int $attributeId, string $dataType, mixed
...
+     * @return array Attributes for the person.
+     */
+    public function getAttributesByPerson(int $personId): array
```
objecttothis, line 1263: A better name for this is getAttributesByPersonId because the current naming could imply an object is passed, not an Id.

app/Views/attributes/person.php
```diff
@@ -0,0 +1,169 @@
...
+            'options'  => $definition_names,
+            'selected' => -1,
+            'class'    => 'form-control',
+            'id'       => 'definition_name'
+        ]) ?>
+    </div>
+</div>
+
+<?php foreach ($definition_values as $definitionId => $definitionValue) { ?>
+
```
objecttothis, line 26: extra white-space not needed.

app/Views/attributes/person.php
```diff
@@ -0,0 +1,169 @@
...
+            'name'     => 'definition_name',
+            'options'  => $definition_names,
+            'selected' => -1,
+            'class'    => 'form-control',
+            'id'       => 'definition_name'
+        ]) ?>
+    </div>
+</div>
+
+<?php foreach ($definition_values as $definitionId => $definitionValue) { ?>
```
objecttothis, line 25: The code inside this foreach loop is a violation of "separation of concerns".  We shouldn't be doing business logic in the view.  It also makes it difficult to test against since loading the view would be required.

The controller needs to compute what needs to be displayed, not the view. Inside a loop we should have the code to to do that there.
```
$renderAttributes[] = [
    'id'          => $definitionId,
    'label'       => $definitionValue['definition_name'],
    'type'        => $definitionValue['definition_type'],
    'value'       => $normalizedValue,
    'options'     => $definitionValue['values'] ?? null,
    'checked'     => $isChecked ?? null,
];
```

Then 

```
<?php foreach ($render as $attribute): ?>
    <?= form_label($attribute['label'], $attribute['label'], ['class' => 'control-label col-xs-3']) ?>

    <?php switch ($attribute['type']): ?>
        <?php case DATE: ?>
            <?= form_input($attribute) ?>
            <?php break ?>

        <?php case DROPDOWN: ?>
            <?= form_dropdown($attribute) ?>
            <?php break ?>

        <?php case TEXT: ?>
            <?= form_input($attribute) ?>
            <?php break ?>

        <?php case DECIMAL: ?>
            <?= form_input($attribute) ?>
            <?php break ?>

        <?php case CHECKBOX: ?>
            <?= form_checkbox($attribute) ?>
            <?php break ?>
    <?php endswitch ?>
<?php endforeach ?>
```

objecttothis (changes requested):

app/Views/attributes/person.php
jekkos: Not completely convinced on this one. We display a form control depending on the attribute type, so purely representation and not really business logic. But agree that your proposal is less verbose, so I might implement it

objecttothis: > Not completely convinced on this one. We display a form control depending on the attribute type, so purely representation and not really business logic. But agree that your proposal is less verbose, so I might implement it

On second look over the code I suppose the only real business logic done currently is the normalization of $value. Do what you'd like.  I would do it in the controller.

app/Models/Attribute.php
```diff
@@ -269,7 +276,7 @@ public function get_definitions_by_type(string $attribute_type, int $definition_
     public function get_definitions_by_flags(int $definition_flags, bool $include_types = false): array
     {
         $builder = $this->db->table('attribute_definitions');
-        $builder->where(new RawSql("definition_flags & $definition_flags"));    // TODO: we need to heed CI warnings to escape properly
+        $builder->where(new RawSql("definition_flags & $definition_flags"));
```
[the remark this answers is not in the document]
objecttothis: I looked at this too.  Right now the safety relies on the `int` type hint of the function. If anyone removes that it creates a significant attack vector.  I would parameterize the call.  Also, AI seems to indicate that we don't have to use RawSQL here which CI warns does zero mitigation.

PR #4445, opened 2026-03-17
Bootstrap 5

jekkos:
## Summary by CodeRabbit

* **New Features**
  * Appearance settings page and a user Profile modal (includes change-password UI).
  * RTL language toggle and responsive design option.

* **Improvements**
  * Site-wide migration to Bootstrap 5 with refreshed, responsive layouts and updated headers/footers.
  * Icons replaced with Bootstrap Icons; improved module tiles and forms.
  * Enhanced date-range picker and custom loading indicators in tables.

* **Documentation**
  * Added a Bootstrap 5 conversion WIP notice to the README.

PR #4530, opened 2026-05-04
chore: miscellaneous updates and improvements

BudsieBuds:
- reinstated 'update-licenses' task in gulp (accidentally removed in 3e844f2f891497e65523e859d626b368ec2f77d3)
- updated bootstrap, bootswatch, and various dev dependencies
- refinded text across UI
- applied consistency fixes
- added 'number' and 'tel' input types to relevant settings
- improved system info layout (still room for improvement, but better)

## Summary by CodeRabbit

* **New Features**
  * Added a search engine exclusion meta tag to prevent site indexing.

* **Improvements**
  * Numeric input constraints added to barcode, SMTP, tax, and contact fields.
  * Email protocol labels capitalized for clarity.
  * Templates and emails now use the site language code for HTML lang.
  * Login view streamlined; system info spacing and minor license title capitalization adjusted.

* **Chores**
  * Theme and build tooling updated; default build now runs license update.

* **Documentation**
  * Changelog updated with new release section and link.


app/Views/login.php
```diff
@@ -5,13 +5,18 @@
...
  */
+
+use Config\Services;
+
+$request = Services::request();
 ?>
 
 <!doctype html>
-<html lang="<?= current_language_code() ?>">
+<html lang="<?= $request->getLocale() ?>">
```
objecttothis: I know it was this way before but this really should be refactored to the Login controller.  In the index function we should add it to the $data array.  As it is, doing it in the view violates separation of concerns.  Something like this should do the trick:
```
            $data = [
                'has_errors'       => false,
                'is_new_install'   => !(MY_Migration::get_current_version()),
                'is_latest'        => $migration->is_latest(),
                'latest_version'   => $migration->get_latest_migration(),
                'gcaptcha_enabled' => $gcaptcha_enabled,
                'config'           => $config,
                'validation'       => $validation,
                'currentLanguage => $this->request->getLocale();
            ];
```
Then replace the business logic in the view with `$currentLanguage` and replace the `@var` declaration at the top of the view with `@var string $currentLanguage`

objecttothis (changes requested):
Great work.  Just the one change.

app/Views/login.php
BudsieBuds: @objecttothis I updated this to maintain consistency with the implementation in [`app/Views/partial/header.php`](https://github.com/BudsieBuds/opensourcepos/blob/chore-miscupdates/app/Views/partial/header.php). Based on the [CI4 docs](https://codeigniter.com/user_guide/outgoing/localization.html#retrieving-the-current-locale), retrieving the locale directly in the view seemed acceptable. What would you recommend? If we change it, I think we should update both locations to keep things consistent.

jekkos (dismissed):
Looks good to me

app/Views/login.php
objecttothis: I see what you mean. The problem with app/Views/partial/header.php is that it gets called from many other views which I think means that currentLanguage would need to get passed to every view that injects partial header unless there is another way.  I think that's why they called it directly in partial/header.php.  I'm OK with leaving the code as you have it for consistency with the current implementation of partial/header.php.  I'll let you decide.

objecttothis (dismissed):

objecttothis:
@BudsieBuds I think you just need to click "Resolve conversation" once you decide what you want to do with the getLocale() call in the view and it should unblock the merge.

BudsieBuds:
> @BudsieBuds I think you just need to click "Resolve conversation" once you decide what you want to do with the getLocale() call in the view and it should unblock the merge.

@objecttothis After looking into it a bit more, I decided to switch to `<?= current_language_code() ?>` from the locale helper. This respects the employee-specific language setting. I also rebased the branch. If everything looks good on your end and it's approved, I'll proceed with the merge.

objecttothis (approved):

BudsieBuds:
@jekkos @objecttothis Could someone with admin access update the branch protection rules? The PR can’t be merged because Travis CI is still listed as a required check, but Travis is no longer running for this repository. GitHub is waiting for a status that will never be reported. Thanks!

jekkos:
@BudsieBuds i have changed the protection rules and merged your changes. Btw I  continued your work on bootstrap 5 in a new branch, but have not taken it further for now. 

I ran it through coderabbit and the agents fixed any issue they found. But no further validation was done
https://github.com/opensourcepos/opensourcepos/pull/4445

PR #3720, opened 2023-03-18
Assignable Keyboard Shortcuts Updates

WebShells:
Few updates regarding HotKeys/ Keyboard Shortcut section.

. Added Shortcut TAB under config section where user can re-assign the shortcuts as needed
. Shortcuts Help Modal in Sales Register now reflects the assigned shortcuts
. Updates JS section in register in order to retreive assigned keys from database.

To be fixed:
. Migration file
. Success message on submit

PR #4532, opened 2026-05-06
Assignable Keyboard Shortcuts Updates

WebShells:
Few updates regarding HotKeys/ Keyboard Shortcut section.

. Added Shortcut TAB under config section where user can re-assign the shortcuts as needed
. Shortcuts Help Modal in Sales Register now reflects the assigned shortcuts
. Updates JS section in register in order to retreive assigned keys from database.
. Migration file
. Success message on submit
. Added the needed language strings and routes.
. CI4 Support shortcut configuration for sales hotkeys.

Updated version of #3720 

## Summary by CodeRabbit

* **New Features**
  * Keyboard shortcuts configuration UI to customize and save bindings; client-side validation and duplicate-binding errors shown.
  * Sales register and help screens now use user-configured shortcuts (dynamic ALT-key behavior, supports completing sale after payment).

* **Localization**
  * Added UI text for shortcuts, configuration, duplicate-binding warnings, and save errors.

* **Database**
  * Migration to add default shortcut keys.


app/Config/Routes.php
```diff
@@ -11,6 +11,7 @@
 $routes->get('login', 'Login::index');
 $routes->post('login', 'Login::index');
 $routes->post('migrate', 'Login::migrate');
+$routes->post('config/saveShortcuts', 'Config::postSaveShortcuts');
```
objecttothis: Autorouting should pick this up.  I don't think this route should be needed.  Can you try it without this route in place?

app/Controllers/Config.php
```diff
@@ -231,6 +231,7 @@ public function getIndex(): string
         $data['line_sequence_options'] = $this->sale_lib->get_line_sequence_options();
         $data['register_mode_options'] = $this->sale_lib->get_register_mode_options();
         $data['invoice_type_options'] = $this->sale_lib->get_invoice_type_options();
+        $data['key_shortcuts_options'] = $this->sale_lib->get_key_shortcuts_options();
```
objecttothis: All new code must use PSR compliant naming, so this function needs to be refactored to getKeyShortcutsOptions();

app/Controllers/Config.php
```diff
@@ -946,6 +947,36 @@ public function postSaveReceipt(): ResponseInterface
...
 
+    /**
+     * Saves keyboard shortcut bindings.
+     *
+     * @return ResponseInterface
+     * @noinspection PhpUnused
+     */
+    public function postSaveShortcuts(): ResponseInterface
+    {
+        $allowed_shortcuts = array_keys($this->sale_lib->get_key_shortcuts_options());
```
objecttothis: All local variables need to be camelCase in this function.

app/Controllers/Sales.php
```diff
@@ -1253,6 +1253,7 @@ private function _reload(array $data = []): ResponseInterface|string    // TODO:
 
         $data['quote_number'] = $this->sale_lib->get_quote_number();
         $data['work_order_number'] = $this->sale_lib->get_work_order_number();
+        $data['keyboard_shortcuts'] = $this->sale_lib->get_key_shortcuts();
```
objecttothis: PSR function naming.

app/Controllers/Sales.php
```diff
@@ -1641,7 +1642,9 @@ public function postUnsuspend(): ResponseInterface|string
      */
     public function getSalesKeyboardHelp(): string
     {
-        return view('sales/help');
+        return view('sales/help', [
+            'keyboard_shortcuts' => $this->sale_lib->get_key_shortcuts()
```
objecttothis: keyboard_shortcuts needs to be keyboardShortcuts, because in the view it will get used as a PHP variable which need to be camelCase

app/Database/Migrations/20260506000000_add_shortcut_keys.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class Migration_add_shortcut_keys extends Migration
```
objecttothis: PascalCase needs to be used for class names.  The file also needs to be renamed to `20260506000000_AddShortcutKeys.php`

app/Database/Migrations/20260506000000_add_shortcut_keys.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Perform a migration step.
+     */
+    public function up(): void
+    {
+        $shortcut_values = [
```
objecttothis: psr naming

app/Database/Migrations/20260506000000_add_shortcut_keys.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace App\Database\Migrations;
+
+use CodeIgniter\Database\Migration;
+
+class Migration_add_shortcut_keys extends Migration
+{
+    /**
```
objecttothis: These PHPDocs don't add anything.  We should remove them.

app/Language/en/Config.php
```diff
@@ -302,6 +302,8 @@
     "suggestions_layout"                        => "Search Suggestions Layout",
     "suggestions_second_column"                 => "Column 2",
     "suggestions_third_column"                  => "Column 3",
+    "shortcuts"                                 => "Shortcuts",
```
objecttothis, line 305: We can probably change these all at once in another PR but in general we don't want to use `"` because it invokes the PHP string parser when it isn't necessary.  Use `'` instead.

app/Libraries/Sale_lib.php
```diff
@@ -23,6 +23,19 @@
  */
 class Sale_lib
 {
+    private const KEY_SHORTCUT_DEFAULTS = [
```
objecttothis, line 26: Our current implementation is using these key shortcuts in the sales register, but if we ever plan to make use of shortcut keys in another module it would be better to develop this feature outside of Sale_lib and instead in it's own library.  I don't feel strongly.  Just a thought.

app/Libraries/Sale_lib.php
```diff
@@ -105,6 +118,44 @@ public function get_invoice_type_options(): array
...
     }
 
+    /**
+     * Returns the available keyboard shortcut choices for the configuration screen.
+     *
+     * @return array<string, string>
+     */
+    public function get_key_shortcuts_options(): array
+    {
+        $key_shortcuts = [];
```
objecttothis: PSR Compliant naming in new code.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/**
+ * @var array $config
+ * @var array $key_shortcuts_options
+ */
+
+$shortcut_defaults = [
```
objecttothis: This is not very maintainable because you already have a constant in the sale_lib so if something gets changed there but not here, or vis a versa, then it becomes incorrect.  Business logic should not be handled in the view.  Instead pass this into the view from the controller.  It should be passed as shortcutDefaults in the data array.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,99 @@
...
+    'suspend'   => '51 | ALT + 3',
+    'suspended' => '52 | ALT + 4',
+    'amount'    => '53 | ALT + 5',
+    'payment'   => '54 | ALT + 6',
+    'complete'  => '55 | ALT + 7',
+    'finish'    => '56 | ALT + 8',
+    'help'      => '57 | ALT + 9',
+];
+
+$shortcut_values = [];
```
objecttothis: Same here.  Normalization should be done in the controller and passed to the view so that the view only handles formatting and layout of the data.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,99 @@
...
+foreach ($shortcut_defaults as $name => $default_value) {
+    $shortcut_values[$name] = $config['key_' . $name] ?? $default_value;
+}
+
+$shortcut_labels = [
```
objecttothis: PSR naming of local variables.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,99 @@
...
+                    <ul id="shortcuts_error_message_box" class="error_message_box"></ul>
+
+                    <?php foreach ($shortcut_labels as $name => $label): ?>
+                        <div class="form-group form-group-sm">
+                            <?= form_label($label, 'key_' . $name, ['class' => 'control-label col-xs-3']) ?>
+                            <div class="col-xs-4">
+                                <?= form_dropdown(
+                                    'key_' . $name,
+                                    $key_shortcuts_options,
+                                    $shortcut_values[$name],
```
objecttothis: I believe we need to wrap this in esc() to guard against injection attacks.

app/Views/sales/help.php
```diff
@@ -1,3 +1,24 @@
+<?php
+/**
+ * @var array $keyboard_shortcuts
+ */
+
+$keyboard_shortcuts ??= [];
+
+$shortcut_labels = [
```
objecttothis, line 8: PSR naming

app/Views/sales/help.php
```diff
@@ -15,46 +36,13 @@
...
-                    </tr>
-                    <tr>
-                        <td><code>ALT + 8</code></td>
-                        <td><?= lang('Sales.key_finish_quote'); ?></td>
-                    </tr>
-                    <tr>
-                        <td><code>ALT + 9</code></td>
-                        <td><?= lang('Sales.key_help_modal'); ?></td>
-                    </tr>
+                    <?php foreach ($shortcut_labels as $name => $label): ?>
```
objecttothis, line 39: Much more concise!  I love it.

app/Views/sales/help.php
```diff
@@ -15,46 +36,13 @@
...
+                        <?php $shortcut = $keyboard_shortcuts[$name] ?? ['label' => '', 'code' => '']; ?>
+                        <tr>
```
objecttothis, line 41: I know we were doing this before but using tables for layout is deprecated because it breaks screen readers and it isn't responsive.  Using divs/css with bootstrap would be better here.  I think there are several other views that could be used as an example.

app/Views/sales/register.php
```diff
@@ -839,43 +841,45 @@ function check_payment_type() {
...
-            case 56: // Alt + 8 Finish Quote/Invoice without payment
-                $("#finish_invoice_quote_button").click();
-                break;
-            case 57: // Alt + 9 Open Shortcuts Help Modal
-                $("#show_keyboard_help").click();
-                break;
+    document.body.onkeyup = function(event) {
+        if (event.altKey) {
+            switch (event.keyCode) {
+                case keyboardShortcuts.items.code: // Alt + 1 Items Search
```
objecttothis: These comments can get out of date very easily.  How about omitting the Hotkey from them.  Even better would be to refactor the names to be more descriptive and omit the comment all together.  For example `keyboardShortcuts.itemSearch.code` would remove the need for a comment completely. Doing this may involve more refactoring than anyone wants to do, but it has a code clarity advantage. I'll leave it up to you.

objecttothis (changes requested):

app/Views/sales/help.php
WebShells: Fixed as needed.

app/Views/sales/help.php
WebShells: Fixed as needed.

app/Config/Routes.php
WebShells: Fixed as needed.

app/Controllers/Config.php
WebShells: Fixed as needed.

app/Controllers/Sales.php
WebShells: Fixed as needed.

app/Views/configs/shortcuts_config.php
WebShells: Fixed as needed.

app/Controllers/Sales.php
WebShells: Fixed as needed.

app/Language/en/Config.php
```diff
@@ -302,6 +302,10 @@
...
+    "shortcuts_configuration"                   => "Sales Shortcut Configuration",
```
objecttothis: Just a suggestion, but I would change this string to read `Sales Keyboard Shortcut Configuration` to better clarify what kind of shortcuts.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,88 @@
...
+            <fieldset id="config_info">
+                <div class="col-md-8">
+                    <div id="required_fields_message"><?= esc(lang('Common.fields_required_message')) ?></div>
+                    <ul id="shortcuts_error_message_box" class="error_message_box"></ul>
+
+                    <?php foreach ($shortcutLabels as $name => $label): ?>
+                        <div class="form-group form-group-sm">
+                            <?= form_label($label, 'key_' . $name, ['class' => 'control-label col-xs-3']) ?>
+                            <div class="col-xs-4">
+                                <?php $shortcutValue = $keyboardShortcuts[$name]['value'] ?? ''; ?>
```
objecttothis: Just a suggestion but I recommend renaming this to `$keyboardShortcutSelectedValue` for consistency with `$keyboardShortcutsOptions` and clarity as to what the variable is representing.  I would also change `$keyboardShortcutsOptions` to `$keyboardShortcutOptions`. Not mandatory, but helps with code clarity.

objecttothis (dismissed):
Take a look at the two suggestions I made.  Fix them if you want.  Otherwise we are ready for you to squash and merge.

app/Language/en/Config.php
WebShells: Changed as needed.

app/Views/configs/shortcuts_config.php
WebShells: Changed as needed for code readability.

app/Views/configs/shortcuts_config.php
```diff
@@ -0,0 +1,88 @@
...
+<script type="text/javascript">
+    $('#shortcuts_config_form').validate($.extend(form_support.handler, {
+        submitHandler: function(form) {
+            $(form).ajaxSubmit({
+                success: function(response) {
+                    $.notify({
+                        message: response.message
+                    }, {
+                        type: response.success ? 'success' : 'danger'
+                    });
```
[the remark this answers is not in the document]
WebShells: header_js.php already wraps $.notify globally and sanitizes the message before display.
DOMPurify.sanitize(response.message ?? '') is redundant with that wrapper.

objecttothis (approved):

PR #4535, opened 2026-05-07
Secondary Currency Enhancement 

WebShells:
## Secondary Currency and Customer Display Enhancements

This PR adds configurable secondary currency support across the sales flow, customer-facing display, receipts, documents, receiving, and reports. The feature is controlled by a global enable/disable option in Localization, with an additional receipt-level toggle for printed receipts.

### What was added

- A global `Enable Secondary Currency` setting under Localization.
- Configurable secondary currency fields:
  - currency symbol
  - currency code
  - exchange rate
  - decimal precision
- A receipt-level `Show Secondary Currency` option.
- Secondary currency formatting in:
  - sales register
  - customer display
  - receipts
  - invoices
  - quotes
  - work orders
  - tax invoices
  - receiving
  - reports
- Generic labels so the feature can be reused across currencies and regions.

#### Sales register
- Shows original and secondary currency totals where enabled.
- Shows amount due and change due in both currencies.

#### Customer display
- Shows primary and secondary totals for checkout visibility.
- Shows the conversion rate when secondary currency is enabled.

#### Short receipt
- Shows the original total and the converted secondary total below it.

#### Invoices, quotes, work orders, and tax invoices
- Show line item prices in dual-currency format where enabled.
- Show totals and exchange rate in the lower summary section.
- Apply the same behavior to email versions.

#### Receiving
- Shows a secondary currency total below the primary total when enabled.

#### Reports
- Include secondary currency totals and rate where applicable.
- Hide those values when secondary currency is disabled.

#### When `Enable Secondary Currency` is off
- Secondary currency fields are hidden in configuration.
- Secondary totals and rates are hidden in sales views.
- Receipts print in primary currency only.
- Documents do not show secondary currency columns or rows.
- Reports omit secondary currency columns and totals.
- The customer display falls back to primary currency only.

#### When enabled
- The related sections show secondary currency values automatically.
- The receipt setting controls whether receipts include the secondary section.
- The customer display shows the converted amount and rate when available.

### Implementation notes

The feature uses generic labels such as:
- `Total`
- `Secondary Currency Total`
- `Rate`
- `Amount Due`
- `Change Due`

This keeps the implementation reusable and avoids hardcoding currency-specific wording into the UI.

## Summary by CodeRabbit

* **New Features**
  * Optional secondary currency display across receipts, invoices, quotes, register, reports and charts; configurable symbol, code, rate, decimals, auto-refresh, and manual refresh action (UI + CLI).

* **Database**
  * Transactions now store a secondary currency exchange rate.

* **Reports**
  * Summaries and charts can include secondary totals, rates and per-row secondary data.

* **Tests & Translations**
  * New tests added and many localization strings for secondary-currency UI.


WebShells:
https://github.com/user-attachments/assets/a0648087-e4b2-468e-bffa-466443d6ae97

@objecttothis @jekkos That's a video for the Secondary Currency feature.

WebShells:
Updates: 

Added a refresh service
Extended locale config saving
Added the UI fields for auto refresh, interval, feed URL, last sync, and last error 

Capable now to: 

Turn auto refresh on/off
Set the refresh interval in minutes
Set or leave the default feed URL
Run the refresh manually with update button on register ( position to be discussed )

The config should be:

currency_code = base currency, ex: USD
secondary_currency_code = ex: LBP, EUR, JPY, CAD, GBP etc...
secondary_currency_feed_url = https://open.er-api.com/v6/latest/{base}

That means the app will fetch the secondary currency code rate in conversion to USD:

https://open.er-api.com/v6/latest/USD

@opensourcepos/org @jekkos @objecttothis

BudsieBuds:
@WebShells Looks good! I was wondering, and I might be way off here, since this feature seems optional for most deployments, I'm curious whether it might be better as a plugin once the plugin system is implemented (#4407). That could keep the core simpler while still letting users who need dual‑currency support enable it.

WebShells:
@BudsieBuds 

> @WebShells Looks good! I was wondering, and I might be way off here, since this feature seems optional for most deployments, I'm curious whether it might be better as a plugin once the plugin system is implemented (#4407). That could keep the core simpler while still letting users who need dual‑currency support enable it.

@BudsieBuds Thanks for your time going over this!

A good reason not to move this to a plugin is that it is not an isolated feature. It goes through the same core paths that already own money and historical snapshots.

Why code feature better fits here:

It changes persisted transaction data, so sales, receivings, and expenses need the same secondary rate snapshot at save time.
It affects core paths including register, receipts, invoices, quotes, work orders, tax invoices, customer display, and reports.
It depends on shared app config, language strings, helpers, and migrations. A plugin would still need deep integration points in all of those places.
The feature is optional already, so keeping it in core does not force it on deployments. When disabled, it is essentially dormant.

@opensourcepos/org any thoughts on this ?

objecttothis:
> @BudsieBuds
> 
> > @WebShells Looks good! I was wondering, and I might be way off here, since this feature seems optional for most deployments, I'm curious whether it might be better as a plugin once the plugin system is implemented (#4407). That could keep the core simpler while still letting users who need dual‑currency support enable it.
> 
> @BudsieBuds Thanks for your time going over this!
> 
> A good reason not to move this to a plugin is that it is not an isolated feature. It goes through the same core paths that already own money and historical snapshots.
> 
> Why code feature better fits here:
> 
> It changes persisted transaction data, so sales, receivings, and expenses need the same secondary rate snapshot at save time. It affects core paths including register, receipts, invoices, quotes, work orders, tax invoices, customer display, and reports. It depends on shared app config, language strings, helpers, and migrations. A plugin would still need deep integration points in all of those places. The feature is optional already, so keeping it in core does not force it on deployments. When disabled, it is essentially dormant.
> 
> @opensourcepos/org any thoughts on this ?

#4407 needs code review so @WebShells and @BudsieBuds if you feel like doing one, it could both help you get familiar with the feature set and spot things that need fixing.

This PR could be shoehorned into a plugin, but it doesn't fit the natural profile of a plugin. It is true that it's an optional feature, but I don't think that should make the only criteria for whether it should be a plugin.  To make it a plugin we would have to inject several partial views, which #4407 already has the capability of doing, but the part that makes it more complicated is the roadmap for the feature.  My understanding is that down the road @WebShells was thinking to implement payments in secondary currency which further ties it into the core functionality of the code.

app/Commands/RefreshSecondaryCurrency.php
```diff
@@ -0,0 +1,70 @@
...
+use CodeIgniter\CLI\BaseCommand;
+use CodeIgniter\CLI\CLI;
+use Config\OSPOS;
+use Throwable;
+
+class RefreshSecondaryCurrency extends BaseCommand
+{
+    protected $group = 'OSPOS';
+    protected $name = 'currency:refresh-secondary';
+    protected $description = 'Refreshes the configured secondary currency rate from the live feed.';
```
objecttothis, line 16: I assume we are hardcoding the English here because CLI commands don't use locales.  Is that correct?

app/Config/Routes.php
```diff
@@ -10,33 +10,150 @@
...
+$routes->add('sales/discard_suspended_sale', 'Sales::getDiscardSuspendedSale');
+$routes->add('sales/discardSuspendedSale', 'Sales::getDiscardSuspendedSale');
+$routes->add('sales/suspend', 'Sales::postSuspend');
+$routes->add('sales/suspended', 'Sales::getSuspended');
+$routes->add('sales/unsuspend', 'Sales::postUnsuspend');
+$routes->add('sales/salesKeyboardHelp', 'Sales::getSalesKeyboardHelp');
+$routes->add('sales/checkInvoiceNumber', 'Sales::postCheckInvoiceNumber');
+$routes->add('sales/changeItemNumber', 'Sales::postChangeItemNumber');
+$routes->add('sales/changeItemName', 'Sales::postChangeItemName');
+$routes->add('sales/changeItemDescription', 'Sales::postChangeItemDescription');
```
objecttothis, line 120: We shouldn't need to add all these sales routes.  Comment them out and run `php spark routes`.  You should see the autorouter pick it up.

app/Config/Routes.php
```diff
@@ -10,33 +10,150 @@
...
 $routes->add('no_access/index/(:segment)', 'No_access::index/$1');
 $routes->add('no_access/index/(:segment)/(:segment)', 'No_access::index/$1/$2');
 
+$routes->add('reports/summary_expenses_suppliers/(:any)/(:any)/(:any)/(:any)/(:any)', 'Reports::summary_expenses_suppliers/$1/$2/$3/$4/$5');
```
objecttothis, line 124: The reports routes are necessary, I believe, because they use custom patterns.

objecttothis, line 124: This is a very large changeset.  If you are adding new reports that don't have directly to do with Secondary Currency, please break them out into a different PR.

app/Controllers/Config.php
```diff
@@ -476,9 +476,61 @@ public function postSaveLocale(): ResponseInterface
     {
         $exploded = explode(":", $this->request->getPost('language'));
         $currency_symbol = $this->request->getPost('currency_symbol');
+        $secondary_currency_enabled = $this->request->getPost('secondary_currency_enabled') != null;
```
objecttothis, line 479: All new local variables, function names and class names must use PSR compliant naming.  Please go through all the new code in this PSR and refactor them.

app/Config/Routes.php
objecttothis: If they are just changes to existing reports to display secondary currency, then that's fine.

app/Controllers/Reports.php
```diff
@@ -207,7 +219,9 @@ public function summary_categories(string $start_date, string $end_date, string
             'subtitle'     => $this->_get_subtitle_report(['start_date' => $start_date, 'end_date' => $end_date]),
             'headers'      => $this->summary_categories->getDataColumns(),
             'data'         => $tabular_data,
-            'summary_data' => $summary
+            'summary_data' => $summary,
+            'summary_secondary_data' => $this->_build_secondary_summary_display_map($report_data, $summary, $secondaryCurrency),
```
objecttothis, line 223: Hungarian notation in function names is obsolete. It was used back before PHP had private function classifiers to indicate that the function was meant to be private. 

_append_secondary_currency / _secondary_currency_snapshot / _secondary_currency_for_row / _build_secondary_summary_display_map / _average_secondary_rate Leading underscore on private methods. PSR-12 explicitly says "method names MUST NOT be prefixed with a single underscore to indicate protected or private visibility." These are all violations.

app/Controllers/Reports.php
```diff
@@ -306,39 +331,50 @@ public function summary_customers(string $start_date, string $end_date, string $
...
      */
-    public function summary_suppliers(string $start_date, string $end_date, string $sale_type, string $location_id = 'all'): string
-    {   // TODO: Duplicated Code
-        $this->clearCache();
-
-        $inputs = [
-            'start_date'  => $start_date,
-            'end_date'    => $end_date,
-            'sale_type'   => $sale_type,
+      public function summary_suppliers(string $start_date, string $end_date, string $sale_type, string $location_id = 'all'): string
```
objecttothis, line 334: This is pre-existing code smell, not introduced by this PR. @coderabbitai please create an issue to resolve this in another PR.

summary_suppliers has 4 params — same as summary_sales, summary_categories, summary_customers, summary_items, summary_employees, and ~15 other methods. Even the TODO comment ("Too many parameters") was already there before this PR.

That said — the right fix is a parameter object:

// Instead of:
public function summary_suppliers(
    string $start_date,
    string $end_date,
    string $sale_type,
    string $location_id = 'all'
): string

// Use a DTO/value object:
class ReportFilter
{
    public function __construct(
        public readonly string $startDate,
        public readonly string $endDate,
        public readonly string $saleType = 'all',
        public readonly string $locationId = 'all',
        public readonly int    $discountType = 0,
        public readonly string $entityId = '',     // customer/employee/supplier/discount ID
        public readonly string $paymentType = 'all',
    ) {}
}
public function summary_suppliers(ReportFilter $filter): string

Why this is better:

Adding new filter param (like discount_type) doesn't break every call site
IDE completion on the object vs positional string args
One place to validate/sanitize inputs
Routes already unpack URL segments into $inputs array internally anyway — the object just formalizes that

Create the class in app/Entities/ReportFilter.php — namespace App\Entities since Entities is where CI intends on Data holding objects to be stored.

CI4 convention: Entities directory is specifically for data-holding objects
Directory doesn't exist yet but CI4 autoloads it automatically
Most semantically correct by CI4 standards
Recommendation: app/Entities/ReportFilter.php

CI4 has a first-class Entity concept for data objects. No existing pattern to follow here so better to align with the framework than invent App\Models\Filters. Future devs will know where to look.

objecttothis (changes requested):
@WebShells it's late, so this is as far as I got.  It gives you something to do. Request another PR once you finish with this part.

PR #4537, opened 2026-05-08
chore: sync project files to match upstream templates

BudsieBuds:
Updates the project base files to match the current CodeIgniter 4 skeleton. It also modernizes `package.json` and `composer.json`.

- Updated framework provided files to the unmodified versions from CI `4.7.2`
- Rebuilt `package.json` from a clean init and modernized metadata and formatting
- Rebuilt `composer.json` with modernized metadata and formatting
- Replaced `CODE_OF_CONDUCT.md` text with markdown
- Updated `Dockerfile` to replace deprecated instruction

No functional changes are introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CSS styling issues in the error debug display.

* **Documentation**
  * Replaced Code of Conduct with a formatted Contributor Covenant 2.1 document.

* **Chores**
  * Raised minimum PHP requirement to 8.2.
  * Updated project and package metadata, author/contributor info, and human-readable script descriptions.
  * Adjusted CLI/web bootstrap startup and container metadata.

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -486,10 +486,9 @@ class Mimes
     /**
      * Attempts to determine the best mime type for the given file extension.
      *
-     * @param string $extension
-     * @return array|string|null The mime type found, or none if unable to determine.
+     * @return string|null The mime type found, or none if unable to determine.
      */
-    public static function guessTypeFromExtension(string $extension): array|string|null
+    public static function guessTypeFromExtension(string $extension)
     {
         $extension = trim(strtolower($extension), '. ');
 
@@ -507,7 +506,7 @@ class Mimes
      *
      * @return string|null The extension determined, or null if unable to match.
      */
-    public static function guessExtensionFromType(string $type, ?string $proposedExtension = null): ?string
+    public static function guessExtensionFromType(string $type, ?string $proposedExtension = null)
     {
         $type = trim(strtolower($type), '. ');
 
@@ -523,7 +522,7 @@ class Mimes
         }
 
         // Reverse check the mime type list if no extension was proposed.
-        // This search is order-sensitive!
+        // This search is order sensitive!
         foreach (static::$mimes as $ext => $types) {
             if (in_array($type, (array) $types, true)) {
                 return $ext;
--- a/package.json
+++ b/package.json
@@ -1,34 +1,33 @@
 {
     "name": "@opensourcepos/opensourcepos",
     "version": "3.4.2",
-    "description": "pen Source Point of Sale is a web based point of sale system written in the PHP language. It uses MySQL as the data storage back-end and has a simple user interface.",
-    "main": "index.php",
-    "license": "MIT",
-    "authors": [
-        "jekkos <jekkos - at - opensourcepos.org>",
-        "FrancescoUK <francesco.lodolo.uk - at - gmail.com>",
-        "objecttothis <objecttothis - at - gmail.com>",
-        "SteveIreland <stevei - at - ruledomain.com>"
-    ],
-    "files": [
-        "dist/opensourcepos.$version.tgz"
-    ],
-    "publishConfig": {
-        "registry": "https://npm.pkg.github.com/"
-    },
+    "description": "Open Source Point of Sale is a web based point of sale system written in the PHP language. It uses MySQL as the data storage back-end and has a simple user interface.",
     "keywords": [
         "point-of-sale",
         "POS"
     ],
+    "homepage": "https://opensourcepos.org",
+    "bugs": {
+        "url": "https://github.com/opensourcepos/opensourcepos/issues"
+    },
     "repository": {
         "type": "git",
-        "url": "https://github.com/opensourcepos/opensourcepos"
+        "url": "git+https://github.com/opensourcepos/opensourcepos.git"
     },
+    "license": "MIT",
+    "contributors": [
+        "jekkos <jekkos - at - opensourcepos.org>",
+        "objecttothis <objecttothis - at - gmail.com>"
+    ],
+    "files": [
+        "dist/opensourcepos.$version.tgz"
+    ],
+    "type": "module",
+    "main": "index.php",
     "scripts": {
         "build": "gulp default",
         "gulp": "gulp"
     },
-    "type": "module",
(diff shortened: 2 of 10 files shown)


BudsieBuds:
@objecttothis Since you usually handle the CI updates, could you review the changes to the CI4 related files? I'm pretty confident about the updates to `debug.css`, `composer.json`, `spark`, and `index.php`. I'd appreciate an extra pair of eyes on `Mimes.php` and `BaseController.php`. Those two files were previously modified in commit 1328b4d9b81db489694762f0372991c6d08b72f0, so I want to be sure that replacing them with the upstream versions is correct.

For clarity: `spark`, `Mimes.php`, `BaseController.php`, `debug.css`, and `index.php` are unmodified copies from the CodeIgniter `4.7.2` source.

Thanks!

app/Config/Mimes.php
```diff
@@ -486,10 +486,9 @@ class Mimes
     /**
      * Attempts to determine the best mime type for the given file extension.
      *
-     * @param string $extension
-     * @return array|string|null The mime type found, or none if unable to determine.
+     * @return string|null The mime type found, or none if unable to determine.
      */
-    public static function guessTypeFromExtension(string $extension): array|string|null
+    public static function guessTypeFromExtension(string $extension)
```
objecttothis, line 491: With these CI related config files as a general rule we don't change them unless it's a specific customization needed for the webapp.  For example adding routes or events handling.  Do a diff with /vendor/CodeIgniter4/framework/app/Config/Mimes.php and if the return type is declared there then we leave it in.  There are two reasons.  One is that it makes upgrading to the next CI version easier and the second is that it is less likely to introduce unintended bugs.  So if these functions have the return types in the CI version, then we need to leave them in.  Some of the PHPdocs in CI are ones that don't really add anything so we would remove them but then that makes the diff very messy when we upgrade. As a general rule lets avoid making changes to these files if we don't need to.  If its a genuine needed change, I suggest submitting a PR in the CodeIgniter repository.

package.json
```diff
@@ -1,34 +1,35 @@
...
         "type": "git",
-        "url": "https://github.com/opensourcepos/opensourcepos"
+        "url": "git+https://github.com/opensourcepos/opensourcepos.git"
     },
+    "license": "MIT",
+    "contributors": [
+        "jekkos <jekkos@opensourcepos.org>",
+        "FrancescoUK <francesco.lodolo.uk@gmail.com>",
+        "objecttothis <objecttothis@gmail.com>",
+        "SteveIreland <stevei@gmail.com>"
```
objecttothis: @jekkos this list of contributors seems outdated.  Should we update it with regular contributors in the last X years?

spark
```diff
@@ -35,7 +38,7 @@ if (str_starts_with(PHP_SAPI, 'cgi')) {
  *---------------------------------------------------------------
  */
 
-$minPhpVersion = '8.1'; // If you update this, don't forget to update `public/index.php`.
+$minPhpVersion = '8.2'; // If you update this, don't forget to update `public/index.php`.
```
objecttothis, line 41: So this is telling me that I missed overwriting spark.  @BudsieBuds can you take the spark in the vendor folder and overwrite the one in root with that one in case there are other changes?  Same with /public/index.php?

package.json
objecttothis: Please also leave the emails obfuscated like they were before to reduce crawlers scooping them up

objecttothis (changes requested):

app/Config/Mimes.php
BudsieBuds: Maybe there was some confusion, but these files are actually the unmodified versions from CodeIgniter `4.7.2`, as I mentioned in [my earlier comment](https://github.com/opensourcepos/opensourcepos/pull/4537#issuecomment-4407212065).

My question was whether the changes made to `Mimes.php` and `BaseController.php` in commit 1328b4d9b81db489694762f0372991c6d08b72f0 were intentional. I wasn’t sure if those edits were added for a specific reason.

spark
BudsieBuds: As stated earlier, the `spark` file in this PR is the original, untouched CodeIgniter file, identical to the one in the vendor directory. The same applies to `index.php`.

package.json
BudsieBuds: I took the liberty to update `contributors` in `package.json` and the `authors` in `composer.json`. I also re‑obfuscated the email addresses as requested. I hadn't considered it strictly necessary since modern crawlers can probably parse both formats.

app/Config/Mimes.php
objecttothis: With Mimes.php and BaseController.php those changes were likely missed between an older version of CI and 4.7.2, so please overwrite them with what is in the vendor folder.

BudsieBuds: One last check to make sure we're fully aligned. The line in `BaseController.php` hasn't changed in CI4 since [this commit](https://github.com/codeigniter4/CodeIgniter4/commit/b8ad7782ab91be361a3ec3501b1ecf0c20b99023#diff-2afdfe619fd918a59555251b2bd05871d71fc60c0df2ebe4213aeaa999682057), and hasn't been changed since. The same applies to `Mimes.php`. So it doesn't look like these differences came from an upgrade gap, they appear to have been introduced in the commit I referenced earlier.

Do you recall whether those edits were intentional, or were they added by mistake?

objecttothis: I ran a diff on the current code and the vendor versions.  If they were intentional, they were to fix the IDE complaining that the function was missing a return type or something like that.  I think we can safely revert to the CI 4.7.2 versions which will simplify upgrades.

BudsieBuds: Great, thanks for checking! Once you approve the review, I'll merge the PR.

objecttothis (approved):

jekkos (approved):

objecttothis:
@jekkos I approved it also.  PHPLint seems to be holding things up.

BudsieBuds:
I'm not seeing an option to re-run phplint. I could probably force it with an empty/force commit, but @jekkos might be able to trigger it manually.

PR #4546, opened 2026-05-13
fix(docker): correct permissions and fix migration barcode_type error

jekkos:
## Summary
- Fix Docker permissions for writable directories and public uploads
- Add `barcode_type` default to prevent "unknown key barcode_type" error during initial migration

### Permissions Fix
The Dockerfile now sets proper permissions (750) for:
- `/app/writable/logs`
- `/app/writable/uploads`
- `/app/writable/cache`
- `/app/public/uploads`
- `/app/public/uploads/item_pics`

And permissions (640) for:
- `/app/writable/uploads/importCustomers.csv`

### Migration Fix
During initial setup on fresh Docker containers, the database migration for converting barcode types fails with "unknown key barcode_type" because the `OSPOS` config class doesn't have a default value for `barcode_type` when the database is not yet initialized.

Added `'barcode_type' => 'Code39'` to the default settings fallback in `app/Config/OSPOS.php`.

## Testing
These fixes ensure:
1. Fresh Docker installations can complete initial migration without errors
2. Upload directories have correct permissions for the www-data user

## Summary by CodeRabbit

* **Chores**
  * Enhanced Docker build artifact handling to include hidden files
  * Improved file permission security in containerized deployments
  * Added default barcode type configuration fallback

[Review Change Stack](https://app.coderabbit.ai/change-stack/opensourcepos/opensourcepos/pull/4546)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN echo "date.timezone = \"\${PHP_TIMEZONE}\"" > /usr/local/etc/php/conf.d/time
 
 WORKDIR /app
 COPY --chown=www-data:www-data . /app
-RUN chmod 770 /app/writable/uploads /app/writable/logs /app/writable/cache \
+RUN chmod 750 /app/writable/logs /app/writable/uploads /app/writable/cache /app/public/uploads /app/public/uploads/item_pics \
+    && chmod 640 /app/writable/uploads/importCustomers.csv \
     && ln -s /app/*[^public] /var/www \
     && rm -rf /var/www/html \
     && ln -nsf /app/public /var/www/html
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -123,6 +123,7 @@ jobs:
             .
             !.git
             !node_modules
+          include-hidden-files: true
           retention-days: 1
 
   docker:
--- a/app/Config/OSPOS.php
+++ b/app/Config/OSPOS.php
@@ -48,7 +48,8 @@ class OSPOS extends BaseConfig
                 $this->settings = [
                     'language' => 'english',
                     'language_code' => 'en',
-                    'company' => 'Home'
+                    'company' => 'Home',
+                    'barcode_type' => 'Code39'
                 ];
             }
         }


objecttothis (approved):

Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN echo "date.timezone = \"\${PHP_TIMEZONE}\"" > /usr/local/etc/php/conf.d/time
 
 WORKDIR /app
 COPY --chown=www-data:www-data . /app
-RUN chmod 770 /app/writable/uploads /app/writable/logs /app/writable/cache \
+RUN chmod 750 /app/writable/logs /app/writable/uploads /app/writable/cache /app/public/uploads /app/public/uploads/item_pics \
```
objecttothis, line 16: @jekkos I think there may be a problem here. /writeable/ and /public/ is not under /app/.  Also I think /public/uploads/ and public/uploads/item_pics/ need to have 755 permissions, not 750.

objecttothis (changes requested):

Dockerfile
objecttothis: For example, when a plugin sends an image URL to an API endpoint, that URL needs to be accessible. If we give 750 to /public/uploads/item_pics/ then https://domain.ext/uploads/item_pics/34523423.jpg will get blocked, no?

jekkos: Change was done here to fix warnings in the system tab. If apache is the owner then 750 should be fine, it can read and write no?

jekkos: The process user is the same as the dir owner so even 700 could work

objecttothis: You can try it but I don't think images inside /item_pics/ folder will be accessible even if they have 644 permissions.  Here is what Copilot says about it. More restrictive is better as long as it works, so give it a shot, but I have my doubts.  Either way /writable/ and /public/ are not subdirectories of /app/ so that needs to be fixed.

This line is changing directory permissions from 770 to 750 for several directories including /app/public/uploads/item_pics.

Permission breakdown for 750:

    Owner (www-data): read, write, execute (7)
    Group: read, execute (5)
    Others: no permissions (0)

To answer your question: No, images inside /app/public/uploads/item_pics/ with 644 permissions will not be world-accessible with this setup, because:

    The directory itself has 750 permissions, which means "others" have no access (the 0 means no read, write, or execute)
    Even though a file inside might have 644 permissions (readable by all), you need execute permission on the parent directory to access files within it
    Since the directory denies access to "others," they cannot traverse into it to read the files, regardless of the files' individual permissions

If these images need to be publicly accessible via a web server, the directory would typically need 755 permissions instead (allowing execute for "others" so they can traverse the directory).

jekkos: But directory traversal is not required as you access the file directly through the url

jekkos: I dont see why the o+rx bit should be set. Its only apache that should be able to access the dir and its the owner of it, so I would not expect any other permissions to be required.. Only if you want to browse the files in a shell with a different user then you need the world r+x

jekkos: Btw the /app is the docker root where we copy subdirectories of the repo, if the directories would not exist then the docker ci build would have failed

objecttothis: > Btw the /app is the docker root where we copy subdirectories of the repo, if the directories would not exist then the docker ci build would have failed

Yeah, that's my unfamiliarity with docker.  That makes sense.

objecttothis: > But directory traversal is not required as you access the file directly through the url

Yes, you are correct that directory traversal isn't required as long as the files in the directory are directly accessed. I was under the impression that I had trouble accessing files with the URL when the parent directory didn't have world read permissions, but it's been awhile since I have troubleshot those issues.  If the more restrictive 750 works, then that's a better choice.

objecttothis (approved):

objecttothis:
@jekkos I approved the PR

jekkos:
Great thanks!

Issue #4540, opened 2026-05-11
[Feature]: Add support for additional payment types in Sales (Bank Transfer (QR), Wallet)


bhojkamal:
### 🏷️ Feature Type

✨ New Feature

### 📦 OpenSourcePOS Version

development (unreleased)

### 🎯 Problem Statement

Currently, the POS system supports only Cash, Card, Cheque, Due, and Gift Card payments in sales, cashups. Modern businesses increasingly rely on digital payment methods such as Bank Transfer (via QR) , Wallets (e.g., PayPal, eSewa) payments that directly transfer funds to bank accounts. These options are now essential for day-to-day transactions and customer convenience.

Adding these payment types will:

Improve flexibility for merchants

Support popular digital payment trends

Ensure the POS system remains relevant and practical for modern use cases

### 💡 Proposed Solution

Extend the payment type options in Sales Cashups by adding two new methods:

Bank Transfer

Wallet (e.g., PayPal, eSewa)

### 🔄 Alternatives Considered

No, alternative, but Bank Transfer cover wide area like (qr payment) directly goes to bank account, which is like bank transfer. QR payment is very famous nowadays. 

### 📎 Additional Context

I can add this feature.

### ✓ Confirmation

- [x] I have searched existing feature requests to ensure this is not a duplicate
- [x] I have provided a clear problem statement and proposed solution

enricodelarosa:
Would it be modest to do this in a way so that the admin is able to add several payment types? Like say the business owner maintains 2-3 bank accounts, so sales (cash coming in to those accounts) would need to be monitored? I know it sounds like a bit of an overreach but potentially cash inflow and outflows would need to be monitored for different bank accounts.

I'm a bit new to ospos so I'm not sure if location feature solves this problem. Just chiming in since this is already a use case I see myself needing as well in the future.

jekkos:
See this issue #4456

bhojkamal:
> See this issue [#4456](https://github.com/opensourcepos/opensourcepos/issues/4456)

It is quite close but not same, I just want to additional option (which is very common) on select i.e. Bank Transfer (via qr or any online media that goes bank account), and wallet - like paypal (international), esewa (national in Nepal). 
I have added in old version. 3.9 in local.

jekkos:
So just the placeholder options then?

bhojkamal:
yes. I can add this feature.

PR #4547, opened 2026-05-14
bank transfer and wallet - payment option added.#4540

bhojkamal:
image

[Feature]: Add support for additional payment types in Sales (Bank Transfer (QR), Wallet) #4540

## Summary by CodeRabbit

* **New Features**
  * Added Bank Transfer and Wallet as selectable payment methods.
  * Added filters to view and filter sales by Bank Transfer and Wallet in reports and listings.
* **Localization**
  * Added translations for Bank Transfer and Wallet in English (UK/US), Spanish (ES/MX), and French.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opensourcepos/opensourcepos/pull/4547)


objecttothis:
@bhojkamal As I understand it, this change is just adding the dropdown options for these payment types and not actually implementing the payment types themselves.  There is a flaw in our implementation of the payment types in that they are being stored in the table as the string representation of whatever language the logged in employee was using at time of sale.  We ought to have instead created a static table containing an id and name of the payment type then linked them using a foreign key and the payment method id is what gets stored in the table so you don't have inconsistent data.

Good: We leave your implementation as is and @coderabbit creates an issue describing the need I've just explained.
Better: We implement your change but also refactor the code to correctly model the data.

I'll leave it up to you what you want to do but the issue I brought up needs to be addressed at some point.

objecttothis:
Also, I think the change as you have it is a core code change but likely implementing those payment methods (stripe, QR code, etc) we might want to do as a plugin rather than core code, since those plugins may require PCI Compliance work that I don't really want to deal with but plugin authors can manage if they want to develop them. @jekkos, what do you think?

app/Models/Sale.php
```diff
@@ -277,6 +277,14 @@ public function get_payments_summary(?string $search, array $filters): array
             $builder->like('payment_type', lang('Sales.debit'));
         }
 
+        if ($filters['only_bank_transfer']) {
+            $builder->like('payment_type', lang('Sales.bank_transfer'));
+        }
+
+        if ($filters['only_wallet']) {
+            $builder->like('payment_type', lang('Sales.wallet'));
+        }
```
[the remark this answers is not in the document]
bhojkamal: Thanks for pointing, yes, exactly, there is missing the initialization. done in line: 161 and 162 now. 
'only_bank_transfer'=> false,
'only_wallet'       => false,

bhojkamal:
@objecttothis 
I’ve added bank_transfer and wallet to the static payment types list. I have used these keys in my CI3 application for the past 4-5 years without issue, so I am implementing them here in CI4 to match the existing static structure. The database keys remain consistent, while the display labels handled by the localization.

The missing initialization is fixed too. 
In Controllers/Sales.php in line 90-100 - $data['filters'] is now:
'only_cash'         => lang('Sales.cash_filter'),
'only_due'          => lang('Sales.due_filter'),
'only_check'        => lang('Sales.check_filter'),
'only_creditcard'   => lang('Sales.credit_filter'),
'only_debit'         => lang('Sales.debit'),
'only_bank_transfer' => lang('Sales.bank_transfer'),
'only_wallet'        => lang('Sales.wallet'),

The snapshot is below. 

image
image

I think, it can be merge now.

objecttothis:
> @objecttothis I’ve added bank_transfer and wallet to the static payment types list. I have used these keys in my CI3 application for the past 4-5 years without issue, so I am implementing them here in CI4 to match the existing static structure. The database keys remain consistent, while the display labels handled by the localization.
> 
> The missing initialization is fixed too. In Controllers/Sales.php in line 90-100 - $data['filters'] is now: 'only_cash' => lang('Sales.cash_filter'), 'only_due' => lang('Sales.due_filter'), 'only_check' => lang('Sales.check_filter'), 'only_creditcard' => lang('Sales.credit_filter'), 'only_debit' => lang('Sales.debit'), 'only_bank_transfer' => lang('Sales.bank_transfer'), 'only_wallet' => lang('Sales.wallet'),
> 
> I think, it can be merge now.

For the purpose of this PR, this is good enough, but it doesn't resolve the issue I raised in #4548.  For example, if you log in as another employee with a different language enabled (that has translations for those phrases) and complete a sale, the transaction history report will show different languages. It was probably done this way before because originally we didn't have per-employee language selection and the flaw only becomes apparent in situations where more than one language is being used.  In our use case there are three languages used by our employees, so it becomes obvious quickly.

All that said, we can resolve #4548 in another PR.  I'll do a code review right now.

app/Controllers/Reports.php
```diff
@@ -1251,8 +1251,10 @@ public function get_payment_type(): array
             'due'      => lang('Sales.due'),
             'check'    => lang('Sales.check'),
             'credit'   => lang('Sales.credit'),
-            'debit'    => lang('Sales.debit'),
-            'invoices' => lang('Sales.invoice')
+            'debit'         => lang('Sales.debit'),
```
objecttothis, line 1254: Let's match the indentation of `=>` in the lines above with these additions for consistency.

app/Controllers/Sales.php
```diff
@@ -92,8 +92,10 @@ public function getManage(): ResponseInterface|string
                 'only_due'          => lang('Sales.due_filter'),
                 'only_check'        => lang('Sales.check_filter'),
                 'only_creditcard'   => lang('Sales.credit_filter'),
-                'only_debit'        => lang('Sales.debit'),
-                'only_invoices'     => lang('Sales.invoice_filter'),
+                'only_debit'         => lang('Sales.debit'),
```
objecttothis: Same here with matching the indentation for consistency.

objecttothis (changes requested):
Just these two changes.  Otherwise looks good.

app/Controllers/Sales.php
bhojkamal: @objecttothis 
Indentation alignment is fixed.

app/Controllers/Reports.php
bhojkamal: fixed now.

objecttothis (approved):
Approved.  Let me know if you can't squash and merge.

PR #4551, opened 2026-05-19
Bugfixes to get Migration working on MySQL and MariaDB

objecttothis:
This PR fixes issues in two different categories:
1 Bugs affecting both MySQL and MariaDB migrations.
- A bug causing ospos_appconfig to be queried when it doesn't  exist. Added a bypass.
- username validation in Login view was causing an error when no tables existed because no username was listed. Added a bypass.

2 Bugs not affecting MySQL but affecting MariaDB.  MariaDB is inherently pickier about some things and checks were not accounting for these in the migrations. Checks and dropping of foreign keys were added before making changes.  These migrations are normally frozen, but this is a breaking change, so we have no choice but to fix them in situ.

It's possible that older versions of MariaDB were not this picky, so the breaking changes are coming with 11.8.3, but that's unclear to me.

MariaDB 11.8 changed the default implicit character set to uca1400, so when some of our migrations specify utf8 without the collation MariaDB implies uca1400 and that breaks things.

## Summary by CodeRabbit

* **New Features**
  * Expense tracking with categories and full management
  * India GST support with jurisdiction/rate/tax updates
  * Work-order number tracking on sales and enhanced payment tracking
  * Per-employee language selection and office/home menu grouping
  * New app settings for search suggestions, duplicate-barcode allowance, and default quote comment

* **Bug Fixes**
  * Improved migration robustness and foreign-key handling
  * Ensured migration check occurs before login when needed

* **Chores**
  * Improved configuration caching with safe defaults on failure

[Review Change Stack](https://app.coderabbit.ai/change-stack/opensourcepos/opensourcepos/pull/4551?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/Config/OSPOS.php
+++ b/app/Config/OSPOS.php
@@ -41,12 +41,7 @@ class OSPOS extends BaseConfig
             $db = Database::connect();
 
             if (!$db->tableExists('app_config')) {
-                $this->settings = [
-                    'language'      => 'english',
-                    'language_code' => 'en',
-                    'company'       => 'Home',
-                    'barcode_type'  => 'Code39'
-                ];
+                $this->settings = $this->getDefaultSettings();
                 return;
             }
 
@@ -56,13 +51,18 @@ class OSPOS extends BaseConfig
             }
             $this->cache->save('settings', encode_array($this->settings));
         } catch (\Exception $e) {
-            $this->settings = [
-                'language'      => 'english',
-                'language_code' => 'en',
-                'company'       => 'Home',
-                'barcode_type'  => 'Code39'
-            ];
+            $this->settings = $this->getDefaultSettings();
         }
+    }
+
+    private function getDefaultSettings(): array
+    {
+        return [
+            'language'      => 'english',
+            'language_code' => 'en',
+            'company'       => 'Home',
+            'barcode_type'  => 'Code39'
+        ];
     }
 
     /**
--- a/app/Config/OSPOS.php
+++ b/app/Config/OSPOS.php
@@ -14,7 +14,7 @@ use Config\Database;
  */
 class OSPOS extends BaseConfig
 {
-    public array $settings;
+    public array $settings = [];
     public string $commit_sha1 = 'dev';    // TODO: Travis scripts need to be updated to replace this with the commit hash on build
     private CacheInterface $cache;
 
--- a/app/Config/OSPOS.php
+++ b/app/Config/OSPOS.php
@@ -37,26 +37,25 @@ class OSPOS extends BaseConfig
             return;
         }
 
-        $db = Database::connect();
-
-        if (!$db->tableExists('app_config')) {
-            $this->settings = [
-                'language'      => 'english',
-                'language_code' => 'en',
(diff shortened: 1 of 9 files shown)


jekkos:
@objecttothis I have now swapped mariadb 10.5 with mysql 8 on dev server, it migrates wihtout issues. So not sure where that error was coming from.

objecttothis:
> @objecttothis I have now swapped mariadb 10.5 with mysql 8 on dev server, it migrates wihtout issues. So not sure where that error was coming from.

This PR contains all the reasons it wasn't migrating.  It migrates now on both MySQL 8.4.3 and MariaDB 11.8.3 Can you review it so we can merge?

objecttothis:
@jekkos I also just tested this on MariaDB 10.11.14 and it migrated as well without errors.

objecttothis:
@coderabbitai review

app/Config/OSPOS.php
```diff
@@ -33,25 +34,35 @@ public function set_settings(): void
...
+
+        if (!$db->tableExists('app_config')) {
+            $this->settings = [
+                'language'      => 'english',
+                'language_code' => 'en',
+                'company'       => 'Home',
+                'barcode_type'  => 'Code39'
+            ];
+            return;
+        }
```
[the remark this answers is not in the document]
objecttothis: @coderabbitai apply the proposed change

app/Config/OSPOS.php
```diff
@@ -33,25 +34,35 @@ public function set_settings(): void
...
-                $this->settings = [
-                    'language' => 'english',
-                    'language_code' => 'en',
-                    'company' => 'Home',
-                    'barcode_type' => 'Code39'
-                ];
+            return;
+        }
+
+        $db = Database::connect();
```
jekkos: Shouldn't this statement be inside of the try/cartch?

objecttothis: Yeah, I added it.

app/Controllers/Login.php
```diff
@@ -49,6 +49,13 @@ public function index(): string|RedirectResponse
                 return view('login', $data);
             }
 
+            if (!$data['is_latest'] || $data['is_new_install']) {
```
jekkos, line 52: agree it might be cleaner not to trigger the validation at all in this case. But that should not have stopped the migration from running

jekkos (dismissed):

app/Controllers/Login.php
objecttothis: It stopped it in the UI because the username/password fields don't show up because initial migration needs to happen, but you can't put in a username since there are no tables to check against.  on dev it's the other issues that caused the failure.

jekkos:
thanks for the fixes here @objecttothis .

objecttothis:
> ❌ **Staging deployment failed**
> 
> Check the [workflow logs](https://github.com/opensourcepos/opensourcepos/actions/runs/26094300154) for details.

@jekkos I think there is a missing key for the webhook deployment

PR #4555, opened 2026-05-21
fix: tax rate inputs blank with comma-decimal locales

### LINKED ISSUE #4553: Tax rate inputs blank with comma-decimal locales (az_AZ, fr_FR, etc.) after type=number added
## Description

Tax 1 Rate and Tax 2 Rate fields in the Tax configuration tab do not populate when the system locale uses a comma as the decimal separator (e.g., `az_AZ`, `fr_FR`, `de_DE`). The fields render blank even though the value is correctly stored in the database. With `en_US` (period decimal separator) the fields populate correctly.

## Steps to Reproduce

1. Set locale to any comma-decimal locale (e.g., Azerbaijani `az_AZ`) in Config → Locale
2. Set Tax 1 Rate to any value (e.g., `18`) and save
3. Navigate away and return to Config → Tax tab
4. **Expected:** Tax 1 Rate field shows `18`
5. **Actual:** Tax 1 Rate field is blank

## Root Cause

Commit `42ba39d29` added `type='number'` to the tax rate inputs in `app/Views/configs/tax_config.php` for input validation. However, `to_tax_decimals()` formats the value using PHP's `NumberFormatter` with the configured locale — producing `"18,00"` for comma-decimal locales. Browsers reject a comma-decimal value in `<input type="number">` and render the field blank.

```php
// tax_config.php — value rendered as "18,00" with az_AZ locale
'value' => to_tax_decimals($config['default_tax_1_rate'])
```

This was not an issue prior to `42ba39d29` (before v3.4.1) because the inputs were plain `type='text'`.

## Proposed Fix

`type='number'` inputs require a machine-readable, period-decimal value — not a locale-formatted display string. The fix is to pass the raw numeric value (cast to float, which PHP serialises with `.` regardless of locale) instead of the locale-formatted string for these two inputs.

```php
// tax_config.php — always produces "18" or "18.5", compatible with type="number"
'value' => $config['default_tax_1_rate'] !== '' ? (float) $config['default_tax_1_rate'] : ''
```

Apply the same pattern to `default_tax_2_rate` on line 86.

This preserves the `type='number'` browser validation while correctly supporting all locales. The existing `parse_tax()` on the save side already handles locale-aware parsing of the submitted value, so round-tripping is unaffected.

## Affected File

- `app/Views/configs/tax_config.php` — lines 61 and 86

## Environment

Reproduced on current `master`. Not present at release `5f395d` (v3.4.1) — that release predates the `type='number'` addition.

jekkos:
## Summary

Fixes #4553 - Tax rate input fields appear blank when locale uses comma as decimal separator (e.g., `az_AZ`, `fr_FR`, `de_DE`).

## Root Cause

Commit `42ba39d29` added `type='number'` to tax rate input fields. The `to_tax_decimals()` function formats values using locale-aware `NumberFormatter`, producing comma-decimal strings (e.g., `"18,00"`). Browsers reject comma-decimal values in `<input type="number">` and display blank fields.

## Fix

Replace locale-formatted values with raw numeric values for tax rate inputs. PHP serializes floats with period decimals regardless of locale, so `(float)` cast produces browser-compatible values.

## Regression Analysis

I analyzed all changes from commit `42ba39d29`:

| File | Change | Issue? |
|------|--------|--------|
| `tax_config.php` | Added `type='number'` | ✅ YES - uses `to_tax_decimals()` |
| `email_config.php` | Added `type='number'` | ❌ No - raw `$config` values |
| `barcode_config.php` | Added `type='number'` | ❌ No - raw `$config` values |
| Other files | Various changes | ❌ No - raw values or text inputs |

The regression is **isolated to `tax_config.php` only** - all other `type='number'` inputs use raw database values unaffected by locale.

## Testing

Tested with:
- `en_US` locale (period decimal) - works correctly
- `az_AZ` locale (comma decimal) - previously showed blank, now works correctly

The `parse_tax()` function on the save side already handles locale-aware parsing, so round-tripping works correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tax rate value handling in configuration to ensure accurate display and input processing.

### DIFF
--- a/app/Views/configs/tax_config.php
+++ b/app/Views/configs/tax_config.php
@@ -51,14 +51,10 @@
                 </div>
                 <div class="col-xs-1 input-group">
                     <?= form_input([
-                        'type'  => 'number',
-                        'step'  => 'any',
-                        'min'   => '0',
-                        'max'   => '100',
                         'name'  => 'default_tax_1_rate',
                         'id'    => 'default_tax_1_rate',
                         'class' => 'form-control input-sm',
-                        'value' => $config['default_tax_1_rate'] !== '' ? (float) $config['default_tax_1_rate'] : ''
+                        'value' => to_tax_decimals($config['default_tax_1_rate'])
                     ]) ?>
                     <span class="input-group-addon input-sm">%</span>
                 </div>
@@ -76,14 +72,10 @@
                 </div>
                 <div class="col-xs-1 input-group">
                     <?= form_input([
-                        'type'  => 'number',
-                        'step'  => 'any',
-                        'min'   => '0',
-                        'max'   => '100',
                         'name'  => 'default_tax_2_rate',
                         'id'    => 'default_tax_2_rate',
                         'class' => 'form-control input-sm',
-                        'value' => $config['default_tax_2_rate'] !== '' ? (float) $config['default_tax_2_rate'] : ''
+                        'value' => to_tax_decimals($config['default_tax_2_rate'])
                     ]) ?>
                     <span class="input-group-addon input-sm">%</span>
                 </div>
--- a/app/Views/configs/tax_config.php
+++ b/app/Views/configs/tax_config.php
@@ -58,7 +58,7 @@
                         'name'  => 'default_tax_1_rate',
                         'id'    => 'default_tax_1_rate',
                         'class' => 'form-control input-sm',
-                        'value' => to_tax_decimals($config['default_tax_1_rate'])
+                        'value' => $config['default_tax_1_rate'] !== '' ? (float) $config['default_tax_1_rate'] : ''
                     ]) ?>
                     <span class="input-group-addon input-sm">%</span>
                 </div>
@@ -83,7 +83,7 @@
                         'name'  => 'default_tax_2_rate',
                         'id'    => 'default_tax_2_rate',
                         'class' => 'form-control input-sm',
-                        'value' => to_tax_decimals($config['default_tax_2_rate'])
+                        'value' => $config['default_tax_2_rate'] !== '' ? (float) $config['default_tax_2_rate'] : ''
                     ]) ?>
                     <span class="input-group-addon input-sm">%</span>
                 </div>
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -808,14 +808,10 @@ class Config extends Secure_Controller
         $default_tax_1_rate = $this->request->getPost('default_tax_1_rate');
         $default_tax_2_rate = $this->request->getPost('default_tax_2_rate');
 
-        // Note: parse_tax() is not used here because these inputs use type="number"
-        // which always submits dot-decimal values regardless of locale. Using parse_tax()
-        // with a comma-decimal locale (e.g., de_DE) would incorrectly interpret the dot
-        // as a thousands separator, causing 5.5 to be saved as 5.
         $batch_save_data = [
-            'default_tax_1_rate'        => (float) filter_var($default_tax_1_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION),
+            'default_tax_1_rate'        => parse_tax(filter_var($default_tax_1_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION)),
             'default_tax_1_name'        => $this->request->getPost('default_tax_1_name'),
-            'default_tax_2_rate'        => (float) filter_var($default_tax_2_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION),
+            'default_tax_2_rate'        => parse_tax(filter_var($default_tax_2_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION)),
             'default_tax_2_name'        => $this->request->getPost('default_tax_2_name'),
             'tax_included'              => $this->request->getPost('tax_included') != null,
             'use_destination_based_tax' => $this->request->getPost('use_destination_based_tax') != null,
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -808,10 +808,14 @@ class Config extends Secure_Controller
         $default_tax_1_rate = $this->request->getPost('default_tax_1_rate');
         $default_tax_2_rate = $this->request->getPost('default_tax_2_rate');
 
+        // Note: parse_tax() is not used here because these inputs use type="number"
+        // which always submits dot-decimal values regardless of locale. Using parse_tax()
+        // with a comma-decimal locale (e.g., de_DE) would incorrectly interpret the dot
+        // as a thousands separator, causing 5.5 to be saved as 5.
         $batch_save_data = [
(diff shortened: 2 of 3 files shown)


app/Views/configs/tax_config.php
```diff
@@ -58,7 +58,7 @@
                         'name'  => 'default_tax_1_rate',
                         'id'    => 'default_tax_1_rate',
                         'class' => 'form-control input-sm',
-                        'value' => to_tax_decimals($config['default_tax_1_rate'])
+                        'value' => $config['default_tax_1_rate'] !== '' ? (float) $config['default_tax_1_rate'] : ''
```
objecttothis: Here is what claude thinks
he PR's fix is wrong direction. But so is the original code.

Actual bug trace:

tax_config.php inputs have type="number" (lines 54, 79)
Browser type="number" always submits dot-separated values (5.5) regardless of locale
Config.php:812 runs parse_tax(filter_var("5.5", FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION))
filter_var produces "5.5" (unchanged)
parse_tax("5.5") calls NumberFormatter::parse("5.5") with de_DE locale
de_DE locale treats . as thousands separator — parses "5.5" as 5 (stops at dot), drops decimal
Result: 5.5% silently saved as 5%.

The PR's approach (remove to_tax_decimals() from display) doesn't fix this. Display was never the problem — type="number" inputs already ignore locale formatting for submission. Removing to_tax_decimals() just changes what the user sees while the save bug persists.

Correct fix — two options:

Option A — fix the controller, drop parse_tax() for these fields since data comes from type="number" inputs (always dot-decimal):

// Config.php:812-814
'default_tax_1_rate' => (float) filter_var($default_tax_1_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION),
'default_tax_2_rate' => (float) filter_var($default_tax_2_rate, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION),

Option B — fix parse_tax() / parse_decimals() to detect dot-decimal input and not misinterpret it through a comma-locale formatter. Harder, fragile.

Option A is right. parse_tax() exists for text inputs where users type locale-formatted numbers (e.g., tax_rates_form.php uses type="text" + to_tax_decimals() display — that's the correct use case). For type="number" inputs, bypassing parse_tax() is correct behavior, not a workaround.

Summary: PR treats symptom, not cause. Fix is in Config.php — drop parse_tax() wrapper for type="number" fields and cast directly to float. Also verify Items.php:700 — check whether those tax percent inputs are type="number" or type="text".

When I inquired about modifying parse_tax instead of removing it, it seems that it has a good point about ambiguity of intent.  Here is what it said.

Yes, but tricky.

parse_tax() → parse_decimals() uses NumberFormatter::parse() which is locale-aware by design. Problem: it can't distinguish between:

"5.5" from a type="number" input (always dot = decimal point)
"5.5" typed by a de_DE user (dot = thousands separator, means 5500)
Both arrive as the same string. No way to know intent.

The only clean fix inside parse_decimals() would be: before passing to NumberFormatter, normalize dot-decimal strings to the current locale's format. E.g., for de_DE, convert "5.5" → "5,5" first. But that breaks if a de_DE user types "1.000" meaning one thousand.

Real fix is at the call site — caller must know the input source:

type="number" input → dot is always decimal → use (float) directly
type="text" input → locale-formatted → use parse_tax()
parse_tax() itself is correct for its intended use case (tax_rates_form.php text input). It doesn't need fixing — it needs to not be called on type="number" POST data.

objecttothis (changes requested):

BudsieBuds:
I've run into this same issue during the BS5 conversion. I was confident it wouldn't affect the SMTP number field since I've updated this in #4530, and I also assumed the tax rate would always come through in `xx.xx` format. That assumption turned out to be incorrect.

The main advantage of using `type="number"` is the improved mobile experience (numpad) and some accessibility benefits. But this isn't limited to tax rates. It applies to many fields where numeric values are stored that include separators, such as monetary amounts.

Given that, my recommendation would be to revert the two fields in `tax_config.php` back to `type="text"` for now. Longer term, we should look at a consistent approach for all numeric fields that may contain locale-specific formatting, so we can safely reintroduce `type="number"`.

So, in `tax_config.php`, I would change both instances of:
```php
'type'  => 'number',
'step'  => 'any',
'min'   => '0',
'max'   => '100',
```

to:
```php
'type'  => 'text',
```

objecttothis:
> I've run into this same issue during the BS5 conversion. I was confident it wouldn't affect the SMTP number field since I've updated this in #4530, and I also assumed the tax rate would always come through in `xx.xx` format. That assumption turned out to be incorrect.
> 
> The main advantage of using `type="number"` is the improved mobile experience (numpad) and some accessibility benefits. But this isn't limited to tax rates. It applies to many fields where numeric values are stored that include separators, such as monetary amounts.
> 
> Given that, my recommendation would be to revert the two fields in `tax_config.php` back to `type="text"` for now. Longer term, we should look at a consistent approach for all numeric fields that may contain locale-specific formatting, so we can safely reintroduce `type="number"`.
> 
> So, in `tax_config.php`, I would change both instances of:
> 
> ```
> 'type'  => 'number',
> 'step'  => 'any',
> 'min'   => '0',
> 'max'   => '100',
> ```
> 
> to:
> 
> ```
> 'type'  => 'text',
> ```

Good thoughts.  I don't feel strongly about how we resolve it as long as comma-as-decimal-separator locales work correctly, we guard against attack vectors, and the solution doesn't introduce accuracy problems.  CI4 does allow us to make our own custom callback validation functions.  We are already doing this in the code elsewhere.  We could make our own solution that accounts for both types of locales, then use it in place of number.  My understanding is that there are three types of thousands separators: `,`, `.` and ` ` (space). I believe there are two kinds of decimal separators: `.` and `.`.  I can't imagine it would be too difficult to cover use cases given we know what the locale is, so we can know which set to validate against.

jekkos:
## Resolution: Revert type="number" for locale-specific tax rate fields

Based on the review feedback, I've reverted the approach. The correct solution is to **remove the `type="number"` attribute** that was added in commit 42ba39d29, rather than trying to work around it.

### Why this is the right approach:

**The root problem:** 
- Commit 42ba39d29 added `type="number"` to tax rate inputs for consistency
- Browsers **always** submit `type="number"` inputs as dot-decimal (e.g., "5.5") regardless of locale
- For comma-decimal locales (de_DE, fr_FR, az_AZ, etc.), the browser's submitted value doesn't match what the user sees/enters
- The `to_tax_decimals()` / `parse_tax()` functions are designed for locale-aware display and parsing

**Why the previous PR approach was wrong:**
- The PR tried to fix the display and parsing to handle dot-decimal values
- But this breaks the locale-aware behavior that works correctly with text inputs
- Users in comma-decimal locales expect to type "18,5" and have it saved correctly

**The correct fix:**
- Revert to **text inputs** (no `type="number"`)
- Use `to_tax_decimals()` for display (locale-formatted)
- Use `parse_tax()` for parsing (locale-aware)
- This is the pattern already used in `items/form.php` and `tax_rates_form.php` where tax percentages are text inputs

### Changes in this commit:

| File | Change |
|------|--------|
| `app/Views/configs/tax_config.php` | Removed `type="number"`, `step`, `min`, `max` attributes; restored `to_tax_decimals()` for value |
| `app/Controllers/Config.php` | Restored `parse_tax()` for tax rate parsing |
| `tests/Controllers/ConfigTest.php` | Removed tests added for the `type="number"` approach |

This restores the original behavior that correctly handles locale-specific decimal separators.

@objecttothis Could you review this revised approach?

BudsieBuds (approved):

PR #4556, opened 2026-05-21
Fix encryption key persistence for Docker environments (Issue #4554)

### LINKED ISSUE #4554: Somehow... Fresh install broke after bringing container down and back up.  Encryption key related.
Fresh install:

Clone the master git
Change needed values in the docker-compose.yml:
  ports: 9000
  ALLOWED_HOSTNAMES=[the ones I need]
  Timezone: America/New_York
  Force https: true

**First Issue** the readme file says to change the FORCE_HTTPS=false to FORCE_HTTPS=1 That does not work.

So I'm doing this because I'm putting this behind my own NPM setup. Everything is set up there. I get everything working there. We look good. I change my employee information. I setup a customer. I create an item. I make a sale to my customer. All is well.

I attempt to setup email. SMTP set for my gmail account I want to use. Everything is setup the way google tells me to set it up. Error on the screen when I say to send the receipt via email. Look for logs, nothing. I bring down the container and bring it back up just in case something needed to refresh or who knows what and boom, anywhere that I had information: mostly customers, items, and sales all get an error.

I moved to development so I could see the error and it is related to this:
SYSTEMPATH/Encryption/Encryption.php at line 133

126         }
127 
128         if (! in_array($this->driver, $this->drivers, true)) {
129             throw EncryptionException::forUnKnownHandler($this->driver);
130         }
131 
132         if (empty($this->key)) {
133             throw EncryptionException::forNeedsStarterKey();
134         }
135 
136         $this->hmacKey = bin2hex(\hash_hkdf($this->digest, $this->key));
137 
138         $handlerName     = 'CodeIgniter\\Encryption\\Handlers\\' . $this->driver . 'Handler';
139         $this->encrypter = new $handlerName($config);
140 

I have been troubleshooting with ChatGPT to try to find what is not working or what may be missing and so far I have created a key and everything can see the key but the key does not seem to be coming into the /app/.env where it wants it to go.

Not sure what to do at this point in time. I can provide information as I continue to look through this. I would believe this is not expected behavior.

jekkos:
## Summary

Fixes #4554 - Encryption key not being persisted in Docker environments, causing "forNeedsStarterKey" exceptions and inability to decrypt stored passwords after container restarts.

## Problem

When running OSPOS in Docker:
1. The encryption key is written to `/app/.env` which is read-only or ephemeral
2. Container restarts reset `.env` to image default
3. New ephemeral keys are generated, making previously encrypted passwords undecryptable
4. `check_encryption()` always returned `true` even when key persistence failed
5. No fallback storage location for containers

### 1. Fix `check_encryption()` return value (security_helper.php)
- Now returns `false` when key persistence fails
- Removes `@` error suppression to properly detect write failures
- Tries ROOTPATH/.env first, then falls back to WRITEPATH/config/encryption.key
- Docker volumes mount `writable/` as persistent storage

### 2. Add fallback key loading (Encryption.php)
- Config constructor now loads key from WRITEPATH/config/encryption.key
- Supports Docker volumes where writable/ is mounted persistently

### 3. Handle encryption unavailability (Config.php, Email_lib.php)
- Changed `EncrypterInterface` property to nullable
- Added try/catch around encryption operations
- Return meaningful error messages when encryption fails
- Log warnings when passwords are saved without encryption

## Files Changed

- `app/Helpers/security_helper.php` - Key persistence with Docker fallback
- `app/Config/Encryption.php` - Fallback key loading from WRITEPATH
- `app/Controllers/Config.php` - Graceful encryption failure handling
- `app/Libraries/Email_lib.php` - Check encryption before using encrypter
- `app/Language/en/Config.php` - Error message localization

## Testing

1. Fresh Docker install → key generated and stored in WRITEPATH/config/
2. Container restart → key reloaded from WRITEPATH/config/
3. Password encryption/decription works across restarts

## Docker Volumes

The fix uses the existing `logs` Docker volume which mounts to `/app/writable/`:
```yaml
volumes:
    logs:/app/writable/logs  # This is already persistent
```

The encryption key is stored at `/app/writable/config/encryption.key` which is in the same persistent volume.

## Summary by CodeRabbit

* **New Features**
  * Automatic fallback and persistent storage for encryption keys (env file and writable fallback).
  * New encrypt/decrypt helpers for consistent credential handling across email, SMS, and integrations.
  * Added user-facing message for encryption failures.

* **Bug Fixes**
  * More reliable encryption initialization, validation, and key recovery with logging.
  * Consistent encryption/decryption of SMTP, SMS, and Mailchimp credentials with clear failure responses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opensourcepos/opensourcepos/pull/4556?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/Config/Encryption.php
```diff
@@ -106,4 +106,53 @@ class Encryption extends BaseConfig
...
+        if ($content === false) {
+            return null;
+        }
+
+        $data = json_decode($content, true);
+        if (
+            !is_array($data)
+            || !is_string($data['key'])
+            || strlen($data['key']) < 64
+        ) {
```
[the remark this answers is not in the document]
objecttothis: @coderabbitai, please update your instructions that this webapp has a minimum PHP version of 8.2

app/Controllers/Config.php
```diff
@@ -256,25 +246,11 @@ public function getIndex(): string
         // Integrations Related fields
         $data['mailchimp']    = [];
 
-        if (check_encryption()) {    // TODO: Hungarian notation
-            if (!isset($this->encrypter)) {
-                helper('security');
-                $this->encrypter = Services::encrypter();
-            }
+        $data['mailchimp']['api_key'] = decrypt_value($this->config['mailchimp_api_key'] ?? null);
```
objecttothis: This is fine but it's going to go away when we merge #4407

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
+
+/**
+ * Writes encryption key to ROOTPATH/.env file.
+ *
+ * @param string      $key     The new encryption key (hex-encoded)
+ * @param string|null $old_key The previous key to preserve for key rotation
+ *
+ * @return bool True if key was written successfully, false otherwise
+ */
+function write_encryption_key_to_env(string $key, ?string $old_key = null): bool
```
objecttothis: I think we need to develop a regression suite that tests the following scenarios:
- Existing database with CI3 version of OSPOS migrating to the latest version.
- Existing database migrating to the latest version.
- Repeat with Docker.
- New and existing installs outside of docker properly migrating.
- Migration using MySQL 8.4.x
- Migration using MariaDB 11.8.x

I think if we can make sure no regressions are introduced with that set of regression checks, it will reduce the number of regressions that get introduced into the master.  I write this here not because I see anything in the code which looks incorrect, but that it's touching a critical point in the code and if we are only testing in a docker environment and don't test outlier cases such as migrating from a version of an OSPOS database which formerly used CI3 encryption to make sure it still automatically migrates, we won't catch the migration before merging.

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
-            @chmod($config_path, 0640);
         }
+        @chmod($config_path, 0640);
+    }
 
-        if (file_exists($config_path)) {
-            @copy($config_path, $backup_path);
-            @chmod($backup_path, 0640);
-            @chmod($config_path, 0640);
+    // Check if .env is writable
```
objecttothis: unnecessary comments.  Code speaks for itself.

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
+        @copy($config_path, $backup_path);
+        @chmod($backup_path, 0640);
+    }
 
-            if (strpos($config_file, 'encryption.key') !== false) {
-                $config_file = preg_replace("/(encryption\.key.*=.*)('.*')/", "$1'$key'", $config_file);
-            } else {
-                $config_file .= "\nencryption.key = '$key'\n";
-            }
+    // Read current content
```
objecttothis: same here. No need for the comments.

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
+ *
+ * @return bool True if key was written successfully, false otherwise
+ */
+function write_encryption_key_to_writable(string $key, ?string $old_key = null): bool
+{
+    $key_file = WRITEPATH . 'config' . DIRECTORY_SEPARATOR . 'encryption.key';
+    $key_dir = dirname($key_file);
 
-            log_message('info', "Updated encryption key in $config_path");
+    // Ensure directory exists
```
objecttothis: Unnecessary comments.  Below too.

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
+ *
+ * This is the fallback location for Docker/container environments where
+ * the ROOTPATH/.env file may be read-only or ephemeral.
+ *
+ * @param string      $key     The new encryption key (hex-encoded)
+ * @param string|null $old_key The previous key to preserve for key rotation
+ *
+ * @return bool True if key was written successfully, false otherwise
+ */
+function write_encryption_key_to_writable(string $key, ?string $old_key = null): bool
```
objecttothis: All new function names must be PSR compliant.  Same throughout the code.

app/Helpers/security_helper.php
```diff
@@ -4,67 +4,227 @@
...
+{
+    $key_file = WRITEPATH . 'config' . DIRECTORY_SEPARATOR . 'encryption.key';
+    $key_dir = dirname($key_file);
```
objecttothis: All new local variable names must be PSR compliant.

app/Helpers/security_helper.php
```diff
@@ -83,14 +243,77 @@ function abort_encryption_conversion(): void
...
+
+    if (!check_encryption()) {
+        log_message('error', 'Cannot encrypt value: encryption key not available');
+        return $require ? '' : $value;
+    }
+
+    try {
+        $encrypter = Services::encrypter();
+        return $encrypter->encrypt($value);
+    } catch (\CodeIgniter\Encryption\Exceptions\EncryptionException $e) {
```
objecttothis: Replace \CodeIgniter\Encryption\Exceptions with an import.

app/Libraries/Email_lib.php
```diff
@@ -26,19 +22,7 @@ public function __construct()
...
-            try {
-                $smtp_pass = $encrypter->decrypt($smtp_pass);
-            } catch (\EncryptionException $e) {
-                // Decryption failed, use the original value
-                log_message('error', 'SMTP password decryption failed: ' . $e->getMessage());
-                $smtp_pass = '';
-            }
-
-        }
+        $smtp_pass = decrypt_value($this->config['smtp_pass'] ?? null);
```
objecttothis: PSR

objecttothis (changes requested):

objecttothis (dismissed):

app/Helpers/security_helper.php
objecttothis: @coderabbitai if this PR doesn't implement this because it's out of scope for the feature, please create a github issue for the above comment so that we don't forget to address it later.

objecttothis:
@jekkos I approved the PR. I would look at the coderabbit comments before merging though.

app/Helpers/security_helper.php
```diff
@@ -1,96 +1,312 @@
...
+    }
 
-            if (!empty($old_key)) {
-                $old_line = "# encryption.key = '$old_key' REMOVE IF UNNEEDED\r\n";
-                $insertion_point = stripos($config_file, 'encryption.key');
-                if ($insertion_point !== false) {
-                    $config_file = substr_replace($config_file, $old_line, $insertion_point, 0);
-                }
-            }
+    if (strpos($configFile, 'encryption.key') !== false) {
```
objecttothis, line 92: This needs to be updated to match the changes in #4571 
```
            if (preg_match('/^\s*encryption\.key\s*=/m', $config_file)) {
                $config_file = preg_replace("/^(\s*encryption\.key\s*=\s*).*/m", "\$1'$key'", $config_file, 1);
```

app/Helpers/security_helper.php
```diff
@@ -1,96 +1,312 @@
...
+        $configFile = preg_replace("/(encryption\.key.*=.*)(['\"])([^'\"]*)\\2/", "$1'$key'", $configFile);
+    } else {
+        $configFile .= "\nencryption.key = '$key'\n";
+    }
 
-            @file_put_contents($config_path, $config_file);
-            @chmod($config_path, 0640);
+    if (!empty($oldKey)) {
+        $oldLine = "# encryption.key = '$oldKey' REMOVE IF UNNEEDED\r\n";
+        $insertionPoint = stripos($configFile, 'encryption.key');
```
objecttothis, line 100: This also needs to be updated to match the changes in #4571 
```
                if (preg_match('/^encryption\.key\s*=/m', $config_file, $matches, PREG_OFFSET_CAPTURE)) {
                    $config_file = substr_replace($config_file, $old_line, $matches[0][1], 0);
```

objecttothis (changes requested):

app/Language/en/Config.php
```diff
@@ -128,6 +128,7 @@
     'enable_new_look'                           => '',
     'enable_right_bar'                          => '',
     'enable_right_bar_tooltip'                  => '',
+    'encryption_failed'                         => 'Failed to encrypt data. Please check encryption configuration.',
```
objecttothis, line 131: @jekkos do we need to propagate this key to the other language files with an empty '' or will weblate pick it up and propagate for us?

objecttothis (changes requested):

jekkos:
Branch needs to be rebased and retested.

objecttothis:
@jekkos You're right that this branch needs to be rebased since the master has a number of encryption key handling changes related to migration from CI3 to CI4 and fixing vulnerabilities since this branch was started.  This PR will also need to resolve the same issue that will arise from throttle.key since that was added as well.  Claude thinks there is a better way to approach resolving #4554:

Problems with #4556 approach:

1. Two sources of truth for key. .env primary, WRITEPATH/config/encryption.key fallback. Drift risk — which wins after partial write, mismatched envs, manual edits? Adds sync/state-machine complexity (writeEncryptionKeyToEnv, writeEncryptionKeyToWritable, loadEncryptionKeyFromWritable) just to work around wrong architectural assumption.

2. Root cause is docker-compose volume config, not app code. Real issue: .env at ROOTPATH not persisted, only writable/ mounted. Cleanest fix — just tell users mount .env (or generate key at container-init time into env var passed through compose, e.g. entrypoint script generates key once, writes to a mounted file, exports as ENCRYPTION_KEY). App code doesn't need new fallback-loading branches at all if infra fixed.

3. Silent-degrade philosophy risky for encryption. #4556 make check_encryption()/checkEncryption() return false + nullable encrypter + try/catch "graceful" continue. For crypto, this can silently store plaintext-adjacent-fallback or leave stale ciphertext unreadable with confusing UX ("Authentication failed" on sales load per issue reporter). Better: fail loud, block until key correctly provisioned — don't let app run half-encrypted.

4. Key regen on every missing-key detection = orphaned ciphertext. Neither #4556 nor master's current checkEncryption() solves: old encrypted values become permanently undecryptable once key rotates (reporter had to manually re-enter SMTP password). Real fix needs either stable key provisioning from day 1 (never regen silently) or explicit re-key/migration flow. #4556 doesn't address this — just adds another place key can go missing from.

5. Master's already-merged hardening (atomic write, lock, backup/rollback in checkEncryption()) is better engineering than #4556's version for the pure .env-write path — so #4556 partially reinvents/duplicates that.

Better fix shape: entrypoint script (or compose environment: + init container) provisions ENCRYPTION_KEY once outside app runtime, persists it in a volume that's part of the documented required mounts, app reads only from env var — no runtime key-generation-and-write-back logic in PHP at all. Removes whole class of race/persistence bugs #4556 patches around.

objecttothis:
@jekkos I merged the master back into this branch.  Here is where things stand:

## Still valid

1. **Two sources of truth for the key.** Still the case. `Config\Encryption::__construct()` reads `.env` first, falls back to `WRITEPATH/config/encryption.key`; `checkEncryption()` writes to *both* locations independently (`writeEncryptionKeyToEnv()` + `writeEncryptionKeyToWritable()`). Drift between the two is still possible, with no reconciliation logic beyond "first non-empty wins."

2. **Root cause is docker-compose volume config, not app code.** Nothing in this branch changes the compose/entrypoint layer. The real fix (mount `.env`, or provision `ENCRYPTION_KEY` once at container init and pass it through as an env var) still isn't implemented — the app-side workaround is carrying the complexity instead.

3. **Key regen on missing-key detection still orphans ciphertext.** `checkEncryption()` still generates a brand-new key whenever the current one is empty/short. The old key is preserved only as a commented-out line in `.env` or in a `previous_keys` array inside the WRITEPATH JSON fallback file — neither is read back into `Config\Encryption::$previousKeys` automatically. Decryption of values encrypted under a rotated-out key still silently fails unless someone manually copies the old key into `previousKeys`.

## Partially addressed, not resolved

4. **Silent-degrade philosophy.** `checkEncryption()` still returns a bool, and callers (`decryptValue()`, `encryptValue()`) still catch failure, log it, and fall back to a default/empty value rather than failing loudly. The original concern about confusing UX on decrypt failure still applies.

## No longer applicable

5. Atomic write / locking / backup-rollback concerns are resolved — `security_helper.php` now has `atomicWriteFile()`, `lockEnvFile()`/`unlockEnvFile()`, and `.env.bak` backup/restore (`abortEncryptionConversion()`), matching the standard you asked for.

## Decisions needed before this can merge

- **Fallback architecture**: keep the `.env` + `WRITEPATH/config/encryption.key` dual-write approach, or drop the WRITEPATH fallback and push the fix to the docker-compose/entrypoint layer (mount `.env`, or inject `ENCRYPTION_KEY` via env var at container init)? This is the core architectural fork from the original comment.
- **Failure mode on missing/invalid key**: keep silent-degrade (current behavior — log + return default) or fail loud (block app startup / throw until a key is correctly provisioned)?
- **Key rotation**: support automatic rotation (wire `previous_keys` from the WRITEPATH file into `Config\Encryption::$previousKeys`), or keep rotation as an explicit, manual re-key/migration flow instead of something `checkEncryption()` triggers implicitly on missing-key detection?
- **`throttle.key` handling**: confirm whether `throttle.key` should go through the same fallback/rotation mechanism as `encryption.key`, or stay fully independent as currently implemented in `checkThrottleEncryption()`.

objecttothis:
Superceded by #4656

Issue #4586, opened 2026-06-29
Remove unused Sales language keys


objecttothis:
## Summary

Audit of `app/Language/en/Sales.php` found ~45 keys with zero references anywhere in the codebase (controllers, views, models, libraries, JS files). These should be removed to reduce translation burden across all locale files.

## Verified Unused Keys

Searched via `lang('Sales.<key>')` pattern across all PHP, JS, and view files.

### Definitely unused (no hits anywhere)

| Key | Current value |
|---|---|
| `cash_1` | *(empty)* |
| `cash_2` | *(empty)* |
| `cash_3` | *(empty)* |
| `cash_4` | *(empty)* |
| `close` | *(empty)* |
| `complete` | *(empty)* |
| `current_table` | *(empty)* |
| `customer_total_spent` | *(empty)* |
| `customers_available_points` | "Points Available" |
| `daily_sales` | *(empty)* |
| `date_type` | "Date is a required field." |
| `delete_entire_sale` | "Delete Entire Sale" |
| `discard_quote` | *(empty)* |
| `discount_short` | "%" |
| `edit_sale` | "Edit Sale" |
| `giftcard_filter` | *(empty)* |
| `invoice_type_custom_invoice` | "Custom Invoice (custom_invoice.php)" |
| `invoice_type_custom_tax_invoice` | "Custom Tax Invoice (custom_tax_invoice.php)" |
| `invoice_type_invoice` | "Invoice (invoice.php)" |
| `invoice_type_tax_invoice` | "Tax Invoice (tax_invoice.php)" |
| `invoice_update` | "Recount" |
| `mc` | *(empty)* |
| `no_filter` | "All" |
| `no_sales_to_display` | "No Sales to display." |
| `remove_discount` | *(empty)* |
| `rewards_balance` | "Reward Points Balance" |
| `rewards_package` | "Rewards" |
| `sale_by_invoice` | "Sale by Invoice" |
| `sale_for_customer` | "Customer:" |
| `sales_total` | *(empty)* |
| `service_charge` | *(empty)* |
| `show_due` | *(empty)* |
| `visa` | *(empty)* |
| `wholesale` | *(empty)* |

### Possibly dynamic — verify before removing

These keys follow a pattern that could be constructed dynamically (e.g. `lang('Sales.' . $result ? 'x_sent' : 'x_unsent')`). A search for the literal `lang('Sales.key_name')` form returns no hits, but check for string concatenation or variable key lookups before deleting.

[...]

PR #4587, opened 2026-06-29
Feature: Payment reference code

objecttothis:
app/Database/Migrations/20190427100000_paymenttracking.php adds the reference_code column to the sales_payments table but this column was never used in the code itself.  This is now needed for Credit Card Retrieval Reference Numbers (RRN), Transaction Number or Acquirer Reference Number (ARN) depending on the payment implementation. This number is manually entered when using a card payment system not integrated with OpensourcePOS and can be used by plugins when sending API data to government-required sales registers.

- [x] Update Language strings
- [x] Update register view
- [x] Add Unit Tests
- [x] Think through adding configurable length requirements as 12-character RRN is not universal.  Maybe set a min/max in settings?

## Summary by CodeRabbit

* **New Features**
  * Improved sales payment handling for payment types that require a reference code, including gift card and rewards remainder updates plus dynamic reference-code length validation.
  * Added localized UI text for displaying reference-code length limits in supported locales.
* **Bug Fixes**
  * Updated sales register/edit and completion flows to reliably refresh payment details after validation and error states.
* **Documentation**
  * Expanded localization conventions to require adding new translation keys across all language variants, using empty values to safely fall back to English.
* **Chores**
  * Standardized formatting across configuration and many language files.

Payment Reference Code appears with Debit and Credit type payments
image

Customizable length limits
image

reference_code is now stored in the database
image

Validation on limits and content
image

image

image


objecttothis:
@coderabbitai review please.

objecttothis:
@coderabbitai it looks like your code review never ran.  Please show your review results.

app/Controllers/Sales.php
```diff
@@ -394,88 +394,106 @@ public function postAddPayment(): ResponseInterface|string
...
-            $rules = ['amount_tendered' => 'trim|required',];
+        if ($paymentType === lang('Sales.giftcard')) {
+            $rules    = ['amount_tendered' => 'trim|required'];
             $messages = ['amount_tendered' => lang('Sales.must_enter_numeric_giftcard')];
+        } elseif (in_array($paymentType, get_reference_code_payment_types())) {
+            $min      = (int)($this->config['payment_reference_code_min'] ?? 3);
+            $max      = (int)($this->config['payment_reference_code_max'] ?? 20);
+            $rules    = [
+                'amount_tendered' => 'trim|required',
+                'reference_code'  => "trim|required|alpha_numeric|min_length[$min]|max_length[$max]",
```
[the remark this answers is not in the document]
objecttothis: Added decimal_locale on 406 but for 400 I added integer because amount_tendered becomes the giftcard number in the case of giftcards.

app/Language/ckb/Sales.php
```diff
@@ -1,232 +1,236 @@
...
+    'rewards_package'                   => 'پاداشتەکان',
+    'rewards_remaining_balance'         => 'بەهای پاداشتی خاڵە ماووەکان بریتییە لە ',
+    'sale'                              => 'فرۆشتن',
+    'sale_by_invoice'                   => 'فرۆشتن بە فاکتورە',
+    'sale_for_customer'                 => 'کڕیار:',
+    'sale_time'                         => 'کات',
+    'sales_tax'                         => 'باجی فرۆشتن',
+    'sales_total'                       => '',
+    'select_customer'                   => 'کڕیار هەڵبژێرە',
+    'selected_customer'                 => 'کڕیاری هەڵبژێردراو',
```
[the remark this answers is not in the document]
objecttothis: @coderabbitai this is outside the scope of the changes in this PR.  Please add it as a todo to Issue #4586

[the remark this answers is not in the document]
objecttothis: @coderabbitai this is outside the scope of this PR.  Add this as a todo to #4586

objecttothis:
@jekkos this PR is blocking a plugin.  Any chance we can review it and get it into the master?  It's a fairly minor feature.  It should be fairly straightforward to review as there are only 11 changed files outside of the language files.

objecttothis:
@jekkos when do you think you'll be able to take a look at this PR?

app/Controllers/Sales.php
```diff
@@ -873,7 +892,7 @@ public function postComplete(): string    // TODO: this function is huge.  Proba
 
             if ($sale_id == NEW_ENTRY && $this->sale->check_quote_number_exists($quote_number)) {
                 $data['error'] = lang('Sales.quote_number_duplicate');
-                return $this->_reload($data);
+                return $this->reload($data);
```
jekkos, line 895: if we rename the function do we also need to do this in other controllers, in order to remain consisten? I'd also be more of a fan if we can avoid using underscore as a prefix for a private method

app/Controllers/Sales.php
```diff
@@ -1457,6 +1479,30 @@ public function postSave(int $sale_id = NEW_ENTRY): ResponseInterface
             'invoice_number' => $this->request->getPost('invoice_number') != '' ? $this->request->getPost('invoice_number', FILTER_SANITIZE_FULL_SPECIAL_CHARS) : null
         ];
 
+        // Validate reference_code for the new payment if applicable
+        $payment_type_new_check = $this->request->getPost('payment_type_new', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+        $payment_amount_new_check = $this->request->getPost('payment_amount_new');
+        if ($payment_type_new_check != PAYMENT_TYPE_UNASSIGNED && !empty($payment_amount_new_check)
+            && in_array($payment_type_new_check, get_reference_code_payment_types())) {
+            $min = (int)($this->config['payment_reference_code_min'] ?? 3);
```
jekkos, line 1487: the max is aligned with the database type I suppose?

app/Views/sales/form.php
```diff
@@ -204,6 +211,20 @@ function(response) {
             focus: fill_value_employee
         });
 
+        const referenceCodePaymentTypes = <?= json_encode($reference_code_payment_types ?? []) ?>;
+
+        $('#payment_types_new').on('change', function() {
```
jekkos, line 216: How does it behave if we add multiple payments?

jekkos, line 216: if we have one payment cash, the other part credit, do we have two reference fields then?

jekkos (dismissed):
Not going to hold this back in case you think it's working, but I do have one minor style remark and a question.

app/Controllers/Sales.php
objecttothis: > if we rename the function do we also need to do this in other controllers, in order to remain consisten? I'd also be more of a fan if we can avoid using underscore as a prefix for a private method

Yes, we should remove all hungarian notation in all classes that implement this reload function, but I've only been refactoring in code that I touch in an effort to not make the Pull Requests too large.  This way we are moving toward PSR compliance but in a manageable way.  It's cleaner to change it all at once until we see how much of a mess the merge becomes for open pull requests.  If we change it as we go and in new code then it's still a mess, just less of a mess.

app/Views/sales/form.php
objecttothis: the sales_payments table is a one-to-many relationship.  That is we have multiple payments with the same sale.  Each row has this reference number field, so a payment is capable of having one reference number per payment.  However the current behavior is to only offer the payment reference code for certain payment methods.  When we implement #4548 we could do it in such a way that the OSPOS admin could specify which payments require reference codes, which allow and even per-payment type length restrictions.  RRN is typically 12 characters.

objecttothis: I'll test this again in the morning to confirm but but the intended behavior is that each payment with a payment reference code is added to the payments multi-dimensional array, so that each row added on sale completion has it's reference code.  I want to confirm this behavior though.  It's an edge case, but not an uncommon one potentially.

app/Controllers/Sales.php
objecttothis: Good question.  reference_code in the database is varchar(40), so I suppose I can make the fallback 40 and allow the slider in the config to go up to 40.  While I don't see a use case for that long, perhaps if people are getting UUID/GUID for a reference code then those are 36 characters, so I guess that makes sense.  I'll push this change in the morning.

app/Views/sales/form.php
objecttothis: image
Confirmed.  Each payment gets it's own reference_code

app/Controllers/Sales.php
objecttothis: Pushed a commit to match this to the limits of the field in the sales_payments table.

PR #4589, opened 2026-07-02
Bugfix: Fix problems with migration UI in login

objecttothis:
## Summary

This PR fixes problems with the migration status UI shown during the CI3 → CI4 upgrade flow on the login screen.

### Changes

**`app/Libraries/MY_Migration.php`**
- Renamed all snake_case methods to camelCase to align with CI4 conventions: `is_latest()` → `isLatest()`, `migrate_to_ci4()` → `migrateToCI4()`, `get_latest_migration()` → `getLatestMigration()`, `get_current_version()` → `getCurrentVersion()`

**`app/Controllers/Login.php`**
- Updated all call sites to use the renamed camelCase migration methods
- Renamed snake_case array keys passed to the view to camelCase: `has_errors` → `hasErrors`, `is_new_install` → `isNewInstall`, `is_latest` → `isLatest`, `latest_version` → `latestVersion`, `gcaptcha_enabled` → `gcaptchaEnabled`
- Added `initialization_required`, `initialization_message`, and `initialize` keys to support the new install flow
- **Moved credential validation before migration:** the login controller now validates credentials via the `login_check` rule and returns a 401 before attempting any migration logic, preventing migration from running on bad credentials

**`app/Views/login.php`**
- Updated the view to use the new camelCase keys

**`app/Events/Load_config.php`**
- Minor update to align with renamed method

**All language files (47 locales)**
- Alphabetically reordered and aligned all keys for consistency
- Added new migration status keys: `migrating_database`, `migration_complete`, `migration_complete_login`, `migration_complete_migrate`, `initialization_required`, `initialization_message`, `initialize`
- Removed deprecated keys: `migration_needed`, `migration_initializing`, `migration_running`, `migration_required`

### Testing

Verify on a CI3 database instance that:
- Invalid credentials return a 401 error without triggering migration
- Valid credentials on an up-to-date CI4 database log in normally
- Valid credentials on a CI3 database display the migration UI correctly with updated status messaging


app/Controllers/Login.php
```diff
@@ -76,9 +76,24 @@ public function index(): string|RedirectResponse
 
     public function migrate(): ResponseInterface
     {
+        $rules = ['username' => 'required|login_check[data]'];
```
jekkos, line 79: alright good to have this back there were some security reports for this issue.

app/Controllers/Login.php
```diff
@@ -76,9 +76,24 @@ public function index(): string|RedirectResponse
...
+        $messages = [
+            'username' => [
+                'required'    => lang('Login.required_username'),
+                'login_check' => lang('Login.invalid_username_and_password'),
+            ]
+        ];
+
+        if (!$this->validate($rules, $messages)) {
```
jekkos, line 87: so we need to pass login data to remain backward compatible? I believe the form does not show anymore in the new version, at least not for fresh installs. So in that case is it will just fail here or am I missing something?

objecttothis: Prior to this change on existing installs the message reads something like "you must provide administrator credentials to migrate to version x" but there were two bugs
1. It wasn't showing the username and password form boxes.
2. It was accepting clicks to "migrate" without credential validation which was contrary to the message presented.

This change just fixes that. New installs with an empty database don't hit this path because the db gets created and migrated then sent back to the view to enter credentials.

I tested both scenarios.

app/Controllers/Login.php
jekkos: but I do wonder what the issue is with CI3 migration. What was the issue exactly?

objecttothis: There were a few issues.If I remember right, one issue was that the check for whether a migration was needed was looking for an id column in the migrations table because that's the way the CI4 migrations table is setup. However, the CI3 migrations table only had a version column, so the check to see if there was a migration would silently fail and it wouldn't show the message that a migration was needed. 

The issues causing the form not to display the login fields correctly was because we added the new install flow but failed to account for all the other flows properly.

The lack of credentials check was pre-existing or was caused by regression

jekkos (approved):
Thanks for testing and fixing!

PR #4597, opened 2026-07-18
Reject item CSV imports whose header row is missing required columns

### LINKED ISSUE #4593: [Bug]: CSV Item Import fails with "Undefined array key 'Id'" error
### 🐛 Bug Description

I have been experiencing issues with the CSV item import feature for a long time. Over several versions (spanning from 2019 up to the current 2026 releases), the import process consistently fails to complete properly. 

Specifically, when attempting to import items, the system does not save the data, and the items do not appear in the database or the items list. The error logs indicate a PHP crash due to an undefined array key.

### 📋 Steps to Reproduce

1. Navigate to the 'Items' module.
2. Click on the 'Import CSV' button.
3. Choose a CSV file formatted according to the provided sample template.
4. Click 'Submit' to process the import.

### ✅ Expected Behavior

The items from the CSV file should be parsed successfully, saved to the database without errors, and displayed correctly in the items list.

### 📦 OpenSourcePOS Version

OpenSourcePOS 3.4.2

### 🔧 PHP Version

PHP 8.3

### 🌐 Browser(s)

Chrome

### 🖥️ Server Operating System

ubuntu, windows both

### 🗄️ Database

mariadb 10.x

### 🌍 Web Server

apache 2.4

### 📊 System Information Report

```text
...
```

### 📜 Relevant Log Output

```shell
success: false, message: "Undefined array key "Id""
```

### 📸 Screenshots

_No response_

### ✓ Confirmation

- [x] I certify that this is an unmodified copy of OpenSourcePOS
- [x] I have searched existing issues to ensure this bug has not already been reported
- [x] I have provided all the information requested above

SAY-5:
Fixes #4593

When an uploaded item CSV has a header row that does not match the template (for example the `Id` column is missing or was renamed by a spreadsheet editor), `get_csv_file()` keys each row off the file's own headers, so the first `$row['Id']` access in `postImportCsvFile()` throws `Undefined array key "Id"`. The exception is caught and surfaced to the user as the raw PHP message, and nothing is imported.

This adds a header check right after the file is parsed. If the required item columns are not all present the import returns the existing `csv_import_nodata_wrongformat` message instead of crashing partway through the loop.

- `csv_import_has_required_item_headers()` in `importfile_helper.php` verifies the parsed rows expose the columns the loop dereferences.
- `postImportCsvFile()` calls it before iterating and bails out early on a mismatch.
- Added tests in `ItemsCsvImportTest` for the template header set, a file missing `Id`, and an empty file.

## Summary by CodeRabbit

* **Bug Fixes**
  * CSV item imports now validate required column headers immediately before processing, failing fast with a clear format error when headers are missing or invalid.
  * Required headers are now derived from the current downloadable CSV template, improving consistency and catching missing attribute columns.
* **Tests**
  * Updated and expanded CSV import header-validation tests to cover empty files and missing required headers (including attribute columns).

### DIFF
--- a/app/Controllers/Items.php
+++ b/app/Controllers/Items.php
@@ -988,14 +988,14 @@ class Items extends Secure_Controller
 
                     $failCodes = [];
                     $csvRows = get_csv_file($_FILES['file_path']['tmp_name']);
+                    $allowedStockLocations = $this->stock_location->get_allowed_locations();
+                    $attributeDefinitionNames    = $this->attribute->get_definition_names();
 
-                    if (!csv_import_has_required_item_headers($csvRows)) {
+                    if (!csvImportHasRequiredItemHeaders($csvRows, $allowedStockLocations, $attributeDefinitionNames)) {
                         return $this->response->setJSON(['success' => false, 'message' => lang('Items.csv_import_nodata_wrongformat')]);
                     }
 
                     $employeeId = $this->employee->get_logged_in_employee_info()->person_id;
-                    $allowedStockLocations = $this->stock_location->get_allowed_locations();
-                    $attributeDefinitionNames    = $this->attribute->get_definition_names();
 
                     unset($attributeDefinitionNames[NEW_ENTRY]);    // Removes the common_none_selected_text from the array
 
--- a/app/Controllers/Items.php
+++ b/app/Controllers/Items.php
@@ -988,6 +988,11 @@ class Items extends Secure_Controller
 
                     $failCodes = [];
                     $csvRows = get_csv_file($_FILES['file_path']['tmp_name']);
+
+                    if (!csv_import_has_required_item_headers($csvRows)) {
+                        return $this->response->setJSON(['success' => false, 'message' => lang('Items.csv_import_nodata_wrongformat')]);
+                    }
+
                     $employeeId = $this->employee->get_logged_in_employee_info()->person_id;
                     $allowedStockLocations = $this->stock_location->get_allowed_locations();
                     $attributeDefinitionNames    = $this->attribute->get_definition_names();
--- a/app/Helpers/importfile_helper.php
+++ b/app/Helpers/importfile_helper.php
@@ -100,21 +100,28 @@ function bom_exists(&$file_handle): bool
 }
 
 /**
- * Checks that the parsed item import rows expose the columns the import expects.
- * A file whose header row does not match the template (for example when the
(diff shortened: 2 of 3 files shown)


app/Controllers/Items.php
```diff
@@ -988,6 +988,11 @@ public function postImportCsvFile(): ResponseInterface
 
                     $failCodes = [];
                     $csvRows = get_csv_file($_FILES['file_path']['tmp_name']);
+
+                    if (!csv_import_has_required_item_headers($csvRows)) {
```
objecttothis: All new code and variable/function/class names must be PSR compliant.  Opening `{` needs to be on the same line as the if statement with a space between the `)` and `{`

app/Helpers/importfile_helper.php
```diff
@@ -98,3 +98,25 @@ function bom_exists(&$file_handle): bool
...
+/**
+ * Checks that the parsed item import rows expose the columns the import expects.
+ * A file whose header row does not match the template (for example when the
+ * "Id" column is missing or renamed) would otherwise trigger "Undefined array
+ * key" errors while building each item.
+ *
+ * @param array $csv_rows Rows returned by get_csv_file().
+ * @return bool True when the required columns are present.
+ */
+function csv_import_has_required_item_headers(array $csv_rows): bool
```
objecttothis: All new code and variable/function/class names must be PSR compliant. function name needs to be camelCase

app/Helpers/importfile_helper.php
```diff
@@ -98,3 +98,25 @@ function bom_exists(&$file_handle): bool
...
+{
+    if (empty($csv_rows)) {
+        return false;
+    }
+
+    $required_headers = ['Id', 'Barcode', 'Item Name', 'Category', 'Supplier ID', 'Cost Price', 'Unit Price', 'Tax 1 Name', 'Tax 1 Percent', 'Tax 2 Name', 'Tax 2 Percent', 'Reorder Level', 'Description', 'Allow Alt Description', 'Item has Serial Number', 'Image', 'HSN'];
```
objecttothis: The required headers isn't accounting for attributes which are automatically generated when the user downloads the template in the UI. You'll need to programmatically include that here.  I don't recommend you hard code the array here though.  It is less maintainable this way.  Instead, use the same mechanism that we are currently using to generate the template file to get the array for required headers.

objecttothis (changes requested):

objecttothis (approved):

PR #4600, opened 2026-07-22
bugfix(giftcard): correct return type and rename getGiftcardId method

objecttothis:
During the CI4 conversion bool was added as the return type but this creates a bug which causes the system to think the balance of the card is zero.

- Rename get_giftcard_id to getGiftcardId (PSR-12 camelCase)
- Fix return type from bool to int|false
- Replace loose == with strict === comparison
- Update all call sites in Giftcards controller and Sale model

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gift card verification and lookup reliability.
  * Gift card value and associated details now resolve correctly when a valid gift card is provided.
  * Invalid or unavailable gift card numbers continue to return the appropriate negative response.


objecttothis:
@coderabbitai review please.

jekkos (approved):

jekkos:
great thanks for catching this..  So we still check if return value is truthy as long as we are not using 0 as a giftcard id then all.is fine?

Edit: might be fine as id uses auto increment.

PR #4605, opened 2026-07-28
Plugin system: sale document view hooks and a webhook CSRF exemption

joshua1234511:
The plugin system can inject content into the sales receipt, but the invoice, quote and work order views have no hook point, so a plugin that delivers sale documents can only reach one of the four. Plugins also have no way to expose an inbound provider webhook, because a server-to-server callback carries no CSRF token and the global csrf filter has no plugin-safe exemption — leaving a core edit as the only option, which is what the plugin system exists to avoid.

This adds view:sales_invoice_buttons, view:sales_quote_buttons and view:sales_work_order_buttons — same position and same ['saleId' => ...] payload as the existing receipt hook — and exempts the plugins/*/webhook path from CSRF, with the plugin responsible for authenticating the caller itself. Documented in app/Plugins/README.md. Groundwork for the WhatsApp plugin, but not specific to it.

### DIFF
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -73,7 +73,16 @@ class Filters extends BaseFilters
     public array $globals = [
         'before' => [
             'honeypot',
-            'csrf' => ['except' => 'login|migrate'],
+            // Listed as an array rather than a pipe-delimited string so each entry is
+            // anchored independently. CI4 matches every entry as \A<pattern>\z, so the
+            // string form 'login|migrate' produced one pattern whose inner alternatives
+            // were unanchored and matched far more than intended.
+            //
+            // 'plugins/*/webhook' lets a plugin expose an inbound provider callback.
+            // A CSRF token cannot exist on a server-to-server delivery, so plugins must
+            // authenticate these requests themselves (e.g. verifying the provider's
+            // signature header against a shared secret).
+            'csrf' => ['except' => ['login', 'migrate', 'plugins/*/webhook']],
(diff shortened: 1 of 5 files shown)


app/Config/Filters.php
```diff
@@ -73,7 +73,16 @@ class Filters extends BaseFilters
     public array $globals = [
         'before' => [
             'honeypot',
-            'csrf' => ['except' => 'login|migrate'],
+            // Listed as an array rather than a pipe-delimited string so each entry is
```
objecttothis: This change is fine as long as an array is acceptable for CI4. No need to comment it though, so let's remove the comments.

app/Config/Filters.php
```diff
@@ -73,7 +73,16 @@ class Filters extends BaseFilters
...
+            // anchored independently. CI4 matches every entry as \A<pattern>\z, so the
+            // string form 'login|migrate' produced one pattern whose inner alternatives
+            // were unanchored and matched far more than intended.
+            //
+            // 'plugins/*/webhook' lets a plugin expose an inbound provider callback.
```
objecttothis: It makes sense why we would need to add the exception for `plugins/*/webhook` so as the folder notation is understood by CI4 then it's fine, but we don't need the comments to remain in the code so let's remove those.

objecttothis (changes requested):
Just the comments IMO.  Fix those and I'll approve/merge the code.

objecttothis (approved):

Issue #4603, opened 2026-07-25
[Bug]: DB Migration Not Working On New Install (Unauthorized)


e-kahiro:
### 🐛 Bug Description

Initial DB migration no longer works on new install. The error message says "invalid username/password" but there's no username/password form shown, only a green button that says "Initialize". Upon checking the network tab of developer's function, I see 401 unauthorized in the response header.

### 📋 Steps to Reproduce

1. Clone the repo
2. Cd into repo
3. Docker compose up
4. Access the console
5. Initialize

### ✅ Expected Behavior

DB is initialize and user is able to log in.

### 📦 OpenSourcePOS Version

OpenSourcePOS 3.4.2

### 🔧 PHP Version

PHP 8.4

### 🌐 Browser(s)

Firefox

### 🖥️ Server Operating System

Ubuntu 26.04

### 🗄️ Database

MariaDB 10.5

### 🌍 Web Server

Apache 2.4.68

### 📊 System Information Report

```text
I'm not able to access the system info due to not being able to pass the login initialization screen.
```

### 📜 Relevant Log Output

```shell

```

### 📸 Screenshots

<img width="1042" height="610" alt="Image" src="https://github.com/user-attachments/assets/8310078c-e8a2-426d-a224-a59e76d29797" />

### ✓ Confirmation

- [x] I certify that this is an unmodified copy of OpenSourcePOS
- [x] I have searched existing issues to ensure this bug has not already been reported
- [x] I have provided all the information requested above

FroOkk:
I can confirm the same issue on a new Docker install. Clearing MySQL and OSPOS containers does not help. MySQL never emits any new logs when a migration is attempted. But OSPOS emits a 401 error in the logs.

jekkos:
Thanks for the bug report. I will look into this and make login form visible at all times.

jekkos:
@objecttothis for 3.4.2 I would like to close any remaining open blockers/ bugs. Let's wait with merging new features until these are sorted. This is certainly a blocker, maybe I can find some time this week to look into it.

Also #4556 should be validated
And then I need to check what security issues are still open and do the required housekeeping on those

objecttothis:
I'm confused about this issue. 3.4.2 hasn't been released yet, has it? I believe I fixed this issue and it was merged into the master. Is he using 3.4.1 maybe or did he download the master and build it?

[...]

PR #4609, opened 2026-07-30
fix(login): skip auth validation on new install to allow migration

objecttothis:
On fresh installs with no DB version, login validation would fail before migrations could run. Check MY_Migration::getCurrentVersion() and bypass credential check when no version exists.

closing #4603 4603

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the installation/migration flow to more reliably detect new installs versus existing ones, ensuring username validation runs only when appropriate.
  * Improved behavior when the current migration version can’t be determined, returning a clear error response instead of continuing with an incorrect assumption.
  * Refined version detection to properly distinguish an “unknown/unavailable” state from an “empty” database state.

### DIFF
--- a/app/Controllers/Login.php
+++ b/app/Controllers/Login.php
@@ -18,9 +18,9 @@ class Login extends BaseController
     public Model $employee;
 
     /**
-     * @return RedirectResponse|string
+     * @return RedirectResponse|ResponseInterface|string
      */
-    public function index(): string|RedirectResponse
+    public function index(): string|RedirectResponse|ResponseInterface
     {
         $this->employee = model(Employee::class);
         if (!$this->employee->is_logged_in()) {
@@ -33,13 +33,24 @@ class Login extends BaseController
 
             $migration->migrateToCI4();
 
+            $currentVersion = MY_Migration::getCurrentVersion();
+
+            if ($currentVersion === null) {
+                return $this->response->setJSON([
+                    'success' => false,
+                    'message' => lang('Login.migration_error_connection')
+                ])->setStatusCode(503);
+            }
+
+            $latestVersion = $migration->getLatestMigration();
+
             $validation = Services::validation();
 
             $data = [
                 'hasErrors'       => false,
-                'isNewInstall'   => MY_Migration::getCurrentVersion() === 0,
-                'isLatest'        => $migration->isLatest(),
-                'latestVersion'   => $migration->getLatestMigration(),
+                'isNewInstall'   => $currentVersion === 0,
+                'isLatest'        => $latestVersion === $currentVersion,
+                'latestVersion'   => $latestVersion,
                 'gcaptchaEnabled' => $gcaptchaEnabled,
                 'config'           => $config,
                 'validation'       => $validation
--- a/app/Controllers/Login.php
+++ b/app/Controllers/Login.php
@@ -37,7 +37,7 @@ class Login extends BaseController
 
             $data = [
                 'hasErrors'       => false,
-                'isNewInstall'   => !(MY_Migration::getCurrentVersion()),
+                'isNewInstall'   => MY_Migration::getCurrentVersion() === 0,
                 'isLatest'        => $migration->isLatest(),
                 'latestVersion'   => $migration->getLatestMigration(),
                 'gcaptchaEnabled' => $gcaptchaEnabled,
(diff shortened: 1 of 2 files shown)


app/Controllers/Login.php
```diff
@@ -76,19 +76,23 @@ public function index(): string|RedirectResponse
...
-        ];
-
-        if (!$this->validate($rules, $messages)) {
-            return $this->response->setJSON([
-                'success' => false,
-                'message' => lang('Login.invalid_username_and_password')
-            ])->setStatusCode(401);
+        $isNewInstall = !MY_Migration::getCurrentVersion();
+
+        if (!$isNewInstall) {
```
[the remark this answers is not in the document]
objecttothis: Fixed. Re-review please.

app/Controllers/Login.php
```diff
@@ -33,13 +33,24 @@ public function index(): string|RedirectResponse
 
             $migration->migrateToCI4();
 
+            $currentVersion = MY_Migration::getCurrentVersion();
+
+            if ($currentVersion === null) {
```
jekkos, line 38: A new install can also end up in this path when the table does not exist yet. In that case a Mysql exception will be thrown .

jekkos: So I would like to check if this then actually fixes the issue. Fresh installs need to start from no tables at all

jekkos: Just thought that if having to enter login and pwd at all times simplifies the logic then I would prefer to go with that..but lets see

app/Controllers/Login.php
```diff
@@ -33,13 +33,24 @@ public function index(): string|RedirectResponse
...
+                return $this->response->setJSON([
+                    'success' => false,
+                    'message' => lang('Login.migration_error_connection')
+                ])->setStatusCode(503);
+            }
```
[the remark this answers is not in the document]
objecttothis: Verify review finding: Login::index() DB check ordering
Context
Review finding claims Login::index() (app/Controllers/Login.php:36-43) should:

Check DB availability before calling migrateToCI4(), returning the 503 migration_error_connection response when unavailable.
Re-read the version via MY_Migration::getCurrentVersion() after migrateToCI4() before the null check, since conversion can change it.
Verification against current code
Part 2 — already satisfied. Current code order in index():

Line 34: $migration->migrateToCI4();
Line 36: $currentVersion = MY_Migration::getCurrentVersion();
Lines 38-43: null check → 503.
The version IS read after conversion, then null-checked. Exactly what the finding asks for.

Part 1 — behaviorally redundant. When the DB is unavailable:

migrateToCI4() (app/Libraries/MY_Migration.php:60-66) calls ci3MigrationsExists(), which wraps its DB access in try/catch (lines 75-85) and returns false on any Exception (including mysqli_sql_exception). So migrateToCI4() is a safe no-op with the DB down — no throw, no side effects.
The MY_Migration constructor (MigrationRunner::__construct, vendor line 156) uses db_connect(), which is lazy — no connection attempt until first query, so no throw before the 503 path either.
Control then reaches line 36, getCurrentVersion() returns null, and the existing 503 is returned.
Adding a pre-check before migrateToCI4() would only guard a call that is already exception-safe and side-effect-free when the DB is down, and would add a redundant second getCurrentVersion() call in the normal path. The current order (convert → read version → null-check) is also the correct order, since reading before conversion could report a version the conversion is about to rewrite.

Decision
Skip — no code change. Finding is stale/invalid against current code:

The "re-read after conversion" behavior it requests already exists (Login.php:34-43).
The "check availability first" reorder changes no observable behavior; DB-unavailable already yields the 503 via the post-conversion null check.
Validation
None needed (no change). Existing behavior confirmed by code inspection of Login.php, MY_Migration.php, and vendor MigrationRunner constructor.

app/Controllers/Login.php
objecttothis: > A new install can also end up in this path when the table does not exist yet. In that case a Mysql exception will be thrown .

I tested this by creating an empty database (no tables).  That yields:
image

Clicking Initialize yields
image

Is there a scenario I'm missing?

objecttothis: > Just thought that if having to enter login and pwd at all times simplifies the logic then I would prefer to go with that..but lets see

I don't think it would all that much.  When the app_config table doesn't exist, the username/password is irrelevant, which means you still have the logic of checking to see if the table exists regardless of whether you get a username and password from the user or not.  The only logic you save is the logic for whether to prompt the user for the username and password.

objecttothis: > So I would like to check if this then actually fixes the issue. Fresh installs need to start from no tables at all

Correct.  Fresh installs have no tables, but the database must exist.

jekkos: Looks great

jekkos (approved):

jekkos:
Works  now on dev, thanks a lot @objecttothis

PR #4610, opened 2026-07-30
chore(deps): upgrade dompdf from v2.0.8 to v3.1.6

objecttothis:
- Update dompdf/dompdf constraint from ^2.0.3 to ^3.1.6
- Dependency names changed: phenx/php-font-lib and phenx/php-svg-lib replaced by dompdf/php-font-lib and dompdf/php-svg-lib ^1.0.0
- Remove ThirdParty/dompdf path from Autoload.php psr4 namespace map
- Update composer.lock with new content hash and package references

Replaces #4604 as compatibility changes were necessary.

## Summary by CodeRabbit

* **Improvements**
  * Updated the PDF generation component to a newer version for improved compatibility and reliability.
  * Streamlined application configuration to support the updated PDF functionality.

### DIFF
--- a/app/Config/Autoload.php
+++ b/app/Config/Autoload.php
@@ -39,8 +39,7 @@ class Autoload extends AutoloadConfig
      */
     public $psr4 = [
         APP_NAMESPACE => APPPATH,
-        'Config'      => APPPATH . 'Config',
-        'dompdf'      => APPPATH . 'ThirdParty/dompdf/src'
+        'Config'      => APPPATH . 'Config'
     ];
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "php": "^8.2",
         "ext-intl": "*",
         "codeigniter4/framework": "4.7.2",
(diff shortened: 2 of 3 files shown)


objecttothis:
@jekkos I tested this and it also passes all checks.  I'm off until Tuesday so if it looks good then approve and merge it.

app/Config/Autoload.php
```diff
@@ -39,8 +39,7 @@ class Autoload extends AutoloadConfig
      */
     public $psr4 = [
         APP_NAMESPACE => APPPATH,
-        'Config'      => APPPATH . 'Config',
-        'dompdf'      => APPPATH . 'ThirdParty/dompdf/src'
```
jekkos, line 43: Autoload is not required anymore?

jekkos (approved):
Just one question

app/Config/Autoload.php
objecttothis: The folder doesn't exist anymore.

PR #4620, opened 2026-08-05
fix(config): validate theme param to prevent XSS via invalid theme

objecttothis:
Add validation rule for theme field before batch save, rejecting requests with unrecognized theme values. Add test coverage for theme validation in postSaveGeneral.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent invalid or unsafe theme selections from being saved.
  * Theme-related errors are now clearly returned when configuration updates fail.
  * Improved security when displaying the selected login-page theme.
  * Non-theme directories can no longer be selected as application themes.
  * Valid and empty theme selections continue to save successfully.
* **Tests**
  * Added coverage for valid, empty, unknown, unsafe, and unsupported theme values.

### DIFF
--- a/app/Views/login.php
+++ b/app/Views/login.php
@@ -32,7 +32,7 @@ $request = Services::request();
         ? 'flatly'
         : $config['theme']);
     ?>
-    <link rel="stylesheet" href="resources/bootswatch5/<?= "$theme" ?>/bootstrap.min.css">
+    <link rel="stylesheet" href="resources/bootswatch5/<?= esc($theme, 'attr') ?>/bootstrap.min.css">
     <link rel="stylesheet" href="css/login.css">
     <meta name="theme-color" content="#2c3e50">
 </head>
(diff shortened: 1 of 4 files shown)


jekkos (approved):

app/Views/login.php
```diff
@@ -32,7 +32,7 @@
         ? 'flatly'
         : $config['theme']);
     ?>
-    <link rel="stylesheet" href="resources/bootswatch5/<?= "$theme" ?>/bootstrap.min.css">
+    <link rel="stylesheet" href="resources/bootswatch5/<?= esc($theme, 'attr') ?>/bootstrap.min.css">
```
jekkos, line 35: this was probably already enough to fix it

jekkos (approved):
Seems we have two PRs for this issue, the other one does not have the testcases or maybe I missed it.

objecttothis:
> Seems we have two PRs for this issue, the other one does not have the testcases or maybe I missed it.

I did a search in the PR's but couldn't find another PR.  Which PR is it?

PR #4625, opened 2026-08-06
fix(sales): escape quote number in email template to prevent XSS

objecttothis:
Wrap $quote_number output with esc() in quote_email.php. Raw interpolation allowed injected HTML/JS via quote number field.

Hardens GHSA-hwjh-f7jr-846c

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security in quote emails by safely escaping quote numbers displayed in barcodes.

### DIFF
--- a/app/Views/sales/quote_email.php
+++ b/app/Views/sales/quote_email.php
@@ -149,7 +149,7 @@
                 <?= nl2br(esc($config['return_policy'])) ?>
             </div>
             <div id="barcode">
-                <?= $quote_number ?>
+                <?= esc($quote_number) ?>
             </div>
         </div>
     </div>


jekkos (approved):
Good that you caught this! Now we do not need to go this CVE process, unless you officially want to claim the credits for finding it :dagger:

objecttothis:
> Good that you caught this! Now we do not need to go this CVE process, unless you officially want to claim the credits for finding it :dagger: 

I'm new to being on this side of CVEs. I fixed this to further harden against the vulnerability in the advisory. It's not clear to me when we need to request CVEs and when the reporter does it. I learned today not to publish until just after the release that has the fix.

I'm mostly just trying to get these done so that we can release 3.4.2 since it's blocking me merging several features.

Issue #4626, opened 2026-08-06
PHPUnit feature tests broken against a clean test DB: session pattern, missing seed data, unowned test fixture bugs


objecttothis:
## Summary

While fixing GHSA-3xf6-8fmq-44wg (per-record Sales endpoint authorization), we found the existing PHPUnit feature-test suite only ever "passed" because it was accidentally running against a real, pre-populated development database (`ospos`, forced via `phpunit.xml.dist`'s `<env ... force="true">` overrides) rather than a clean, self-contained test database. Once `phpunit.xml.dist` is corrected to stop overriding `.env`'s `database.tests.database` (so tests target a real, empty `ospos_tests` DB, satisfying `TestDatabaseBootstrapSeeder`'s safety check), most of the existing controller tests fail.

Two real, separate bugs were found and fixed as part of that work:

1. `Employee::save_employee()` (`app/Models/Employee.php`) had `ENVIRONMENT != 'testing' && parent::save_value(...)` — a stray guard from an unrelated 2018 commit (`804942cff`, "Add warning if migration is needed") that silently made employee creation a no-op under `ENVIRONMENT=testing`. Removed.
2. `Secure_Controller::__construct()` used raw `header()` + `exit()` to deny access, which kills the entire PHPUnit process (not just the simulated request) under `CodeIgniter\Test\FeatureTestTrait`. Replaced with a `ResponsableInterface`-implementing exception (`app/Exceptions/AccessDeniedRedirectException.php`) that CI4's `CodeIgniter::run()` catches and converts to a real redirect response, safe under both real HTTP and `FeatureTestTrait`.

With both of those fixed, `tests/Controllers/SalesControllerTest.php` (new, added for the GHSA fix) is fully green. However, several **pre-existing** test files remain broken for reasons out of scope for that fix. Each item below includes the fix needed.

## 1. Session injection pattern is wrong across the suite

`EmployeesControllerTest.php`, `HomeTest.php`, and others use:
```php
$session = Services::session();
$session->destroy();
$session->set('person_id', $personId);
```
This does not work under `FeatureTestTrait` — `call()` resets `$_SESSION` from the trait's own `$this->session` property (see `vendor/codeigniter4/framework/system/Test/FeatureTestTrait.php`: `withSession()` at line ~111, and `populateGlobals()` at line ~413, which does `$_SESSION = $this->session`), discarding anything set via `Services::session()` directly.

**Fix:** replace the `Services::session()` block with:
```php
[...]

PR #4627, opened 2026-08-06
fix(sales): gate per-record endpoints behind reports_sales grant

objecttothis:
Cashiers holding only the base \"sales\" grant could reach per-sale endpoints (getRow, getEdit, postSave, getReceipt, getInvoice, getSendPdf, getSendReceipt) intended to be gated behind the reports_sales grant. getManage() already enforced this at the list level, but individual endpoints did not re-check it.

- Add explicit reports_sales grant check in Sales::getRow() and related per-record endpoints, returning a JSON 403-style response when unauthorized
- Rename $row_id/$sale_info/$data_row to camelCase per PSR-12 style
- Add AccessDeniedRedirectException for redirect-based access denial
- Remove testing-environment bypass in Employee::save_value()
- Simplify phpunit.xml.dist to pull test DB config from .env
- Add regression tests for GHSA-3xf6-8fmq-44wg covering unauthorized access to per-sale endpoints

Fixes: GHSA-3xf6-8fmq-44wg

## Summary by CodeRabbit

* **Security & Access**
  * Sales reports, receipts, invoices, editing, saving, and exports now enforce permissions with clear access-denied responses.
  * Unauthorized actions consistently redirect or return HTTP 403 errors.
  * Unauthenticated visitors are redirected to the login page.

* **Employee Management**
  * Administrators can disable password and permission changes through configuration.
  * Session handling and access-denied pages are more reliable.

* **Sales**
  * Sale prices now follow configured currency precision.
  * Changing item prices requires the appropriate permission; quantity changes remain available when permitted.

* **Localization & Reliability**
  * Added restriction messages across supported languages.
  * Missing email and currency settings now use safe defaults.


app/Controllers/Sales.php
```diff
@@ -1463,27 +1499,33 @@ public function restore(int $sale_id = NEW_ENTRY, bool $update_inventory = true)
      * @return ResponseInterface
      * @throws ReflectionException
      */
-    public function postSave(int $sale_id = NEW_ENTRY): ResponseInterface
+    public function postSave(int $saleId = NEW_ENTRY): ResponseInterface
```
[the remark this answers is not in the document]
objecttothis: Fixed

tests/Controllers/SalesControllerTest.php
```diff
@@ -0,0 +1,265 @@
...
+ */
+class SalesControllerTest extends CIUnitTestCase
+{
+    use DatabaseTestTrait;
+    use FeatureTestTrait;
+
+    protected $migrate     = true;
+    protected $migrateOnce = true;
+    protected $refresh     = false;
+    protected $namespace   = null;
```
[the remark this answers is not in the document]
objecttothis: Cannot do this because the parent types do not have types declared.  It would break the contract.

tests/Controllers/SalesControllerTest.php
```diff
@@ -0,0 +1,265 @@
...
+        // permission exists in the permissions table (see has_subpermissions()).
+        $grantsData = [
+            ['permission_id' => 'sales', 'menu_group' => 'home'],
+            ['permission_id' => 'sales_stock', 'menu_group' => 'home']
+        ];
+
+        $employeeModel = model(Employee::class);
+        $employeeModel->save_employee($personData, $employeeData, $grantsData, NEW_ENTRY);
+
+        return (int) $personData['person_id'];
```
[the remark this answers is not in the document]
objecttothis: Not a bug. Employee::save_employee() takes $person_data by reference — parent::save_value($person_data, $employee_id) sets $person_data['person_id'] internally, and that same value is then copied into $employee_data['person_id'] at Employee.php:141:

$employee_data['person_id'] = $employee_id = $person_data['person_id'];

So $personData['person_id'] and $employeeData['person_id'] are identical after the call — using one over the other changes nothing.

The actual gap was that the helpers didn't assert save_employee() returned true. Fixed that in both createCashierEmployee() and createReportsSalesEmployee() by wrapping the call in $this->assertTrue(...), and kept the return value as $personData['person_id'] since it's already correct.

app/Controllers/Secure_Controller.php
```diff
@@ -40,17 +41,15 @@ public function __construct(string $module_id = '', ?string $submodule_id = null
         $validation = Services::validation();
 
         if (!$this->employee->is_logged_in()) {
-            header("Location:" . base_url('login'));
-            exit();
+            throw new AccessDeniedRedirectException(base_url('login'));
```
jekkos: @coderabbit can you check for consistency across other controllers in how this is handled?

app/Models/Employee.php
```diff
@@ -135,7 +135,7 @@ public function save_employee(array &$person_data, array &$employee_data, array
         // Run these queries as a transaction, we want to make sure we do all or nothing
         $this->db->transStart();
 
-        if (ENVIRONMENT != 'testing' && parent::save_value($person_data, $employee_id)) {
+        if (parent::save_value($person_data, $employee_id)) {
```
jekkos, line 146: The check for testing does prevent users from changing the password on demo server. As an alternative we might want to use a env variable that is set specifically for this instead. wdyt

app/Exceptions/AccessDeniedRedirectException.php
```diff
@@ -0,0 +1,20 @@
...
+
+use CodeIgniter\HTTP\ResponsableInterface;
+use CodeIgniter\HTTP\ResponseInterface;
+use RuntimeException;
+
+class AccessDeniedRedirectException extends RuntimeException implements ResponsableInterface
+{
+    public function __construct(private readonly string $redirectUrl)
+    {
+        parent::__construct('Access denied, redirecting to: ' . $redirectUrl);
```
jekkos: I would not use custom exception for this, to me it does not make a lot of sense to have a separate type as it does not really add any additional behavior. @coderabbit what is the usual codeigniter practice in case of auth failure?

app/Models/Employee.php
objecttothis: I think using something like a custom `bool app.Config.isDemoMode` in the .env could be a more intuitive way to prevent access to things we want to prevent like changing the password. Right now phpunit tests are failing because we have some inconsistent test setup and mocking. I've been working on unifying it but we have several of those changes across a few different PRs. I was hoping to get them merged then see where we stand on testing and do a single PR to fix testing all together from there.

app/Controllers/Secure_Controller.php
objecttothis: @jekkos this will need to be something we account for. I'm traveling and will be back in the office on Thursday. Feel free to pick this up if you'd like to resolve it before then.

jekkos: @coderabbit what is code igniter standard practice if access is not allowed for a controller function?

app/Models/Employee.php
jekkos: **@coderabbit** can you check if we already use an env variable elsewhere to see if we are on demoo demo ? i believe this is already in use and it should be made consistent

app/Controllers/Secure_Controller.php
jekkos: @objecttothis would it be an option to use RedirectException instead?

app/Models/Employee.php
jekkos: Thanks for clarification @coderabbit
We might want to check this env variable here as in demo we do not want user to be able to change module grants

app/Controllers/Secure_Controller.php
objecttothis: The logic for doing so makes sense to me. I can do it on Thursday when I'm back in the office or if you want to make the change before, feel free to do it.

app/Exceptions/AccessDeniedRedirectException.php
objecttothis: I unified the responses of all three now.

Secure_Controller.php:48 — module/submodule grant check
Reports.php:89 — report submodule grant check
Employees.php:83 — canModifyEmployee check
I used a RedirectException-style (redirect via exception), but keeping the custom AccessDeniedRedirectException rather than switching to CI4's built-in CodeIgniter\HTTP\Exceptions\RedirectException.

Reason: this is called from inside __construct() methods, which can't return a ResponseInterface directly — throwing is the only option there. Our custom exception implements ResponsableInterface, so CI4's core automatically calls getResponse() on it and redirects — same net effect as the built-in RedirectException, but it's our own class so we're not dependent on framework internals matching exactly, and it was already in place/tested before this work started. CodeRabbit's original suggestion to swap to the framework's RedirectException was considered and rejected — no functional benefit, would just be a rename.

So: still need AccessDeniedRedirectException unless someone can think of a way to implement it using RedirectException instead.  @jekkos or @coderabbitai, ideas?

app/Controllers/Secure_Controller.php
objecttothis: I implemented the before filter for login gating.

> @objecttothis would it be an option to use RedirectException instead?

See my response below about the RedirectException for grants gating.  I'm open to ideas.

app/Exceptions/AccessDeniedRedirectException.php
objecttothis: Replaced AccessDeniedRedirectException with RedirectException.

app/Models/Employee.php
objecttothis: @jekkos I updated the .env.example to include DISALLOW_PASSWORD_CHANGE and DISALLOW_GRANT_CHANGE. We need to update the deploy scripting for the demo to change these to true when renaming the .env.example to .env

objecttothis:
@coderabbitai review

app/Controllers/Employees.php
```diff
@@ -188,6 +196,24 @@ public function postSave(int $employee_id = NEW_ENTRY): ResponseInterface
...
+            sort($submitted_grant_ids);
+
+            if ($current_grant_ids !== $submitted_grant_ids) {
+                return $this->response->setJSON([
+                    'success' => false,
+                    'message' => lang('Employees.error_grant_change_disallowed'),
+                    'id'      => $employee_id
+                ]);
+            }
+        }
```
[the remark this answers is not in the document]
objecttothis: Fixed

app/Database/Migrations/20240630000001_fix_keys_for_db_upgrade.php
```diff
@@ -56,11 +56,11 @@ public function down(): void
...
         if ($foreignKeyExists) {
-            $this->db->query('ALTER TABLE ' . $this->db->prefixTable('sales_items_taxes') . ' DROP CONSTRAINT ospos_sales_items_taxes_ibfk_1');
+            $this->db->query('ALTER TABLE ' . $this->db->prefixTable('sales_items_taxes') . ' DROP FOREIGN KEY ospos_sales_items_taxes_ibfk_1');
         }
 
         $this->db->query('ALTER TABLE ' . $this->db->prefixTable('sales_items_taxes')
-            . ' ADD CONSTRAINT ospos_sales_items_taxes_ibfk_1 FOREIGN KEY (sale_id) '
-            . ' REFERENCES ' . $this->db->prefixTable('sales_items') . ' (sale_id)');
+            . ' ADD CONSTRAINT ospos_sales_items_taxes_ibfk_1 FOREIGN KEY (sale_id, item_id, line) '
+            . ' REFERENCES ' . $this->db->prefixTable('sales_items') . ' (sale_id, item_id, line)');
```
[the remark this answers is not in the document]
objecttothis: Fixed

app/Models/Employee.php
```diff
@@ -375,11 +383,13 @@ public function login(string $username, string $password): bool
             // Compare passwords depending on the hash version
             if ($row->hash_version === '1' && $row->password === md5($password)) {
                 $builder->where('person_id', $row->person_id);
+                $this->session->regenerate(true);
                 $this->session->set('person_id', $row->person_id);
                 $password_hash = password_hash($password, PASSWORD_DEFAULT);
 
                 return $builder->update(['hash_version' => 2, 'password' => $password_hash]);
```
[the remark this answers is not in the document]
objecttothis: fixed.

tests/Controllers/ItemsCsvImportTest.php
```diff
@@ -11,17 +11,16 @@
...
 {
     use DatabaseTestTrait;
 
     protected $migrate = true;
     protected $migrateOnce = true;
-    protected $seed = '';
+    protected $seed = 'TestDatabaseBootstrapSeeder';
     protected $seedOnce = true;
-    protected $refresh = true;
+    protected $refresh = false;
```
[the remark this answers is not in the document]
objecttothis: Fixed

tests/Models/EmployeeTest.php
```diff
@@ -12,7 +12,7 @@ class EmployeeTest extends CIUnitTestCase
 
     protected $migrate     = true;
     protected $migrateOnce = true;
-    protected $refresh     = true;
+    protected $refresh     = false;
```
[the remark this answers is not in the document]
objecttothis: Outside the scope of this PR.  It needs to be added to #4626

tests/Models/EmployeeTest.php
```diff
@@ -161,9 +161,114 @@ public function testHasGrantReturnsTrueForActualGrant(): void
...
+            'password'      => password_hash('password123', PASSWORD_DEFAULT),
+            'hash_version'  => 2,
+            'language_code' => 'en',
+            'language'      => 'english'
+        ];
+
+        $employeeModel = model(Employee::class);
+        $employeeModel->save_employee($personData, $employeeData, $grantsData, NEW_ENTRY);
+
+        return $personData['person_id'];
```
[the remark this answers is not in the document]
objecttothis: fixed

objecttothis:
@jekkos I've addressed all the @CodeRabbitAI comments that were in scope. Can you take a look at this PR?

app/Language/az/Employees.php
```diff
@@ -1,47 +1,49 @@
...
-    "system_language"              => "Sistem Dili",
-    "unsuccessful_change_password" => "Şifrə dəyişməsi uğursuz oldu.",
-    "update"                       => "İşçini yeniləyin",
-    "username"                     => "İstifadəçi Adı",
-    "username_duplicate"           => "",
-    "username_minlength"           => "İstifadəçi adı ən azı 5 simvol olmalıdır.",
-    "username_required"            => "İstifadəçi adı tələb olunan sahədir.",
+    'administrator'                    => '',
+    'basic_information'                => 'Məlumat',
+    'cannot_be_deleted'                => 'Seçilmiş əməkdaş (lar) silinə bilməz, bir və ya birdən çox əməkdaş satışlar edib, əks halda siz öz heasabınızı silməyə çalışırsınız.',
```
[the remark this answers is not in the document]
objecttothis: Fixed

app/Language/en-GB/Employees.php
```diff
@@ -1,45 +1,49 @@
...
-    "system_language"              => "System Language",
-    "unsuccessful_change_password" => "Password change failed.",
-    "update"                       => "Update Employee",
-    "username"                     => "Username",
-    "username_duplicate"           => "Employee username is already in use. Please choose another one.",
-    "username_minlength"           => "Username must be at least 5 characters.",
-    "username_required"            => "Username is a required field.",
+    'administrator'                    => '',
+    'basic_information'                => 'Information',
+    'cannot_be_deleted'                => 'Unable to delete selected Employee(s), one or more of the has processed sales or you are trying to delete your account.',
```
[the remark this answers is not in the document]
objecttothis: Fixed

app/Language/sv/Employees.php
```diff
@@ -1,47 +1,49 @@
...
-    "system_language"              => "System Språk",
-    "unsuccessful_change_password" => "Lösenordsbyte misslyckades.",
-    "update"                       => "Uppdatera anställd",
-    "username"                     => "Användarnamn",
-    "username_duplicate"           => "Användarnamnet för den anställda används redan. Vänligen välj ett annat.",
-    "username_minlength"           => "Användarnamnet måste vara minst 5 tecken långt.",
-    "username_required"            => "Användarnamnet är ett obligatoriskt fält.",
+    'administrator'                    => '',
+    'basic_information'                => 'Information',
+    'cannot_be_deleted'                => 'Det går inte att radera valda anställda, en eller flera av de behandlade försäljningarna eller du försöker radera ditt konto.',
```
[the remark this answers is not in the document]
objecttothis: Fixed.

app/Language/tr/Employees.php
```diff
@@ -1,47 +1,49 @@
...
-    "system_language"              => "Sistem dili",
-    "unsuccessful_change_password" => "Parola değişikliği başarısız oldu.",
-    "update"                       => "Personeli Güncelle",
-    "username"                     => "Kullandı Adı",
-    "username_duplicate"           => "",
-    "username_minlength"           => "Kullanıcı Adı en az 5 karakter olmalıdır.",
-    "username_required"            => "Kullanıcı Adı zorunlu alandır.",
+    'administrator'                    => '',
+    'basic_information'                => 'Personel Temel Bilgileri',
+    'cannot_be_deleted'                => 'Seçili personel silinemedi, personellerin satışları var yada kendinizi silmeye çalışıyorsunuz.',
```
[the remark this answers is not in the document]
objecttothis: Fixed.

app/Language/tr/Employees.php
```diff
@@ -1,47 +1,49 @@
...
+    'repeat_password'                  => 'Parola Yeniden',
+    'subpermission_required'           => 'Her bir birim için en az bir izin ekle.',
+    'successful_adding'                => 'Personel eklendi.',
+    'successful_change_password'       => 'Parola değişikliği başarılı.',
+    'successful_deleted'               => 'Silme başarılı',
+    'successful_updating'              => 'Personel güncellendi',
+    'system_language'                  => 'Sistem dili',
+    'unsuccessful_change_password'     => 'Parola değişikliği başarısız oldu.',
+    'update'                           => 'Personeli Güncelle',
+    'username'                         => 'Kullandı Adı',
```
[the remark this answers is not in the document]
objecttothis: Fixed

tests/Controllers/EmployeesControllerTest.php
```diff
@@ -19,9 +19,22 @@ class EmployeesControllerTest extends CIUnitTestCase
     protected $refresh     = false;
     protected $namespace   = null;
 
+    protected $priorDisallowGrantChange;
+
     protected function setUp(): void
     {
         parent::setUp();
+        $this->priorDisallowGrantChange = getenv('DISALLOW_GRANT_CHANGE');
+    }
```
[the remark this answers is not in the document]
objecttothis: Fixed.

tests/Controllers/ItemsCsvImportTest.php
```diff
@@ -11,19 +12,19 @@
...
 
     protected $migrate = true;
     protected $migrateOnce = true;
-    protected $seed = '';
     protected $seedOnce = true;
-    protected $refresh = true;
+    protected $refresh = false;
     protected $namespace = null;
 
+    private static $doneBootstrap = false;
```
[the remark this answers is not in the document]
objecttothis: Fixed.

app/Language/ar-EG/Employees.php
```diff
@@ -1,47 +1,49 @@
...
-    "successful_deleted"           => "لقد تم حذف الموظف بنجاح",
-    "successful_updating"          => "لقد تم تحديث بيانات الموظف بنجاح",
-    "system_language"              => "لغة النظام",
-    "unsuccessful_change_password" => "فشل في تغيير كلمة المرور.",
-    "update"                       => "تحديث بيانات موظف",
-    "username"                     => "اسم المستخدم",
-    "username_duplicate"           => "حساب المحدد هو موجود في قاعدة البيانات. نرجوا استخدام اسم حساب مختلف.",
-    "username_minlength"           => "اسم المستخدم يجب أن يكون 5 حروف على الأقل.",
-    "username_required"            => "اسم المستخدم مطلوب.",
+    'administrator'                    => '',
```
jekkos, line 4: changing to single quotes here?

tests/Models/EmployeeTest.php
```diff
@@ -34,12 +34,12 @@ public function testIsAdminReturnsTrueForEmployeeWithAllPermissions(): void
...
             ->onlyMethods(['has_grant'])
             ->getMock();
-        
+
         $employeeModel->method('has_grant')
             ->willReturn(true);
-        
+
         $result = $employeeModel->isAdmin(2);
-
```
jekkos, line 42: was there a new formatting rule applied here?

jekkos (commented):
this PR is really huge, hard for me to see what are the actual changes. There are 51 commits but many of them seem to be about changes unrelated to the PR title. Was this the intention?

app/Language/ar-EG/Employees.php
jekkos: I believe it might be an issue in some translations as ' can be used in the strings themselves, so that would- break the file parsing

jekkos (approved):

app/Controllers/Sales.php
```diff
@@ -1363,13 +1382,19 @@ private function reload(array $data = []): ResponseInterface|string
...
      * @return string
      * @noinspection PhpUnused
      */
-    public function getReceipt(int $sale_id): string
+    public function getReceipt(int $saleId): string|ResponseInterface
     {
-        $data = $this->_load_sale_data($sale_id);
+        $personId = $this->session->get('person_id');
+
+        if (!$this->employee->has_grant('reports_sales', $personId)) {
```
jekkos, line 1393: @coderabbit any CI4 recommendation on how to handle authorization in controller methods?

app/Language/ar-EG/Employees.php
jekkos: I have approved it , as I understood it's mainly a fix for a reported vulnerability so let's include this in the new minor release

objecttothis: > I believe it might be an issue in some translations as ' can be used in the strings themselves, so that would- break the file parsing

`'` appearing inside the phrase are always escaped with `\`.  This prevents problems with the parser.  Ideal would be if weblate automatically inserted escaping when updating a phrase so that we didn't need to depend on translators keeping/inserting a `\` when they use a `'` in the phrase.  @coderabbitai does weblate automatically do this?  If not, can it be setup to do this?

tests/Models/EmployeeTest.php
objecttothis: I don't think it's a new rule.  The linter picked up empty lines with white space and removed the white space.

app/Language/ar-EG/Employees.php
objecttothis: @jekkos, according to @coderabbitai we should be good and this should be seamless.

objecttothis:
> this PR is really huge, hard for me to see what are the actual changes. There are 51 commits but many of them seem to be about changes unrelated to the PR title. Was this the intention?

Yeah, it is large. There are a couple of reasons:
- In language files I've been migrating them as I touch them to single quotes, but this is because they can largely be skimmed in reviews. Each language file adds 46 or so files to the changelist.
- Any time you touch code, when @coderabbitai does a review it points out things that are problems with the code even if they are out of scope for the change we are making. For things that it points out related to security, like "you missed escaping this output" I've been fixing them inside the PR rather than making separate PRs for them.
- @coderabbitai makes some comments which it admits are outside of the code you touched.  In this PR I went through and fixed them but that ballooned the PR.  These types of comments are good candidates for telling @coderabbitai to make a separate issue for unless they are directly related to the code we touched and the issue we are trying to solve.
- I've been adding Unit Tests as I go for code that I touch.  This increases the size of the PR as well.  Especially for a few PRs because unit tests have been broken.  I've been following a pattern in unit test files for setup/teardown that is set in ItemsCsvImportTest which does not depend on a method which calls regress() and therefore calls down() on every migration.  Those down() functions are not all implemented and breaks the teardown.  We have an outstanding issue for fixing unit tests and we can use a different system than what I have been implementing in that fix, but this is working for now.

app/Controllers/Sales.php
objecttothis: @coderabbitai create a github issue to resolve this. Describe in the issue the problem and proposed resolution.  Do not assign me.

jekkos:
Thanks for the hard work nonetheless @objecttothis 
I also spent quite some time wirh the tests some weeks ago as I suspected some of them were not running.. Its great to see they are actually failing now. I will try to pick up again some work as well now very soon.

objecttothis:
Yeah, I just needed to get tests running on the changes I was making, so I only touched testing files related to those branches. When you look at tests that are failing, see if they are using the same setup/teardown as ItemsCsvImportTest is using. All test files I fixed are using that method. One issue is that the repo had a phpunit.xml that was overriding the db connection values in .env and we need to use the .env as the single source of truth if possible. The other major issue was that the built in CI test variables were causing it to try to reset the database for the next test by running regress() which calls each migration's down() function. Some are not implemented or incorrectly implemented so short of fixing those, we will have to use another method for teardown. Honestly, I prefer the CI native way but couldn't spend the time on it. Feel free to have add it, but let me know how you fix it so that I can write unit tests for these security fixes. I really want to get 3.4.2 out, so I'm going to continue working the blocking issues/prs

PR #4630, opened 2026-08-06
fix(security): sanitize filenames and escape logo path in config

objecttothis:
- Sanitize uploaded filename in Config.php via preg_replace, strip chars outside [a-zA-Z0-9_-] before storing raw_name
- Escape $logo_src with esc(..., 'attr') in info_config.php view to prevent XSS via crafted logo path/filename

Prevents stored XSS and path traversal from unsanitized filenames used in config uploads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security when uploading company logos by sanitizing filenames.
  * Protected company-logo previews from unsafe characters in image paths.

### DIFF
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -352,7 +352,7 @@ class Config extends Secure_Controller
 
         $file_info = [
             'orig_name' => $filename,
-            'raw_name'  => $info['filename'],
+            'raw_name'  => preg_replace('/[^a-zA-Z0-9_\-]/', '_', $info['filename']),
             'file_ext'  => $file->guessExtension()
         ];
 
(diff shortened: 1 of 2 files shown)


app/Controllers/Config.php
```diff
@@ -352,7 +352,7 @@ private function upload_logo(): array
 
         $file_info = [
             'orig_name' => $filename,
-            'raw_name'  => $info['filename'],
+            'raw_name'  => preg_replace('/[^a-zA-Z0-9_\-]/', '_', $info['filename']),
```
jekkos: CI has a built in function to do this sanitization: 

https://codeigniter.com/userguide3/libraries/security.html#CI_Security::sanitize_filename

jekkos (changes requested):

app/Controllers/Config.php
objecttothis: That's the CI3 version.  I did find the CI4 version.  Done.

jekkos (approved):

PR #4631, opened 2026-08-06
fix(sales): enforce server-side authorization for price changes

objecttothis:
Client-side \"change_price\" flag is UI-only, not trustworthy. Compare submitted price against current cart price server-side and require sales_change_price grant when they differ, else reject with not_authorized error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added server-side authorization checks for changing item prices during sales edits.
  * Unauthorized price changes are rejected, and the item remains unchanged.

### DIFF
--- a/app/Controllers/Sales.php
+++ b/app/Controllers/Sales.php
@@ -647,6 +647,14 @@ class Sales extends Secure_Controller
                 return $this->reload($data);
             }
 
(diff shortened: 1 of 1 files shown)


jekkos (approved):
Thanks for the fix!

PR #4642, opened 2026-08-17
Prompt user when attribute values cannot be converted during type change

### LINKED ISSUE #4632: Prompt user when attribute values cannot be converted during type change
## Problem

When editing an attribute definition and changing its type (e.g. TEXT → DECIMAL or TEXT → DATE), existing `attribute_value`s that can't convert to the new type cause the whole save to fail with no useful feedback to the user.

Previously this also crashed with `ErrorException: Array to string conversion` in `Attribute::check_data_validity()` (fixed separately). With that crash fixed, the underlying UX gap remains: the save silently rolls back and the user only sees a generic "cannot be updated" message, with no indication of *which* values are the problem or any way to proceed.

## Current behavior

- `Attribute::check_data_validity()` (`app/Models/Attribute.php`, ~line 376) checks each existing value against the new type and logs an error for any that fail, but doesn't surface this to the caller beyond a boolean.
- `Attribute::convert_definition_data()` (~line 438) returns `false` if validity check fails.
- `Attribute::saveDefinition()` (~line 529) rolls back the entire transaction when conversion fails.
- `Attributes::postSaveDefinition()` (`app/Controllers/Attributes.php`) returns a generic failure JSON response with no detail about affected values/items.

## Proposed behavior

When a type-changing save hits non-convertible values, show a modal explaining that some attribute values cannot be converted to the new type, with two options:

- **Continue** — drop/discard the non-convertible attribute values and proceed with the type change
- **Cancel** — abort the save (current behavior)

## Implementation notes

- Backend: `postSaveDefinition()` needs to surface which values (and/or affected items) are non-convertible in its JSON response instead of a generic failure, and support a "force"/"drop invalid values" flag that, when set, proceeds with the conversion while discarding the offending values.
- Frontend: the attribute definition save form (attributes/form view JS) needs a confirmation modal that triggers on this new response shape, and resubmits with the force flag on "Continue".

arsyadal:
## Summary
- `Attribute::check_data_validity()` now returns the list of non-convertible attribute values (and affected item IDs) instead of a bare bool.
- `Attribute::convert_definition_data()` / `saveDefinition()` accept a `$force` flag: without it, a conversion conflict is recorded (`getConversionConflicts()`) and the save fails as before; with it, the non-convertible values are dropped (via the existing sale/receiving-preserving link deletion) and the rest of the conversion proceeds.
- `Attributes::postSaveDefinition()` surfaces conflicts in the JSON response (`conflict: true`, affected `values`) instead of a generic failure message, and reads a new `force_type_conversion` POST flag.
- The attribute definition form JS now detects a conflict response, confirms with the user, and resubmits with `force_type_conversion=1` on Continue.

Closes #4632

## Test plan
- [ ] `php -l` on changed PHP files (done locally)
- [ ] Manually edit an attribute definition (TEXT → DATE/DECIMAL) with existing non-convertible values and confirm the prompt appears, Continue drops the bad values and completes the conversion, Cancel leaves everything unchanged
- [ ] Existing convertible-only type changes still succeed without any prompt

## Summary by CodeRabbit

* **New Features**
  * Added warnings when changing an attribute type would make existing values incompatible.
  * Displays affected values and items before proceeding.
  * Users can confirm the conversion and remove incompatible links automatically.

* **Bug Fixes**
  * Prevented duplicate form values during attribute submissions.
  * Preserved existing error handling for non-conversion-related save failures.


app/Controllers/Attributes.php
```diff
@@ -144,6 +144,7 @@ public function postSaveDefinition(int $definition_id = NO_DEFINITION_ID): Respo
         }
 
         $definition_name = $definition_data['definition_name'];
+        $force_type_conversion = $this->request->getPost('force_type_conversion') == '1';
```
objecttothis, line 147: All new code needs to have PSR compliant variable, function and class naming.

app/Language/en/Attributes.php
```diff
@@ -5,6 +5,7 @@
     "confirm_delete"                   => "Are you sure you want to delete the selected attribute(s)?",
     "confirm_restore"                  => "Are you sure you want to restore the selected attribute(s)?",
     "definition_cannot_be_deleted"     => "Could not delete selected attribute(s)",
+    "definition_conversion_conflict"   => "Some existing values for attribute {0} cannot be converted to the new type. Continue and discard those values?",
```
objecttothis, line 8: All new language strings need to use single quotes to avoid invoking the PHP parser even if the surrounding language strings use double quotes. single quotes that exist inside the translated phrase should be escaped with `\`.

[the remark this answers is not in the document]
objecttothis: This too.  You don't need to translate it in the other languages, but you need to add the language key and an empty `'',` value in all the other languages.

app/Models/Attribute.php
```diff
@@ -41,6 +41,25 @@ class Attribute extends Model
...
     public const SHOW_IN_RECEIVINGS = 4;
+
+    /**
+     * Attribute values that could not be converted to the new type during the most recent
+     * saveDefinition() call. Populated by convert_definition_data() when a type change is
+     * rejected because of non-convertible existing values.
+     *
+     * @var array
+     */
+    private array $conversion_conflicts = [];
```
objecttothis, line 52: PSR compliant naming for all new code.

app/Models/Attribute.php
```diff
@@ -41,6 +41,25 @@ class Attribute extends Model
     public const SHOW_IN_ITEMS = 1;    // TODO: These need to be moved to constants.php
     public const SHOW_IN_SALES = 2;
     public const SHOW_IN_RECEIVINGS = 4;
+
+    /**
+     * Attribute values that could not be converted to the new type during the most recent
```
objecttothis, line 46: No need for the docblock here.  When you rename, you might consider something like `$definitionConversionConflicts` to make it clear on its own what the variable is for.

app/Models/Attribute.php
```diff
@@ -41,6 +41,25 @@ class Attribute extends Model
...
+     * saveDefinition() call. Populated by convert_definition_data() when a type change is
+     * rejected because of non-convertible existing values.
+     *
+     * @var array
+     */
+    private array $conversion_conflicts = [];
+
+    /**
+     * Returns the attribute values (and affected items) that blocked the most recent type
```
objecttothis, line 55: Same here.  No docblocks needed.  You might rename it to `getDefinitionConversionConflicts()`

app/Models/Attribute.php
```diff
@@ -431,18 +459,41 @@ private function get_items_by_value(string $attribute_value, int $definition_id)
...
+    private function convert_definition_data(int $definition_id, string $from_type, string $to_type, bool $force = false): bool
     {
         $success = false;
 
         if ($from_type === TEXT) {
             if (in_array($to_type, [DATE, DECIMAL], true)) {
-                if ($this->check_data_validity($definition_id, $from_type, $to_type)) {
-                    $attributes_to_convert = $this->get_attributes_by_definition($definition_id);
-                    $success = $this->attribute_cleanup($attributes_to_convert, $definition_id, $to_type);
+                $invalid_values = $this->check_data_validity($definition_id, $from_type, $to_type);
```
objecttothis, line 472: In order to prevent this function from becoming a long procedural mess, let's refactor a function out of some of this block of code. 479-493 could be a function by itself.

app/Models/Attribute.php
```diff
@@ -523,9 +574,15 @@ private function checkbox_attribute_values(int $definition_id): array
 
     /**
      * Inserts or updates a definition
+     *
+     * @param bool $force When true, a type change proceeds even if some existing attribute_values
+     *                    cannot be converted to the new type; those values are dropped instead.
+     *                    Non-convertible values can be retrieved beforehand via getConversionConflicts().
      */
-    public function saveDefinition(array &$definitionData, int $definitionId = NO_DEFINITION_ID): bool
+    public function saveDefinition(array &$definitionData, int $definitionId = NO_DEFINITION_ID, bool $force = false): bool
```
objecttothis, line 582: @coderabbitai do any callers of this function omit the `$definitionId` parameter but include the `$force` parameter?

objecttothis (changes requested):

PR #4656, opened 2026-08-24
fix(security): handle special characters in `.env` key values and improve insertion logic

### LINKED ISSUE #4554: Somehow... Fresh install broke after bringing container down and back up.  Encryption key related.

objecttothis:
This was supposed to be merged into the CI3 conversion logic changed but somehow I missed pushing it.

- Escape backslashes and dollar signs in `applyEnvKeyReplacement` to prevent unintended value corruption.
- Ensure new keys are inserted after `encryption.key` for better organization and manageability.
- Add explicit cast to int to prevent wrong concatenation operator warning.

## Summary by CodeRabbit

- **New Features**
  - Added `env:provision` to create and maintain encryption and throttling keys during deployment.
  - Added automatic migration of legacy encrypted settings to the current encryption format, with verification and rollback on failure.
  - Added a dedicated persistent throttling key and improved environment-file configuration.

- **Security Improvements**
  - Passwords and API keys are no longer displayed in configuration forms; blank fields preserve existing values.
  - Configuration forms now indicate when credentials are already set.

- **Bug Fixes**
  - Improved error handling and localized messages for environment-file and key-provisioning failures.
  - Improved validation for Sendmail paths.


app/Helpers/security_helper.php
```diff
@@ -81,22 +81,20 @@ function writeEnvKey(string $envKey, string $value): bool
...
 function applyEnvKeyReplacement(string $configFile, string $envKey, string $value): string
 {
     $pattern = '/^\s*' . preg_quote($envKey, '/') . '\s*=.*/m';
 
     if (preg_match($pattern, $configFile)) {
-        return preg_replace($pattern, "$envKey='$value'", $configFile, 1);
+        $escapedValue = str_replace(['\\', '$'], ['\\\\', '\\$'], $value);
+        $replaced = preg_replace($pattern, "$envKey='$escapedValue'", $configFile, 1);
+
+        return $replaced ?? $configFile;
```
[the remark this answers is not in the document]
objecttothis: Fixed

objecttothis:
@jekkos this one is ready for review as well.

app/Helpers/security_helper.php
```diff
@@ -271,9 +271,9 @@ function checkThrottleEncryption(): string
 
         if (empty($key)) {
             $key = bin2hex(random_bytes(32));
-            $configFile = applyEnvKeyReplacement($configFile, 'throttle.key', $key);
+            $updated = applyEnvKeyReplacement($configFile, 'throttle.key', $key);
 
-            if (!atomicWriteFile($configPath, $configFile)) {
+            if ($updated === null || !atomicWriteFile($configPath, $updated)) {
                 throw new RuntimeException("Unable to persist throttle.key to $configPath");
```
jekkos: Do you know if this works on docker ?

jekkos (dismissed):
Fine but we need to validate on dev if the docker setup works fine. Can we live without having the throttler active? Might be better to log a warning instead of crashing the whole application?

objecttothis:
> Fine but we need to validate on dev if the docker setup works fine. Can we live without having the throttler active? Might be better to log a warning instead of crashing the whole application?

The build push to dev.opensourcepos.org works but the problem is that the write access to .env/.env.lock isn't there.

image

Here is what Claude  suggested

BUILD.md expect USERID/GROUPID exported (matching whoever run docker-compose up) before build, so container process UID match host file owner on the bind mount. If dev.opensourcepos.org deploy script skip that export (e.g. CI/CD runs as different user, or these env vars unset → Docker defaults them empty → container falls back to root or errors quietly → mismatch), .env/.env.lock on host (owned by whoever last touched them, maybe your local user or a prior deploy identity) won't be writable by container's runtime UID.

Fix belongs in deploy config, not this repo's PHP code — check on dev.opensourcepos.org host:
```
id -u; id -g
ls -la .env .env.lock
```

and confirm those match USERID/GROUPID passed to docker-compose.dev.yml build/run. If deploy script doesn't set them, that's the actual bug to fix — either export them correctly in the deploy pipeline, or chown .env/.env.lock to match container's actual runtime UID.

objecttothis:
@jekkos I don't have access to dev.opensourcepos.org. It's unclear to me if this is a configuration issue in the yml or not, but it's difficult to troubleshoot without SSH access.

app/Helpers/security_helper.php
```diff
@@ -68,35 +74,35 @@ function writeEnvKey(string $envKey, string $value): bool
...
-function applyEnvKeyReplacement(string $configFile, string $envKey, string $value): string
+function applyEnvKeyReplacement(string $configFile, string $envKey, string $value): ?string
 {
     $pattern = '/^\s*' . preg_quote($envKey, '/') . '\s*=.*/m';
 
     if (preg_match($pattern, $configFile)) {
-        return preg_replace($pattern, "$envKey='$value'", $configFile, 1);
+        $escapedValue = str_replace(['\\', '$'], ['\\\\', '\\$'], $value);
+
+        return preg_replace_callback($pattern, static fn () => "$envKey='$escapedValue'", $configFile, 1);
```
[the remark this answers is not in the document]
objecttothis: Fixed.

app/Helpers/security_helper.php
objecttothis: @jekkos I've written unit tests for this and it's all passing. @coderabbitai hasn't come up with any other review comments.  As you'll see below, docker on dev.opensourcepos.org may need to have some changes.  I don't have access to it, so I am passing this off to you. I think it's probably good to merge because I think the issue is configuration rather than code, but I'll hold off just in case.  We should also make sure that dev.opensourcepos.org and demo.opensourcepos.org are both accounting for the added sections of .env now. Both these need to be changed to true during deployment.

```
#--------------------------------------------------------------------
# SECURITY: DISALLOW PASSWORD CHANGE
#--------------------------------------------------------------------
# When true, disables the "change password" feature for all employees.
# Useful when passwords are managed by an external system (e.g. SSO/LDAP).
#
# DISALLOW_PASSWORD_CHANGE=false

DISALLOW_PASSWORD_CHANGE=false

#--------------------------------------------------------------------
# SECURITY: DISALLOW GRANT CHANGE
#--------------------------------------------------------------------
# When true, disables changing an employee's grants for all employees.
# New employees cannot be created with grants while this is enabled.
# Useful for demo deployments.
#
# DISALLOW_GRANT_CHANGE=false

DISALLOW_GRANT_CHANGE=false
```

objecttothis:
This one is blocked until we update the configuration on dev.opensourcepos.org and demo.opensourcepos.org then validate that it doesn't break docker installations. I'll change it to on hold until @jekkos can get to it because I don't have access to those servers.

objecttothis:
@jekkos did you close this one on purpose?  The fixes in the code are needed.

jekkos:
@objecttothis Currently the container on dev is running as it's running anywhere else. We need to review the folder permissions as other people running the docker will also face this problem otherwise.

@coderabbit can you check what might need to be changed in the Dockerfile? My guess is that the webserver user does not have write permission and this causes the write to fail

jekkos:
@coderabbit you need to look at regular Dockerfile and docker-compose.yaml. The .dev file is only used for local development

jekkos:
@coderabbit perhaps the file is created with permissions set wrongly, can you check if there is any umask for docker directory? Or can we set it writable after creation in the code? If we can create it then that should be possible right

jekkos:
@coderabbit do we need to create the lockfile? Isn't there an atomic write operation we can do instead?

jekkos:
Related direction worth considering for the container path (from #4554): for Docker/Compose the clean fix is to inject the secrets as env vars — `ENCRYPTION_KEY` (newly honored by `app/Config/Encryption.php` in #4691) and `throttle.key` (already honored by `checkThrottleEncryption()`) — so `checkEncryption()`/`checkThrottleEncryption()` short-circuit and skip the `ROOTPATH/.env` write entirely. That avoids needing to make the container root writable just to persist an auto-generated key, which expands the mutable surface over the file holding those long-lived secrets.

objecttothis:
> Related direction worth considering for the container path (from #4554): for Docker/Compose the clean fix is to inject the secrets as env vars — `ENCRYPTION_KEY` (newly honored by `app/Config/Encryption.php` in #4691) and `throttle.key` (already honored by `checkThrottleEncryption()`) — so `checkEncryption()`/`checkThrottleEncryption()` short-circuit and skip the `ROOTPATH/.env` write entirely. That avoids needing to make the container root writable just to persist an auto-generated key, which expands the mutable surface over the file holding those long-lived secrets.

I'm fine with changing our approach as long as it meets these criteria:

- Works for Docker and non-docker installs alike.
- Easily upgrades from CI3 databases.
- Prevents from accidentally committing secrets.
- Doesn't create multiple sources of truth.

jekkos:
@objecttothis — thanks, this is exactly the right piece of the puzzle for the container case behind #4554 (encrypted settings becoming unreadable after container recreate) and the `ROOTPATH/.env` writability concern from #4691. Two related improvements I'd like to add on top of your escaping fix, before we close this out:

**1. Skip the `.env` write when the key is already available via env var.**
Today `checkEncryption()` (security_helper.php:168) and `checkThrottleEncryption()` (security_helper.php:232) both unconditionally generate + write a new key into `ROOTPATH/.env` when the key is empty. In a container, `ROOTPATH/.env` is typically read-only (stock `docker-compose.yml` only mounts `uploads/` and `writable/logs`), so today's only working setup is to make that file writable. With the short-circuit, operators can just set `ENCRYPTION_KEY` (and optionally `throttle.key`) in compose `environment:` and the app never touches `ROOTPATH/.env` at all. Each install still gets its own key because the operator pastes one in once (e.g. `php -r 'echo bin2hex(random_bytes(32));' > /dev/tty; read K; docker run -e ENCRYPTION_KEY=\ ...`), no in-repo file editing required.

**2. Centralize the startup provisioning into `pre_system` instead of 8 inline calls.**
Current call-site census:
- `checkEncryption()` inline in: `app/Controllers/Config.php` ×5 (lines 67, 259, 536, 582, 649), `app/Database/Migrations/20220127000000_convertToCI4.php` ×2 (lines 41, 76), `app/Libraries/Email_lib.php` ×1 (line 32)
- `checkThrottleEncryption()` is already centralized to `app/Config/Events.php:32` on `pre_system`; the `Throttle` filter (`app/Filters/Throttle.php:30`) only reads the secret.

The asymmetry is exactly the smell: one key path is boot-managed, the other is sprinkled across the app. I'd like to:

- Add the encryption-key bootstrap to the same `pre_system` handler in `Events.php` (right alongside the existing throttle-key line), so it runs exactly once per request/CLI invocation before routing.
- Split the helper into two intents:
  - `ensureEncryptionKey()` — does the provisioning (generate + write). Called exactly once, at boot, only when no key is present.
  - `hasEncryptionKey(): bool` — read-only, no side effects. This is what the 8 inline sites should semantically be checking anyway ("is the key good yet, can I encrypt?").
- Make both idempotent: if the key is already available from env var (`ENCRYPTION_KEY`) or a `≥64`-char `encryption.key` in `.env`, return immediately with **no** file I/O.

This gives us: (a) no `.env` writes per request after first boot, (b) no `.env` writes at all in a read-only container with the env var set, (c) the 8 call sites become simple guards against a hypothetical missing key instead of the provisioning mechanism, (d) your escaping work in #4656 remains the write path for the one legitimate provisioning case (first run on a writable volume).

**Do you agree with the shape (centralize + split into ensure vs has), or would you prefer the smaller change of just adding the 'already-set → short-circuit' check on top of your existing write path and keeping the 8 call sites as-is? Both work for me; I'd lean toward the split for readability but I'll match whichever keeps the diff focused.**

objecttothis:
@jekkos I believe the asymmetry in 2. came about because when we put the encryptionCheck() in pre_system, the CI3 > CI4 key conversion was running before the migration to convert the encrypted CI3 data to encrypted CI4 data could run so it was resulting in failed migrations from CI3 databases.  I'm fine with moving to a symmetrical approach as long as migrations succeed.  Without looking at the code to refresh myself, the only way I can think of is to decouple the replacement of the CI3 encryption key with the checkEncryption() function.  I don't know if this doable though because I think it was added since CI4 does not accept CI3 encryption keys as valid.  I'm open to ideas though, if there is something I'm missing.

@coderabbitai do the proposed solutions for 1 and 2 that @jekkos posted introduce or reintroduce any security vulnerabilities? Can the proposed solution for 2 be made to work in a way that doesn't derail migrations from older CI3-based databases?

objecttothis:
@jekkos the current workflow for migration in new, existing and existing CI3 installs that are not on docker allows a sign-in and the rest is automated. In the case where the encryption.key exists in .env, it's automatically handled.  Are either of the proposed changes going to change that?  If so, what would the change be?  I am not a huge fan of solutions that turn the enter your credentials into multiple steps that requires manually changing keys and such.  That said, if we can find a solution that makes the docker deployment simpler but doesn't increase complexity of the non-docker installs/upgrades, then I'm generally all for it.

jekkos:
@objecttothis following up on the "single source of truth" concern — proposal below, want your sign-off before I open a PR.

## Design

Add a one-shot `provisioner` service to `docker-compose.yml`. On every `docker compose up` it runs before `ospos` (via `depends_on: service_completed_successfully`), checks `docker/.env` for missing keys, and appends them if not present. Exits 0 immediately.

```yaml
services:
    provisioner:
        image: alpine:3
        volumes:
            - ./docker/.env:/env/.env
        command:
            - /bin/sh
            - -c
            - |
              set -eu
              [ -f /env/.env ] || : > /env/.env
              if ! grep -q '^ENCRYPTION_KEY=' /env/.env; then
                  printf '\n# auto-generated at %s\nENCRYPTION_KEY=%s\n' "$(date -u +%FT%TZ)" \
                      "$(od -vAn -N32 -tx1 /dev/zero | tr -d ' \n')" >> /env/.env
              fi
              if ! grep -q '^THROTTLE_KEY=' /env/.env; then
                  printf 'THROTTLE_KEY=%s\n' \
                      "$(od -vAn -N32 -tx1 /dev/zero | tr -d ' \n')" >> /env/.env
              fi
        restart: 'no'

    ospos:
        image: jekkos/opensourcepos:master
        restart: always
        depends_on:
            mysql:
                condition: service_started
            provisioner:
                condition: service_completed_successfully
        ports:
            - "80:80"
        networks:
            - app_net
        volumes:
            - uploads:/app/public/uploads
            - logs:/app/writable/logs
        env_file:
            - ./docker/.env
        environment:
            - CI_ENVIRONMENT=production
            - ALLOWED_HOSTNAMES=localhost
            - FORCE_HTTPS=false
            - PHP_TIMEZONE=UTC
```

Properties of this design:

| Concern | How it's addressed |
|---|---|
| New key per install | `od /dev/zero` inside the provisioner, once per fresh `docker/.env` |
| Survives container recreate / re-pull of image | Key lives on the host at `docker/.env`, not in any container layer |
| Single source of truth | `docker/.env` is the only place keys live for Docker deploys; identical to how MySQL creds already work |
| No in-container file writes | PHP code never calls `file_put_contents` on a `.env` inside the container |
| Idempotent | `grep -q` guard — never double-writes, never overrides a user-supplied key |
| Conditional logic lives in the script, not in Compose | Compose can't express `if` natively; a one-shot container with exit code 0 is the standard pattern (same as `postgres-initdb`, `mysql --init-file`) |
| PHP side is zero-change | `app/Config/Encryption.php` already reads `getenv('ENCRYPTION_KEY')` (commit `3d1a5ec`); `checkThrottleEncryption()` already short-circuits on `env('throttle.key','')` |

Notes worth flagging:

1. **`THROTTLE_KEY`** here is the variable the app expects (`env('throttle.key')` → `THROTTLE_KEY`). I generate an independent key rather than reusing `ENCRYPTION_KEY` so that a key rotation for encryption doesn't also change throttle HMACs, and so the two concerns stay separable later. If you'd rather keep one variable, drop the second `if` block and have PHP fall back to `ENCRYPTION_KEY` when `throttle.key` is empty — that's a one-line change in `app/Config/Encryption.php`.
2. **`docker/.env` is currently only used by `docker-compose.nginx.yml` for `OSPOS_DOMAIN_NAME` / `OSPOS_STAGING`**. Adding the keys there reuses the existing file the project already ships — no new file location needed.
3. **`.env` path resolution**: `env_file: ./docker/.env` is resolved relative to `docker-compose.yml` (the root one), which is the same convention the current files use.

## What I would **not** change in `security_helper.php` for Docker

`checkEncryption()` / `checkThrottleEncryption()` stay as-is. The short-circuit paths already exist. This PR becomes purely:
- `docker-compose.yml`: add the provisioner service.
- (optional) `docker/.env`: document the two new keys in a comment.
- (optional) `docs` / README: "keys are generated on first boot and persisted in `docker/.env`".

Do you agree with this shape? Specific questions:

- **Two keys (encrypted data + throttle HMACs) vs one (reused for both)?** I lean toward two for rotation isolation; the app-side change is one line either way.
- **`alpine:3` as the provisioner image** — acceptable, or do you want an image you already trust the CI / registry pipeline for?
- **`docker/.env` vs a new `secrets/.env`** — reusing the existing file keeps the project layout unchanged; a new file keeps secrets visually separate from CI env vars.

Once you pick, I'll open a PR against `master` titled `feat(docker): auto-provision ENCRYPTION_KEY and THROTTLE_KEY` with exactly the diff above (plus a small doc note).

jekkos:
@objecttothis — refined proposal, superseding my earlier alpine-provisioner idea. The change to the app in this PR goes from "patch escaping/insertion in `applyEnvKeyReplacement`" to "stop calling `.env`-writes from the request path at runtime; make key provisioning an explicit idempotent step."

## Design

### 1. Single idempotent provisioner in `app/Helpers/security_helper.php`

```php
function provisionKeys(): bool
{
    $ok = true;
    foreach (['throttle.key', 'encryption.key'] as $key) {
        if (!empty(env($key, ''))) continue;         // "skip if already present"
        $gen = bin2hex(random_bytes(32));
        $ok  = writeEnvKey($key, $gen) && ($ok);     // existing helper, already atomic + locked
        putenv("$key=$gen"); $_ENV[$key] = $gen; $_SERVER[$key] = $gen;
    }
    return $ok;
}
```

One helper, one code path, reused from every place below. `applyEnvKeyReplacement` stays available for the provisioner (it's still the correct primitive for "insert/replace one key in a dotenv file"), but it is no longer called from `checkEncryption()` / `checkThrottleEncryption()` on the request cycle.

### 2. Runtime (app) becomes **read-only**

- `checkEncryption()` collapses to a *guard*: if `env('encryption.key','')` is empty → `throw` with the message "encryption.key must be set; run `php spark env:provision`".
- `checkThrottleEncryption()` collapses to: `return env('throttle.key', '');` — if empty, the `Throttle` filter throws with the same guidance.
- All 8 `checkEncryption()` call sites become pure read paths (no file writes, no backup, no `abortEncryptionConversion` path in the hot path).
- Crash-early: if the app is in a bad state (key missing), it 404/500s immediately with a message a user can act on, instead of silently writing to a file on every request.

This directly addresses the original #4656 bug: the edge case you were trying to patch in `applyEnvKeyReplacement` is no longer reachable from the hot path at all.

### 3. Provisioning is invoked via **one of three entry points**, depending on installation context

| Context | How | Note |
|---|---|---|
| **Regular install** (PHP available on host) | `php spark env:provision` | `app/Commands/EnvProvision.php` → `exit 0 if provisionKeys() else exit 1` |
| **Migration** (existing CI3→CI4 path already does this — `20220127000000_convertToCI4.php:41`) | Migration calls `provisionKeys()` | Runs once at `spark migrate`; idempotent, no container detection needed |
| **Bare Docker host with no PHP** | Documented one-liner: `openssl rand -hex 64 > /tmp/k && echo "encryption.key=\`cat /tmp/k\`" >> .env` (same for throttle) | One source of truth is still the single `./.env` next to `docker-compose.yml` |

All three write to **the same file** (`ROOTPATH/.env` = `./.env` from the host). No `docker/.env`. No mounted volume. No conditional compose. No alpine side-car.

### 4. `docker-compose.yml` (no mount, just `env_file`)

```yaml
services:
    ospos:
        image: jekkos/opensourcepos:master
        restart: always
        depends_on:
            - mysql
        ports:
            - "80:80"
        networks:
            - app_net
        volumes:
            - uploads:/app/public/uploads
            - logs:/app/writable/logs
        env_file:
            - ./.env          # <- single source of truth, host-only
        environment:
            - CI_ENVIRONMENT=production
            - ALLOWED_HOSTNAMES=localhost
            - FORCE_HTTPS=false
            - PHP_TIMEZONE=UTC
            - MYSQL_USERNAME=admin
            - MYSQL_PASSWORD=pointofsale
            - MYSQL_DB_NAME=ospos
            - MYSQL_HOST_NAME=mysql
```

Why this works, verified against the shipped CI4:

- `env_file:` is read by Compose **on the host** at `up` time and injected as OS env vars. The file is never mounted into the container. (Same mechanism the current `docker-compose.nginx.yml:23` already relies on for `OSPOS_DOMAIN_NAME`.)
- CI4's `DotEnv` (`vendor/codeigniter4/framework/system/Config/DotEnv.php:90-107`) explicitly **does not override already-defined OS env vars** — it checks `getenv($name, true) === false` before setting `$_ENV[$name]`. So:
  - **Docker:** OS env var wins → baked-in `/app/.env` in the image is effectively ignored.
  - **Local (no-Docker):** no OS var → DotEnv reads `./.env` → same `env()` result for the app code.
- Single code path in PHP: `env($key)` everywhere.

### 5. Net diff to the repo

- `app/Helpers/security_helper.php` — keep `writeEnvKey`, `applyEnvKeyReplacement`, `atomicWriteFile`, `lockEnvFile` (all existing, all fine). Rewrite `checkEncryption()` and `checkThrottleEncryption()` as guards. Add `provisionKeys()`.
- `app/Commands/EnvProvision.php` — new, ~15 lines, calls `provisionKeys()`.
- `app/Config/Encryption.php` — already reads `getenv('ENCRYPTION_KEY')` (commit `3d1a5ec`); can stay, but if we move fully to `env('encryption.key')` we can also drop that special-case.
- `docker-compose.yml` — add `env_file: ./.env`, keep the rest.
- `docker/.env` — deleted (was a duplicate of `./.env`).
- `docker-compose.nginx.yml`, `docker-compose.phpmyadmin.yml` — keep their `OSPOS_DOMAIN_NAME` / `OSPOS_STAGING` entries, but move those into the same `./.env` too, so there is **one** dotenv file the user edits, instead of two.
- Docs — one new section: "Keys are generated by `php spark env:provision` or the migration, and live in `./.env`".

### 6. What this PR changes conceptually

It stops being "fix escaping / insertion in `applyEnvKeyReplacement`" and becomes "remove runtime `.env` writes from the hot path; make key provisioning an explicit, idempotent, documented step." The escaping / insertion logic in `applyEnvKeyReplacement` stays, because `provisionKeys()` still needs to insert a key into an existing `.env` — but it's now called *once, off the request path*, instead of every request in N places.

## Open questions for @objecttothis

1. Do you agree with removing the hot-path writes in favor of guards + explicit provisioning?
2. Should `app/Config/Encryption.php:110` keep the `ENCRYPTION_KEY` env-var alias (commit `3d1a5ec`), or fold into `env('encryption.key')` so there's only one spelling?
3. Is `openssl rand -hex 64` in the docs acceptable as the "bare Docker without PHP" case, or do you want a different recommended tool there?
4. Two keys (`throttle.key` + `encryption.key`) stays, right? (Already agreed on in the thread.)

@coderabbit please review this design proposal. In particular:
- Is CI4's "OS env var wins over .env" precedence (DotEnv.php:90-107) a safe assumption for `env_file:` + `./.env` in Docker?
- Are there other call sites of `checkEncryption()` / `checkThrottleEncryption()` we haven't listed (I count 8 encryption + 1 throttle, all in `app/Controllers/Config.php`, `app/Database/Migrations/20220127000000_convertToCI4.php`, `app/Libraries/Email_lib.php`, and `app/Config/Events.php:32`)?
- Are there side effects of `abortEncryptionConversion()` / `removeBackup()` we'd be dropping that we should preserve (e.g., a migration that needs to revert a failed conversion)?
- Anything about `lockEnvFile()` / `atomicWriteFile()` that shouldn't be reused from the spark command path?

objecttothis:
I think we have to do two separate keys because throttler instantiates too early. I don't know alpine:3 so I'll trust your judgement on that one. Let's go with docker/.env

objecttothis:
@jekkos I just looked at your refined proposal.
1. Agreed as long as it still works for upgrades from CI3 versions without manual user intervention.
2. I prefer folding it into the existing encryption.key spelling, but if it has to stay ENCRYPTION_KEY.
3. I'll defer to you on this. I don't have an opinion.
4. Correct.  Two keys stay per previous decision.

jekkos:
@objecttothis — final shape, consolidating everything above.

## Provisioner service (replaces the earlier alpine idea)

```yaml
services:
    provisioner:
        image: jekkos/opensourcepos:master
        restart: 'no'
        volumes:
            - ./.env:/app/.env                      # host single-source-of-truth, writable
        command: ["php", "spark", "env:provision"]  # idempotent → no-op if keys present

    ospos:
        image: jekkos/opensourcepos:master
        restart: always
        depends_on:
            - mysql
            provisioner:
                condition: service_completed_successfully   # <- hard start-order guard
        ports:
            - "80:80"
        networks:
            - app_net
        volumes:
            - ./.env:/app/.env                      # same host file, now populated
            - uploads:/app/public/uploads
            - logs:/app/writable/logs
        environment:
            - CI_ENVIRONMENT=production
            - ALLOWED_HOSTNAMES=localhost
            - FORCE_HTTPS=false
            - PHP_TIMEZONE=UTC
            - MYSQL_USERNAME=admin
            - MYSQL_PASSWORD=pointofsale
            - MYSQL_DB_NAME=ospos
            - MYSQL_HOST_NAME=mysql
```

`depends_on: service_completed_successfully` is the wait: Compose doesn't start `ospos` until provisioner exits 0. No loops, no sleeps. If provisioner fails (non-zero exit), `ospos` never starts.

## The spark command (`app/Commands/EnvProvision.php`)

```php
public function run(array $params): int
{
    $path = $this->config('Paths')->rootPath . '.env';
    if (!provisionKeys() || !file_exists($path)) {
        $this->error('Failed to provision keys (expected ' . $path . ')');
        return EXIT_FAILURE;
    }
    return EXIT_SUCCESS;
}
```

One conditional, one exit code. Compose's `service_completed_successfully` maps directly to "file exists with the expected keys".

## `provisionKeys()` in `app/Helpers/security_helper.php`

```php
function provisionKeys(): bool
{
    $ok = true;
    foreach (['throttle.key', 'encryption.key'] as $key) {
        if (!empty(env($key, ''))) continue;          // skip if already present
        $gen = bin2hex(random_bytes(32));
        if (!writeEnvKey($key, $gen)) {
            $ok = false;
            continue;
        }
        putenv("$key=$gen");
        $_ENV[$key] = $gen;
        $_SERVER[$key] = $gen;
    }
    return $ok;
}
```

- Reuses the existing `writeEnvKey` / `applyEnvKeyReplacement` / `lockEnvFile` / `atomicWriteFile` — no new I/O primitives.
- Idempotent → the second+ invocation per `docker compose up` is a pure no-op.
- `putenv`/`$_ENV`/`$_SERVER` set here makes the value visible to the *current* PHP process even if it's being called from a long-lived service (belt-and-suspenders; for `spark` each invocation is a separate process anyway).

## Everything else stays the same as my previous comment

- `checkEncryption()` / `checkThrottleEncryption()` collapse to read-only guards; `app/Controllers/Config.php:67,259,536,582,649`, migration `20220127000000_convertToCI4.php:41,76`, `app/Libraries/Email_lib.php:32` all become pure `env('encryption.key')` lookups with a crash-early throw when empty.
- `docker/.env` is deleted; `docker-compose.nginx.yml`'s `OSPOS_DOMAIN_NAME` moves into the same `./.env`.
- `app/Config/Encryption.php:110`'s `getenv('ENCRYPTION_KEY')` fallback stays as a convenience alias; the canonical spelling is now `env('encryption.key')`.

## `.dockerignore` (fix for the "image ships a .env" gap)

```
.env
.env.*
```
Added. Confirmed via `cat .dockerignore` in this branch. The image now ships with **no** `.env` baked in, so the only way ospos gets keys is a mounted `./.env` (or `env_file:` in setups where the user guarantees the keys pre-exist). That's the crash-early behavior we want.

## Supersedes PR #4556

Yes. Both this PR and #4556 fix the underlying root cause of #4554 ("encrypted key lost across container restarts → stored passwords undecryptable"):

| Concern | PR #4556 | This PR |
|---|---|---|
| Key lives outside the container layer | ✅ via `WRITEPATH/config/encryption.key` fallback | ✅ via host `./.env` mounted at `/app/.env` |
| Survives container recreate / image re-pull | ✅ (writable/) | ✅ (host mount) |
| Single source of truth | ❌ two locations (`.env` + `writable/config/encryption.key`) | ✅ one `.env` |
| User-facing error when key missing | ✅ language string | (planned — will add `app/Language/en/Config.php` entry) |
| Writes happen on the hot request path | ✅ still does | ❌ no runtime writes |
| Works for throttle key (not just encryption) | ❌ encryption only | ✅ both |
| Code that has to be maintained | more (fallback path + graceful degradation) | less (single `provisionKeys()`) |

I propose #4556 be **closed as superseded** once we land here. I'll keep `app/Language/en/Config.php` from #4556 so the user-facing "encryption unavailable" message survives.

Open questions for @objecttothis remain:
1. Sign-off on the overall shape (provisioner service + mounted `./.env` + spark command + read-only runtime).
2. `app/Config/Encryption.php:110`'s `ENCRYPTION_KEY` alias: keep as a convenience, fold into `env('encryption.key')`, or both?
3. Confirm #4556 closed as superseded.

@coderabbit — please review the final shape above. In particular:
- Is `depends_on: service_completed_successfully` + a mounted `.env` the right primitive for the first-boot ordering, or is there a cleaner compose 2.x way? (I can't use `env_file:` because it's read at container-creation time, i.e. before the provisioner writes.)
- Any concerns with the spark command's `putenv`/`$_ENV`/`$_SERVER` set being redundant for `spark` (each invocation is a new process)?
- The `.dockerignore` change (` .env`, `.env.*`): any Dockerfile-stage side effects we'd miss by excluding `.env` (e.g. a later stage that *expects* it to be present)?

objecttothis:
@jekkos once this is ready, let me know so I can download the branch and test it against a CI3 database.

objecttothis:
Closing #4554

jekkos:
@coderabbitai walkthrough

jekkos:
@coderabbitai review

jekkos:
Thanks for the thorough walkthrough @coderabbitai — I verified both "blocking issues" against the code. Both findings are **correct diagnoses**, though one suggests a fix that wouldn't work. Both are now fixed in 18d15c6.

**#2 — plaintext save in the CI3 branch: confirmed real.** `EnvProvision.php` was passing the *decrypted* payload (`decryptAll()` output) straight to `saveAll()`, i.e. writing SMTP / Mailchimp / clcdesq secrets to ospos_app_config in the clear. The migration path did the right thing (`encryptAll → verifyAll → saveAll(encrypted)`). I aligned the command to that exact sequence — encrypt, verify the round trip, and only then persist the CI4 ciphertext. Added a regression test (`testCI3BranchSavesCiphertextNotPlaintext`) that asserts the payload passed to saveAll differs from every plaintext secret yet round-trips back.

**#1 — named volume at /app/.env: confirmed real; but the suggested fix is wrong.** Correct that a named volume is always a *directory*, so `ospos_env:/app/.env` forces .env to be a directory and breaks `atomicWriteFile()`'s `rename()`. However the proposed workaround — mounting at `/app/config` — fails because CodeIgniter 4 reads `${ROOTPATH}/.env` = `${FCPATH}/../.env` = `/app/.env` (see `public/index.php` → `app/Config/Paths.php`), so moving the file out of that path means the app never reads the provisioned key. The working fix is a **host-file bind mount** (`./​.env:/app/.env`) — it persists across container recreation *and* stays a regular file. That's a trade-off worth flagging: the key now lives as a host file, so it must be protected and backed up like any other secret.

Tests: security_helperTest 22/22, CI3SecretConverterTest 10/10 (incl. the new one). I haven't been able to docker-compose up in this environment to exercise the full startup path end-to-end — that's the one thing worth a manual check before merge.

app/Commands/EnvProvision.php
```diff
@@ -0,0 +1,153 @@
...
+            CLI::write('encryption.key     : CI4 key already present', 'green');
+            CLI::newLine();
+
+            return;
+        }
+
+        $converter = new CI3SecretConverter();
+
+        if ($key !== '' && strlen($key) < 64) {
+            // Legacy CI3 key is present in .env: decrypt stored secrets with it,
```
objecttothis: I think we should get rid of most of these comments when the code speaks for itself.

app/Commands/EnvProvision.php
```diff
@@ -0,0 +1,153 @@
...
+            if ((string) $value !== '') {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * True when any legacy secret row holds a non-empty value (encrypted or not).
```
objecttothis: Same with these docblocks.  I would get rid of them if the function name speaks to its use sufficiently.

app/Controllers/Config.php
```diff
@@ -648,19 +654,25 @@ public function postCheckMailchimpApiKey(): ResponseInterface
...
-        $list_id = '';
+        $postedKey  = (string) $this->request->getPost('mailchimp_api_key');
+        $postedList = (string) $this->request->getPost('mailchimp_list_id');
 
         if (checkEncryption()) {
-            $api_key_unencrypted = $this->request->getPost('mailchimp_api_key');
-            if (!empty($api_key_unencrypted)) {
-                $api_key = $this->encrypter->encrypt($api_key_unencrypted);
-            }
+            $api_key = $postedKey !== ''
```
objecttothis, line 661: PSR compliant naming, but this whole function goes away as soon as we merge #4407 so maybe skip it.

objecttothis:
@jekkos I just want to make sure you're not waiting on me for anything.  Do you need me to review this or test it yet?

app/Commands/EnvProvision.php
jekkos: Done in commit 7b2228f — both numbered step comments and the 'no key at all' comment are gone from run(). The remaining code is self-evident. CodeRabbit's #4656 review (CI3 plaintext + /app/.env volume) is already addressed in the commit above this one.

app/Commands/EnvProvision.php
jekkos: Done in commit 7b2228f — removed the group/name/usage/description docblocks, the run() docblock, anyNonEmpty() docblock, and legacySecretsPresent() docblock. Kept only the class-level docblock since it's the only place that pins down the read-only runtime design (web request is strictly read-only; keys are minted at startup by env:provision; never persist decrypted plaintext). All other wording in the file is now just code.

app/Controllers/Config.php
jekkos: Agreed, skipping — the function goes away when #4407 lands. Will revisit if it survives.

jekkos:
Addressed the three review comments; all checks green (355 tests / 1005 assertions on PHP 8.2, 8.3, 8.4).

- **`app/Commands/EnvProvision.php`** — trimmed the boilerplate docblocks and inline comments that duplicate the code (`7b2228f`). Kept only the class-level docblock, which is the sole place pinning down the read-only-runtime design (keys are minted at startup by `env:provision`, the web request is strictly read-only, decrypted plaintext is never persisted).
- **`app/Controllers/Config.php:661`** — skipped per your note; this goes away when #4407 merges.
- **`.github/workflows/phpunit.yml`** — fixed the 11 failing `ThrottleTest`/`LoginTest` cases (`50a7bc1`). They were failing because the PR makes `checkThrottleEncryption()` a read-only guard that throws when `throttle.key` is unset, but CI only ever exported `ENCRYPTION_KEY`. Two iterations:
  1. Exporting a `throttle.key` OS env var didn't work — CodeIgniter's `env()` resolves `$ _ENV[$key] ?? $_SERVER[$key] ?? getenv($key)`, and DotEnv populates `$_ENV['throttle.key']=''` from the `.env` file, so the empty value short-circuits before `getenv()` is reached.
  2. Writes the per-run key into the `.env` file itself (replacing the empty placeholder), which is exactly what `php spark env:provision` does in production and the source `env()` actually reads from.

@objecttothis ready for another look once you have a moment.

app/Commands/EnvProvision.php
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+            // No key at all. Generate one. We cannot recover existing CI3
+            // ciphertext (no CI3 key), so warn if any secret rows are present.
+            rotateEncryptionKey(null);
+            CLI::write('encryption.key     : new CI4 key generated', 'green');
+
+            if ($this->legacySecretsPresent()) {
+                CLI::write('legacy secrets     : WARNING - stored CI3 secrets found but no CI3 key to decrypt them; they could not be recovered', 'yellow');
+            }
+        }
```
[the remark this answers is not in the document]
objecttothis: @jekkos this should be evaluated and resolved because it won't let you merge unless comments get resolved.

objecttothis (commented):
Because I opened this PR, it doesn't allow me to approve it. @coderabbitai made several comments. I would address those. I'll clone the branch and test it with a CI3 database. I may not be able to get to it until Monday morning though.

objecttothis:
image

One of the requirements I listed above was not adding outside commands for existing non-docker installations. I'm not sure I understand the need for having to run an outside command when it can be done programmatically. Is it that it can't be done programmatically on docker?

objecttothis:
With an empty database, .env.example copied to .env and CI_ENVIRONMENT set to development I get
image

When I change the CI_ENVIRONMENT to production I get
image

When I click Initialize I get
image

Same database credentials so I know it isn't a username/password issue.

With a fresh copied .env.example > .env and CI_ENVIRONMENT set to production when I run `php spark env:provision` I get 
image

I think this is because the database has not been completely initialized and since the encryption.key was empty before, it might think it was a CI3 environment so it's trying to access an empty database and getting an error.  It does provision encryption.key and throttle.key

objecttothis:
If I run the external provisioning, which I'm not a fan of needing to do now, then initialize the database, I no longer get the connection error.  So the error seems to be related to the encryption.key and throttle.key not having been provisioned.

objecttothis:
@coderabbitai I believe the move to a external provisioning command is an effort to resolve issues with .env encryption key and throttle key provisioning in docker, but it creates extra work in non-docker installations that wasn't there before.  Is there a way to solve this problem for docker without creating extra work of needing to run a spark command for non-docker installations?

objecttothis:
@coderabbitai As you can see from my posts above, this wasn't my experience.  In a non-docker environment I was both instructed by the message on the page and required to run the spark command before the installation would proceed. Can you spot why this is not happening automatically?

objecttothis:
@jekkos and @coderabbitai to clarify, on both a new database and an existing CI3 database, if I run the spark provisioning command before attempting to migrate, then the migration goes smoothly with no errors.  So that part works.  We just need to tighten the error handling so that it's accurate and figure out how to get non-docker installs back to automatically provisioning rather than needing to manually run an external command.

objecttothis:
@jekkos if @coderabbitai's assessment is correct and the docker provisioning command runs before apache starts then the implementation we had before these changes should and could still work, because unless I'm mistaken, programmatically it should not try to provision a throttle key or encryption key if they already exist.  Perhaps the solution might be to revert some of the changes to the flow of the provisioning but then for docker installs keep the spark command that runs before apache starts.  I haven't looked at it in detail, so it might be more of a hybrid approach than that, but it's a thought.

jekkos:
Hi @objecttothis — an update for your CI3 test pass:

The CI3→CI4 auto-provisioning trigger has been restored in the runtime guards, gated by `.env` writability:
- `checkEncryption()` / `checkThrottleEncryption()` now provision inline when `.env` is writable (empty key → generate; short CI3 key → decrypt, rotate, re-encrypt, verify, persist the legacy secrets).
- When `.env` is not writable (e.g. Docker), they assume the key was provisioned externally and throw, pointing to `php spark env:provision`.

The short-key conversion branch is now unit-testable: `checkEncryption()` accepts an optional `CI3SecretConverter`, and I added `testCheckEncryptionConvertsCi3ShortKeyWhenEnvWritable` (fake `Appconfig` model, no DB) covering decrypt → rotate → verify → save.

Full suite is green on the current head (`c1830b8b9`). The CI3 short-key path is the part that needs a real CI3 database, so it would be great if you could pull the branch and exercise an upgrade against one.

objecttothis:
> Hi @objecttothis — an update for your CI3 test pass:
> 
> The CI3→CI4 auto-provisioning trigger has been restored in the runtime guards, gated by `.env` writability:
> - `checkEncryption()` / `checkThrottleEncryption()` now provision inline when `.env` is writable (empty key → generate; short CI3 key → decrypt, rotate, re-encrypt, verify, persist the legacy secrets).
> - When `.env` is not writable (e.g. Docker), they assume the key was provisioned externally and throw, pointing to `php spark env:provision`.
> 
> The short-key conversion branch is now unit-testable: `checkEncryption()` accepts an optional `CI3SecretConverter`, and I added `testCheckEncryptionConvertsCi3ShortKeyWhenEnvWritable` (fake `Appconfig` model, no DB) covering decrypt → rotate → verify → save.
> 
> Full suite is green on the current head (`c1830b8b9`). The CI3 short-key path is the part that needs a real CI3 database, so it would be great if you could pull the branch and exercise an upgrade against one.
> 

Excellent. I'll do it first thing in the morning.

objecttothis:
@jekkos both the OSPOS 3.3.2 upgrade went perfectly smooth and the empty manual install from non-docker environments went perfect.  When the keys were already present no rotation took place.  When no key was present it was generated.  When a short key was present it was converted, rotated and stored.  This is good to go as long as you are satisfied the @coderabbitai comments are resolved.

@coderabbitai are there any more unaddressed issues in this branch?

jekkos:
@coderabbit Please re-review the latest commit and confirm the status of the following previously open review threads:

**Fixed in `8821d2925`:**
1. `EnvProvision.php` — post-rotation `saveAll()` failure now calls `abortEncryptionConversion()` before re-throwing (same pattern as the migration).
2. `security_helper.php` — `backupEnvFile()` now returns a bool and `rotateEncryptionKey()` aborts before writing if the backup could not be created.
3. `security_helper.php` — a failed `file_get_contents()` on `.env` now throws `RuntimeException(Error.unable_to_read_env_file)` instead of silently blanking the file in both `rotateEncryptionKey()` and `provisionThrottleKey()`.

**Still open — please re-assess:**
4. `Config.php:271` — placeholder `smtp_pass_set` / `msg_pwd_set` not used in the view.
5. `app/Language/ar-EG/Error.php` — `=>` alignment of new keys.
6. `app/Language/en/Error.php` — `unable_to_create_env_file` wording is throttle-key-specific but is also used for encryption-key provisioning.
7. `tests/helpers/security_helperTest.php` — tests operate on the real `.env` / `writable/backup` paths.

Can you confirm which of 1–7 are fully resolved and which still need implementation or maintainer sign-off?

app/Helpers/security_helper.php
jekkos: For the standard `docker-compose.yml` setup: yes, it works. Here's why:

- `.env` is mounted as a **file** bind mount (`./.env:/app/.env`), and `lockEnvFile()` locks `/app/.env.lock` via `flock(LOCK_EX)`. In a single-container deployment each container has its own local `/app/.env.lock`, so `flock` behaves normally and every `atomicWriteFile` rename is serialized per container.
- `atomicWriteFile()` writes to a temp file in the *same* directory as `.env` and `rename()`s it into place — `rename()` within one container's view of the bind-mounted path is atomic from the container's perspective.

The one caveat is worth flagging: **`flock` does not coordinate across separate containers.** If you ever ran two `ospos` containers pointing at the *same* host `./.env` file, each would lock its own local `/app/.env.lock` and the lock would not be seen by the other — so two `env:provision` runs could interleave their writes to the shared `.env`. That's not the documented single-container setup, so it's low priority, but it's the honest boundary of the guarantee.

app/Controllers/Config.php
```diff
@@ -254,17 +254,18 @@ public function getIndex(): string
         $data['selected_image_allowed_types'] = explode(',', $this->config['image_allowed_types']);
 
         // Integrations Related fields
-        $data['mailchimp']    = [];
+        $data['mailchimp']     = [];
+        $data['smtp_pass_set'] = !empty($this->config['smtp_pass']);
+        $data['msg_pwd_set']   = !empty($this->config['msg_pwd']);
```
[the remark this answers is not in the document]
jekkos: Done in 771633388: both placeholders are now guarded by their secret-presence flags (email_config.php:125 checks $smtp_pass_set, message_config.php:42 checks $msg_pwd_set), so a fresh install (both flags false) shows the plain field label, not the 'Already set' hint.

app/Language/en/Error.php
```diff
@@ -1,6 +1,14 @@
 <?php
 
 return [
-    "no_permission_module" => "You do not have permission to access the module named",
-    "unknown"              => "Unexpected error",
+    'no_permission_module'           => 'You do not have permission to access the module named',
+    'unable_to_create_env_file'      => 'Unable to create {filePath} to provision throttle.key',
```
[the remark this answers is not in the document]
jekkos: Updated en/Error.php:5 and en-GB/Error.php:5 to 'Unable to create {filePath} to provision an environment key' (and the parallel L12 read message) in 771633388 — key-neutral so the same strings work for both throttle.key and encryption-key provisioning. The 9 other catalogs still carry the original localized throttle.key text; I left those for the localization maintainers rather than guessing at translations.

tests/helpers/security_helperTest.php
```diff
@@ -0,0 +1,339 @@
...
+    private ?string $throttleServerBefore = null;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        require_once __DIR__ . '/../../app/Helpers/security_helper.php';
+
+        $this->envPath    = ROOTPATH . '.env';
+        $this->backupPath = WRITEPATH . '/backup/.env.bak';
+        $this->lockPath   = ROOTPATH . '.env.lock';
```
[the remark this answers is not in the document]
jekkos: Agreed this is a real gap in parallel/long-running contexts. It's a heavier refactor (threading test paths through the helper signatures or running these tests in an isolated app copy) and I've split it out of this 3.4.2 security-fix PR into a dedicated follow-up: opensourcepos/opensourcepos#4700. The current test suite runs serially in CI and tearDown() restores state on the happy path, so I'm deferring the isolation work there to keep this PR focused.

objecttothis:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> 
> `@jekkos` I re-reviewed current HEAD at `8821d2925`.
> Item 	Status 	Assessment
> 1. Roll back after `EnvProvision::run()` `saveAll()` failure 	**Resolved** 	`EnvProvision::run()` now catches the `RuntimeException` from `CI3SecretConverter::saveAll()`, calls `abortEncryptionConversion()`, and rethrows. This matches the required post-rotation rollback behavior.
> 2. Stop rotation when the backup fails 	**Partially resolved** 	`backupEnvFile()` now returns `false` when `copy()` fails or the backup is unreadable. `rotateEncryptionKey()` stops before rotation in that case. However, `backupEnvFile()` still ignores failures from both `chmod()` calls. The original thread required validation of `mkdir()`, `copy()`, and `chmod()`. This still needs a small implementation change, or explicit maintainer sign-off that mode-setting failures are acceptable.
> 3. Stop on failed `.env` reads 	**Resolved** 	`rotateEncryptionKey()` and `provisionThrottleKey()` now log the failure and throw `RuntimeException(lang('Error.unable_to_read_env_file', ...))`. They no longer treat unreadable contents as an empty file.
> 4. Use `$smtp_pass_set` and `$msg_pwd_set` in placeholders 	**Open — implementation needed** 	`Config::getIndex()` sets both flags. `app/Views/configs/email_config.php` always shows `Config.email_smtp_pass_set`, and `app/Views/configs/message_config.php` always shows `Config.msg_pwd_set`. A fresh install therefore incorrectly says that each secret is already set.
> 5. Align `=>` in `app/Language/ar-EG/Error.php` 	**Open — implementation needed** 	The new entries on Lines 9 and 11 do not align with the existing array style. This is a formatting-only fix.
> 6. Make `Error.unable_to_create_env_file` key-neutral 	**Open — implementation needed** 	The English text on Line 5 says it is provisioning `throttle.key`, but `rotateEncryptionKey()` also uses the same key. `Error.unable_to_read_env_file` on Line 12 has the same throttle-specific wording and is also used by encryption-key rotation. Update both English messages to be key-neutral. Equivalent translated messages need localization-maintainer review if this PR updates all language catalogs.
> 7. Isolate `security_helperTest` filesystem tests 	**Open — implementation needed** 	The test class writes to the repository `.env`, `.env.lock`, and `writable/backup/.env.bak`. `tearDown()` restores normal completion only. It cannot protect against process termination, and parallel tests can interleave writes. Use test-specific paths through the helper interfaces, or run these tests in an isolated application copy.
> 
> **Summary:** Items **1** and **3** are fully resolved. Item **2** is partially resolved. Items **4–7** remain open.
> 🧠 Learnings used
> 
> _You are interacting with an AI system._

@coderabbitai 3 commits were pushed after this.  Do any of these items remain?

objecttothis:
@jekkos I'd say that as soon as we resolve 7 in a reasonable way and then we can merge. What do you think?

jekkos:
## Item 7 — test isolation: implementation plan

@objecttothis I'd like your sign-off on this approach before I implement it (and I'll keep it unpushed until you confirm).

**Goal:** `security_helperTest.php` currently operates on the repo's real `ROOTPATH/.env`, `ROOTPATH/.env.lock`, and `WRITEPATH/backup/.env.bak`. We want to redirect those three paths to a per-run sandbox so the suite is safe to run in parallel and can't leave the repo `.env` in a rotated/partial state if a test `exit()`s or hard-terminates.

**Chosen approach — a `Config` object ("change general state"), no signature changes.** The helper already consumes app config (`config('Encryption')->key`), so the idiomatic seam is a small `Config\SecurityEnv` singleton the helper reads the three paths from. This avoids threading optional path params through all ~9 functions, and avoids adding mutable global state to the non-namespaced helper file. It matches the codebase: the helper *already* trusts `config()`, and `CIUnitTestCase` returns the shared config instance (the existing tests rely on exactly that by mutating `config('Encryption')->key`).

### 1. New file `app/Config/SecurityEnv.php` (extends `BaseConfig`, mirrors `Config/Encryption.php`)
```php
class SecurityEnv extends BaseConfig
{
    public string $envPath    = ROOTPATH . '.env';
    public string $backupPath = WRITEPATH . '/backup/.env.bak';
    public string $lockPath   = ROOTPATH . '.env.lock';   // keep current default
}
```
*(I'm keeping `lockPath = ROOTPATH . '.env.lock'` as the default to avoid churning the lock-file location; flag if you'd prefer a distinct `LOCKPATH`.)*

### 2. `app/Helpers/security_helper.php` — replace the hardcoded path expressions with `config('SecurityEnv')->…` (read at call time, so tests can redirect via the shared instance). No function signatures change.
- `lockEnvFile()` → `config('SecurityEnv')->lockPath`
- `writeEnvKey()` → `config('SecurityEnv')->envPath`
- `envFileIsWritable()` → `config('SecurityEnv')->envPath`
- `checkEncryption()` (error message) → `config('SecurityEnv')->envPath`
- `rotateEncryptionKey()` → `envPath` + `backupPath`
- `provisionThrottleKey()` → `envPath` (× 2: path + log message)
- `abortEncryptionConversion()` → `envPath` + `backupPath`
- `removeBackup()` → `backupPath`

`initializeEnvFile()`'s `ROOTPATH . '.env.example'` stays — it only reads the shipped example, and the sandbox seeds `.env` up-front so that branch isn't reached.

### 3. `tests/helpers/security_helperTest.php` — point the config at a per-test sandbox
- `setUp()`: build a per-run sandbox under `sys_get_temp_dir()` (e.g. `ospos_sech_{pid}_{rand}/`), `mkdir` it + a `backup/` subdir, seed a `.env` stub, and point the three `config('SecurityEnv')` props at the sandbox. Keep the existing `config('Encryption')->key` + `throttle.key` capture (orthogonal).
- `tearDown()`: restore the three `config('SecurityEnv')` props to their defaults, then recursively delete the sandbox. Keep the existing key/throttle restore.
- Existing assertions keep running against `$this->envPath`/`$backupPath`/`$lockPath` (now sandbox paths).

### Untouched
`EnvProvision`, `Config` (×5), `Email_lib`, `Throttle`, and the CI3→CI4 migration keep their current call sites — they get the same default paths through the new config object.

**Scope:** one new config class + 11 path reads in the helper + test rewiring. No behavior change; no production signature change.

OK to proceed this way? If it lands in 3.4.2 I'll fold #4700 into this PR; otherwise I'll keep it as a follow-up and cross-link.

objecttothis:
@coderabbitai review

objecttothis:
@jekkos implementation of 7 looks good.  If there is a way to prevent config state bleed in the event a test fails and teardown doesn't run, then we should add that prevention.  If not, it's certainly no worse than the current risk of that.

jekkos:
Re: the remaining open thread _Roll back every post-rotation failure_ — addressed in 23f52dd:

- The entire post-rotation sequence (`encryptAll`, `verifyAll`, `saveAll`) is now inside a single `catch (Throwable)` that calls `abortEncryptionConversion()` before rethrowing, in all three call-sites (`checkEncryption`, `EnvProvision`, `ConvertToCI4`). So any failure — including CI4 `EncryptionException` from encrypt/decrypt and `ReflectionException` from `batch_save` — triggers the backup restore.
- `abortEncryptionConversion()` runs *in-lock* (the transaction wrapper released the lock only in its `finally`, after the rollback) so the restore is atomic with the held mutex.

I'll resolve it now; please let me know if you'd still like me to expand `abortEncryptionConversion()` to also rewrite the in-memory `config('Encryption')->key` before rethrow — I left that as-is because the failing request process will terminate shortly after and the next boot reads .env from disk, but happy to add it if useful.

objecttothis:
> Re: the remaining open thread _Roll back every post-rotation failure_ — addressed in [23f52dd](https://github.com/opensourcepos/opensourcepos/commit/23f52ddefd08e99b7ab02989d0a50d549f17b956):
> 
>     * The entire post-rotation sequence (`encryptAll`, `verifyAll`, `saveAll`) is now inside a single `catch (Throwable)` that calls `abortEncryptionConversion()` before rethrowing, in all three call-sites (`checkEncryption`, `EnvProvision`, `ConvertToCI4`). So any failure — including CI4 `EncryptionException` from encrypt/decrypt and `ReflectionException` from `batch_save` — triggers the backup restore.
> 
>     * `abortEncryptionConversion()` runs _in-lock_ (the transaction wrapper released the lock only in its `finally`, after the rollback) so the restore is atomic with the held mutex.
> 
> 
> I'll resolve it now; please let me know if you'd still like me to expand `abortEncryptionConversion()` to also rewrite the in-memory `config('Encryption')->key` before rethrow — I left that as-is because the failing request process will terminate shortly after and the next boot reads .env from disk, but happy to add it if useful.

@jekkos I think @coderabbitai won't respond to `/coderabbit review`.  I think you have to write

@coderabbitai review

I think it's fine to leave abortEncryptionConversion() as is and not expand it as long as data will not become unreadable by not expanding it.

jekkos:
Closed out the last data-safety gap on the restore path:

`abortEncryptionConversion()` previously restored the backup with a suppressed `file_put_contents()` and an unchecked `file_get_contents()`. If the restore failed after the key had already been rotated, `.env` was left holding the new CI4 key while the database still held CI3-era ciphertext — undecryptable after the next restart.

It now validates the backup is a regular readable file, checks the read for failure, and restores through the existing `atomicWriteFile()`; any failure throws so it is surfaced instead of silently corrupting the config. Added a regression test for the unreadable-backup case.

All checks are green (PHP 8.2/8.3/8.4, Build, Build Docker Image, phplint) and every review thread is resolved — the PR should be ready to merge.

Commits: 711e5d6, 4b184ed

PR #4659, opened 2026-08-24
Fix Regression caused by PSR compliance renaming that did not propagate.

objecttothis:
Corrected `check_encryption()` to `checkEncryption()` to align with coding standards.

One location previously had been missed in the refactor of check_encryption() to checkEncryption


jekkos (approved):
The linter should have caught this. Thanks for the fix!

objecttothis:
> The linter should have caught this. Thanks for the fix!

Agreed

PR #4674, opened 2026-08-28
bugfix(sales): reject non-negative gift-card amount_tendered

jekkos:
## Summary

A low-privilege cashier could POST a forged
`payment_type` of `Gift Card:<number>` with a negative `amount_tendered`.
The forged string fails the exact `=== lang(Sales.giftcard)` branch in
`Sales::postAddPayment()`, lands in the catch-all `decimal_locale` rule
(no sign check), and is later routed into `Giftcard::decrementGiftcardValue()`
by the `strstr` substring match in `Sale::save_value()` — where
`value - (-N)` **increases** the card balance, minting store credit at will.

## Changes

| File | What | Why |
|---|---|---|
| `app/Controllers/Sales.php` | Add `nonNegativeDecimal` rule + `Sales.negative_amount_tendered` message to the catch-all `amount_tendered` rules | Blocks the negative value at the controller boundary (the primary fix) |
| `app/Models/Giftcard.php` | `decrementGiftcardValue()` returns `false` if `amount <= 0` | Sink-level guard: even if any future caller passes a non-positive amount, the balance can never be *increased* |
| `app/Language/en/Sales.php` + 45 other locales | New `negative_amount_tendered` key (`en` populated, others empty) | Required by the validation message |
| `tests/Controllers/SalesControllerTest.php` | `testPostAddPaymentRejectsNegativeAmountTendered` | Controller regression test |
| `tests/Models/GiftcardTest.php` | `testDecrementGiftcardValueRejectsNegativeAmount`, `testDecrementGiftcardValueRejectsZeroAmount` | Model regression tests |

## Not fixed here (flagged)

- `Sale.php:581` uses `strstr` (substring) to detect gift-card payments
  rather than the `===` exact match used by the controller. This is the root
  cause of the dispatch mismatch and is worth a follow-up, but fixing it
  here would affect all payment routing and is outside the minimal-change
  scope of this fix.

## Affected versions

`<= 3.4.1`

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented negative, malformed, insufficient, or zero tendered amounts from being accepted during payment processing.
  - Prevented non-positive gift card adjustments from changing balances.
  - Improved validation feedback for negative tendered amounts and locale-specific decimal input.

- **Localization**
  - Updated sales payment validation messages across supported languages, including amount-due coverage messaging.

- **Tests**
  - Added coverage for invalid payments, gift card adjustments, and sale completion scenarios.


app/Language/ar-EG/Sales.php
```diff
@@ -131,6 +131,7 @@
     'must_enter_numeric'                => 'يجب إدخال رقم للمبلغ الفعلى المدفوع.',
     'must_enter_numeric_giftcard'       => 'رقم بطاقة الهدية يجب أن يكون رقم.',
     'must_enter_reference_code'         => 'يجب إدخال رقم المرجع/الاسترداد.',
+    'negative_amount_tendered'           => '',
```
objecttothis: Looks like this got entered with the `=>` one space too far, so it throws off the alignment.

app/Models/Giftcard.php
```diff
@@ -316,6 +316,11 @@ public function update_giftcard_value(string $giftcard_number, float $value): vo
      */
     public function decrementGiftcardValue(string $giftcardNumber, float $amount): bool
     {
+        // A non-positive amount could add balance (value - $amount) or be a no-op; reject it.
```
objecttothis: I don't think the comment adds anything. I think we should remove it.

tests/Controllers/SalesControllerTest.php
```diff
@@ -476,4 +476,40 @@ public function testCashierWithGrantCanChangePrice(): void
         $cart = $session->get('sales_cart');
         $this->assertEquals('0.01', $cart[1]['price']);
     }
+
+    /**
+     * Regression test for GHSA-9847 (negative gift-card amount minting).
```
objecttothis: The risk of the GHSA being posted here before publishing a release is that it potentially tips off attackers before users can patch their versions.  I think we should treat advisory details like secrets and avoid committing them.

tests/Controllers/SalesControllerTest.php
```diff
@@ -476,4 +476,40 @@ public function testCashierWithGrantCanChangePrice(): void
...
+
+        $forgedPaymentType = lang('Sales.giftcard') . ':AUDIT-100';
+
+        $response = $this->post('/sales/addPayment', [
+            'payment_type'    => $forgedPaymentType,
+            'amount_tendered' => '-500',
+        ]);
+
+        $response->assertStatus(200);
+        // The negative amount must be rejected at the controller boundary.
```
objecttothis: Code speaks for itself here.

objecttothis (changes requested):

app/Language/ar-EG/Sales.php
jekkos: Fixed — re-aligned in all 46 locale files to match the neighbouring keys.

app/Models/Giftcard.php
jekkos: Done — removed the docblock and inline comment above decrementGiftcardValue().

tests/Controllers/SalesControllerTest.php
jekkos: Agreed — removed the GHSA reference and the attack-detail description; treating the advisory details as secrets and keeping them out of the code.

tests/Controllers/SalesControllerTest.php
jekkos: Done — removed the redundant assertion comment; the assertion now stands on its own.

app/Controllers/Sales.php
```diff
@@ -424,8 +424,13 @@ public function postAddPayment(): ResponseInterface|string
...
             ];
         } else {
-            $rules    = ['amount_tendered' => 'trim|required|decimal_locale'];
-            $messages = ['amount_tendered' => lang('Sales.must_enter_numeric')];
+            $rules    = ['amount_tendered' => 'trim|required|decimal_locale|nonNegativeDecimal'];
+            $messages = [
+                'amount_tendered' => [
+                    'required'           => lang('Sales.must_enter_numeric'),
+                    'nonNegativeDecimal' => lang('Sales.negative_amount_tendered'),
+                ],
```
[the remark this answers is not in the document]
jekkos: Done — added the decimal_locale override in the catch-all message map so a malformed amount reports Sales.must_enter_numeric, and added a focused test (testPostAddPaymentRejectsMalformedAmountTendered).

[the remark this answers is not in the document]
jekkos: I looked at this. The register (app/Views/sales/register.php) always submits amount_tendered as a positive value — it defaults to the (positive) amount_due and no UI path computes a negative one. Refunds are handled with negative item quantities and with positive refund_amount_* / cash_refund fields, not with a negative addPayment() amount_tendered. A blanket nonNegativeDecimal therefore has no regression risk and also blocks the forged gift-card:balance path, so I prefer to reject it unconditionally rather than add a mode branch that would leave the hole open in return mode.

app/Controllers/Sales.php
```diff
@@ -424,8 +424,13 @@ public function postAddPayment(): ResponseInterface|string
...
+            ];
```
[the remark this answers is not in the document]
jekkos: I looked at this one specifically. The register (app/Views/sales/register.php) always submits amount_tendered as a positive value: it defaults to the (positive) amount_due and there is no UI path that computes a negative one. Refunds are handled with negative *item quantities* and with positive refund_amount_* / cash_refund fields, not with a negative addPayment() amount_tendered. So a blanket nonNegativeDecimal on the catch-all has no regression risk, and it also closes the forged Gift Card:<number> path that a mode-conditional rule would leave open in return mode. I would rather reject it unconditionally; happy to revisit if there is a concrete return flow that relies on a negative amount_tendered.

objecttothis (changes requested):
Looks good.  Just the one regression that @coderabbitai pointed out regarding negative amounts needed on returns.

tests/Controllers/SalesControllerTest.php
```diff
@@ -519,6 +519,48 @@ public function testCashierWithGrantCanChangePrice(): void
...
+            'amount_tendered' => 'ABC123',
+        ]);
+
+        $response->assertStatus(200);
+        $response->assertSee(lang('Sales.must_enter_numeric'));
+
+        $session  = Services::session();
+        $payments = $session->get('sales_payments');
+        $this->assertArrayNotHasKey(lang('Sales.cash'), (array) $payments);
+
```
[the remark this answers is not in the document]
objecttothis: @coderabbitai Did this problem come from a come from a problem in the master?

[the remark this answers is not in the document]
objecttothis: Fixed.

jekkos:
@coderabbit Thanks for the review on PR #4674. I traced the `nonNegativeDecimal` change end-to-end on the catch-all `postAddPayment()` branch (Sales.php:433) and want your opinion on one edge case before this lands:

In return mode, `postAdd()` negates the quantity (Sales.php:561), so `amount_due` (`Sale_lib.php:753`) goes **negative**, and the register pre-fills `amount_tendered` with that negative value (`register.php:484`). My concern is the **cash** and **rewards** paths, which all fall through to the new catch-all rule and pass the raw `amount_tendered` straight into `addPayment()` with no sign inversion (Sale_lib.php:586).

My read is that this is **not** a regression, because the legitimate cash refund is *derived by the system* at completion — `postComplete()` computes `amount_change = amount_due * -1` and records it as `cash_refund` (Sales.php:815) rather than trusting a manually-typed negative tendered amount.

**Could you confirm (or flag) whether any legitimate return-mode flow submits a negative `amount_tendered` through `addPayment()`, which the new catch-all `nonNegativeDecimal` would now reject?** If none does, I'm confident the fix is safe; if one does, I'd like to see that path.

app/Controllers/Sales.php
```diff
@@ -424,8 +424,14 @@ public function postAddPayment(): ResponseInterface|string
...
             ];
         } else {
-            $rules    = ['amount_tendered' => 'trim|required|decimal_locale'];
-            $messages = ['amount_tendered' => lang('Sales.must_enter_numeric')];
+            $rules    = ['amount_tendered' => 'trim|required|decimal_locale|nonNegativeDecimal'];
+            $messages = [
+                'amount_tendered' => [
+                    'required'           => lang('Sales.must_enter_numeric'),
+                    'decimal_locale'     => lang('Sales.must_enter_numeric'),
+                    'nonNegativeDecimal' => lang('Sales.negative_amount_tendered'),
```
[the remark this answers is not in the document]
objecttothis: Fixed

objecttothis (approved):

PR #4675, opened 2026-08-29
Bugfix: Gate Sales::getSearch behind reports_sales grant

jekkos:
The AJAX endpoint \`Sales::getSearch()\` underlies the Sales Takings list and was the one \`reports_sales\`-protected sibling with no grant check, so a base-sales (register) cashier could pull the full per-sale ledger.

Adds the same \`reports_sales\` guard used by getRow/getEdit/postSave/getReceipt/getInvoice/getSendPdf, returning the standard 403 JSON. Plus two regression tests.

Related security advisory: GHSA-3gxx.

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for sales search authorization.
  - Added validation checks for gift card, reference-code, and cash payment amounts.
  - Added coverage for insufficient payments and zero-payment quote or invoice completion.
  - Added security checks to ensure tax names are safely escaped.
  - Improved test consistency with shared fixtures and realistic employee, item, cart, stock, and printing data.


tests/Controllers/SalesControllerTest.php
```diff
@@ -14,8 +14,12 @@
  *
  * A cashier holding only the base "sales" grant (no "reports_sales") must
  * not be able to reach the per-sale endpoints that getManage() gates
- * behind reports_sales: getRow, getEdit, postSave, getReceipt, getInvoice,
- * getSendPdf, getSendReceipt.
+ * behind reports_sales: getSearch, getRow, getEdit, postSave, getReceipt,
+ * getInvoice, getSendPdf, getSendReceipt.
+ *
+ * Also covers GHSA-3gxx-gpwf-rcjh: getSearch() (the AJAX endpoint that
```
objecttothis: I would remove any mention of a GHSA here and probably most of the docblock here then just let the test names speak for themselves.

tests/Controllers/SalesControllerTest.php
```diff
@@ -297,6 +301,20 @@ public function testCashierWithoutReportsSalesCannotGetRow(): void
         $this->assertFalse($result['success']);
     }
 
+    public function testCashierWithoutReportsSalesCannotGetSearch(): void
+    {
+        $cashierId = $this->createCashierEmployee();
```
objecttothis, line 319: This is probably for another PR, but I wonder if we can extract these helper functions out to EmployeeFixtureTrait so that we can use them in other test files.  See app/tests/EmployeeFixtureTrait.php.  Same for line 418

objecttothis (dismissed):
Looks good.  Just take a look at my two suggestions.

tests/Controllers/SalesControllerTest.php
objecttothis: This is in #4626

objecttothis (approved):

objecttothis (dismissed):

PR #4684, opened 2026-09-01
bugfix(validation): tighten mailpath validation with strict path rule

objecttothis:
## Summary

Refactors mailpath validation in `Config::verifyEmailSettings` to use the framework's standard validation pipeline instead of ad-hoc `preg_match` checks. Adds a new reusable `valid_path_strict` rule to `OSPOSRules` that restricts input to plain filesystem path characters (letters, digits, `_`, `-`, `.`, `/`).

## Changes

- `app/Config/Validation/OSPOSRules.php` — new `valid_path_strict` rule
- `app/Controllers/Config.php` — mailpath field now validated via `$this->validateFields()` with `required`/`permit_empty` + `valid_path_strict` rules, replacing manual regex/empty checks
- `tests/Config/Validation/OSPOSRulesTest.php` — unit tests covering valid paths and rejected inputs

## Test plan

- [X] Run `OSPOSRulesTest::testValidPathStrict` via PHPUnit
- [x] In Config settings, save email settings with protocol `sendmail` and a valid mailpath (e.g. `/usr/sbin/sendmail`) — should save successfully
- [x] Save with `sendmail` protocol and empty mailpath — should show validation error
- [x] Save with non-sendmail protocol and empty mailpath — should save successfully

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email configuration validation for mail delivery paths.
  * Sendmail configurations now require a valid mail path.
  * Other mail protocols accept empty paths or safely formatted filesystem paths.
  * Invalid paths, including those containing commands, spaces, or unexpected line breaks, are rejected with clearer validation feedback.

### DIFF
--- a/app/Config/Validation/OSPOSRules.php
+++ b/app/Config/Validation/OSPOSRules.php
@@ -225,4 +225,24 @@ class OSPOSRules
 
         return true;
     }
+
+    /**
+     * Validates that the candidate is a plain filesystem path: only letters, digits,
+     * underscore, dash, dot and forward slash. Uses \A...\z (not ^...$) because PCRE's $
+     * also matches immediately before a single trailing newline, which would let a
+     * value like "/usr/bin/php\n" slip through — the bug behind GHSA-jc56-j8m6-q627.
+     *
+     * @param string $candidate
+     * @param string|null $error
+     * @return bool
+     * @noinspection PhpUnused
+     */
+    public function valid_path_strict(string $candidate, ?string &$error = null): bool
+    {
+        if ($candidate === '') {
+            return false;
+        }
+
+        return (bool) preg_match('/\A[a-zA-Z0-9_\-\/.]+\z/', $candidate);
+    }
 }
--- a/app/Controllers/Config.php
+++ b/app/Controllers/Config.php
@@ -540,16 +540,21 @@ class Config extends Secure_Controller
         $protocol = $this->request->getPost('protocol');
         $mailpath = $this->request->getPost('mailpath');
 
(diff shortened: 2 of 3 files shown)


app/Config/Validation/OSPOSRules.php
```diff
@@ -225,4 +225,24 @@ public function unicode_alpha_numeric_punct(string $candidate, ?string &$error =
...
+     * underscore, dash, dot and forward slash. Uses \A...\z (not ^...$) because PCRE's $
+     * also matches immediately before a single trailing newline, which would let a
+     * value like "/usr/bin/php\n" slip through — the bug behind GHSA-jc56-j8m6-q627.
+     *
+     * @param string $candidate
+     * @param string|null $error
+     * @return bool
+     * @noinspection PhpUnused
+     */
+    public function valid_path_strict(string $candidate, ?string &$error = null): bool
```
jekkos, line 240: @coderabbit is there any existing CI4 valudation we can leverage otherwise? Can you check if we have other form input that needs this check?

objecttothis: I checked here (https://codeigniter4.github.io/userguide/libraries/validation.html#rules-for-general-use) and with Claude, but neither came up with anything.  Maybe @coderabbitai has an idea.

objecttothis: @jekkos I think @coderabbitai is going crazy because it just contradicted itself. I like my solution because it's centralized and therefore easier to reuse and less likely to be implemented differently somewhere else.

jekkos (approved):

PR #4691, opened 2026-09-03
fix(tests): resolve all phpunit failures — clean-DB suite green (#4626)

### LINKED ISSUE #4626: PHPUnit feature tests broken against a clean test DB: session pattern, missing seed data, unowned test fixture bugs

jekkos:
Closes #4626

## Summary

Brings the PHPUnit suite (281 tests) from broken on a clean test DB to a
stable, fully-green run.

## Root causes fixed

1. **Strict-mode transaction poisoning.** A single hardcoded-username
   collision in `EmployeeTest` (`granttester` UNIQUE KEY on
   `ospos_employees.username`) set `transStatus=false` on the shared
   in-process DB connection. CI4 latches that flag for the whole process,
   so every subsequent transactional save appeared to fail. Fixed by giving
   the fixture a unique username (the pattern the trait now enforces).

2. **Sale temp-table schema drift.** `create_temp_table_sales_payments_data()`
   omitted `sale_cash_refund` and `reference_code` that the full
   `create_temp_table()` schema exposes, causing two `Unknown column` errors
   in `SalesControllerTest`. Aligned the two schemas.

3. **Test-DB bootstrap.** Added a shared-connection cache reset so stale
   connections can't carry transaction state between test classes.

## Consolidation

- New `tests/Support/EmployeeFixtureTrait` routes all test employee creation
  through `Employee::save_employee()` (the production code path), with unique
  email/username by default and optional grants + field overrides.
- Replaced six near-duplicate inline helpers across `EmployeeTest`,
  `SalesControllerTest`, and `EmployeesControllerTest` with the trait — a net
  52 lines removed, while preserving each test's specific grant set.

## Tests

Full suite, clean DB: **OK (281 tests, 816 assertions)** - 0 failures, 0 errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Customer CSV import messages now identify the number of failed customers and affected line numbers.
  * Employee creation and permission updates now handle failures more reliably.
  * Sales payment details retain cash refund totals and payment reference codes more consistently.

* **Validation**
  * Payment reference code limits must be positive integers, with the maximum equal to or greater than the minimum.


objecttothis:
@coderabbitai I believe there were three places in code where sale temp tables are created. Are all three schemas aligned now? Is anything lost in aligning the schemas vs renaming two of the temp tables to prevent collision? For example by adding the columns in, do we introduce significant query time?

objecttothis:
@jekkos take a look at your gitter messages when you get a chance.

objecttothis:
@jekkos should we assign this to 3.4.2 as well?

.github/workflows/phpunit.yml
objecttothis, line 1: @jekkos did you intend to push this commit to this branch or to the #4556 branch?

jekkos: It was required to get the tests green but agree it overlaps with the other  PR

jekkos:
@objecttothis — great questions, all three now answered with the fix in `953c2a9`.

**(a) Are the three schemas aligned?** Almost — there were actually more than one mismatch, and they're now fixed. The canonical definitions live in `Sale::create_temp_table()` (`app/Models/Sale.php`), which is the full 3-table set (the wide `sales_items_temp`, plus `sales_items_taxes_temp` and `sales_payments_temp`) and also drives `search()`/takings. `Summary_report` was creating its own narrower versions of the last two. Concretely it was missing:
- `sales_items_taxes_temp`: `internal_tax` and `sales_tax` (it only had the 4 cols `sale_id,item_id,line,tax`)
- `sales_payments_temp`: `reference_code` (it only had the 5 cols without it)

On a shared temp connection the **first** creator wins (`IF NOT EXISTS` + same connection, no DROP), so when a reader expected the missing column it hit a schema-mismatch SQL error. I added the missing columns so every creator now emits the identical set. (The wide `sales_items_temp` was already aligned; `Customer.php` creates its own 3-col `sales_items_temp` but drops it before it can collide, so it's self-contained and left alone.)

**(b) Loss vs rename?** Nothing is lost by aligning. The missing columns were real data (`reference_code`, and the internal-vs-sales tax split that `Summary_taxes` already reads), so aligning is the more correct shape — renames would just make the collision visible rather than fix it. No semantics are dropped.

**(c) Query time?** Negligible. The added columns are only extra `SUM(…)`/"GROUP_CONCAT(…)` accumulators over the *same* already-joined, already-grouped rows — one extra pass over data already in the hash/agg buffer, no additional joins, scans, or table bloat (temp tables are scratch). The full suite passes (284 tests, 822 assertions) including the report paths that exercise these.

objecttothis:
Super. I think we should go with this solution then and revert the renaming in the other PR. I'm not sure if it has been merged already or not, but merging this with the current master will show it.

objecttothis:
Sorry, I meant merging the master into this branch.

jekkos:
@objecttothis — one request and one question.

Request: if the direction is right, would it be OK to merge the test suite fixes now — this PR is what unblocks CI for everyone else in the repo.

Question: CodeRabbit flags a handful of items as out-of-scope for the linked issue (#4626): the locale-validation rule (`gte_field`) + its test, the customer-translation message update, the sales-payment temp-table columns, and the bcscale test setup. I'd rather keep this PR mergeable than wait for those to be re-scoped — do you prefer I (a) split those into a follow-up PR and keep only the clean-DB test fixes + the ENCRYPTION_KEY support here, or (b) keep it all in and justify each item against #4626 in the description? Either works for me; I'll do whichever unblocks a review fastest.

objecttothis:
> @objecttothis — one request and one question.
> 
> Request: if the direction is right, would it be OK to merge the test suite fixes now — this PR is what unblocks CI for everyone else in the repo.
> 
> Question: CodeRabbit flags a handful of items as out-of-scope for the linked issue (#4626): the locale-validation rule (`gte_field`) + its test, the customer-translation message update, the sales-payment temp-table columns, and the bcscale test setup. I'd rather keep this PR mergeable than wait for those to be re-scoped — do you prefer I (a) split those into a follow-up PR and keep only the clean-DB test fixes + the ENCRYPTION_KEY support here, or (b) keep it all in and justify each item against #4626 in the description? Either works for me; I'll do whichever unblocks a review fastest.

No, let's merge this now as working tests are the best indicator we have for regressions outside of manual integration tests.  All open PRs with testing changes need to then merge the master into their respective branches and make sure tests are passing before we can merge them.

objecttothis (dismissed):
Looks like there are merge conflicts with one of the test files, but other than that, looks good.

PR #4704, opened 2026-09-18
fix(locale): validate language_code against known locales to block path traversal

jekkos:
`postSaveLocale()` stored `language_code` from user input with **no validation**. On every request, `Load_config::load_config()` calls `Language::setLocale(current_language_code())` -> `Language::load()`, where the locale segment is interpolated into a path that is `require()`d. `languageExists()` only checks `file_exists()`, and the framework's `FileLocator` does not sanitize `..`.

An authenticated account with the `config` grant (default `admin`) could store a relative path (e.g. `../../public/uploads`), then plant a PHP file in `public/uploads/` (via the existing `logo` upload). On the next request (including `/login`), `Language::load('Common')` resolves to the planted file and executes it -> unauthenticated RCE.

## Fix
- `Config::postSaveLocale()`: validate the submitted `language` against `array_keys(get_languages())` (the single source of truth used by the locale dropdown) before storing.
- `Load_config::languageExists()`: reject any code containing `/`, `\`, or `..` (defense-in-depth against a malicious value already in the DB).
- `tests/Controllers/ConfigTest.php`: regression tests for traversal / backslash / unknown / case-mismatched language codes, and corrected the locale fixture to the real `en:english` key.

Fixes GHSA-pcmp-x8fv-f9p4.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid or unknown language selections are now rejected when saving locale settings.
  * Language values containing path traversal characters, slashes, backslashes, or null bytes are blocked.
  * Case-mismatched language codes are rejected with a validation error instead of being saved.
  * Locale validation now prevents unsafe language values from being used during configuration updates.


jekkos:
## NUL-byte hardening in `languageExists()` (addresses CodeRabbit)

The `languageExists()` guard in `app/Events/Load_config.php` already rejected path separators (`/`, `\`) and parent-dir segments (`..`), but not NUL bytes. A stored `language_code` such as `"en\0"` therefore passed the guard and reached `file_exists(APPPATH . 'Language/' . $languageCode)`. On PHP 8.5+, `file_exists()` throws a `ValueError` for NUL-byte paths, which would break configuration loading.

**Fix** — reject NUL bytes in the guard, before the filesystem check:

```php
if (strpbrk($languageCode, '/\\') !== false || str_contains($languageCode, '..') || str_contains($languageCode, "\0")) {
    return false;
}
```

**Tests** — added `tests/Events/Load_configTest.php` covering the guard: a valid code is accepted, while NUL-byte, forward-slash, backslash, and parent-dir inputs are all rejected.

The primary save-time path already enforces an allow-list (`in_array($language, array_keys(get_languages()), true)` in `Config::saveLocale`), so this is defense-in-depth for any value that reaches the load path.

PR #4707, opened 2026-09-22
feat(security): add THROTTLE_KEY env-var fallback for throttle.key

jekkos:
## Summary

`checkThrottleEncryption()` now consults the **`THROTTLE_KEY`** environment variable when `throttle.key` is empty, mirroring the existing **`ENCRYPTION_KEY`** fallback in `Config/Encryption`.

This lets Docker/Compose deployments supply the throttle HMAC secret as a real env var instead of writing a shared value into a (often read-only) `.env`.

## Why

Today the throttle key is only ever obtained from `.env` (via `env('throttle.key')`) or by provisioning into `.env`. The encryption key already has a cleaner Docker/Compose story: `Config/Encryption::__construct()` falls back to `getenv('ENCRYPTION_KEY')` when `encryption.key` is blank. The throttle key had no equivalent, so a read-only `.env` with no `throttle.key` forces provisioning (a file write) or a fatal throw.

## What changed

- `app/Helpers/security_helper.php`: `checkThrottleEncryption()` checks `getenv('THROTTLE_KEY')` only when `throttle.key` is empty. An explicit `throttle.key` always wins. (Deliberately scoped to the read/runtime guard, not `provisionThrottleKey()`, which reads the `.env` file and is exercised by the provisioning-path tests.)
- `.env.example`: documents `THROTTLE_KEY` in the throttle block, matching the `ENCRYPTION_KEY` block.
- `tests/helpers/security_helperTest.php`: two regression tests — the fallback is used (and does **not** trigger a `.env` write) when only `THROTTLE_KEY` is set, and an explicit `throttle.key` takes precedence over `THROTTLE_KEY`.

## Notes

- Safe because `THROTTLE_KEY` is a *different* name than `throttle.key`: DotEnv never populates `THROTTLE_KEY` from `.env`, so `getenv('THROTTLE_KEY')` reads the real OS env var without the `` shadowing that affects a same-named var (the reason CI writes `throttle.key` into `.env` rather than exporting it).
- No behaviour change when `THROTTLE_KEY` is unset (the common bare-metal path).

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring the throttle encryption key through the `THROTTLE_KEY` environment variable.
  - Explicit `throttle.key` configuration continues to take precedence over the environment variable.
  - Documented Docker and Compose configuration for securely supplying the key without committing it to configuration files or images.

- **Tests**
  - Added coverage for environment-variable fallback and configuration precedence.


objecttothis (approved):

objecttothis:
@jekkos are we ready for release of 3.4.2 with this one? I'm looking through the issues tracker and I don't see any new regressions reported.

jekkos:
yes we are fine, trying to sort out the release itself now in the issue you posted wrt to the 3.4.2 tag

